PR #52, opened 2014-12-19
Flaskify 2

kerrydc:
I've replaced the werkzeug calls with Flask calls, and added two flask extensions that make this easier. Tests pass.

### DIFF
--- a/ga4gh/server/__init__.py
+++ b/ga4gh/server/__init__.py
@@ -1,5 +1,9 @@
 from flask import Flask
+from flask.ext.cors import CORS
 
 app = Flask(__name__)
+CORS(app, allow_headers='Content-Type')
+
+app.config['MAX_CONTENT_LENGTH'] = 2 * 1024 * 1024 # 2MB 
 
 import views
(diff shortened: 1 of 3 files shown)


jeromekelleher:
Looks great, thanks @kerrydc! Two questions:

1. Have you tested this against the compliance tester (https://github.com/ga4gh/compliance)? There is a slight irritation at the moment as the compliance tester uses coordinates that are not in our example data. The simplest workaround for me was to just change the coordinates in the JS code for the compliance tester.
2. Would it be possible to add some automated tests for CORS compliance? If this is a big deal then don't worry about it, but I thought that this might be a good opportunity to start using the framework that @melaniedc started.

skeenan:
Is this ready to merge?

kerrydc:
Hi Everyone,
Melanie is working on front end tests, so CORS will be part of that. The
compliance tests are currently broken (in this PR and also in master), but
I will submit a PR to fix them on Friday. So, ready to merge.
Kerry

Issue #10, opened 2014-09-03
Response generation model is poor


jeromekelleher:
The current model for generating responses in the server is to construct a complete `GASearchVariantsResponse` object and to then convert this to a JSON string. This is a poor model because it entails creating lots of intermediate objects and lists of `GAVariants` and `GACalls`. This can consume large amounts of memory on the server and is a factor in the current poor performance when returning large numbers of calls. It would be much better if we could generate the JSON directly and write this to a buffer, without generating the lists of `GAVariants` and `GACalls`.

We need:

- An abstraction that lets us write JSON encoded attributes directly to the buffer. We should also bear in mind we'll want to implement [partial responses](https://developers.google.com/discovery/v1/performance) at some point. This also feeds into issue #5 where we should keep this problem in mind when generating the protocol element classes from Avro. 
- Management of the number of records written to the buffer according to the requested page size and `nextPageToken` handling.
- An approximate limit in bytes over which the response cannot grow. If we exceed this limit while generating a record, we terminate the response once the current record has been completed. This allows us to limit server memory resources and manage latency.

We should also keep the ability to directly convert protocol elements to JSON though, as this is useful for clients.

jeromekelleher:
I'm a big fan of generators @bartgrantham, and I definitely would like to see them used here. As I outlined above, I think a function `getVariants(request)` which is a generator over all the variants in a request (for example) would simplify things a lot. However, I worry about having generators all the way down, as I don't know how we would handle errors in this case. Suppose an error occured while generating the 100th variant in a list, and Flask had already started writing the HTTP response to the socket. How would we handle this error then? 

For this reason I think it's better to fill a buffer with the response which we only write back to the client when it's complete. This way we can properly handle errors whenever they occur.

Thanks for the links above though; I think this is exactly what we'll want when we're serving reference sequences. The approach in the protocol is quite clunky at the moment, and HTTP streaming would seem to be a much simpler solution (as discussed with @cassiedoll and @calbach a while back).

pgrosu:
The HTTP streaming discussion seems to skip my memory.  Could you point me to it so I refresh my mind? :)

Thanks,
~p

jeromekelleher:
Discussion happened offline @pgrosu --- it'll become a PR on the schemas repo at some point.

pgrosu:
Ah!  Sounds good, I was just wondering if it had anything to do with something along these lines:

http://flask.pocoo.org/docs/0.10/patterns/streaming/

[...]

PR #57, opened 2014-12-31
Style

kerrydc:
Please look over this and make sure all of my variable name changes are helping to make the code more readable. I will have one more follow up commit to fix line length, but wanted feedback so that I could add other changes while doing that.

### DIFF
--- a/ga4gh/backends.py
+++ b/ga4gh/backends.py
@@ -159,8 +159,8 @@ class WormtableVariantSet(object):
                 else:
                     if row[rowPosition] is not None:
                         # Missing values are not included in the info array
-                        call.info[info] = \
-                            self.convertInfoField(row[rowPosition])
+                        call.info[info] = self.convertInfoField(
+                            row[rowPosition])
             variant.calls.append(call)
         return variant
 
@@ -341,22 +341,22 @@ class TabixVariantSet(object):
         if request.end is None:
             request.end = request.start + 1
 
-        cursor = self._tabixMap[request.referenceName]\
-            .fetch(str(request.referenceName), j-1, int(request.end))
+        cursor = self._tabixMap[request.referenceName].fetch(
+            str(request.referenceName), j-1, int(request.end))
         # Weird with the -1 but seems required. 0 based i guess?
-        while len(variant) < request.pageSize:
+        while len(variants) < request.pageSize:
             try:
                 record = cursor.next()
             except StopIteration:
                 break
-            variant.append(self.convertVariant(record))
+            variants.append(self.convertVariant(record))
         try:
             record = cursor.next()
         except StopIteration:
             pass
         else:
             response.nextPageToken = self.convertVariant(record).start
-        response.variants = variant
+        response.variants = variants
 
         return response
 
@@ -370,9 +370,10 @@ class Backend(object):
     """
     Superclass of GA4GH protocol backends.
     """
-
-    def __init__(self, dataDir, variantSet):
+    # TODO: create real parent class for VariantSet.
+    def __init__(self, dataDir, variantSetClass):
         self._dataDir = dataDir
+        print(dataDir)
         self._variantSetMap = {}
         # All directories in datadir are assumed to correspond to VariantSets.
         for variantSetId in os.listdir(self._dataDir):
@@ -381,8 +382,9 @@ class Backend(object):
             # so we skip non-directory files.
             if os.path.isfile(relativePath):
                 continue
-            self._variantSetMap[variantSetId] = variantSet(variantSetId,
-                                                           relativePath)
+            print(variantSetId)
+            self._variantSetMap[variantSetId] = variantSetClass(
+                variantSetId, relativePath)
         self._variantSetIds = sorted(self._variantSetMap.keys())
 
     def searchVariants(self, request):
@@ -397,8 +399,8 @@ class Backend(object):
             else:
                 pageToken = int(pageToken)
             request.pageToken = pageToken
-        variantSet = self\
-            ._variantSetMap[request.variantSetIds[variantSetIndex]]
+        variantSet = self._variantSetMap[
+            request.variantSetIds[variantSetIndex]]
         response = variantSet.searchVariants(request)
         # Add the index of the variant set of the next
         # page to the nextPageToken
@@ -434,8 +436,8 @@ class Backend(object):
             variantSet = protocol.GAVariantSet()
             variantSet.id = self._variantSetIds[currentIndex]
             variantSet.datasetId = "NotImplemented"
-            variantSet.metadata = self._variantSetMap[variantSet.id]\
-                .getMetadata()
+            variantSet.metadata = self._variantSetMap[
+                variantSet.id].getMetadata()
             variantSetList.append(variantSet)
             currentIndex += 1
         if currentIndex < len(self._variantSetIds):
--- a/ga4gh/backends.py
+++ b/ga4gh/backends.py
@@ -150,7 +150,8 @@ class WormtableVariantSet(object):
             # to call back to searchCallSets to get more info?
             call.callSetId = callSetId
             call.callSetName = callSetId
-            for (info, col), rowPosition in zip(self._sampleCols[callSetId], rowPositions):
+            for (info, col), rowPosition in zip(self._sampleCols[callSetId],
+                                                rowPositions):
                 if info == self.GENOTYPE_LIKELIHOOD_NAME:
                     call.genotypeLikelihood = row[rowPosition]
                 elif info == self.GENOTYPE_NAME:
@@ -158,7 +159,8 @@ class WormtableVariantSet(object):
                 else:
                     if row[rowPosition] is not None:
                         # Missing values are not included in the info array
-                        call.info[info] = self.convertInfoField(row[rowPosition])
+                        call.info[info] = \
+                            self.convertInfoField(row[rowPosition])
             variant.calls.append(call)
         return variant
 
@@ -219,7 +221,8 @@ class WormtableVariantSet(object):
             # end the iteration early when we have generated enough results.
             currentRow = next(cursor, None)
             while currentRow is not None and len(variants) < request.pageSize:
-                variants.append(self.convertVariant(currentRow, sampleRowPositions))
+                variants.append(self.convertVariant(currentRow,
+                                                    sampleRowPositions))
                 currentRow = next(cursor, None)
             if currentRow is not None:
                 response.nextPageToken = currentRow[self.POS_COL]
@@ -235,9 +238,10 @@ class WormtableVariantSet(object):
                 # The result must still be within the range and must match
                 # the specified name exactly. The cursor is positioned at
                 # the first row >= the specified key.
-                if request.start <= currentRow[self.POS_COL] < request.end \
-                        and currentRow[self.ID_COL] == name:
-                    variants.append(self.convertVariant(currentRow, sampleRowPositions))
+                if (request.start <= currentRow[self.POS_COL] < request.end and
+                        currentRow[self.ID_COL] == name):
+                    variants.append(self.convertVariant(currentRow,
+                                                        sampleRowPositions))
         response.variants = variants
         return response
 
@@ -245,7 +249,7 @@ class WormtableVariantSet(object):
         """
         Returns a list of GAVariantSetMetadata objects for this variant set.
         """
-        def getMetadata(infoField, col):
+        def f(infoField, col):
             metadata = protocol.GAVariantSetMetadata()
             metadata.key = infoField
             metadata.value = ""
@@ -258,13 +262,13 @@ class WormtableVariantSet(object):
             return metadata
         ret = []
         for infoField, col in self._infoCols:
-            ret.append(getMetadata(infoField, col))
+            ret.append(f(infoField, col))
         if len(self._sampleCols) > 0:
             # TODO this is pretty nasty, making a list just to take the head.
             sampleName = list(self._sampleCols.keys())[0]
             for infoField, col in self._sampleCols[sampleName]:
                 if infoField != self.GENOTYPE_NAME:
-                    ret.append(getMetadata(infoField, col))
+                    ret.append(f(infoField, col))
         return ret
 
 
@@ -292,7 +296,8 @@ class TabixVariantSet(object):
         record = record.split('\t')
         position = int(record[1])
 
-        variant.id = "{0}:{1}:{2}".format(self._variantSetId, record[0], position)
+        variant.id = "{0}:{1}:{2}".format(self._variantSetId,
+                                          record[0], position)
         variant.variantSetId = self._variantSetId
         variant.referenceName = record[0]
         variant.names = []  # What's a good model to generate these?
@@ -329,7 +334,7 @@ class TabixVariantSet(object):
         response = protocol.GASearchVariantsResponse()
         response.variantSetId = self._variantSetId
 
-        variant = []
+        variants = []
         j = int(request.start)
         if request.pageToken is not None:
             j = int(request.pageToken)
@@ -366,18 +371,19 @@ class Backend(object):
     Superclass of GA4GH protocol backends.
     """
 
-    def __init__(self, dataDir, VariantSet):
-        # We will use dataDir as the Dataset Id.
+    def __init__(self, dataDir, variantSet):
         self._dataDir = dataDir
-        self._variantSetDict = {}
+        self._variantSetMap = {}
         # All directories in datadir are assumed to correspond to VariantSets.
         for variantSetId in os.listdir(self._dataDir):
             relativePath = os.path.join(self._dataDir, variantSetId)
-            # A variant set is all files under a directory, so we skip non-directory files.
+            # A variant set is all files under a directory,
+            # so we skip non-directory files.
             if os.path.isfile(relativePath):
                 continue
-            self._variantSetDict[variantSetId] = VariantSet(variantSetId, relativePath)
-        self._variantSetIds = sorted(self._variantSetDict.keys())
+            self._variantSetMap[variantSetId] = variantSet(variantSetId,
+                                                           relativePath)
+        self._variantSetIds = sorted(self._variantSetMap.keys())
 
     def searchVariants(self, request):
         assert len(request.variantSetIds) > 0
@@ -391,7 +397,8 @@ class Backend(object):
             else:
                 pageToken = int(pageToken)
             request.pageToken = pageToken
-        variantSet = self._variantSetDict[request.variantSetIds[variantSetIndex]]
+        variantSet = self\
+            ._variantSetMap[request.variantSetIds[variantSetIndex]]
         response = variantSet.searchVariants(request)
         # Add the index of the variant set of the next
         # page to the nextPageToken
@@ -413,6 +420,7 @@ class Backend(object):
         """
         # This is temporary until we do this properly based on the output
         # buffer size
+        print(request)
         if request.pageSize is None:
             request.pageSize = 100
         response = protocol.GASearchVariantSetsResponse()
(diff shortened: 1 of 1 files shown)


ga4gh/backends.py
```diff
@@ -186,19 +186,19 @@ def searchVariants(self, request):
         else:
             readCols = self._table.columns()[:self._firstSamplePosition]
             callSetIds = request.callSetIds
-            for csid in request.callSetIds:
-                cols = [c for name, c in self._sampleCols[csid]]
+            for callSetId in request.callSetIds:
+                cols = [col for name, col in self._sampleCols[callSetId]]
                 readCols.extend(cols)
         # Now we get the row positions for the sample columns
```
Naburimannu: Nit: the nested for loops are obscure enough we should consider commenting *why* we're doing this, or perhaps how, or looking for a better control structure.

ga4gh/backends.py
```diff
@@ -186,19 +186,19 @@ def searchVariants(self, request):
...
-            l = []
-            for c in self._sampleCols[csid]:
-                l.append(j)
-                j += 1
-            sampleRowPositions[csid] = l
-        v = []
+        currentRowPosition = self._firstSamplePosition
+        for callSetId in callSetIds:
+            sampleRowPositionsList = []
+            for col in self._sampleCols[callSetId]:
```
Naburimannu: (Maybe there's a pythonic range generator? All we're doing is appending [currentRowPosition...(currentRowPosition+#cols-1)] to the list, right? But this is totally off-topic for this PR)

ga4gh/backends.py
```diff
@@ -329,7 +329,7 @@ def searchVariants(self, request):
         response = protocol.GASearchVariantsResponse()
         response.variantSetId = self._variantSetId
 
-        v = []
+        variant = []
```
jeromekelleher: variants would be better here - this is the list of variants we're building up.

ga4gh/backends.py
```diff
@@ -365,18 +365,23 @@ class Backend(object):
...
+    def __init__(self, dataDir, VariantSet):
+        # We will use dataDir as the Dataset Id.
         self._dataDir = dataDir
-        self._variantSets = {}
-        # All files in datadir are assumed to correspond to Datasets.
-        for vsid in os.listdir(self._dataDir):
-            f = os.path.join(self._dataDir, vsid)
-            self._variantSets[vsid] = Dataset(vsid, f)
-        self._variantSetIds = sorted(self._variantSets.keys())
+        self._variantSetDict = {}
```
jeromekelleher: ``self._variantSetMap`` perhaps? We've used ``_tabixMap`` above, so we should probably settle on one convention if we're going to encode some type hints in the variable name. Should we convey that we're mapping ids to objects in there (i.e. ``self._variantSetIdMap``)? This might be a good convention to adopt to make it clear when we're mapping an id to an object?

ga4gh/backends.py
```diff
@@ -411,20 +416,20 @@ def searchVariantSets(self, request):
...
-            j += 1
-        if j < len(self._variantSetIds):
-            response.nextPageToken = j
-        response.variantSets = v
+            currentIndex = int(request.pageToken)
+        variantSetList = []
+        while currentIndex < len(self._variantSetIds) and len(variantSetList) < request.pageSize:
+            variantSet = protocol.GAVariantSet()
+            variantSet.id = self._variantSetIds[currentIndex]
+            variantSet.datasetId = self._dataDir
```
jeromekelleher: Is this a good idea? dataDir might contain an arbitrary file system path, which I don't think we should leak. Plus it's not very good as an ID. Maybe we should just have an attribute ``self._id`` which is derived when we instantiate the VariantSet? This is probably outside the context of this PR, so perhaps it's better to revert to the static string for now.

jeromekelleher:
This looks great, thanks for going to the effort of understanding my needlessly terse variable names! I completely agree with the spirit of this, and I accept that single letter variables are over used. I still like them, and I think it's quite a Pythonic thing to do, but we should probably only use single letter variables within a very limited range (say 5 lines), so one can easily look back to see how the variable is defined. 

I agree with pretty much all the changes you've made though.

In general, all of this code needs a cleanup, as there is lots of shared code between the VariantSet subclasses and in other places. These are beyond the scope of this PR though I think.

Would it be possible to separate out the commits for just this PR please? It's confusing to see the commits for other PRs included.

Naburimannu:
+1 modulo nits and Jerome's comments; you're being slightly more verbose than I would, but not absurdly so.

ga4gh/backends.py
```diff
@@ -280,30 +280,30 @@ def __init__(self, variantSetId, vcfPath):
...
-            tabixfile = pysam.Tabixfile(fn)
+        for vcfFile in glob.glob(os.path.join(vcfPath, "*.vcf.gz")):
+            tabixfile = pysam.Tabixfile(vcfFile)
             for chrom in tabixfile.contigs:
                 if chrom in self._tabixMap:
                     raise Exception("cannot have overlapping VCF files.")
                 self._tabixMap[chrom] = tabixfile
 
     def convertVariant(self, record):
-        v = protocol.GAVariant()
```
hershman: Going from v to variant adds a lot of verbosity to this function.  Given that it's called convertVariaint is v is really ambiguous?

ga4gh/backends.py
pashields: +1 to variantSetIdMap

ga4gh/backends.py
```diff
@@ -411,20 +416,20 @@ def searchVariantSets(self, request):
...
-            vs.datasetId = "NotImplemented"
-            vs.metadata = self._variantSets[vs.id].getMetadata()
-            v.append(vs)
-            j += 1
-        if j < len(self._variantSetIds):
-            response.nextPageToken = j
-        response.variantSets = v
+            currentIndex = int(request.pageToken)
+        variantSetList = []
+        while currentIndex < len(self._variantSetIds) and len(variantSetList) < request.pageSize:
```
pashields: pretty sure you'd want to use slices here. Am I missing something? Something like:

```python
start = int(request.pageToken)
end = start + request.pageSize
variantSetIdsInPage = self._variantSetIds[start:end]
for variantSetId in variantSetIdsInPage:
  ...
response.nextPageToken = start + len(variantSetList)
```

ga4gh/backends.py
pashields: https://docs.python.org/2/library/itertools.html#itertools.count I believe is the best way to do that.

ga4gh/backends.py
pashields: I default to saying it's ambiguous and the verbosity is limited. Particularly for people who are used to a particular shorthand that is different. For instance, above, there was a variable ```fn``` which I would assume is a reference to a function, but it's actually suppose to be filename. I don't have any associations with v, but I'm sure somebody does.

That being said, I can imagine a style proposal (that would need to be documented and voted upon, presumably) where we reserve a few one letter variables like:

variable name|type
-------------------|-----
v|GAVariant
vs|GAVariantSet

Etc, etc. This allows a controlled level of terseness and keeps everyone pretty sync'd about what names are appropriate.

ga4gh/backends.py
pgrosu: @Naburimannu, wouldn't it be easier to just select it by index like this:

```
cols = list(self._sampleCols[callSetId][0])[1:]
```

The code looks really nice, but it would be great to add more documentation regarding the process that happens behind the scenes.  Maybe something like this might be nice:

```
# The selection is performed to return a list(tuple) for the column of interest
#   for the specificed callSetId from self._sampleCols, as such:
#  
#   [(u'GT', <wormtable.Column object at 0x2aaab7d70090>)]
#
# Then the required information is a list containing only the wormtable column, as follows:
#
#   [<wormtable.Column object at 0x2aaab7d70090>]
```

Just a thought,
~p

ga4gh/backends.py
jeromekelleher: Yeah, this is cleaner all right. What I'm really keen to do though is abstract away all of this paging logic away from the individual methods. Paging should be implemented exactly once, and I think a nice way to do it is to make  accessor methods (`getVariants`, `getVariantSets`, etc) that are generators over all the objects specified in the request. There are some details about how we deal with pageTokens, but I'm sure we can figure this out. I outlined this idea in the discussion for #10 earlier.

I guess the larger point here is that I don't see much of this code as 'final'; a lot of it needs to be aggressively refactored as we generalise to support the rest of the protocol. As we refactor it we should definitely examine the style and make sure that it's as easy as possible to understand.

ga4gh/backends.py
```diff
@@ -217,60 +217,60 @@ def searchVariants(self, request):
...
     def getMetadata(self):
         """
         Returns a list of GAVariantSetMetadata objects for this variant set.
         """
-        def f(infoField, col):
-            md = protocol.GAVariantSetMetadata()
-            md.key = infoField
-            md.value = ""
-            md.id = ""
+        def getMetadata(infoField, col):
```
jeromekelleher: Is it confusing to have a function called `getMetadata` embedded in a method called `getMetadata`? I'd be keen to keep the convention of using `f` as a locally defined, single-use function. I'm happy to go with the consensus here though, if others disagree - how does everyone else feel about this?

Thanks for doing this again @kerrydc, this is a great exercise for us to make our coding style more explicit. We should probably write up the consensus, which we can add to over time.

ga4gh/backends.py
jeromekelleher: +1 to a style proposal - we should have a document that we can point people to when these sort of issues arise.

ga4gh/backends.py
```diff
@@ -365,18 +365,23 @@ class Backend(object):
     """
     Superclass of GA4GH protocol backends.
     """
-    def __init__(self, dataDir, Dataset):
+
+    def __init__(self, dataDir, VariantSet):
```
pashields: lowercase this pls

jeromekelleher: I find this a bit confusing all right; what about `variantSetClass` to make the type explicit?

Naburimannu: I don't understand what the confusion is?

pashields: the capital V. This is a variable name, not a class name, and as such should not be capitalized based on the conventions of the rest of the code.

At least, that was my nitpick.

ga4gh/backends.py
pgrosu: +1 I definitely agree.  For clarity, we can have it in both of the of the following locations, with an additional Readme file:

1) At the beginning of the file as in the following for defining reads, though it would pertinent to the `[server]` implementation:

https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/reads.avdl

2) Before each function definition if it is local to that scope.

I feel that each function could benefit with an additional function contract for better clarity.  This would entail something as follows, besides what is already written as a description.  If the functions are too complex, then we can modularize them:

```
 /*   
  *   convertVariant: ( self, record ) -> GAVariant
  *
  *   Definition: It will create a GAVariant from information related in
  *               self, and the variant and call information in the
  *               record.  GAVariant will also store all the relevant
  *               GACall information associated to it, that is extracted
  *               from record.
  *    
  *   Inputs:
  *       self   : is a ...
  *       record : is a ...
  *    
  *   Output: GAVariant
  *
  *   Returns: A populated GAVariant object.
  */
```
 ~p

skeenan:
Is this ready to merge?

kerrydc:
Not yet, one more round of discussion and a follow-up commit are needed.
Thanks!
Kerry

bioinformed:
Not to rush you, but I'm waiting on this to land before adding a new
backend for htslib VCF/BCF.

-Kevin


On Tue, Jan 6, 2015 at 11:23 AM, Kerry Donny-Clark <notifications@github.com

kerrydc:
Sure, I'll try to get this wrapped up tonight/tomorrow AM.

ga4gh/backends.py
kerrydc: Done.

ga4gh/backends.py
kerrydc: Done.

ga4gh/backends.py
kerrydc: Done.

ga4gh/backends.py
kerrydc: I will leave project-wide style guides and such issues for a follow up PR. For a public project I think clarity should trump terseness at all times. In any case it is easier to understand code where variables say clearly what they are. This makes code more self-documenting, and decreases ramp-up time. For example, I had to re-read this code many times to figure out what to rename variables to, indicating that they were not clear to begin with.

ga4gh/backends.py
kerrydc: I'm trying to change style mostly, and leave code changes for a follow up.

ga4gh/backends.py
kerrydc: Good point, I will rename the inner function.

kerrydc: Renaming to f for now, though I strongly disagree with the convention. 'f' as a single letter is obscure, and commonly used for file as well. Better to spell out what each function does.

ga4gh/backends.py
kerrydc: Done.

ga4gh/backends.py
kerrydc: Sounds good on both points, but I'd like to push this off to a follow up PR that addresses code issue in the backends class in general.

ga4gh/backends.py
pgrosu: +1 @kerrydc, I also agree with expanding `f` as well.  The code will flow more clearly that way.

ga4gh/backends.py
kerrydc: See my comment above, this is also something I'd like to postpone.

ga4gh/backends.py
jeromekelleher: OK, lets leave it as `f` for now, but try and come up with a better convention for this sort of thing. I don't feel strongly about using/not using single-letter variables, it'll just take me a little getting used to on my part.

jeromekelleher:
I'm +1 to merge whenever @kerrydc is happy with this. (I'm not sure how much I'll be able to check in over the next few days.)

pashields:
I'll also +1 this, though it looks like your changes from Flaskify are on this branch as well @kerrydc. It would be better if they were not when the merge occurs.

ga4gh/backends.py
pgrosu: +1 @kerrydc :)

ga4gh/backends.py
pgrosu: Understandable and agreed :)

kerrydc: All tests pass, ready for merge.
Kerry

ga4gh/backends.py
jeromekelleher: OK, point conceded! `f` is a terrible name for a function (`buildMetadata` is much better), and we should deprecate the short variable names as we go forward.

ga4gh/backends.py
```diff
@@ -339,19 +344,19 @@ def searchVariants(self, request):
         cursor = self._tabixMap[request.referenceName]\
             .fetch(str(request.referenceName), j-1, int(request.end))
         # Weird with the -1 but seems required. 0 based i guess?
-        while len(v) < request.pageSize:
+        while len(variant) < request.pageSize:
```
jeromekelleher: Shouldn't this be `variants`?

ga4gh/backends.py
```diff
@@ -386,8 +397,9 @@ def searchVariants(self, request):
             else:
                 pageToken = int(pageToken)
             request.pageToken = pageToken
-        ds = self._variantSets[request.variantSetIds[variantSetIndex]]
-        response = ds.searchVariants(request)
+        variantSet = self\
```
jeromekelleher: I would use a hanging/visual indent here instead of breaking the line, e.g.
```python
variantSet = self._variantSetMap[
    request.variantSetIds[variantSetIndex]]
```

ga4gh/backends.py
```diff
@@ -217,60 +219,62 @@ def searchVariants(self, request):
...
     def getMetadata(self):
         """
         Returns a list of GAVariantSetMetadata objects for this variant set.
         """
-        def f(infoField, col):
-            md = protocol.GAVariantSetMetadata()
-            md.key = infoField
-            md.value = ""
-            md.id = ""
+        def buildMetadata(infoField, col):
```
jeromekelleher: Actually, this function is a bit pointless. We should just get rid of it and inline the code into the loop below.

ga4gh/backends.py
```diff
@@ -365,18 +370,24 @@ class Backend(object):
     """
     Superclass of GA4GH protocol backends.
     """
-    def __init__(self, dataDir, Dataset):
+
+    def __init__(self, dataDir, variantSet):
```
jeromekelleher: I would prefer `variantSetClass` here. To me this reads like the parameter is an *instance* of `VariantSet` not a *subclass* of `VariantSet`.

ga4gh/backends.py
```diff
@@ -408,23 +420,27 @@ def searchVariantSets(self, request):
...
-        response.variantSets = v
+            currentIndex = int(request.pageToken)
+        variantSetList = []
+        print(self._variantSetIds)
+        while (currentIndex < len(self._variantSetIds) and
+               len(variantSetList) < request.pageSize):
+            variantSet = protocol.GAVariantSet()
+            variantSet.id = self._variantSetIds[currentIndex]
+            variantSet.datasetId = "NotImplemented"
+            variantSet.metadata = self._variantSetMap[variantSet.id]\
```
jeromekelleher: Again, I think we should prefer hanging indents over breaking lines. [PEP 8](https://www.python.org/dev/peps/pep-0008/) says: "The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation."

Have you run this code through the `pep8` tool?

ga4gh/backends.py
```diff
@@ -122,45 +122,47 @@ def convertVariant(self, row, sampleRowPositions):
...
-                    self.convertGenotype(call, row[j])
+                    self.convertGenotype(call, row[rowPosition])
                 else:
-                    if row[j] is not None:
+                    if row[rowPosition] is not None:
                         # Missing values are not included in the info array
-                        call.info[info] = self.convertInfoField(row[j])
-            v.calls.append(call)
-        return v
+                        call.info[info] = \
```
jeromekelleher: I would prefer (see below for reason)
```python
call.info[info] = self.convertInfoField(
    row[rowPosition])
```

jeromekelleher:
Sorry for the late comments here - I've only just had a chance to look through the latest version. It looks to me like there is a bug in `searchVariants` method of the `TabixBackend` --- there is no test coverage here, which is why it didn't show up.

I think we should fix this, and clear up our policy on breaking lines with `\` (which to me should be done only as a last resort) before merging.

ga4gh/backends.py
kerrydc: Done.
Note: There is no VariantSet superclass. VariantSets just inherit from Object. Here we expect only an object that has a constructor and some methods. A later PR should create a VariantSet superclass for sure.

ga4gh/backends.py
kerrydc: pep8 passed all of the above. I have updated to hanging bracket or parens, and this also passes.

ga4gh/backends.py
kerrydc: Done.

ga4gh/backends.py
kerrydc: For a later PR with code fixing in mind.

ga4gh/backends.py
kerrydc: Done.

kerrydc:
Now all comments addressed, tests pass, calls made to both backends using cli for variants-search and variantsets-search methods.

hershman:
+1

jeromekelleher:
+1. Some `print()` calls have appeared in the `TabixBackend` class which I wouldn't like to see ordinarily, but since these classes need a fairly hefty refactoring (and the tabix backend needs quite a lot of work to get to the same point as the wormtable backend), I don't think it's worth blocking the merge now.

Since we've got two +1s on the latest changes and a lot of previous positive votes I'm merging.

Thanks again for this @kerrydc, this has been a very fruitful and positive process. I think we can iterate very quickly towards a style that we all agree on. I'll work on a cleanup for the test cases soon, which will give us another opportunity to discuss these issues.

PR #59, opened 2015-01-08
Unified paging

jeromekelleher:
This is an early PR for comments and discussion. 

Here is a first pass at implementing paging in a general way. The idea is that the `Backend` object gets a call to `searchX` and then delegates this to `runRequest`, which does the paging in a general way. `runRequest` gets a bunch of parameters which allow it to do this, the key one being `objectGenerator`. This is an iterator over all of the objects specified by the request.

The tricky bit is dealing with page tokens. To deal with this, the generator yields pairs of values `(object, nextPageToken)`, which it's the generators responsibility to manage. It's easy in this case, but it'll be more trick when I move on to variants.

Before doing this, I'd like to get some feedback. What do people think of the general outline?

### DIFF
--- a/ga4gh/backends.py
+++ b/ga4gh/backends.py
@@ -381,7 +381,7 @@ class Backend(object):
     def searchVariants(self, request):
         """
         Returns a GASearchVariantsResponse for the specified
-        GASearchVariantsRequest  object.
+        GASearchVariantsRequest object.
         """
         return self.runSearchRequest(
             request, protocol.GASearchVariantsRequest,
@@ -430,3 +430,16 @@ class Backend(object):
                     nextPageToken = "{0}:{1}".format(
                         variantSetIndex, variant.start + 1)
                     yield variant, nextPageToken
+
+
+class MockBackend(Backend):
+    """
+    A mock Backend class for testing.
+    """
+    def __init__(self, dataDir=None):
+        # TODO make a superclass of backend that does this
+        # automatically without needing to know about the internal
+        # details of the backend.
+        self._dataDir = None
+        self._variantSetIdMap = {}
+        self._variantSetIds = []
--- a/ga4gh/backends.py
+++ b/ga4gh/backends.py
@@ -9,15 +9,22 @@ import os
 import glob
 import datetime
 
-import wormtable as wt
 import pysam
+import wormtable as wt
 
 import ga4gh.protocol as protocol
 
 
-class WormtableVariantSet(object):
+class VariantSet(object):
     """
-    Class representing a single dataset backed by a wormtable directory.
+    Class representing a single VariantSet in the GA4GH data model.
+    """
+    # TODO abstract details shared by wormtable and tabix based backends.
+
+
+class WormtableVariantSet(VariantSet):
+    """
+    Class representing a single variant set backed by a wormtable directory.
     We assume that VCF data has been converted to wormtable format using
     vcf2wt, which maps VCF data to wormtable columns in the following
     way:
@@ -142,7 +149,6 @@ class WormtableVariantSet(object):
             pos = col.get_position()
             if row[pos] is not None:
                 variant.info[infoField] = self.convertInfoField(row[pos])
-        variant.calls = []
         # All of the remaining values in the row correspond to Calls.
         for callSetId, rowPositions in sampleRowPositions.items():
             call = protocol.GACall()
@@ -164,31 +170,24 @@ class WormtableVariantSet(object):
             variant.calls.append(call)
         return variant
 
-    def searchVariants(self, request):
+    def getVariants(self, referenceName, startPosition, endPosition,
+                    variantName, callSetIds):
         """
-        Serves the specified GASearchVariantsRequest and returns a
-        GASearchVariantsResponse. If the number of variants to be returned is
-        greater than request.pageSize then the nextPageToken is set to a
-        non-null value. Subsequent request objects should provide this value in
-        the pageToken attribute to obtain the next page of results.
+        Returns an iterator over the specified variants. The parameters
+        correspond to the attributes of a GASearchVariantsRequest object.
         """
-        # This is temporary until we do this properly based on the output
-        # buffer size
-        if request.pageSize is None:
-            request.pageSize = 100
-        response = protocol.GASearchVariantsResponse()
-        # First we get the set of columns for wormtable to retrieve. We don't
-        # want to waste time reading in and decoding data for columns that
-        # are not needed for the output when we specify callSetIds
-        if request.callSetIds is None:
+        # TODO encoding issues here? Wormtable (and most likely pysam and other
+        # low level libraries) only deal with bytes. We'll need to adopt some
+        # convention here for where we move from unicode to bytes.
+        chrom = referenceName.encode()
+        if callSetIds is None:
             readCols = self._table.columns()
             # These must be in the correct order, so we cannot use the keys of
             # self._sampleCols
             callSetIds = self._sampleNames
         else:
             readCols = self._table.columns()[:self._firstSamplePosition]
-            callSetIds = request.callSetIds
-            for callSetId in request.callSetIds:
+            for callSetId in callSetIds:
                 cols = [col for name, col in self._sampleCols[callSetId]]
                 readCols.extend(cols)
         # Now we get the row positions for the sample columns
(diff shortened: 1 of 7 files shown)


ga4gh/backends.py
```diff
@@ -415,35 +413,50 @@ def searchVariants(self, request):
...
+        Runs the specified request by getting objects from the specified
+        generator, creating a response of the specified class and setting
+        a page of objects to the specified pageListName variable.
+        """
+        response = responseClass()
+        pageList = []
+        pageToken = None
+        for obj, pageToken in objectGenerator(request):
+            pageList.append(obj)
+            if len(pageList) == request.pageSize:
```
pashields: >= would save us from all sorts of slippery cases (pageSize is 0, None, or negative for instance).

jeromekelleher: Good call @pashields, thanks.

jeromekelleher:
This is now ready for discussion and (hopefully!) merging.

I've implemented unified paging, as well as doing a much-needed refactoring on some of the more important classes. Here's a summary:

- Paging is now done centrally on the backend object. This is now seen as the central point for the Flask frontend to send all requests. We get the objects from generators for the objects in question; these generators must return `(object, nextPageToken)` pairs and be able to resume iteration from any point.
- The Backend object has been refactored so that there is only one class (we no longer have the `WormtableBackend` and `TabixBackend` classes). We set a variable on the Flask app (`app.backend`) to allow the calls to get routed to the correct place.
- Various minor tweaks and PEP8 updates.
- The `Backend` API has now been changed so that it takes text versions of the requests and returns text versions of the responses. This simplifies the frontend substantially, and allows us a lot of flexibility in how we generate the string responses.

There is quite a lot in here, and it's probably too big for a single PR. However, I think the cumulative effect of these changes is a codebase that's substantially easier to understand and work with, so I hope it's not too much hassle reviewing it. Apologies also if I've clobbered any one else's work with this update.

ga4gh/backends.py
```diff
@@ -368,99 +311,104 @@ def getMetadata(self):
...
+        Runs the specified request. The request is a string containing
+        a JSON representation of an instance of the specified requestClass
+        in which the page list variable has the specified pageListName.
+        We return a string representation of an instance of the specified
+        responseClass in JSON format. Objects are filled into the page list
+        using the specified object generator, which must return
+        (object, nextPageToken) pairs, and be able to resume iteration from
+        any point using the nextPageToken attribute of the request object.
+        """
+        request = requestClass.fromJSON(requestStr)
```
kerrydc: I would call validate here, and therefore either add a validate JSON string method to protocol or else pass in the JSON dict here.

ga4gh/backends.py
```diff
@@ -368,99 +311,104 @@ def getMetadata(self):
...
-        response = protocol.GASearchVariantSetsResponse()
+        return self.runSearchRequest(
+            request, protocol.GASearchVariantSetsRequest,
+            protocol.GASearchVariantSetsResponse, "variantSets",
+            self.variantSetsGenerator)
+
+    def searchVariants(self, request):
+        """
+        Returns a GASearchVariantsResponse for the specified
+        GASearchVariantsRequest  object.
```
kerrydc: nit: remove one space.

ga4gh/backends.py
```diff
@@ -368,99 +311,104 @@ def getMetadata(self):
...
+            self.variantsGenerator)
+
+    def variantSetsGenerator(self, request):
+        """
+        Returns a generator over the (variantSet, nextPageToken) pairs defined
+        by the speficied request.
+        """
         currentIndex = 0
         if request.pageToken is not None:
             currentIndex = int(request.pageToken)
```
kerrydc: This is assuming that pageToken can be parsed as an int. That is not always the case.

ga4gh/backends.py
```diff
@@ -368,99 +311,104 @@ def getMetadata(self):
...
+        """
+        Returns a generator over the (variant, nextPageToken) pairs defined by
+        the specified request.
+        """
+        variantSetIds = request.variantSetIds
+        startVariantSetIndex = 0
+        startPosition = request.start
+        if request.pageToken is not None:
+            startVariantSetIndex, startPosition = map(
+                int, request.pageToken.split(':'))
```
kerrydc: If we assume pageToken has some structure (str containing ':', for example) we should validate it and throw an error if it does not conform.

ga4gh/server/__init__.py
```diff
@@ -6,6 +10,6 @@
 CORS(app, allow_headers='Content-Type')
 
 # TODO: Replace hard-coded value with a flag or conifg setting.
-app.config['MAX_CONTENT_LENGTH'] = 2 * 1024 * 1024 # 2MB 
+app.config['MAX_CONTENT_LENGTH'] = 2 * 1024 * 1024  # 2MB
```
kerrydc: I think you pulled this in by accident. Please remove.

tests/test_backends.py
```diff
@@ -117,8 +117,8 @@ def resultIterator(
         not_done = True
         request.pageSize = pageSize
         while not_done:
-            response = searchMethod(request)
-            self.assertEqual(type(response), ResponseClass)
+            responseStr = searchMethod(request.toJSON())
+            response = ResponseClass.fromJSON(responseStr)
             l = getattr(response, listMember)
```
kerrydc: Validate the response?

kerrydc:
Mostly looks good, but it needs a lot of unit test coverage. I would suggest that your test cases cover both legal and illegal input, and that you exercise raising all errors that can be raised.

kerrydc:
Mostly looks good, but it needs a lot of unit test coverage. I would suggest that your test cases cover both legal and illegal input, and that you exercise raising all errors that can be raised.

kerrydc:
Also, perhaps while writing tests you could update the style and variable names in test_backends.py. I'm working in there as well, adding Tabix backend tests, so please don't bother adding those.

ga4gh/server/__init__.py
pashields: This looks a pep8 fix @kerrydc. It should be covered by #63 though, so it will just get fixed in the a merge or rebase.

ga4gh/backends.py
jeromekelleher: Agreed.  The required method is being added in #64, so I've put in a TODO.

ga4gh/backends.py
jeromekelleher: From where? Sorry, I'm not seeing it.

ga4gh/backends.py
jeromekelleher: It is in this case though. PageTokens are handled locally here, so that each generator is locally responsible for managing its own. So, for variantSets, the pageToken is always an int, and for variants, the pageToken is always a ":" seperated pair of ints. We'll have to do something a bit more complicated with these when we start supporting Datasets, but I think the extension should be easy enough.

bioinformed: Not trying to rush anyone, but as with the previous round of cleanups, I'm
waiting for this refactoring to land before I implement the new htslib
VCF/BCF backend.

ga4gh/backends.py
jeromekelleher: This is a good idea. I've added a new method `parsePageToken` that verifies and parses the pageToken. It might work out that assuming all parts of the token must be integers is restrictive, but I think it's worth trying to do this for a start. It's a lot harder for an attacker to do nasty injection attacks when they are restricted to using integers.

ga4gh/server/__init__.py
jeromekelleher: This was convergent evolution all right! It'll get fixed when I rebase.

tests/test_backends.py
jeromekelleher: Definitely. Again, the required method is being added in #64, so I've put in a TODO.

jeromekelleher:
Thanks for the feedback @kerrydc, it's been very useful. I've implemented the majority of them, and updated the style on the variant tests. I haven't added any tests on failure conditions, as I don't see much point in this until we define a proper error model (i.e., we define the exceptions we expect and when we should see them; this should be something we tackle quite soon I think). I also think that the test cases need a decent refactoring, and it would be best to wait until you've finished working on the TabixBackend tests before this is undertaken. 

Also, in the interest of not holding @bioinformed up any longer, it would be good to try to wrap this up and vote on it.

ga4gh/backends.py
```diff
@@ -368,99 +311,135 @@ def getMetadata(self):
...
-        return response
+    def parsePageToken(self, pageToken, numValues):
+        """
+        Parses the specified pageToken and returns a list of the specified
+        number of values. Page tokens are assumed to consist of a fixed
+        number of integers seperated by colons. If the page token does
+        not conform to this specification, raise a InvalidPageToken
+        exception.
+        """
+        tokens = pageToken.split(":")
```
kerrydc: What happens when pageToken is an int or None?

ga4gh/backends.py
kerrydc: GASearchVariantsRequest  object.
_____________________^^_____

ga4gh/backends.py
jeromekelleher: Well, `nextPageToken` is a string in the protocol, so we can safely assume that conforms to this (once we start calling `validate()` on our input request). We've already tested that it's not `None` when we call this method.

We're changing the way things are done slightly here, as pageTokens will not be exactly the same form as before. I think this is a lot simpler though, and should generalise nicely to working with dataSets too. We do have reasonably good test coverage on the paging code, so I'm fairly sure it works properly in the nominal case. We definitely need to write some test code to break it though, and deal with these error cases properly.

ga4gh/backends.py
jeromekelleher: Aha, spotted it now! Thanks, was being a bit blind there.

Naburimannu:
I may be missing something, but it looks like this also solves #10? I can't spot the process ever getting larger than ~210 MB.

jeromekelleher:
It shouldn't have made any real difference to #10 - we're still building up the list of all objects and returning them. Perhaps we've change the default `pageSize` or something? I think we had some extra code in there to set it before?

kerrydc:
+1, thanks for the test fixes. We should add more test cases, but that can wait for a future PR.

kerrydc:
Jerome, anything outstanding on this PR? If not, let us vote! I'll merge on another two +1s.

Naburimannu:
Performance / memory footprint has definitely changed since I did my initial measurements for #10, but I verified this morning that it is also changed at tip-of-tree, so something else must be at fault.

+1

pashields:
+1

jeromekelleher:
I'm happy to merge this ASAP. Let's follow up with more test cases in a later PR.

PR #60, opened 2015-01-08
Fill in test cases for Views unit test

melaniedc:
This causes the following error if a field is not set in the request:

File "/usr/local/google/home/melanielc/git/server/ga4gh/server/views.py", line 21, in handleHTTPPost
    protocolRequest = protocolClass.fromJSON(data)
  File "/usr/local/google/home/melanielc/git/server/ga4gh/protocol.py", line 126, in fromJSON
    return cls._decode(d)
  File "/usr/local/google/home/melanielc/git/server/ga4gh/protocol.py", line 153, in _decode
    instance.__dict__[k] = v
UnboundLocalError: local variable 'v' referenced before assignment

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -1,21 +1,83 @@
 """
-Tests for the frontend.
+Unit tests for the frontend code.
 """
 
 import unittest
 
-import ga4gh.server as server
 import ga4gh.cli
+import ga4gh.protocol as protocol
+import ga4gh.server as server
+
+
+class MockBackend(ga4gh.backends.Backend):
+    """
+    A mock Backend class for testing.
+    """
+    def __init__(self, dataDir=None):
+        self._dataDir = dataDir
+
+    def searchVariants(self, request):
+        response = protocol.GASearchVariantsResponse()
+        return response
+
+    def searchVariantSets(self, request):
+        response = protocol.GASearchVariantSetsResponse()
+        return response
+
 
 class TestFrontend(unittest.TestCase):
 
     def setUp(self):
+        server.app.config['TESTING'] = True
+        server.app.config['VariantBackend'] = MockBackend()
         self.app = server.app.test_client()
 
     def testServer(self):
-        self.assertTrue('404' in self.app.get('/').data)
+        self.assertEqual(404, self.app.get('/').status_code)
 
-    # TODO: Fill in actual test cases here...
+    def testRouteReferences(self):
+        for path in ['/references/1', 'references/1/bases', 'referencesets/1']:
+            self.assertEqual(404, self.app.get(path).status_code)
 
-if __name__ == '__main__':
-    unittest.main()
+        for path in ['/referencesets/search', '/references/search']:
+            self.assertEqual(404, self.app.get(path).status_code)
+
+    def testRouteCallsets(self):
+        self.assertEqual(404, self.app.post('/callsets/search').status_code)
+
+    def testRouteReads(self):
+        for path in ['/reads/search', '/readgroupsets/search']:
+            self.assertEqual(404, self.app.post(path).status_code)
+
+    def testRouteVariants(self):
+        for path in ['/variantsets/search', '/variants/search']:
+            # POST gets routed to input checking
+            self.assertEqual(415, self.app.post(path).status_code)
+
+            self.assertEqual(400, self.app.post(
+                path,
+                headers={'Content-type': 'application/json'}).status_code)
+
+            # OPTIONS should return success
+            self.assertEqual(200, self.app.options(path).status_code)
+
+            # GET is not defined
+            self.assertEqual(405, self.app.get(path).status_code)
+
+    def testVariantsSearch(self):
+        response = self.app.post('/variants/search',
+                                 headers={'Content-type': 'application/json'},
+                                 data='{"variantSetIds": [1, 2]}')
+        self.assertEqual(200, response.status_code)
+        responseData = protocol.GASearchVariantsResponse.fromJSON(
+            response.data)
+        self.assertEqual(responseData.variants, [])
+
+    def testVariantSetsSearch(self):
+        response = self.app.post('/variantsets/search',
+                                 headers={'Content-type': 'application/json'},
+                                 data='{"dataSetIds": [1, 2]}')
+        self.assertEqual(200, response.status_code)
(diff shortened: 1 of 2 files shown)


kerrydc:
+1

melaniedc:
Hey, sorry for the weird commit ordering - I'm still getting used to github...

These tests bring coverage of the frontend code up to 100% (from 35% previously). They also uncovered a small bug in the JSON decoding, a fix for which I am including in this PR.

Thanks!
Melanie

pgrosu:
Hey Melanie,

No worries - that's totally fine :)  Many thanks for helping out!

Paul

tests/test_views.py
```diff
@@ -1,21 +1,87 @@
...
+
+class MockBackend(ga4gh.backends.Backend):
+    """
+    A mock Backend class for testing.
+    """
+    def __init__(self, dataDir):
+        self._dataDir = dataDir
+
+    def searchVariants(self, request):
+        response = protocol.GASearchVariantsResponse()
```
jeromekelleher: We should just be able to return `protocol.GASearchVariantsResponse()` as  the default value for `variants` is the empty list.

tests/test_views.py
```diff
@@ -1,21 +1,87 @@
...
+
+            # GET is not defined
+            self.assertTrue('405' in self.app.get(path).data)
+
+    def testVariantsSearch(self):
+        response = self.app.post('/variants/search',
+                                 headers={'Content-type': 'application/json'},
+                                 data='{"variantSetIds": [1, 2]}')
+        self.assertEqual(200, response.status_code)
+        self.assertTrue('"variants": []' in response.data)
```
jeromekelleher: Is this a little bit brittle?  Perhaps we should parse the JSON we're getting back (and verify that it is well formed JSON), and then run this test on the resulting dictionary?

tests/test_views.py
```diff
@@ -1,21 +1,87 @@
...
+        self.assertTrue('404' in self.app.post('/callsets/search').data)
+
+    def testRouteReads(self):
+        for path in ['/reads/search', '/readgroupsets/search']:
+            self.assertTrue('404' in self.app.post(path).data)
+
+    def testRouteVariants(self):
+        for path in ['/variantsets/search', '/variants/search']:
+            # POST gets routed to input checking
+            self.assertTrue('Unsupported media type'
```
jeromekelleher: I think it would be a bit cleaner to check the status code here rather than do string matching.

tests/test_views.py
```diff
@@ -1,21 +1,87 @@
...
+                            in self.app.post(path).data)
+
+            self.assertTrue('Malformed request' in self.app.post(
+                path, headers={'Content-type': 'application/json'}).data)
+
+            # OPTIONS should return success
+            self.assertEqual(200, self.app.options(path).status_code)
+
+            # GET is not defined
+            self.assertTrue('405' in self.app.get(path).data)
```
jeromekelleher: Why are we string matching against the body here and checking the status_code above? I think it would be better for all of these tests to be of the form `self.assertEqual(<HTTP_CODE>, self.app.options(path).status_code)`.

tests/test_views.py
```diff
@@ -1,21 +1,87 @@
 """
-Tests for the frontend.
+Unit tests for the frontend code.
 """
 
+import os
+import tempfile
```
jeromekelleher: I don't think we're using these imports here.

jeromekelleher:
This is great stuff, thanks @melaniedc! I've a few very minor comments about how we run the tests, but the overall approach is great, and I'm very happy that we're getting some proper frontend test cases in place.

melaniedc:
Thanks for the review @jeromekelleher !  I've fixed up the response code/data string mismatch, they should all look at response codes now. Let me know if I've missed anything :)

kerrydc:
+1 once again :)

tests/test_views.py
```diff
@@ -1,21 +1,87 @@
...
+    def searchVariantSets(self, request):
+        response = protocol.GASearchVariantSetsResponse()
+        return response
+
 
 class TestFrontend(unittest.TestCase):
 
     def setUp(self):
+        server.app.config['TESTING'] = True
+        server.app.config['VariantBackend'] = MockBackend('/tmp/')
```
jeromekelleher: I think it would be a bit better if we had `MockBackend()` here instead of providing /tmp as a dummy parameter. I think you can do this by just `def __init__(self, dataDir=None):` for the constructor above.

tests/test_views.py
```diff
@@ -1,21 +1,87 @@
 """
-Tests for the frontend.
+Unit tests for the frontend code.
 """
 
 import unittest
 
-import ga4gh.server as server
+import ga4gh.backends
 import ga4gh.cli
```
jeromekelleher: This isn't being used, please delete it. I've been using flake8 to catch these kind of problems, it's pretty good.

tests/test_views.py
```diff
@@ -1,21 +1,87 @@
...
+    def testVariantSetsSearch(self):
+        response = self.app.post('/variantsets/search',
+                                 headers={'Content-type': 'application/json'},
+                                 data='{"dataSetIds": [1, 2]}')
+        self.assertEqual(200, response.status_code)
+        responseData = protocol.GASearchVariantSetsResponse.fromJSON(
+            response.data)
+        self.assertEqual(responseData.variantSets, [])
 
 if __name__ == '__main__':
```
jeromekelleher: I don't have a strong opinion about this, but we're using nosetests as the preferred test running interface. It might be fine to use unittests.main also, or it might cause strange problems down the line, I don't know. I'm happy with keeping this if you want it, I thought I'd just bring up this point here.

jeromekelleher:
Looks great @melaniedc, thanks for the updates. I've a few other minor requests (which I should have seen on the first review, sorry), and we're good to merge. Could you squash these commits down also please? Ideally, we'd like to have two commits  here: one for the bug fix and another for the test case additions.

melaniedc:
Done, though it looks like I over-squashed somewhat... Thanks!

jeromekelleher:
Thanks @melaniedc, I'm +1 to merge now.

There's a conflict though, so I can't merge automatically. Can you `git merge upstream/master`, fix the conflict and push please?

melaniedc:
Done :)

Issue #61, opened 2015-01-09
Support configuration


Naburimannu:
In order to support [performance benchmarking] (https://github.com/ga4gh/server/issues/46) we should be able control whether or not the server is logging performance statistics, particularly those that are expensive - memory use or detailed profile data. This is a good impetus to start using [Flask configurations] (http://flask.pocoo.org/docs/0.10/config/).

This can be some combination of command-line, environment variables, Python classes, and files.

(Addresses a TODO in ga4gh/server/__init__.py)

dcolligan:
Can this be closed?  Seems like bb79afd80d48df9d3fb8827ffd8c7f532c62620a addressed it.

Naburimannu:
Yep, planned followup is captured in #69 and #70.

PR #62, opened 2015-01-09
Externalize configuration.

Naburimannu:
Read a set of configuration values out of a file (config.py).
Checks the environment variable GA4GH_CONFIGURATION for the name of the configuration object in that file to use; default is ga4gh.server.config:Config.

Addresses issue https://github.com/ga4gh/server/issues/61.

This initial version only provides the default configuration, but in prototyping profiling I've created subclasses of Config, set new values on them, used the environment variable to select between them, and verified that the correct config was selected.

### DIFF
--- a/ga4gh/server/__init__.py
+++ b/ga4gh/server/__init__.py
@@ -1,11 +1,14 @@
 from flask.ext.api import FlaskAPI
 from flask.ext.cors import CORS
+import os
 
 app = FlaskAPI(__name__)
-CORS(app, allow_headers='Content-Type')
 
-# TODO: Replace hard-coded value with a flag or conifg setting.
-app.config['MAX_CONTENT_LENGTH'] = 2 * 1024 * 1024  # 2MB
+app.config.from_object('ga4gh.server.config:DefaultConfig')
+if os.environ.get('GA4GH_CONFIGURATION') is not None:
+    app.config.from_envvar('GA4GH_CONFIGURATION')
+
+CORS(app, allow_headers='Content-Type')
 
 import views
 
--- a/ga4gh/server/config.py
+++ b/ga4gh/server/config.py
@@ -1,0 +1,16 @@
+"""
+Sets of configuration values for a GA4GH server.
+Any of these values may be overridden by a file on the path
+given by the environment variable GA4GH_CONFIGURATION.
+"""
+
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+class DefaultConfig(object):
+    """
+    Simplest default server configuration.
+    """
+    MAX_CONTENT_LENGTH = 2 * 1024 * 1024  # 2MB
--- a/README.txt
+++ b/README.txt
@@ -183,6 +183,9 @@ Deployment
 *TODO* Give simple instructions for deploying the server on common platforms
 like Apache and Nginx.
 
+Configuration parameters are specified in the file ga4gh/server/config.py;
+they can be overridden by setting the absolute path of a file containing
+new values in the environment variable GA4GH_CONFIGURATION. 
 
 +++++++++++++
 Running tests


hershman:
+1

Naburimannu:
Not ready to land - I'm doing something wrong in testing with virtualenv, so test runs are contaminating one another.

Naburimannu:
Should be cleaner and more correct now (modulo change in default value from ga4gh.server.Config to ga4gh.server.config:Config). For the record, necessary steps in testing from inside a virtualenv:
> rm -rf build
python setup.py install
ga4gh_server ...  

setup.py doesn't clean up before installing, so moved files will still be around.

Two future possibilities:
* app.config.from_envvar() exists but doesn't fail cleanly if the environment variable isn't set
* We might want to stick config.py in the root rather than in ga4gh/server/; this is easier to get working.

Naburimannu:
https://github.com/Naburimannu/server/commit/a464e257fdb5e7359ae5993db1ce903996759cfd uses this mechanism to turn basic timing logs on and off, as an example.

pgrosu:
Okay for now, but maybe in the future we can have a text file in a (root) config directory that is read by `config.py`.  That way we don't touch the code.

~p

jeromekelleher:
This looks good, thanks @Naburimannu. We definitely want a configuration interface that's easy to use, and that follows the common Flask idioms. I don't like the idea of the config file being buried inside in the code tree though, since our recommended deployment mechanism will be through PyPI. (It'll be quite hard to find ga4gh/server/config.py then, and editing the source code is not something we want to recommend.) What do you think of the pattern recommended by Flask (http://flask.pocoo.org/docs/0.10/config/#configuring-from-files)?
```python
app = Flask(__name__)
app.config.from_object('yourapplication.default_settings')
app.config.from_envvar('YOURAPPLICATION_SETTINGS')
```
The default settings could come from `ga4gh.server.config.py` and these defaults are overridden by a file specified in an environment var. I'm not crazy about environment vars, but if it's the common idiom then I'm happy to go with it.

How do you see this interacting with `ga4gh_server` for development purposes? I like to have access to options via a CLI for ease of development, so I'd like to add options for most things in here. This doesn't seem to be the flask idiom though, so I'm happy to go with the consensus if people think this is pointless.

Also, how does this interact with our unit tests? What pattern should we adopt for setting up the frontend when we're testing it?

I think this is a really important aspect of usability, so it's worth our while thinking it through properly. Thanks for getting the ball rolling.

Naburimannu:
pgrosu@, jeromekelleher@: Text file in or near root is what I'd expected.

As far as a text file, from the Flask docs this is apparently one of their idioms (http://flask.pocoo.org/docs/0.10/config/#development-production). The alternative jeromekelleher@ linked to is to have a file whose contents just happen to be parseable as a bunch of python fields, but that gives you a single configuration with no structured way to create variants, which I at least wanted to explore with this PR.

As far as being in or near root, it looks like config.from_object() wants the object to be part of a module, not a free-floating file. config.from_pyfile() should avoid that. I'm currently running into all sorts of inscrutable errors dealing with paths through virtualenv and .eggs when I try to switch to config.from_pyfile(), but I'll see if I can figure that out this week.

If what you mean is to have the default configuration at ga4gh/server/config.py and an override user-edited config.py specified by environment variable, yes, I think that should work fine.
 
I certainly want to support command-line arguments. I think we need to use flask.cli and/or flask.script, but the recent PEP8 correctness PR removed our flask.cli include. I look at that as a likely next task.

jeromekelleher:
>If what you mean is to have the default configuration at ga4gh/server/config.py and an override user-edited config.py specified by environment variable, yes, I think that should work fine.

This is what I was getting at all right; sorry, should have been clearer.

Naburimannu:
New version of the PR (which has a messy history, but [clean files](https://github.com/ga4gh/server/pull/62/files)) shows something closer to what we're looking for, then?

Loads a default configuration object. I can create a text file with a bunch of pythonic statements of the form
  KEY = value
and specify the **absolute** file path in the GA4GH_CONFIGURATION environment variable.

ga4gh/server/__init__.py
```diff
@@ -1,11 +1,14 @@
...
 from flask.ext.cors import CORS
+import os
 
 app = FlaskAPI(__name__)
-CORS(app, allow_headers='Content-Type')
 
-# TODO: Replace hard-coded value with a flag or conifg setting.
-app.config['MAX_CONTENT_LENGTH'] = 2 * 1024 * 1024  # 2MB
+app.config.from_object('ga4gh.server.config:DefaultConfig')
+if os.environ.get('GA4GH_CONFIGURATION') != None:
```
jeromekelleher: Nit: we should use `is not None` instead of `!= None`. See https://www.python.org/dev/peps/pep-0008/#programming-recommendations

ga4gh/server/config.py
```diff
@@ -0,0 +1,3 @@
+class DefaultConfig(object):
```
jeromekelleher: Please add a documentation comment at the top of the file and the standard ``__future__`` imports.

jeromekelleher:
Yeah, I think this is good. We can iterate on how the user defines the configuration values (and it would be really nice if we could do this from the server CLI), but I think this is a solid start for setting the defaults. We can then start filling out what our configuration values are and make progress on cleaning up the hard-coded config values that are lying around.

Naburimannu:
Done; also added a note to README.txt.

As for command-line configuration overrides, I think I see three cases, which we can implement as we need them:

1. choose a different *set* of default config values from ga4gh/server/config.py. The Flask doc with subclasses like ProductionConfig, DevelopmentConfig, TestingConfig seems to me to suggest that we can use subclassing from DefaultConfig to make it clear that we're only changing a subset of config values; we could use "--config" to select another name.

2. choose a different *file* of override config values, because we don't like environment variables. We could have something like "--config_file" to specify another path.

3. override individual configuration values; this has to be feature-specific.

If that makes sense I can file low-priority issues for (1) and (2).

(This work turns out to have been a bit of a distraction because as I further understand the Flask idiom I don't think I need full configuration support for performance benchmarking, but it's still something that was a TODO so probably worth doing.)

jeromekelleher:
Looks good to me; please go ahead and file issues for (1) and (2) as you say.

+1 to merge.

kerrydc:
+1

jeromekelleher:
Everyone seems happy with this so I'm merging. Thanks @Naburimannu.

PR #64, opened 2015-01-12
Improve JSON serialization and equality testing.

pashields:
These changes should increase readability and performance. JSON serialization shouldn't have any semantic changes, though equality testing will now exit early. The early exit should yield much better performance when dealing with large and complicated classes. This patch also removes the deprecated toJSON method because we are pre-release and there really isn't a reason to have deprecated code hanging around.

I'm hoping that there might be some other suggestions as to how to improve clarity here. I spent some time trying to create a unified interface to serialization and deserialization, but I was unable to get anything that I thought was actually more clear. The level of duplication here is fairly minimal and so far I've chosen the 15 lines of slight duplication devil vs 30 lines of generic generator code.

### DIFF
--- a/ga4gh/protocol.py
+++ b/ga4gh/protocol.py
@@ -7,6 +7,7 @@ from __future__ import unicode_literals
 
 import json
 import datetime
+import itertools
 
 import avro.io
 
@@ -46,7 +47,7 @@ class ProtocolElement(object):
     # rigorous testing.
 
     def __str__(self):
-        return "{0}({1})".format(self.__class__.__name__, self.toJSON())
+        return "{0}({1})".format(self.__class__.__name__, self.toJSONString())
 
     def __eq__(self, other):
         """
@@ -56,38 +57,14 @@ class ProtocolElement(object):
         TODO This feature is experimental and requires more testing; use
         with caution!!
         """
-        # TODO This implementation is ugly! There must be a nicer way to do
-        # this...
-        ret = False
-        if type(other) == type(self):
-            ret = True
-            for f in self.schema.fields:
-                k = f.name
-                a1 = getattr(self, k)
-                a2 = getattr(other, k)
-                if self.isEmbeddedType(k):
-                    if isinstance(f.type, avro.schema.ArraySchema):
-                        if len(a1) == len(a2):
-                            ret = ret and all(x == y for x, y in zip(a1, a2))
-                        else:
-                            ret = False
-                    else:
-                        ret = ret and a1 == a2
-                else:
-                    ret = ret and a1 == a2
-        return ret
+        if type(other) != type(self):
+            return False
+
+        fieldNames = itertools.imap(lambda f: f.name, self.schema.fields)
+        return all(getattr(self, k) == getattr(other, k) for k in fieldNames)
 
     def __ne__(self, other):
         return not self == other
-
-    def toJSON(self):
-        """
-        Returns a JSON encoded string representation of this ProtocolElement.
-
-        TODO This method is deprecated in favour of the more explicit
-        toJSONString.
-        """
-        return json.dumps(self, cls=ProtocolElementEncoder)
 
     def toJSONString(self):
         """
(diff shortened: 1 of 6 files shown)


ga4gh/protocol.py
```diff
@@ -4,6 +4,7 @@
 from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
+from itertools import imap
```
jeromekelleher: Arbitrary nitpick: I think of the import section as a bunch of groups: (1) the `__future__` section; (2) the standard library; (3) third party libraries; (4) within package imports. I think it's useful to group them this way so we can see what's coming from where. I also like to order the imports within each group in increasing order of line-length. I find it useful, but it might just be an embarrassing aesthetic habit on my part!

I would tend to prefer importing  a module and using fully referenced names in most cases too. I think this makes it clearer what's happening when reading the code. I might not immediately remember what imap does in this case when I'm looking at the code below, whereas `itertools.imap` tells me what sort of thing it's likely to do.

These are all very subjective style issues, and I wouldn't argue in favour of any of them strongly. It would be useful to come up with some convention on imports though, and to document this in our style guide.

ga4gh/protocol.py
```diff
@@ -56,39 +57,15 @@ def __eq__(self, other):
...
-                            ret = False
-                    else:
-                        ret = ret and a1 == a2
-                else:
-                    ret = ret and a1 == a2
-        return ret
+        if type(other) != type(self):
+            return False
+
+        fieldNames = imap(lambda f: f.name, self.schema.fields)
```
jeromekelleher: `f` should probably be `field` here if we're going to be consistent. I'm comfortable with `f`, but I don't know where we draw the line. @kerrydc, what's your position?

ga4gh/protocol.py
```diff
@@ -102,57 +79,69 @@ def toJSONDict(self):
...
         d = json.loads(jsonStr)
-        return cls._decode(d)
+        return cls.fromJSONDict(d)
 
     @classmethod
-    def _decode(cls, d):
-        # TODO This code is needlessly obscure and needs refactoring.
-        instance = cls()
-        schema = cls.schema
+    def fromJSONDict(cls, d):
```
jeromekelleher: Should `d` be `jsonDict` here?

ga4gh/protocol.py
pashields: I would argue that lambdas are a special case given they extremely limited syntactic scoping. I’ll leave this around for comment though.

jeromekelleher:
This is a big improvement, thanks @pashields. There's a few style issues we should nail down so that we all agree, but the code looks very clean.

ga4gh/protocol.py
pashields: I'm good with that. I wasn't sure how we came down on the "v for variant" sorts of things.

ga4gh/protocol.py
pashields: Sounds good. Thanks for the pointer.

pashields:
Thanks Jerome. I'll see what others have to say and then make all final style changes.

pashields:
Addressed Jerome's style concerns with the exception of the f in lambda. I'd like to argue in favor of short var names when the scope is very limited (1-2 lines) such as in a lambda.

jeromekelleher:
Thanks for that, it's looking good. I agree on the short variable names in this case: list comprehensions, lamba functions, etc, lose a lot of their clarity if they're split across lines, so I think this is a legitimate use of single letter variables.

We seem to have some conflicts - could you either merge or rebase off upstream master please?

+1 to merge.

kerrydc:
+1, agreed on lamdba variable names.

ga4gh/protocol.py
```diff
@@ -56,39 +57,15 @@ def __eq__(self, other):
         TODO This feature is experimental and requires more testing; use
         with caution!!
         """
```
Naburimannu: Do we have enough test coverage to make this comment obsolete yet?

Naburimannu:
+1 to merge.

ga4gh/protocol.py
pashields: I'm not sure that we do. I haven't really evaluated that TBH.

pashields:
Rebased.

jeromekelleher:
OK, thanks. Following the conversation with @bioinformed, it would be good if we could land #59 first and try to stabilise the API a bit. Would you mind if we put off merging this for a few days when things have settled down a bit?

pashields:
NP. Just ping me and I’ll bring it up to date.

kerrydc:
+1
Seeing as @bioinformed won't be able to merge his change for a bit, I would vote for getting this in sooner. Jerome, what do you think?

jeromekelleher:
I think I'd rather get #59 in first if @pashields doesn't mind updating this PR afterwards. I think it would be cleaner to push the API fixes in here through after #59, as it's already quite large. I'd like to merge this immediately after #59 goes through, ideally.

pashields:
Brought this up on top of #59.

PR #66, opened 2015-01-12
Add an end to end test for the server.

melaniedc:
This sets up some basic cases for interactions between the backend and the server. It will definitely need more filling in, but this gives us a basic framework to build off of.

### DIFF
--- a/tests/test_end_to_end.py
+++ b/tests/test_end_to_end.py
@@ -1,0 +1,104 @@
+"""
+End-to-end tests for each backend configuration. Sets up a server with
+the backend, sends some basic queries to that server and verifies results
+are as expected.
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import unittest
+
+from tests.test_backends import WormtableTestFixture
+import ga4gh.backend as backend
+import ga4gh.datamodel.variants as variants
+import ga4gh.frontend as frontend
+import ga4gh.protocol as protocol
+
+
+class EndToEndWormtableTest(unittest.TestCase):
+    def setUpServer(self, backend):
+        frontend.app.config['TESTING'] = True
+        frontend.app.backend = backend
+        self.app = frontend.app.test_client()
+
+    def setUp(self):
+        self._wtTestFixture = WormtableTestFixture()
+        self._wtTestFixture.setUp()
+        self.setUpServer(backend.Backend(self._wtTestFixture.dataDir,
+                                         variants.WormtableVariantSet))
+
+    def tearDown(self):
+        self._wtTestFixture.tearDown()
+
+    def sendJSONPostRequest(self, path, data):
+        return self.app.post(path,
+                             headers={'Content-type': 'application/json'},
+                             data=data)
+
+    def testVariantSetsSearch(self):
+        # TODO: Set up the test backend API to return these values rather than
+        # hard-coding them here.
+        expectedIds = ['example_1', 'example_2', 'example_3', 'example_4']
+        request = protocol.GASearchVariantSetsRequest()
+        request.pageSize = len(expectedIds)
+        response = self.sendJSONPostRequest('/variantsets/search',
+                                            request.toJSONString())
+
+        self.assertEqual(200, response.status_code)
+
+        responseData = protocol.GASearchVariantSetsResponse.fromJSONString(
+            response.data)
+        self.assertTrue(protocol.GASearchVariantSetsResponse.validate(
+            responseData.toJSONDict()))
+
+        self.assertIsNone(responseData.nextPageToken)
+        self.assertEqual(len(expectedIds), len(responseData.variantSets))
+        for variantSet in responseData.variantSets:
+            self.assertTrue(variantSet.id in expectedIds)
+
+    def testVariantsSearch(self):
+        # TODO: As above, get these from the test backend API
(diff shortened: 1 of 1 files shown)


tests/test_end_to_end.py
```diff
@@ -0,0 +1,95 @@
...
+        server.app.config["VariantBackend"] = backend
+        self.app = server.app.test_client()
+
+    def sendJSONPostRequest(self, path, data):
+        return self.app.post(path,
+                             headers={'Content-type': 'application/json'},
+                             data=data)
+
+    def runVariantSetsSearch(self):
+        expectedIds = ['example_1', 'example_2', 'example_3', 'example_4']
```
jeromekelleher: This is fine for now, but we should modify our test backend API to return the variant sets we want so we don't need to hard code these values. Could you insert a TODO comment to this effect here please?

tests/test_end_to_end.py
```diff
@@ -0,0 +1,95 @@
...
+        self.assertTrue(protocol.GASearchVariantSetsResponse.validate(
+            responseData.toJSONDict()))
+
+        self.assertIsNone(responseData.nextPageToken)
+        self.assertEqual(4, len(responseData.variantSets))
+        for variantSet in responseData.variantSets:
+            self.assertTrue(variantSet.id in expectedIds)
+
+    def runVariantsSearch(self):
+        expectedIds = ['example_1']
```
jeromekelleher: Please insert a line `referenceName = '1'` and replace the hard-coded values below. Also add a TODO comment stating we should get these values from the test backend when it's being refactored.

tests/test_end_to_end.py
```diff
@@ -0,0 +1,95 @@
...
+
+        self.assertEqual(200, response.status_code)
+
+        responseData = protocol.GASearchVariantSetsResponse.fromJSON(
+            response.data)
+        self.assertTrue(protocol.GASearchVariantSetsResponse.validate(
+            responseData.toJSONDict()))
+
+        self.assertIsNone(responseData.nextPageToken)
+        self.assertEqual(4, len(responseData.variantSets))
```
jeromekelleher: It would be better to use `len(expectedIds)` here rather than 4.

jeromekelleher:
This is excellent, thank @melaniedc! We will need to generalise this to work against unknown test data in the future, and I've added a few requests to make this explicit. There's nothing particularly special about the test data we're using at the moment, and I'm certain we'll have to add more as bugs arise that our current test set won't provoke. 

We're going to need a big refactor of the test system at some point, but I'd rather wait to see what other great stuff is going to land before doing this.

melaniedc:
Agreed. My main goal here is to set up the basic framework so someone with a better grasp of the biology/functionality can fill in more useful scenarios :)

melaniedc:
This now works with the new backend organization :)

kerrydc:
+1, looks good.

jeromekelleher:
Looking at this again, I wonder if perhaps we should reorganise it a little to make the test cases more modular. As it, we only get one test case from this module, `testWormtable`. As we add more and more tests, this one test gets more and more complicated, but we don't get any more test cases (from unittest's perspective). It would be nice to change `runVariantSetsSearch` to `testVariantSetsSearch` and do the same for variants. This way they are tested independently, and failing one test doesn't prevent the execution of the other. Ideally, we'd like to handle the fixtures in the `setUp` and `tearDown` methods of the test class, but since `WormtableTestFixture` is quite heavy we should probably handle it in the module `setUp` and `tearDown` as is done in `test_backends`. Another advantage of this is that `tearDown` is guaranteed to happen --- at the moment, if an error occurs, the WormtableTestFixture will not be cleaned up, leaving a mess behind in /tmp.

melaniedc:
Sounds good - especially with the way the backends stuff is evolving I think that makes more sense. We'll need to re-organize these some more once we work out #82 and the tabix backends tests but for now I'm just making it wormtable specific since that's what's implemented :)

tests/test_end_to_end.py
```diff
@@ -0,0 +1,100 @@
+"""
+End-to-end tests for each backend configuration. Sets up a server with
+the backend, sends some basic queries to that server and verifies results
+are as expected.
+"""
```
jeromekelleher: Can we get the __future__ imports here please? These are important for py2 -> py3 compatibility, and we definitely want to support py3 at some point!

jeromekelleher:
Looks good to me! I'm +1 to merge, modulo the future imports.

melaniedc:
Done and squashed :)

dcolligan:
+1

Naburimannu:
Are we mixing end-to-end tests with unit tests? I like keeping unit tests fast and separating out other things like integration and benchmarking.

jeromekelleher:
I see this as a quick unit test, just to make sure everything is plumbing together correctly. It's a bit slow now because of the test fixtures required, but hopefully we can fix this soon. What bit are you concerned about here @Naburimannu?

Naburimannu:
Sorry I left this idle. +1; I'll file an issue to discuss separating unit, integration, and end-to-end tests. @melaniedc pointed out that we want to be running at least a few end-to-end tests in Travis, and for now having all the tests in one bucket is what we've got.

jeromekelleher:
Merging, thank @melaniedc!

PR #72, opened 2015-01-13
Changed version number.

jeromekelleher:
The version number is currently set arbitrarily to 0.5.0. The idea originally was to follow the API version number that we were implementing. However, this is less important now that we have automated generations of schemas from a specific protocol version, and we can find out what version of the protocol we are implementing using `ga4gh.protocol.version`.

We should therefore choose a different version number. Since we are currently pre-alpha, 0.1.0a1 seems a good choice. Any thoughts?

### DIFF
--- a/ga4gh/__init__.py
+++ b/ga4gh/__init__.py
@@ -2,4 +2,4 @@
 Reference implementation of the GA4GH APIs.
 """
 
-__version__ = '0.5.0'  # Follows API version?
+__version__ = '0.1.0a1'


kerrydc:
We are very early in development and working quickly. I'm happy to be version 0.1, and not change until we at least implement all methods in the API.

Naburimannu:
I wouldn't worry about establishing expectations for versioning until we have more deployability.

jeromekelleher:
Agreed @Naburimannu - I just thought 0.5 was a bit arbitrary and it might be seen as implying that the project was more mature than it is. Also agreed that we shouldn't change any version numbers subsequently until we are shipping a PyPI package. (Here it's very useful to have a1, a2, b1, b2 etc for alpha and beta releases.)

Issue #67, opened 2015-01-13
Generate URL endpoints from schema definitions


jeromekelleher:
It would be very useful to have a list of valid URL endpoints and the types of their arguments and return values for testing purposes. This information should be derivable from the Avro schemas and could be added to the auto generated file `_protocoldefinitions.py`. 

The general pattern is Search[Class]Request maps to the URL /[class]/search which takes a Search[Class]Request and returns a Search[Class]Response. We can scan for these class names in `scripts/generate_schemas.py` using a simple regex, and write a list of (url, requestClass, responseClass) triples at the end of the `_protocoldefinitions.py`.

A different (and inconsistent) pattern holds for GET requests, which should be dealt with later.

dcolligan:
Starting work on this issue

dcolligan:
Fixed in #74

Issue #54, opened 2014-12-22
Fix Travis security issue


kerrydc:
We get this warning from Travis builds:
"/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/pkg_resources.py:1045: UserWarning: /home/travis/.python-eggs is writable by group/others and vulnerable to attack when used with get_resource_filename. Consider a more secure location (set with .set_extraction_path or the PYTHON_EGG_CACHE environment variable)."
We should fix this.

jeromekelleher:
Agreed; this makes life awkward when trying to deploy using WSGI also, so it would be good to get to the bottom of it.

jeromekelleher:
Fixed in #75

PR #74, opened 2015-01-15
Dcolligan 67

dcolligan:
A first attempt at addressing #67

Feedback appreciated.

### DIFF
--- a/ga4gh/_protocol_definitions.py
+++ b/ga4gh/_protocol_definitions.py
@@ -1586,3 +1586,26 @@ This metadata represents VCF header information.
         self.number = None
         self.type = None
         self.value = None
+
+postMethods = \
+    [('/callsets/search',
+      GASearchCallSetsRequest,
+      GASearchCallSetsResponse),
+     ('/readgroupsets/search',
+      GASearchReadGroupSetsRequest,
+      GASearchReadGroupSetsResponse),
+     ('/reads/search',
+      GASearchReadsRequest,
+      GASearchReadsResponse),
+     ('/referencesets/search',
+      GASearchReferenceSetsRequest,
+      GASearchReferenceSetsResponse),
+     ('/references/search',
+      GASearchReferencesRequest,
+      GASearchReferencesResponse),
+     ('/variantsets/search',
+      GASearchVariantSetsRequest,
+      GASearchVariantSetsResponse),
+     ('/variants/search',
+      GASearchVariantsRequest,
+      GASearchVariantsResponse)]
--- a/scripts/generate_schemas.py
+++ b/scripts/generate_schemas.py
@@ -17,6 +17,7 @@ import tempfile
 import requests
 import argparse
 import subprocess
+import re
 
 import avro.schema
 
@@ -83,24 +84,25 @@ class SchemaClass(object):
         stack = [schema]
         # Strip out all the docs
         while len(stack) > 0:
-            d = stack.pop()
-            if "doc" in d:
-                d["doc"] = ""
-            for u in d.values():
-                if isinstance(u, dict):
-                    stack.append(u)
-                elif isinstance(u, list):
-                    for v in u:
-                        if isinstance(v, dict):
-                            stack.append(v)
-        s = json.dumps(schema)
-        with tempfile.TemporaryFile() as f:
-            f.write(s)
-            f.seek(0)
+            elm = stack.pop()
+            if "doc" in elm:
+                elm["doc"] = ""
+            for value in elm.values():
+                if isinstance(value, dict):
+                    stack.append(value)
+                elif isinstance(value, list):
+                    for dic in value:
+                        if isinstance(dic, dict):
+                            stack.append(dic)
+        jsonData = json.dumps(schema)
+        with tempfile.TemporaryFile() as tmp:
+            tmp.write(jsonData)
+            tmp.seek(0)
             # Filter the text through fmt to make it look like acceptable code.
-            p = subprocess.Popen(["fmt"], stdout=subprocess.PIPE, stdin=f)
-        (output, err) = p.communicate()
-        exitStatus = p.wait()
+            subproc = subprocess.Popen(
+                ["fmt"], stdout=subprocess.PIPE, stdin=tmp)
+            (output, err) = subproc.communicate()
+            exitStatus = subproc.wait()
         if exitStatus != 0:
             msg = "Error occured running fmt: {0}: {1}".format(exitStatus, err)
             raise Exception(msg)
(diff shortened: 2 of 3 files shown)


scripts/generate_schemas.py
```diff
@@ -288,22 +334,25 @@ def run(self):
...
         d = os.path.join(self.schemaDir, "src/main/resources/avro")
-        # TODO should we make it an option to download this each time
-        # or to use a local copy? Should we continue using this version?
-        url = "http://www.carfab.com/apachesoftware/avro/stable/java/"\
-            "avro-tools-1.7.7.jar"
-        # it's useful to put a copy in cwd and uncomment this for
-        # developement; should we put this in as an official option?
-        # self.avroJar = os.path.join(os.getcwd(), "avro-tools.jar")
-        self.download(url, self.avroJar)
+        if self.avroJarPath:
```
jeromekelleher: I think we should prefer the more explicit `self.avroJarPath is not None` here. This reads to me like we expect the variable is a boolean, whereas the comparison with None is much more obvious. See also [pep8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations).

scripts/generate_schemas.py
```diff
@@ -277,9 +297,35 @@ def convertAvro(self, avdlFile):
...
-        # capturing stdout/stderr here
-        subprocess.check_call(args)
+        if self.verbosity > 1:
+            subprocess.check_call(args)
+        else:
+            devnull = open(os.devnull, 'w')
+            subprocess.check_call(args, stdout=devnull, stderr=devnull)
+            devnull.close()
+
+    def extract_post_methods(self, avdlFile):
```
jeromekelleher: We're standardising on camelCase throughout the codebase. Personally I prefer underscore delimited like this, but since the protocol is defined using camelCase, it seemed the lesser of two evils to at least have a consistent system.

scripts/generate_schemas.py
```diff
@@ -317,8 +366,9 @@ def main():
     parser.add_argument(
         "version",
         help="The tagged git release to process, e.g., v0.5.1")
-    # TODO is this the right approach? Maybe we should be noisy be
```
jeromekelleher: I'm not sure we should delete this - should we be noisy or quiet by default here? It's not obvious to me what's best. It was actually quite useful to have the output from the java process, as this pointed out a bug in the schema definitions. (See ga4gh/schemas#195 for this issue). It would be good to discuss the best approach here anyway I think.

ga4gh/_protocol_definitions.py
```diff
@@ -1586,3 +1586,29 @@ def __init__(self):
         self.number = None
         self.type = None
         self.value = None
+
+postMethods = \
+    [('/reads/search',
+      'GASearchReadsRequest',
```
jeromekelleher: I think it would be more useful to have the actual classes here, rather than the names. You should just be able to remove the quotes - because this is at the end of the file, all the classes will be defined already, and there won't be any problems.

scripts/generate_schemas.py
```diff
@@ -227,6 +228,24 @@ def write(self):
                 c = classes[name]
                 c.write(f)
 
+            # can't just use pprint library because
```
jeromekelleher: Hadn't heard of `pprint` before, looks really useful! What's the issue with pep8?

scripts/generate_schemas.py
```diff
@@ -277,9 +297,35 @@ def convertAvro(self, avdlFile):
...
+            devnull = open(os.devnull, 'w')
+            subprocess.check_call(args, stdout=devnull, stderr=devnull)
+            devnull.close()
+
+    def extract_post_methods(self, avdlFile):
+        methods = []
+        f = open(avdlFile)
+        lines = f.readlines()
+        url = request = response = None
+        for line in lines:
```
jeromekelleher: This is a neat method, but I worry that it's a little brittle. What would happen if someone is not quite systematic about the comment format? An alternative method (which avoids reparsing the file) is to run a regexp against all of the classes that we have generated at the end of the process. We match something like Search[x]Request, and pull out the middle token, and use this to construct the URL. Does this make sense?

jeromekelleher:
This is a great start, thanks @dcolligan! There's some minor issues with style, but it's definitely the sort of thing that we want here. Regarding style, there's been some extended discussions and we've decided to deprecate the use of short (i.e. single letter) variable names. Would it be possible to update the style in this file while you're in here?  (See #57 for a previous PR updating variable names and style). An official style-guide is in the pipelines, but the only major change to the current style is favour of longer variable names.

scripts/generate_schemas.py
```diff
@@ -277,9 +297,35 @@ def convertAvro(self, avdlFile):
             print("converting", avdlFile)
         if self.verbosity > 1:
             print("running:", " ".join(args))
-        # TODO silence the Java output unless we are in verbose mode by
-        # capturing stdout/stderr here
-        subprocess.check_call(args)
+        if self.verbosity > 1:
+            subprocess.check_call(args)
+        else:
+            devnull = open(os.devnull, 'w')
```
jeromekelleher: It would be nice to use a `with` block here to ensure `devnull` is closed even if an error occurs.

scripts/generate_schemas.py
dcolligan: Depending on the lengths of the names of the classes, pprint may choose to indent in a way that is not pep8-compliant.  But just more generally, pprint doesn't seem to want to respect pep8 -- pprint seems to be a library meant for human readability, not compliance after being piped to a generated code file.

dcolligan:
Yeah, I'll do some more cleanup of the file and incorporate your suggestions and resubmit.  Thanks.

scripts/generate_schemas.py
dcolligan: Using with instead of open/close should probably also be a directive in the future style guide; saves a line of code and is more robust to boot.

jeromekelleher:
Excellent, thanks @dcolligan. Just keep pushing commits to this branch, and we can squash the commits down then once it's ready to merge.

dcolligan:
Style question: should we keep file handles as single letter variables?  For instance:

```
with open('somefile.txt', 'r') as f:
    f.read()
    ...
```

dcolligan:
Also, lambdas:

```
return sorted(self.schema.fields, key=lambda f: f.name)
```

jeromekelleher:
I think lambdas are fine with single letter variable names, but file handles aren't. We should try to say what the file is, I think.

Also, good suggestion re ``with`` for the style guide.

diekhans:
One of the most valuable code style things I have learned (and
embarrassingly it took decades to learn it) is to write lots of
small function rather that a few large ones.  I strive to have
ten line functions.  Seriously.  It makes things like loops,
conditionals, with/open so says to understand.

any, my $0.02 on code org..

mark

kerrydc:
I strongly agree about short functions when possible.

jeromekelleher:
Yep, definitely +1 on the short functions.

dcolligan:
Okay, ready for another look.

Also, is there a way to manually kill a build in travis that you know is going to fail?

scripts/generate_schemas.py
```diff
@@ -83,24 +84,25 @@ def formatSchema(self):
...
+        with tempfile.TemporaryFile() as tmp:
+            tmp.write(jsonData)
+            tmp.seek(0)
             # Filter the text through fmt to make it look like acceptable code.
-            p = subprocess.Popen(["fmt"], stdout=subprocess.PIPE, stdin=f)
-        (output, err) = p.communicate()
-        exitStatus = p.wait()
+            subproc = subprocess.Popen(
+                ["fmt"], stdout=subprocess.PIPE, stdin=tmp)
+        (output, err) = subproc.communicate()
```
jeromekelleher: I think these lines should be indented under the `with` clause, as we're could be using stdin at any point. (My bad in the first place)

scripts/generate_schemas.py
```diff
@@ -200,32 +204,63 @@ def __init__(self, version, schemaDir, outputFile):
...
+        for avscFile in glob.glob(os.path.join(self.schemaDir, "*.avsc")):
+            self.classes.append(SchemaClass(avscFile))
+        self.requestClassNames = \
+            [clazz.name for clazz in self.classes
+                if re.search('Search.*Request', clazz.name)]
+        self.responseClassNames = \
+            [clazz.name for clazz in self.classes
+                if re.search('Search.*Response', clazz.name)]
+        self.postSignatures = []
+        for request, response in \
```
jeromekelleher: I think we should prefer 
```python
 for request, response in zip(
        self.requestClassNames, self.responseClassNames):
```
here, and use line breaks as a last-resort. (See the section in pep8 on maximum line length) We'll definitely put this into the style guide.

jeromekelleher:
Looks great to me @dcolligan, thanks! I'll have another proper look through it in after the weekend.

scripts/generate_schemas.py
```diff
@@ -200,32 +204,63 @@ def __init__(self, version, schemaDir, outputFile):
         self.schemaDir = schemaDir
         self.outputFile = outputFile
         self.classes = []
-        for f in glob.glob(os.path.join(self.schemaDir, "*.avsc")):
-            self.classes.append(SchemaClass(f))
+        for avscFile in glob.glob(os.path.join(self.schemaDir, "*.avsc")):
+            self.classes.append(SchemaClass(avscFile))
+        self.requestClassNames = \
```
jeromekelleher: We should prefer line breaking using brackets and hanging indents here over using the `\`. Here, we'd have
```python
self.requestClassNames = [
    cls.name for cls in self.classes if re.search(
        'Search.*Request', cls.name)]
```
We've used `cls` as the convention for referring to classes elsewhere in the code.

scripts/generate_schemas.py
```diff
@@ -200,32 +204,63 @@ def __init__(self, version, schemaDir, outputFile):
...
         Writes the generated schema classes to the output file.
         """
-        with open(self.outputFile, "w") as f:
-            self.writeHeader(f)
+        with open(self.outputFile, "w") as outputFile:
+            self.writeHeader(outputFile)
             # Get the classnames and sort them to get consistent ordering.
-            names = [c.name for c in self.classes]
-            classes = dict([(c.name, c) for c in self.classes])
+            names = [clazz.name for clazz in self.classes]
```
jeromekelleher: Ditto to preferring `cls` here - if you think it's important though, we can easily change the other places we're following this convention to `clazz` and put a note in the upcoming style guide.

jeromekelleher:
This is perfect @dcolligan - modulo some tiny style nits, I'm happy to merge.

dcolligan:
Okay, fixed your style concerns and squashed down to a single commit.

Btw, going forward I think it would be a good idea to put the issue number in the git commit message (as I've done here) for easy cross-referencing.

jeromekelleher:
Looks great, thanks! I agree on putting the issue # into the git commit message, this is a good convention (one more thing to add to our 'contributing' file?). 

+1.

pashields:
Unless you are manually merging PRs that is unnecessary as the github UI shows this under the commit message. (note #79)

screen shot 2015-01-21 at 10 31 28 am

scripts/generate_schemas.py
```diff
@@ -200,32 +204,63 @@ def __init__(self, version, schemaDir, outputFile):
         self.schemaDir = schemaDir
         self.outputFile = outputFile
         self.classes = []
-        for f in glob.glob(os.path.join(self.schemaDir, "*.avsc")):
-            self.classes.append(SchemaClass(f))
+        for avscFile in glob.glob(os.path.join(self.schemaDir, "*.avsc")):
+            self.classes.append(SchemaClass(avscFile))
+        self.requestClassNames = [
+            cls.name for cls in self.classes
+            if re.search('Search.*Request', cls.name)]
```
pashields: Is there a reason these are 'Search.*' vs 'Search.+'?
Not really important as long as we never have anything called SearchRequest, but still.

dcolligan:
@pashields I think that's the pull request number, not the issue number.

In your example, pull request 79 is here: https://github.com/ga4gh/server/pull/79  AFAICT pull request 79 did not address any tracked issue.  (That being said, I'm not GitHub expert so please correct me if I am wrong about this.)

As another example, my pull request 75 here: https://github.com/ga4gh/server/pull/75
fixed issue 54: https://github.com/ga4gh/server/issues/54

On the other hand, it does look like GitHub uses the same incrementing integer to number pull requests and issues (that is, there will never be an issue number 79).  So distinguishing between the "pull request number" and the "issue number" is not so meaningful... the equivalent issue URL does redirect to the pull request, however: https://github.com/ga4gh/server/issues/79

Also, one could have a pull request that addresses multiple issues.  I still think putting issue numbers in the commit log is a good convention.  This also has the advantage of conveying more information (though, admittedly, less without the aid of cross-referencing on GitHub) when one is browsing git history in a terminal.

dcolligan:
Oh, github's markup rewrote my URLs to the internally-linked issue/PR numbers, so my point is a little less clear... oh well.

dcolligan:
Pushed new commit addressing @pashields concerns about the regex

jeromekelleher:
I think this is ready to merge once it's been brought up to date (can you update and push please @dcolligan?). Any more votes?

pashields:
Sorry, I should've kept up with this PR. I still am unconvinced about the issue number in the commit message, but we can make a separate issue for discussion if we want to consider making standard.

Thanks @dcolligan!

+1

dcolligan:
This should be fine to merge... the only test that is failing is a test that is also failing in master: `tests.test_views.TestFrontend.testCors`

@melaniedc

dcolligan:
Here's the link to the master Travis build, btw: https://travis-ci.org/ga4gh/server

jeromekelleher:
OK, I've manually checked that this merges nicely with #93 and the tests pass, so I'm merging. Thanks @dcolligan, this is a great addition to our protocol parsing tool kit!

Issue #46, opened 2014-12-15
Server performance benchmarking needed


jeromekelleher:
As the reference server evolves, we need to have a mechanism to keep track of how the server's performance changes as we add new features. There is currently a crude performance benchmark built into the client program, but this is not satisfactory as it includes network latency and client side JSON parsing in its timings.

One way that we might get a better handle on query performance is to build a separate benchmarking program that takes a dataset (in the same format as the server program, whatever that turns out to be --- see discussion at #44), and runs some standard queries directly against the Backend implementation. That is, we cut out the HTTP layer entirely and run the queries against the GA4GH endpoints (e.g., `searchVariants(request)`, where request is a `SearchVariantsRequest` object). We should get the response from these functions in the form of JSON strings. 

This should allow us to get some good numbers for how our backend implementation is performing, without the noise of HTTP servers, network latency, JSON parsing etc, etc. We might also track memory usage, following the approach taken by @Naburimannu  in #10.

What are the important aspects that we need to track? What standard queries should we use? @benedictpaten, any thoughts here?

Naburimannu:
Some naive numbers using timer.clock() on my workstation, working one level higher than @jeromekelleher suggested by logging in server/views.py:handleHTTPPost()

ga4gh_client variants-search http://localhost:8000 1000g_2013 -r1 -e 60000 -c HG03279 | wc
 * trivial sample query from the documentation, produces 3 pages of output
 * 0.005 - 0.01 seconds per page (somewhat higher the first time the server gets this query)
 * adding a console dump from heapy (print(hp.heap()), qv https://github.com/Naburimannu/server/tree/heapy) adds 0.25 seconds
 * similar cost of calling print(hp.heap().domisize) for a summary of total heap size

ga4gh_client variants-search http://localhost:8000 1000g_2013 -e 100000 -r2 -c'*' | less -S
 * stress test query motivating https://github.com/ga4gh/server/issues/10
 * 3.9 - 4.6 seconds per page
 * print(hp.heap()) adds about 3.25 seconds
 * print(hp.heap().domisize) adds about 2.5 seconds

Until we address the problem in #10 we can't have always-on memory monitoring, and we probably don't want it always-on.

timer.clock() is deprecated in Python 3, but it looks like time.process_time() is a drop-in replacement.

[...]

PR #78, opened 2015-01-16
ga4gh_benchmark server benchmarking

Naburimannu:
Here's a more substantive strawman for #46.

Open issues:
* How many queries?
* Which queries?

### DIFF
--- a/server_benchmark.py
+++ b/server_benchmark.py
@@ -1,0 +1,99 @@
+"""
+Stand-alone benchmark for the GA4GH reference implementation.
+
+Assumes that wormtable sample data is installed at ./ga4gh-example-data.
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import re
+import time
+import argparse
+
+import ga4gh.backend
+import ga4gh.protocol as protocol
+import ga4gh.datamodel.variants as variants
+
+
+def _heavyQuery():
+    """
+    Very heavy query: all calls on chromosome 2
+    (11 pages, 90 seconds to fetch the entire thing
+    on a high-end desktop machine)
+    """
+    request = protocol.GASearchVariantsRequest()
+    request.referenceName = '2'
+    request.variantSetIds = ['1000g_2013']
+    request.callSetIds = None
+    request.pageSize = 100
+    request.end = 100000
+    return request
+
+
+def timeOneSearch(queryString):
+    """
+    Returns (search result as JSON string, time elapsed during search)
+    """
+    startTime = time.clock()
+    resultString = backend.searchVariants(queryString)
+    endTime = time.clock()
+    elapsedTime = endTime - startTime
+    return resultString, elapsedTime
+
+
+def extractNextPageToken(resultString):
+    """
+    Calling GASearchVariantsResponse.fromJSONString() can be slower
+    than doing the variant search in the first place; instead we use
+    a regexp to extract the next page token.
+    """
+    m = re.search('(?<=nextPageToken": )(?:")?([0-9]*?:[0-9]*)|null',
+                  resultString)
+    if m is not None:
+        return m.group(1)
+    return None
+
+
+def benchmarkOneQuery(request, repeatLimit=3, pageLimit=3):
+    """
+    Repeat the query several times; perhaps don't go through *all* the
+    pages.  Returns minimum time to run backend.searchVariants() to execute
+    the query (as far as pageLimit allows), *not* including JSON
+    processing to prepare queries or parse responses.
+    """
+    times = []
+    queryString = request.toJSONString()
+    for i in range(0, repeatLimit):
+        resultString, elapsedTime = timeOneSearch(queryString)
+        accruedTime = elapsedTime
+        pageCount = 1
+        token = extractNextPageToken(resultString)
+        # Iterate to go beyond the first page of results.
+        while token is not None and pageCount < pageLimit:
+            pageRequest = request
+            pageRequest.pageToken = token
+            pageRequestString = pageRequest.toJSONString()
+            resultString, elapsedTime = timeOneSearch(pageRequestString)
+            accruedTime += elapsedTime
+            pageCount = pageCount + 1
+            token = extractNextPageToken(resultString)
+        times.append(accruedTime)
+
+    # TODO: more sophisticated statistics. Sometimes we want min(),
+    # sometimes mean = sum() / len(), sometimes other measures,
+    # perhaps exclude outliers...
+
+    # If we compute average we should throw out at least the first one.
+    # return sum(times[2:])/len(times[2:])
+    return min(times)
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description="GA4GH reference server benchmark")
+
+    backend = ga4gh.backend.Backend("ga4gh-example-data",
+                                    variants.WormtableVariantSet)
+
+    initialRequest = _heavyQuery()
+    print(benchmarkOneQuery(initialRequest))


Naburimannu:
Just to be explicit: if you have the ga4gh-example-data (Wormtable) dataset installed, you can type 'ga4gh_benchmark' and get a number ~15 minutes later.

Running this verified for me that the memory bloat issue (#10) is still open.

kerrydc:
+1

ga4gh/cli.py
```diff
@@ -63,6 +64,51 @@ def server_main():
...
+    initialRequest.start = 0
+    initialRequest.end = 100000
+    initialRequestString = initialRequest.toJSONString()
+
+    # Repeat the query ~10x; if we compute average we should probably
+    # throw out at least the first one.
+    # TODO: more sophisticated statistics. Sometimes we want min(),
+    # sometimes mean = sum() / len(), sometimes other measures,
+    # perhaps exclude outliers...
+    for i in range(0, 10):
```
jeromekelleher: The `0` isn't necessary here.

ga4gh/cli.py
```diff
@@ -63,6 +64,51 @@ def server_main():
...
+        start_time = time.clock()
```
jeromekelleher: camelCase please.

ga4gh/cli.py
```diff
@@ -63,6 +64,51 @@ def server_main():
         app.backend = backend
         app.run(host="0.0.0.0", port=args.port, debug=True)
 
+
+def server_benchmark():
```
jeromekelleher: This should probably be `server_benchmark_main()`, just to keep the convention that the main entry point for any program will end with `main`. (We should camelCase this convention at some point though)

jeromekelleher:
I like it @Naburimannu, we definitely want this. I wonder though, do we want to ship this as a first-class CLI program installed on every user's machine, or should it be a development tool (say, `server_benchmark.py`)? If it is to be a first-class program, then the input will need to be generalised and we'll need to provide a UI to set the queries and so on. If it's a dev script, then I'm happy for it to depend on example data living in a fixed location. What do you think?

ga4gh/cli.py
Naburimannu: Agreed; when I was chasing down a bug I made everything explicit, but it doesn't need to be shipped.

Naburimannu:
Coverted to a standalone script with a header comment to tell you (briefly) what data we want where. Started parameterizing it for future plans.

Before the b2bd475 commit, it would take ~90 seconds to run on my desktop to time ~30 seconds inside Backend.searchVariants(), because the JSON parsing overhead is huge. Using a regexp instead feels fragile, but 3x speedup for a tool makes this a lot more feasible to consider running regularly, and this is explicitly the "dev script" branch of of @jeromekelleher's question so we might tolerate some fragility.

server_benchmark.py
```diff
@@ -0,0 +1,101 @@
...
+
+def timeOneSearch(queryString):
+    """
+    Returns (search result as JSON string, time elapsed during search)
+    """
+    startTime = time.clock()
+    resultString = backend.searchVariants(queryString)
+    endTime = time.clock()
+    elapsedTime = endTime - startTime
+    return (resultString, elapsedTime)
```
jeromekelleher: The brackets aren't actually needed here. Feel free to keep them if you think it improves readability --- just thought I'd point it out.

server_benchmark.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # TODO: more sophisticated statistics. Sometimes we want min(),
+    # sometimes mean = sum() / len(), sometimes other measures,
+    # perhaps exclude outliers...
+
+    # If we compute average we should throw out at least the first one.
+    # return sum(times[2:])/len(times[2:])
+    return min(times)
+
+
```
jeromekelleher: I'd rather see a `if __name__ == '__main__'` condition here.

jeromekelleher:
Looks great @Naburimannu. Other than a couple of tiny (optional) nits, I think it's perfect. I agree on the use of the regex - JSON parsing is a massive overhead, and just distracts from what we're really trying to measure.

Naburimannu:
Nits addressed and squashed; ready to review? (the CORS test was broken by a previous commit; filed https://github.com/ga4gh/server/issues/94 with my attempt at diagnosis)

jeromekelleher:
+1, thanks @Naburimannu. @dcolligan, would you mind having a quick look over this and +1'ing if you're happy with it?

I'm looking forward to seeing memory profiling and the other great tools you've developed getting integrated into this in later PRs!

server_benchmark.py
```diff
@@ -0,0 +1,100 @@
...
+    # perhaps exclude outliers...
+
+    # If we compute average we should throw out at least the first one.
+    # return sum(times[2:])/len(times[2:])
+    return min(times)
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description="GA4GH reference server benchmark")
+    args = parser.parse_args()
```
dcolligan: nit: eliminate unused args variable

server_benchmark.py
```diff
@@ -0,0 +1,100 @@
...
+        accruedTime = elapsedTime
+        pageCount = 1
+        token = extractNextPageToken(resultString)
+        # Iterate to go beyond the first page of results.
+        while token is not None and pageCount < pageLimit:
+            pageRequest = request
+            pageRequest.pageToken = token
+            pageRequestString = pageRequest.toJSONString()
+            resultString, elapsedTime = timeOneSearch(pageRequestString)
+            accruedTime = accruedTime + elapsedTime
```
dcolligan: nit: could do accruedTime += elapsedTime

ga4gh/cli.py
```diff
@@ -66,6 +66,7 @@ def server_main():
             args.dataDir, args.variantSetClass)
         frontend.app.run(host="0.0.0.0", port=args.port, debug=True)
 
+
```
dcolligan: why this change?  generally, let's try to keep the diff as small as possible, and especially not add unnecessary files to the diff (unless this is some pep8 thing?)

dcolligan:
The only thing that is keeping me from +1'ing this is the file path of `server_benchmark.py` ... do we want it in the root dir?  I think we should keep a few things in the root dir as possible.  What about under scripts/ or perhaps creating a new dir dev_scripts/ ?

Naburimannu:
I have no objection to moving it to scripts/; it's where it is to mimic server_dev.py and client_dev.py.
I'll clean up the nits, and agree that the blank line shouldn't be there; will remove unless it's a PEP8 issue.

jeromekelleher:
It's a lot easier to keep these scripts in the root, as you'll have to install the full `ga4gh` package if you want to run them otherwise (they import ga4gh stuff). The only simple way I found to do it (without installing) was to create a hard link from `scripts/something.py` to `something.py` in the root, and then  run `python something.py`. However, these hard links kept breaking when git changed files around, so I was never quite sure what version of the file I was working with. Unless someone's got a better workaround for the import problem, I'd much rather leave scripts that import `ga4gh` in the root.

dcolligan:
@jeromekelleher hm, I hadn't considered that.

In that case, would it make sense to make it a part of the ga4gh package somewhere?  Hm, but even if we did that, we would still need some launching glue in the root, a la server_dev or client_dev.

Ok, I take it back... leaving it in the root is fine.

dcolligan:
(Python does have a relative import syntax, but it only can be used within a package, not across them https://docs.python.org/2/tutorial/modules.html#intra-package-references )

Naburimannu:
Does anybody recognize this travis failure?

```
Searching for avro
Reading https://pypi.python.org/simple/avro/
No local packages or download links found for avro
error: Could not find suitable distribution for Requirement.parse('avro')
The command "python setup.py install" failed and exited with 1 during .
```

I can't just retry Travis jobs, I think because I don't have the Contributor bit.

jeromekelleher:
Nope, haven't seen that one before. PyPI seems to be a weak link in general here, I've seen it refusing connections from Travis a few times.

jeromekelleher:
I reran it there, and it works fine now. We need some bdist_wheels for our requirements we can download from somewhere reliable, this is going to get very tedious otherwise.

jeromekelleher:
OK, I think this is ready to merge. @dcolligan, do you agree?

dcolligan:
@Naburimannu , please squash your commits down to one commit and rebase on top of the current master before resubmitting the branch

After that, I'll merge it.  (or someone else can... here's my +1 after that: +1)

Naburimannu:
Rebased & squashed.
I notice that I was overly-literal when you suggested removing args, and didn't also remove the argument parser, but I'd like to go ahead and land this soonish; the next CL I'm planning in this area will either use or remove it.

PR #80, opened 2015-01-16
Add cmd line option to switch server config

dcolligan:


Naburimannu:
+1

hershman:
+1

ga4gh/cli.py
```diff
@@ -58,6 +61,7 @@ def server_main():
     if "variantSetClass" not in args:
         parser.print_help()
     else:
+        frontend.configure(args)
```
kerrydc: This only needs the config arg, not args.

ga4gh/frontend.py
```diff
@@ -13,11 +13,14 @@
 import flask.ext.cors as cors
 
 app = api.FlaskAPI(__name__)
-app.config.from_object('ga4gh.server.config:DefaultConfig')
-if os.environ.get('GA4GH_CONFIGURATION') is not None:
-    app.config.from_envvar('GA4GH_CONFIGURATION')
 
-cors.CORS(app, allow_headers='Content-Type')
+
+def configure(args):
```
kerrydc: ditto here.

jeromekelleher:
+1, with @kerrydc's requests.

ga4gh/frontend.py
dcolligan: If a config can not be found the app.config.from_object(configStr) will fail with "ImportError: No module named DummyConfig," so implicitly that's already taken care of.

dcolligan:
Pushed again with @kerrydc's suggestions incorporated

jeromekelleher:
Looks good to me. Let's wait to see what @kerrydc says before we merge.

pashields:
+1

Issue #55, opened 2014-12-29
Clean up code to improve readability


kerrydc:
Rewriting some of the bigger classes to make variable names more readable and improve commenting would make understanding the code and contributing to the project easier.

jeromekelleher:
Agreed - see the comments on PR #57 for my take. Should we try to make the style that we are looking for more explicit? I.e., is there a document somewhere that we could refer to, in terms of choosing class names, variable names, etc?

dcolligan:
IMHO this issue is too general to keep open.  Code style could always be improved, but that in and of itself doesn't need an issue.

Alternatively, we should come up with a list of specific changes we want to complete before this issue gets closed.

jeromekelleher:
I guess we should finish cleaning up all of the existing code in terms of the single letter variable names before closing this. A quick look suggests that the following files need to be cleaned up:

- ga4gh/cli.py
- ga4gh/client.py
- tests/test_schemas.py

I would suggest closing this issue once this is done. Thanks for keeping track of this @dcolligan, we should use issues more systematically!

dcolligan:
Fixed in d35744a9fdde68dc91596e2cf3d68866eea91708

PR #84, opened 2015-01-20
Rename single-character variables

dcolligan:
Issue #55

### DIFF
--- a/ga4gh/client.py
+++ b/ga4gh/client.py
@@ -30,23 +30,23 @@ class HTTPClient(object):
         """
         notDone = True
         while notDone:
-            s = request.toJSONString()
+            jsonData = request.toJSONString()
             headers = {"Content-type": "application/json"}
             # make sure we correctly join with/out trailing slashes
             fullUrl = posixpath.join(self._urlPrefix, url)
             # TODO Can we get requests to output debugging information when
             # debugLevel > 0?
-            r = requests.post(fullUrl, s, headers=headers, verify=False)
-            r.raise_for_status()
-            s = r.text
-            self._bytesRead += len(s)
+            post = requests.post(fullUrl, s, headers=headers, verify=False)
+            post.raise_for_status()
+            jsonData = post.text
+            self._bytesRead += len(jsonData)
             if self._debugLevel > 1:
                 # TODO use a logging output and integrate with HTTP client more
                 # nicely.
                 print("json response:")
-                pp = json.dumps(json.loads(s), sort_keys=True, indent=4)
+                pp = json.dumps(json.loads(jsonData), sort_keys=True, indent=4)
                 print(pp)
-            resp = protocolClass.fromJSONString(s)
+            resp = protocolClass.fromJSONString(jsonData)
             # TODO handle HTTP errors from requests and display.
             for v in getattr(resp, listAttr):
                 yield v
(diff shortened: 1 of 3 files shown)


ga4gh/client.py
```diff
@@ -30,23 +30,23 @@ def runRequest(self, request, url, protocolClass, listAttr):
...
             fullUrl = posixpath.join(self._urlPrefix, url)
             # TODO Can we get requests to output debugging information when
             # debugLevel > 0?
-            r = requests.post(fullUrl, s, headers=headers, verify=False)
-            r.raise_for_status()
-            s = r.text
-            self._bytesRead += len(s)
+            post = requests.post(fullUrl, s, headers=headers, verify=False)
+            post.raise_for_status()
+            json = post.text
```
jeromekelleher: Not sure I like having a variable the same name as an imported package!

tests/test_schemas.py
```diff
@@ -106,35 +107,35 @@ def getRandomValue(self, cls, fieldName):
...
             # TODO put in some probability that return None if
             # we have a union schema
-            t = t.schemas[1]
-        if isinstance(t, avro.schema.MapSchema):
+            typ = typ.schemas[1]
+        if isinstance(typ, avro.schema.MapSchema):
             ret = {}
             for j in range(random.randint(0, maxListSize)):
-                l = [randomString() for j in range(
+                randStr = [randomString() for j in range(
```
jeromekelleher: Maybe `j` should be `_` here? It's actually a bit dangerous as it is, as `j` is defined in the outer context. I think this wouldn't have worked in older versions of Python.

tests/test_schemas.py
```diff
@@ -83,19 +84,19 @@ def getInvalidValue(self, cls, fieldName):
         Returns a value that should trigger a schema validation failure.
         """
         field = self.getAvroSchema(cls, fieldName)
-        t = field.type
-        if isinstance(t, avro.schema.UnionSchema):
-            t = t.schemas[1]
-        if isinstance(t, avro.schema.MapSchema):
+        typ = field.type
```
jeromekelleher: Actually, reading pep8 I see that the recommended way of doing this is to use a trailing underscore:
```
single_trailing_underscore_ : used by convention to avoid conflicts with Python keyword, 
```
By this, we should have `class_` instead of `cls` and `type_` instead of `typ`. What's your preference? I don't have a strong opinion about it, I just thought I should bring it up.

ga4gh/client.py
dcolligan: oops, good catch

tests/test_schemas.py
dcolligan: Yes, it should be.

Something like pylint would catch these things, but I don't know if we want to add that to our build system... pep8 is picky enough!  It's something to consider, however.

tests/test_schemas.py
dcolligan: May as well start now, but this doesn't seem like a convention that needs to be rigorously enforced, IMHO.

jeromekelleher: Yeah, I'm not really bothered about it. It's up to you if you want to bring it in.

jeromekelleher:
Looks good, thanks.

+1.

pashields:
+1

jeromekelleher:
Ok, nobody seems to care about `cls` vs `class_`, so lets go with the current convention for now. Thanks for this @dcolligan.

Issue #48, opened 2014-12-19
Write audit logs for all requests and responses


kerrydc:
Access to data should be logged. Flask provides a logging framework, so that may be a good start. Once we have authentication we should log user IDs along with actions.

dcolligan:
Any preference of where we should put the log file?  It should probably be a command line option, but what should the default be?

jeromekelleher:
There's a question here for deployment also - do we log to the HTTP server logs, or do we make our own separate log? What's the common practice for this sort of thing in Flask?

dcolligan:
Flask looks like it provides a thin wrapper to python's standard logging package: http://flask.pocoo.org/docs/0.10/quickstart/#logging

The Flask docs page for apache config doesn't say anything about logs: http://flask.pocoo.org/docs/0.10/deploying/mod_wsgi/

Here's a Flask docs page on using some of the logging.handlers handlers: http://flask.pocoo.org/docs/0.10/errorhandling/#logging-to-a-file

From this I'm inferring that we make our own log, but if anyone else has an opinion, feel free to chime in.

In any event, the difference between logging to one place or another should be just a change in the initial logger setup configuration.

dcolligan:
Closing for ambiguity / already implemented... see discussion in #85

PR #85, opened 2015-01-20
Add server request / response logging

dcolligan:
Straw man proposal for issue #48 

This code produces the following in log/server.log after server startup and a request:
```
2015-01-20 11:25:15,102 [<module>] INFO: Server started
2015-01-20 11:25:15,597 [_log] INFO:  * Running on http://0.0.0.0:8000/
2015-01-20 11:25:15,607 [_log] INFO:  * Restarting with reloader
2015-01-20 11:25:15,728 [<module>] INFO: Server started
2015-01-20 11:25:17,495 [_log] INFO: 127.0.0.1 - - [20/Jan/2015 11:25:17] "POST /variants/search HTTP/1.1" 200 -
2015-01-20 11:25:17,592 [_log] INFO: 127.0.0.1 - - [20/Jan/2015 11:25:17] "POST /variants/search HTTP/1.1" 200 -
2015-01-20 11:25:17,654 [_log] INFO: 127.0.0.1 - - [20/Jan/2015 11:25:17] "POST /variants/search HTTP/1.1" 200 -
2015-01-20 11:25:17,671 [_log] INFO: 127.0.0.1 - - [20/Jan/2015 11:25:17] "POST /variants/search HTTP/1.1" 200 -
```

It also doesn't send any messages to stdout or stderr anymore.  The POST messages are from the werkzeug logger.  (Still trying to figure out how to eliminate the werkzeug logger's extraneous formatting.)

Feedback appreciated.


jeromekelleher:
Looks good, thanks @dcolligan. Here's some random thoughts on logging

- I think we should log to stdout by default when we are running the development server, and log to a file when running under apace/nginx/whatever. It's not clear to me what the default for this should be, if anything.  What happens if we log to stdout when running under a real webserver? I think there are real advantages to logging directly to the webserver's file if we can (log rotation, etc).
- I'm not a big fan of having a `log` directory in our source tree. When we deploy using pip, this won't be of any use and I think it's more useful to log to stdout for development purposes.
- Should logging all be done in `frontend.py`? It would be nice if this was possible, as we'd avoid bringing in extra dependencies on the frontend from backend.py.

pashields:
+1 on stdout/stderr by default for all the reasons @jeromekelleher described. Let others decide where they want it to go, rotation, etc.

dcolligan:
@jeromekelleher various logging file handlers support log rotation, e.g. https://docs.python.org/2/library/logging.handlers.html#rotatingfilehandler I suggest if we're logging to a file we rotate by default.  Does this change your opinion on a log directory?  If not, where should the log(s) go?  (I'm less convinced than you are about unifying Apache/Flask logs; I'm not sure if people necessarily want GET/POST/GET log msgs intertwined with error msgs.)  Now that I think about it, we should probably have a separate galogging module that any component can import.

Since there seems to be a lot of opinions on logging, perhaps it would be good to also make the logger configurable in the same way that we are making Flask configurable.  Otherwise we're going to be stuck with something like a stdout/err in dev, unified logs and file rotation to a set place default.  Maybe that's fine, but it's something we should consider...

jeromekelleher:
+1 on making logs configurable @dcolligan --- let's see what other people think about this.

pashields:
+ 1 for configurable, though I'd like to see a more detailed proposal.

ga4gh/frontend.py
```diff
@@ -22,6 +22,20 @@ def configure(config):
         app.config.from_envvar('GA4GH_CONFIGURATION')
     cors.CORS(app, allow_headers='Content-Type')
 
+import logging
```
melaniedc: Can you put this up with the other imports?

dcolligan:
TBH, I'm kind of uncertain about how to proceed on this PR.  The original issue stated by @kerrydc was "Access to data should be logged. Flask provides a logging framework, so that may be a good start. Once we have authentication we should log user IDs along with actions."  But the HTTP requests are already logged to stdout by werkzeug, at least in development.  And there's no auth feature yet.  So that seems to satisfy the requirements for development.  No one seems to have a good idea of what is required in production.  People also (rightly) seem to be against producing any log files.  Configurable logs seem to be a second step after having some custom logging infrastructure in place and knowing a bit about what kinds of configurations we want. So AFAICT, #48 is already satisfied by virtue of what the server already does.  Unless anyone has a specific suggestion of how to move forward on this, I would advocate closing this PR and the issue.

jeromekelleher:
I'm with you here @dcolligan --- let's revisit this when we have auth.

Issue #70, opened 2015-01-13
Specify configuration override file from the command line


Naburimannu:
Once #61 is addressed, we would like to be able to specify a configuration override file from the command line ("--config_file=..."?) in the same way we can with the GA4GH_CONFIGURATION environment variable.

dcolligan:
What format should this config file be in?  What data should we be expecting to be in the file?

Naburimannu:
@dcolligan: Although I haven't read the Flask source to be sure, the usual convention (http://flask.pocoo.org/docs/0.10/config/) and our assumption when we merged https://github.com/ga4gh/server/pull/62 is that the file is a sequence of
    KEY = Value
where keys are all-caps underscore-separated identifier, and values are python expressions. The only key which is defined so far is MAX_CONTENT_LENGTH. The config documentation page lists several other keys which Flask respects.

The application's main config.py is a set of classes, each of which looks like a config file, and #69 captures choosing among them.

This is good, straightforward work to do, but the reason I pushed back when you jumped into #69 is that I don't know that we have any examples that you can use for reasonable tests; my tests for https://github.com/ga4gh/server/pull/62 were all artificial. You could try setting DEBUG = True in a file and see how the server behavior changes?

dcolligan:
So, we're going to have three ways of specifying configuration variables:

1) environment variable
2) configuration file
3) cmd line switch to pick from pre-defined configurations

Are all of these necessary?  Specifically, does 1 make sense anymore?

And, in any event, if some of these conflict with each other what should be the resolution steps?

Naburimannu:
We implemented (1) and (2) because it was what Flask seemed to support. I think @jeromekelleher and I would both *think* (1) is unnecessary given (2) and (3), but I'd want to talk to people with actual deployment experience - @pashields, do have any sense of how much people use environment variables in production? Or have an idea who'd be good to ask?

The examples in the Flask doc lead me to expect that the configuration file is there mostly to override particular parts of the pre-defined sets. Since Flask doesn't seem to embrace command-line switches I don't see any guidance there about what you do when both environment variable and command line specify configuration files.

PR #86, opened 2015-01-20
Dcolligan 70 cfg file

dcolligan:
This works, but I'm still not sure what preference we want to override what; see discussion in #70

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -15,11 +15,13 @@ import flask.ext.cors as cors
 app = api.FlaskAPI(__name__)
 
 
-def configure(config):
+def configure(config="DefaultConfig", config_file=None):
     configStr = 'ga4gh.serverconfig:{0}'.format(config)
     app.config.from_object(configStr)
     if os.environ.get('GA4GH_CONFIGURATION') is not None:
         app.config.from_envvar('GA4GH_CONFIGURATION')
+    if config_file is not None:
+        app.config.from_pyfile(args.config_file)
     cors.CORS(app, allow_headers='Content-Type')
 
 
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -30,6 +30,9 @@ def server_main():
     parser.add_argument(
         "--config", "-C", default='DefaultConfig', type=str,
         help="The configuration to use")
+    parser.add_argument(
+        "--config-file", "-F", type=str,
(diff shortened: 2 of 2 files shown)


jeromekelleher:
Let's get #80 merged first before we discuss this.

I think it would be useful to give a little more information in the PR title - I have trouble remembering what issue is what! Also, I'm not sure putting the user name into the title adds much information.

Naburimannu:
Agreed with Jerome's comments.

As I read the flask doc, the implicit expectation seems to be that paths are absolute, which would allow you to not bury your config file inside ga4gh/. I'd prefer not committing example.cfg at all.

I suspect that using DEBUG=True in your example *should* give a visible change in server behavior which would be easy to manually verify.

pashields:
+1

dcolligan:
@jeromekelleher Yes, I agree that we should wait on #80 before moving forward with this.  This pull request was more of a POC for the time being.  Also, good suggestion on more informative PR request titles.  Github seems to parse them automatically from the branch name if not provided.  I put the user name in as a matter of habit; at a previous job we had trouble identifying which branch belonged to whom but with Github's interface that doesn't really seem to be an issue.  I'll adopt a more useful convention.

dcolligan:
Ok, so now that #80 has been merged, can we revisit this PR?

ga4gh/frontend.py
```diff
@@ -15,11 +15,13 @@
 app = api.FlaskAPI(__name__)
 
 
-def configure(config):
+def configure(config, config_file):
```
jeromekelleher: It'll be easier to give this some default values. What about `configure(config="DefaultConfig", configFile=None)`?

tests/test_views.py
```diff
@@ -15,7 +15,7 @@
 class TestFrontend(unittest.TestCase):
 
     def setUp(self):
-        frontend.configure("TestConfig")
+        frontend.configure("TestConfig", None)
```
jeromekelleher: This change isn't needed with the default value for `configFile`.

jeromekelleher:
Looks good to me; it's a useful tool setting up configurations for development. +1, modulo the comments above.

dcolligan:
@jeromekelleher pushed commit with your suggestions

jeromekelleher:
OK, two +1s so I'm merging. Thanks @dcolligan.

Issue #82, opened 2015-01-16
Refactor backends classes for dependency injection


kerrydc:
Our backends package now is difficult to test, due to the dependency the VariantSet classes have on the filesystem. We should change the VariantSets to take objects in their constructors instead of file paths, and create these objects using a factory.

jeromekelleher:
Sounds good @kerrydc.

dcolligan:
Working on this

jeromekelleher:
Closing this, as it has been effectively solved in recent developments.

PR #87, opened 2015-01-21
Refactor of backends to use dependency injection

dcolligan:
First pass at this; @kerrydc let me know if this is what you had in mind.

Issue #82


jeromekelleher:
I don't really understand the rationale behind these changes @dcolligan - could you explain your thinking a bit please?

Naburimannu:
Ditto @jeromekelleher 's request - it seems like it's gone a lot farther than the call in #82 by adding a complete and unnecessary abstraction layer instead of just minimal factories.

Tangentially, I'm also confused by something I see here and think I've seen in other recent PRs - are we using camelCase for all functions? Or was that just for functions that were part of the protocol?

jeromekelleher:
We should use camelCase for every function and variable we define I think. It just gets confusing otherwise, trying to decide when to use what style. There's a few cases of underscore delimited things left around from me making mistakes, but we should correct these as we see them.

dcolligan:
@jeromekelleher @Naburimannu  the rationale is that if we want to make testing the logic of backends easy, we need to throw a layer of indirection between backends and wormtable, etc.  That way in testing devs can provide classes that do whatever the tester wants to pass to the constructors instead of the default Backend* classes I created here, which do call into wormtable.

jeromekelleher:
Ah, OK, that's making more sense now. I see what you're doing here now, and this is a good approach. However, I'm not sure how much effort it's worth putting in to abstracting out Wormtable stuff. We're hoping to ditch it completely soon, so I'd rather focus our efforts at the `VariantSet` level rather than in here. If we make mock VariantSets (such as a SimulatedVariantSet similar to the one that we [started off with](https://github.com/ga4gh/server/blob/3421994c2ef629b7bdb97f8e1cb72291c311311c/ga4gh/server.py)), then we can make test code that works at this level.

So, for now, I think we'd be better off with a simple factory based approach like @Naburimannu suggested. I'm hoping that the details of file handlers can be localised in here when we create the code for handling our general `Dataset` directories. It would also be good to see a `VariantSet` superclass that actually does something, but I suspect @kerrydc will be working on this as he's bringing the TabixVariantSet up to compliance. 

We're still working on processes for sharing work and making sure that we're not cutting across each other @dcolligan, so if you have any ideas on how we can improve this please shout out!

dcolligan:
@jeromekelleher I don't think anyone is stepping on each others' toes here; my and @Naburimannu 's PRs do different things.  In any event, these are relatively small changes.  More generally, I think that just indicating in the issue that one is working on it (and creating an issue for something that one is working on which does not have an issue yet) is fine.  I don't think anything has gone obviously wrong with process.  Perhaps we could give people more insight into what is going on, but I want to avoid a full-blown bug tracking system and also I don't think the overhead of doing that would be very popular with anyone.

pashields:
Are these new classes supposed to be generic to all potential backends? It looks like openTable() still calls wormtable directly. Perhaps some doc describing these classes would be helpful. If these are wormtable specific they should be named accordingly.

dcolligan:
Closing this since it doesn't seem like we're going to be using it, and it's not even the change that was requested in the issue

PR #88, opened 2015-01-21
Server now catches and handles exceptions

dcolligan:
Issue #6

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -8,9 +8,14 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import os
+import traceback
+
 import flask
 import flask.ext.api as api
 import flask.ext.cors as cors
+
+import ga4gh.frontend_exceptions as frontendExceptions
+import ga4gh.protocol as protocol
 
 app = api.FlaskAPI(__name__)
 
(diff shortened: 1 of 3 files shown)


ga4gh/exceptions.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import ga4gh.protocol as protocol
+
+
+class ParentException(Exception):
```
jeromekelleher: I'd probably rather call this `ServerError`, but let's see what the outcome of the Exception vs Error discussion is.

ga4gh/exceptions.py
```diff
@@ -0,0 +1,85 @@
+"""
+Module containing error objects we return to clients
+"""
+
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import ga4gh.protocol as protocol
```
jeromekelleher: Are we using  the protocol module import here?

ga4gh/exceptions.py
```diff
@@ -0,0 +1,85 @@
...
+            "code": self.code,
+            "description": self.description,
+        }
+
+
+class BadRequestException(ParentException):
+
+    def __init__(self):
+        super(BadRequestException, self).__init__()
+        self.status = 400
```
jeromekelleher: I think httpStatus would be a bit clearer here - these numbers might be a bit mysterious otherwise.

ga4gh/exceptions.py
```diff
@@ -0,0 +1,85 @@
...
+        self.message = "Bad request"
+        self.code = "BAD_REQUEST"
```
jeromekelleher: For the protocol object, we need an integer error code. How do we allocate these codes? For now, I'd suggest just putting in 1, 2,... and we'll come up with a better approach later.

ga4gh/frontend.py
```diff
@@ -41,49 +43,58 @@ def handleHTTPOptions():
     return response
 
 
+@app.errorhandler(Exception)
+def handle_error(error):
+    if not isinstance(error, exceptions.ParentException):
+        error = exceptions.ServerException()
+    response = flask.jsonify(error.toDict())
```
jeromekelleher: We should use the `protocol.GAException` class here, and remove the `toDict()` method. Something like
```python
protocolException = protocol.GAException()
# Or should our exception class use code and message for consistencty?
protocolException.errorCode = error.code
protocolException.message = error.description
response = flask.jsonify(error.toJSONDict())
```
We also need to make sure we're not losing the stack trace when unexpected errors occur during development. Maybe a development/debug flag would be useful for this?

jeromekelleher:
Looks good @dcolligan, thanks. There's a few issues we need to make decisions about, but this is a big step in the right direction.

ga4gh/frontend.py
```diff
@@ -41,49 +45,64 @@ def handleHTTPOptions():
...
+def handle_error(error):
+    if not isinstance(error, exceptions.ParentException):
+        if app.config['DEBUG']:
+            print(traceback.format_exc(error))
+        error = exceptions.ServerException()
+
+    protocolException = protocol.GAException()
+    protocolException.errorCode = error.code
+    protocolException.message = error.message
+    response = flask.jsonify(protocolException.toJSONDict())
```
pashields: We should set the content-type here. In the long run, we may want to offer other types.

ga4gh/exceptions.py
```diff
@@ -0,0 +1,83 @@
...
+class ParentException(Exception):
+
+    def __init__(self):
+        super(ParentException, self).__init__()
+        self.message = ""
+        self.code = ""
+        self.description = ""
+        self.httpStatus = None
+
+    def toDict(self):
```
jeromekelleher: I don't think this is being used for anything, so we should remove it.

ga4gh/exceptions.py
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+
+class BadRequestException(ParentException):
+
+    def __init__(self):
+        super(BadRequestException, self).__init__()
+        self.httpStatus = 400
+        self.message = "Bad request"
+        self.code = "GAE1"
```
jeromekelleher: Code has to be an integer for the protocol - if we return a string, then we won't be conformant. I agree we'll want to have a systematic way of handling these in the future and just putting in integers isn't a good idea, but it'll get us started and we can figure out how to do it properly later.

ga4gh/frontend.py
dcolligan: the `jsonify` call does this automatically http://flask.pocoo.org/docs/0.10/api/#flask.json.jsonify

ga4gh/exceptions.py
dcolligan: Yeah, the schema should probably change this type to a string for more flexibility... unless they also want to give us defined codes for defined errors (but I think they probably view this as outside the schema's purview)

ga4gh/frontend.py
pashields: Cool. Apologies for not reading the docs :frowning:

dcolligan: No problem!  I didn't know this until I looked it up, either ;)

jeromekelleher:
I think this is a good place to start, so I'm +1 on this, depending on the outcome of the discussion at #6.

dcolligan:
Just pushed another change which implements my suggestion (2) in #6

melaniedc:
+1

kerrydc:
+1

PR #89, opened 2015-01-22
Simple dependency injection for backend tables

Naburimannu:
Allows us to call different constructors during testing instead of relying on actual on-filesystem databases. This strawman proposal puts all backends' factories on the same class; they could be backend-specific for easier code reading

Addresses https://github.com/ga4gh/server/issues/82.


Naburimannu:
@dcolligan @jeromekelleher I think this is what @kerrydc was getting at, but he's traveling in California this week.

pashields:
@Naburimannu you have some pep8 issues

This does solve the issue kerry and I were discussing that led to the slightly underspecified #82. Kerry was trying avoid creating actual on disk files for tabix testing. I believe this will take care of that. I don't think this mutually exclusive with what @dcolligan is arguing for. Perhaps more importantly, it's a fine step in the right direction.

I'll +1 this even without splitting it out because it's useful now.

Naburimannu:
Fixed PEP8.

Naburimannu:
Thanks for the +1, @pashields , but this really was a strawman - it's not actually integrated, and I don't  think it can be made to work with cli and backend as currently written. We need to move the entire directory traverse out of Backend.__init__() to allow non-filesystem ways of setting up backends.

@dcolligan, I think @pashields is right that you and I are solving different problems with our changes.

dcolligan:
@Naburimannu Yes, I think we're all agreed that our changes are orthogonal.

kerrydc:
+1 to this approach, it is indeed what I was getting at.

jeromekelleher:
OK, let's bring this up to date and merge then. Can you rebase please @Naburimannu?

jeromekelleher:
Do you want to bring this up to date and re-review @Naburimannu, or should we close this PR?

Naburimannu:
I should be able to readdress this Friday.

Naburimannu:
I think this is the right way forward, but I've not been able to work on it in too long. Closing the PR; somebody else can feel free to do the work, or I'll circle back on it when I have the opportunity.

Issue #76, opened 2015-01-16
Extend client library to include full protocol


jeromekelleher:
The client library needs to support the full GA4GH protocol. We should develop and test this against existing implementations.

dcolligan:
Working on this.

jeromekelleher:
Not sure what happened here, but reopening this as we haven't implemented the complete protocol yet,

dcolligan:
I think GitHub automatically closes an issue if it is referenced in a commit that is merged to master... kind of annoying (since it implies that issues can't be long running and addressed in a series of commits)

dcolligan:
The next PR will close this issue... there's still three client endpoints in the 0.5.1 spec to implement

PR #90, opened 2015-01-22
Added referencesets/search client endpoint

dcolligan:
Issue #76 -- obviously very WIP.  Going to be adding client endpoints to this branch as I go along...

Tried to follow the conventions established in the code already.  @jeromekelleher

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -71,27 +71,42 @@ def server_main():
 ##############################################################################
 
 
-class VariantSetSearchRunner(object):
+class AbstractSearchRunner(object):
+    """
+    Abstract base class for search runner classes
+    """
+    def __init__(self, args):
+        self._workarounds = set(args.workarounds.split(','))
+        self._key = args.key
+
+    def usingWorkaroundsFor(self, workaround):
+        return workaround in self._workarounds
+
+
+class VariantSetSearchRunner(AbstractSearchRunner):
     """
     Runner class for the variantsets/search method.
     """
     def __init__(self, args):
+        super(VariantSetSearchRunner, self).__init__(args)
         svsr = protocol.GASearchVariantSetsRequest()
         svsr.pageSize = args.pageSize
         self._request = svsr
         self._verbosity = args.verbose
-        self._httpClient = client.HTTPClient(args.baseUrl, args.verbose)
+        self._httpClient = client.HTTPClient(
+            args.baseUrl, args.verbose, self._workarounds, self._key)
 
     def run(self):
         for variantSet in self._httpClient.searchVariantSets(self._request):
             print(variantSet.datasetId, variantSet.id)
 
 
-class VariantSearchRunner(object):
+class VariantSearchRunner(AbstractSearchRunner):
     """
     Runner class for the variants/search method.
     """
     def __init__(self, args):
+        super(VariantSearchRunner, self).__init__(args)
         svr = protocol.GASearchVariantsRequest()
         svr.referenceName = args.referenceName
         svr.variantName = args.variantName
@@ -107,7 +122,8 @@ class VariantSearchRunner(object):
         svr.variantSetIds = args.variantSetIds.split(",")
         self._request = svr
         self._verbosity = args.verbose
-        self._httpClient = client.HTTPClient(args.baseUrl, args.verbose)
+        self._httpClient = client.HTTPClient(
+            args.baseUrl, args.verbose, self._workarounds, self._key)
 
     def run(self):
         for variant in self._httpClient.searchVariants(self._request):
@@ -130,6 +146,32 @@ class VariantSearchRunner(object):
                 c.callSetId, c.genotype, c.genotypeLikelihood, c.info,
                 c.phaseset, sep=":", end="\t")
         print()
+
+
+class ReferenceSetSearchRunner(AbstractSearchRunner):
+    """
+    Runner class for the referencesets/search method.
+    """
+    def __init__(self, args):
+        super(ReferenceSetSearchRunner, self).__init__(args)
+        request = protocol.GASearchReferenceSetsRequest()
+        if args.accessions is not None:
+            request.accessions = args.accessions.split(",")
+        if self.usingWorkaroundsFor(client.HTTPClient.workaroundGoogle):
+            # google says assemblyId not a valid field
+            del request.__dict__['assemblyId']
+        if args.md5checksums is not None:
+            request.md5checksums = args.md5checksums.split(",")
+        request.pageSize = args.pageSize
+        self._request = request
+        self._verbosity = args.verbose
+        self._httpClient = client.HTTPClient(
+            args.baseUrl, args.verbose, self._workarounds, self._key)
+
+    def run(self):
+        for referenceSetId in self._httpClient.searchReferenceSets(
+                self._request):
+            print(referenceSetId.id)
 
 
 class BenchmarkRunner(VariantSearchRunner):
@@ -200,6 +242,10 @@ def client_main():
     parser = argparse.ArgumentParser(description="GA4GH reference client")
     # Add global options
     parser.add_argument('--verbose', '-v', action='count', default=0)
+    parser.add_argument(
+        "--workarounds", "-w", default='', help="The workarounds to use")
+    parser.add_argument(
+        "--key", "-k", help="The auth key to use")
     subparsers = parser.add_subparsers(title='subcommands',)
 
     # help
@@ -219,8 +265,8 @@ def client_main():
         "benchmark",
         description="Run simple benchmarks on the various methods",
         help="Benchmark server performance")
-    addUrlArgument(bmParser)
     bmParser.set_defaults(runner=BenchmarkRunner)
+    addUrlArgument(bmParser)
     addOptions(bmParser)
     # variantsets/search
     vssParser = subparsers.add_parser(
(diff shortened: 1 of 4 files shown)


auth.cfg
```diff
@@ -0,0 +1,2 @@
+[google]
+key = YOUR_KEY_HERE
```
jeromekelleher: I'm not crazy about adding in conf files into our source tree. What do we do when we install using `pip`? Do we try to distribute these files into standard locations? 

I don't think we really need a conf file here in any case: a CLI argument would seem the simpler choice.

ga4gh/cli.py
```diff
@@ -127,6 +127,31 @@ def printVariant(self, v):
         print()
 
 
+class ReferenceSetSearchRunner(object):
+    """
+    Runner class for the referencesets/search method.
+    """
+    def __init__(self, args):
+        rsr = protocol.GASearchReferenceSetsRequest()
+        if rsr.accessions:
```
jeromekelleher: I think we should prefer `if x is not None` over `if x`, as it's more explicit.

ga4gh/cli.py
```diff
@@ -127,6 +127,31 @@ def printVariant(self, v):
...
+            rsr.accessions = args.accessions.split(",")
+        # TODO google-specific assemblyId not a valid field,
+        # referenceSetId (contrary to goog docs) not implemented
+        del rsr.__dict__['assemblyId']
```
jeromekelleher: It looks like we'll need to put in some workarounds for Google in general here. However, we don't want to make our client Google specific. Perhaps we could add in a command line argument to turn on Google workarounds?

ga4gh/cli.py
```diff
@@ -127,6 +127,31 @@ def printVariant(self, v):
...
+        # rsr.referenceSetId = args.assembly_id
+        if rsr.md5checksums:
```
jeromekelleher: Why are we doing the split here?

ga4gh/cli.py
```diff
@@ -127,6 +127,31 @@ def printVariant(self, v):
         print()
 
 
+class ReferenceSetSearchRunner(object):
+    """
+    Runner class for the referencesets/search method.
+    """
+    def __init__(self, args):
+        rsr = protocol.GASearchReferenceSetsRequest()
```
jeromekelleher: `request` is probably a better name here.

jeromekelleher:
Looks good @dcolligan, thanks. I've made a few comments in line, but they're pretty minor.

We need some test cases for this client code now, and we might as well start putting them in now. Have you any ideas for how we might unit test this code (i.e., the `client.searchX` functions)?

Let's take the compliance tester bit out, and keep that for a separate PR.

auth.cfg
dcolligan: I agree that pip installs make including cfg files a bad idea.

On the other hand, we have run into this problem of configuration before, some of which can get quite complicated (logging conf, server conf, auth conf -- which could be arbitrarily complex for certain sites) and I don't think "add another command line flag" is the solution to all of them.  To me, it makes more sense to have _some_ file(s) somewhere that, yes, a user can point to with a command line argument.

The "pre-populate configs in some .py file and have a cmd line flag to select from amongst them" solution doesn't work in cases where, like auth, we do not know the values in advance.

I'll add a command line flag for this for the time being, but we really need a better solution, IMHO.

dcolligan:
Pushed a change that addresses some of your concerns.

As far as testing, the search* functions don't do much; most of the logic is in `runRequest`.  For testing it, we should probably just mock the request library calls.  Ah, and I see I've made the test coverage fail.

Although, as far as APIs go, I find end-to-end tests generally the most useful since the major problem is testing that all of the endpoints/protocol implementations line up -- there's not a ton of logic to be checked.

auth.cfg
kerrydc: I think we should go with flags, with sensible defaults, and the ability to
set flags via a flag specifying a config file. Note that I may be biased by
regularly setting 50-100 command line flags in a build. Longer term we will
want a guided setup that will prompt the user for various values on
installation.
Kerry

dcolligan:
I pushed some unit tests, @jeromekelleher, so let me know what you think.

ga4gh/client.py
```diff
@@ -11,15 +11,19 @@
 
 import ga4gh.protocol as protocol
 
+requests.packages.urllib3.disable_warnings()
```
jeromekelleher: Can we get a comment to explain why we're disabling warnings here please?

tests/test_client.py
```diff
@@ -0,0 +1,117 @@
+"""
+Tests for the client
+"""
+
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import unittest
+import mock
```
jeromekelleher: The convention I've adopted so far with imports is to have a number of sections. (1) futures; (2) standard libary; (3) third party libraries; (4) package local. Since mock isn't part of that standard library until Python 3.3, which we're not supporting yet, so we should put this into third party stuff. Also, I think we'll need to update the Travis script to install mock for testing.

tests/test_client.py
```diff
@@ -0,0 +1,117 @@
...
+            # invoke SUT
+            url = "referencesets/search"
+            protocolClass = protocol.GASearchReferenceSetsResponse
+            listAttr = "referenceSets"
+            httpClient = makeHttpClient()
+            result = [refSet for refSet in httpClient.runRequest(
+                request, url, protocolClass, listAttr)]
+
+            # verify results of invocation
+            assert len(result) == 2
```
jeromekelleher: These hard-coded values seem a bit brittle and mysterious when reading them. Can we make some constants and use these instead perhaps?

ga4gh/cli.py
```diff
@@ -127,6 +143,32 @@ def printVariant(self, v):
...
+class ReferenceSetSearchRunner(AbstractSearchRunner):
+    """
+    Runner class for the referencesets/search method.
+    """
+    def __init__(self, args):
+        super(ReferenceSetSearchRunner, self).__init__(args)
+        request = protocol.GASearchReferenceSetsRequest()
+        if args.accessions is not None:
+            request.accessions = args.accessions.split(",")
+        if self.usingWorkaroundsFor('google'):
```
jeromekelleher: I'm not crazy about using hard-coded values for program flow, as it's very brittle. Perhaps we could add a constant to `client.py` that we could check against here and in `client.py` instead?

jeromekelleher:
This looks great @dcolligan, thanks. I especially like the testing setup using `mock`, this seems very neat. I'm struggling to get my head around exactly how it works though - do you think you could add a few quick comments explaining the overall testing strategy and how it works? 

I'm like the workarounds API too, minus the hard-coded strings. We'll probably need something a bit more sophisticated later, but this'll get us up and running.

tests/test_client.py
pashields: In this case, they are referring to values inside the dummy response. Testing against specific values is more brittle, but it saves you from figuring out that two incorrect values are equal.

In production code I'd be with you 100%, but in tests I think this way is way safer.

tests/test_client.py
dcolligan: The CI build wouldn't have passed if we needed to explicitly `pip install mock`

dcolligan:
Pushed another commit which squashes the previous ones and addresses some concerns here.

@jeromekelleher as far as the tests go, mocks are essentially used to provide a test stand-in for functionality that may be either expensive or impossible to set up.  Classic examples include database connections, network calls and large file calls.  Here in `TestRunRequest` I'm patching the `requests.post` call to be intercepted by the mock library and return a `DummyResponse` object.  In this way, we both preserve the property that the test suite can be run locally (thereby improving modularity and robustness since relying on a remote resource which may not be available would be a bad idea anyways) and we save some time by not actually waiting for the network.  We also improve modularity by abstracting the details of the actual requests.post library call -- some bug in a change in the requests library implementation now won't affect our test code.

In `TestSearchMethodsCallRunRequest` the idea is similar, except we don't really care to test the return value since there's no logic between us providing a dummy one and us testing what the return value would be.  So we simply assert that the method we wanted to test called the function it should be calling correctly.  Again, we're using mocks to test only the layer of code we want to test and not anything below it.  

Let me know if you still have questions about any of this.

(the only test that failed in CI is the CORS test)

tests/test_client.py
```diff
@@ -0,0 +1,118 @@
+"""
+Tests for the client
+"""
+
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import unittest
+import requests
```
jeromekelleher: requests is a third party lib - could you move it into the next section please?

tests/test_client.py
```diff
@@ -0,0 +1,118 @@
...
+    Test the logic of runRequest
+    """
+    def testRunRequest(self):
+        # setup
+        mockPost = mock.Mock()
+        with mock.patch('requests.post', mockPost):
+            mockPost.side_effect = [DummyResponse(), DummyResponse('{}')]
+            request = DummyRequest()
+
+            # invoke SUT
```
jeromekelleher: What's SUT? Sorry if I'm being a bit dense here!

jeromekelleher:
Thanks for the explanation @dcolligan, this helps a lot. This is a great approach, and gives a nice lightweight way of testing just what we want to test. Very nice!

I'm +1 to merge, modulo a couple of tiny nits above.

tests/test_client.py
pashields: System Under Test

dcolligan:
Ok @jeromekelleher pushed something to address all your concerns (except the one that you and @pashields disagree on... I'm ambivalent about that issue myself)

pashields:
@jeromekelleher you have a strong feeling on the testing issue? I don't feel the need to block this PR on it either way.

jeromekelleher:
Sorry, should has responded to this. I'm happy to go with your view @pashields, I don't feel strongly about this at all.

pashields:
Cool. Thanks.

+1 on this PR

jeromekelleher:
Excellent. I kicked off a fresh travis build as there has been a fair bit of merging today. I'll merge once the build completes - looks like it's failing when we try to download pysam. We really need to get some binary wheels for travis sorted out!

dcolligan:
Yeah, I had to force push again to get Travis working... but it passes now.

jeromekelleher:
Merfged - thanks @dcolligan!

PR #91, opened 2015-01-23
Update test coverage in travis run to include all files and enforce minimum coverage of 80%

melaniedc:


jeromekelleher:
Interesting! I think we should give this a try and see how it goes.

+1.

melaniedc:
Cool, let me know if you think 80% is too harsh - the idea is to make sure people add tests when they add new code :) If it does work well for everyone we can increase the minimum as our coverage goes up.

kerrydc:
+1, this is great.

dcolligan:
What level of test coverage are we at now?  Assuming we are at less than 80% -- which I venture that we are -- isn't this going to fail every TravisCI run until we get above 80%?

melaniedc:
We should be at exactly 80% now when including all files - you can see in the details of the travis run on this PR.

dcolligan:
Awesome, I'll merge.

Naburimannu:
On the Travis bot, my PR is getting an error that suggests not that my coverage is to low but that there's something wrong with our coverage configuration?

    0.61s$ nosetests --with-coverage --cover-package ga4gh --cover-inclusive --cover-min-percentage 80
    usage: nosetests [-h]
    nosetests: error: unrecognized arguments: --with-coverage --cover-package ga4gh --cover-inclusive --cover-min-percentage 80
    The command "nosetests --with-coverage --cover-package ga4gh --cover-inclusive --cover-min-percentage 80" exited with 2.

From https://travis-ci.org/ga4gh/server/builds/48106520#L453

Naburimannu:
Even though nosetests -h lists --cover-inclusive, if I include that in the locally command line it fails (with the same error message seen on Travis); if I remove it from the command line it succeeds.

Naburimannu:
Error messages cleared up (both locally and on the Travis server) once I'd merged with master. No problem here, I guess.

PR #68, opened 2015-01-13
File reorganisation

jeromekelleher:
This PR reorganises the files in the `ga4gh` package into a more logical order and opens the way for adding support for reads and references. Apologies for basing the diff on #59 - this was already quite a large change, so I thought it simpler to go with this route.  Hopefully that will be merged soon and we'll be able to see the changes here more clearly.

The basic gist of this PR is that we layout the code as follows:
```
ga4gh/
     __init__.py
     backend.py
     frontend.py
     protocol.py
     datamodel/
           __init__.py
          variants.py
          reads.py
          references.py
```
(I've left out a few files here that haven't been  affected.).

The idea is that we pull the classes in `backends.py` that are dealing with the GA4GH 'data model' and put them into their own package. We further split this package up into `variants.py`, `reads.py`, and `references.py`. I think it's a good idea to split these up as they are logically distinct and now that there's a lot of us working on the code, it'll be good to avoid merge issues.

I've renamed `backends` to `backend` since there is now a single `Backend` object. Similarly, I've renamed the server package to `frontend`, since it contains the Flask frontend. 

I changed the Flask frontend to use a single file for now. I did argue for separating it out into `views.py` and `__init__.py` when we were making the transition to Flask, but I don't think we're getting a lot of value out of this at the moment. We can always split things out again later if `frontend.py` is getting too complicated. 

I've made the minimal changes to the test cases to make them work. I think we should refactor and reorganise these as a separate PR later.

This is just a proposal to start a discussion, so please weigh in if you disagree with this layout!

PR #93, opened 2015-01-26
Added testing config and moved default config file

jeromekelleher:
This PR moves the `ga4gh.server.config.py` file to `ga4gh.serverconfig.py`, completing the removal of the `server` package. Also adds a stub `TestConfig` class used to set up flask frontend tests.

This fixes the current build breakage in Travis, and should be small and uncontroversial, so it would be good to vote and merge this as soon as possible.

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -16,7 +16,7 @@ app = api.FlaskAPI(__name__)
 
 
 def configure(config):
-    configStr = 'ga4gh.server.config:{0}'.format(config)
+    configStr = 'ga4gh.serverconfig:{0}'.format(config)
     app.config.from_object(configStr)
     if os.environ.get('GA4GH_CONFIGURATION') is not None:
         app.config.from_envvar('GA4GH_CONFIGURATION')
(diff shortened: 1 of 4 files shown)


pashields:
+1

diekhans:
What is the logic behind removing the server package?  As software grows, its help to have things already in logical groupings.

Naburimannu:
@diekhans, I think the main change was at https://github.com/ga4gh/server/pull/68, although I don't know that there was much discussion on the merits.

diekhans:
thanks @Naburimannu, I actually thing we should have a ga4gh.refserver package and have the entire reference server under that.  Hopefuly there will be a lot more ga4gh python code (and we should keep the packages consistent with the namespaces in avro.

jeromekelleher:
This would be good to do @diekhans, and I hope we can rethink the package layout again fairly soon. It was working out to be very complicated with Flask, and there was no longer a clean division between server and client. I thought it best to get rid of the `server` package for the moment, in the interest of making progress as quickly as possible. We definitely need to rethink this when it comes to distributing the code. Hopefully we'll write a bunch of code over the next few months, and we'll be able to see more clearly where we need to make the divisions then.

Naburimannu:
+1, for unit test fix and for deploying an implicit test of configuration files. We'll have to keep thinking about package layout.

Issue #94, opened 2015-01-26
 CORS unit test broken


Naburimannu:
@jeromekelleher @dcolligan : the merge in 6b9ef7e4e seems to have broken @melaniedc 's new CORS unit test? If I check out 399e96 or earlier SHAs, nosetests completes.

I think what's going on here is that we no longer call cors.CORS() for all invocations, but only in frontend.configure(), which is only called when we're entering through server_main().

Does nosetests not execute server_main()?

dcolligan:
I have a fix in a PR... see https://github.com/ga4gh/server/pull/95

dcolligan:
#93 fixes this

PR #95, opened 2015-01-27
Fix broken unit test

dcolligan:
Issue #94 

So this is what happened:
```
$ git log origin/master --oneline --decorate --graph
* 05f346f (HEAD, origin/94_cors_test, 94_cors_test) Fix broken unit test
*   77d9352 (upstream/master, origin/master, origin/HEAD, master) Merge pull req
|\
| * d35744a (origin/dcolligan_55_cleanup) Rename single-character variables
* |   6b9ef7e Merge pull request #80 from dcolligan/dcolligan_69
|\ \
| * | 22df2e1 (origin/dcolligan_69) Add cmd line option to switch server config
| |/
* |   399e966 Merge pull request #91 from melaniedc/travis
|\ \
| * | 434de3f fix linebreak error
| * | 499a5e0 Make coverage include all files and enforce a minimum of 80%.
* | |   6800cf1 Merge pull request #65 from melaniedc/cors
|\ \ \
| |/ /
|/| |
| * | ff9ee5c Add unit test for CORS functionality.
|/ /
* |   e29dc58 Merge pull request #77 from pashields/pep8
```

Neither @melaniedc's change nor my change was a child of the other.  Both PRs passed unit tests in their respective branches, had no conflicts with the current master, and were merged into master, but their interaction caused the failure.

I suppose this problem is unavoidable with the merging model we have now, where commits to be merged in are not mandated to be rebased onto the most recent master.  That is, until we have a linear commit history, we could keep having this problem.

Do we want to have a linear commit history?  That would mean not using the "merge" buttons in the GitHub UI...


dcolligan:
It also means that GitHub's model of travis integration is broken, since it allows people to apply the merge commit to master before CI is run on it... grumble...

Naburimannu:
+1 on the fix, for expedience. Might we want to always configure, however? (It seems to me that once a test starts depending on configuration, it's drifting away from being a unit test?)

And "+1" on the grumble; I'm also used to infrastructure that prevents a commit from landing if it breaks any tests.

jeromekelleher:
Actually, #93 already fixed this issue (sorry, I should have been more clear in the PR message), so I'd vote to merge #93 and close this PR.

+1 on the grumbles though, this is an annoying problem. Any suggestions for ways to fix it?

pashields:
According to travis-ci doc:
```
Rather than test the commits from the branches the pull request is sent from, we test the merge between the origin and the upstream branch.
```

So if that's the case we should just re-run the tests by hand in travis before the merge.

jeromekelleher:
This is a good idea I think. So, in this case, I should have rerun the travis build before I merged the CORS unit test?

Should we close this PR?

dcolligan:
@pashields That would fix the issue, at the cost of adding some additional developer effort on each merge, except I don't think there's any way to manually run tests in Travis, is there?  (It would make sense that there's not, since that kind of defeats the point of CI.)

If there's not, we should have a `run_tests` script that we keep updated for this purpose.

Naburimannu:
Now that #93 is merged, yes, let's close this.
Having to manually run travis (or just manually kick off nosetests) is unfortunate, but that might need to be part of the merge process until we find a better workflow.

jeromekelleher:
We can tell Travis to rerun a build - just click through on the little green tick (or red cross), and on the top right corner of the travis page there is a icon to rerun the build. This might not show up for every user though. I've just kicked off a rebuild for #133 to see if it works.

pashields:
It should show up for all collaborators.

Issue #96, opened 2015-01-27
Create run_tests.py for local Travis imitation


dcolligan:
Created after discussion in #95 

Travis (AFAIK) does not allow manual test runs (see http://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit -- a run can be triggered via a push, and the whole issue is we want a CI run BEFORE a push to master to ensure we don't break it!) and it would be a good practice for merging devs to ensure merges don't break master before pushing them to GitHub.

`run_tests.py` should imitate the Travis CI run.  `travis.yml` and `run_tests.py` should be updated in parallel.

dcolligan:
PR in #97

PR #97, opened 2015-01-27
Add run_tests.py to imitate travis run

dcolligan:
Issue #96

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,8 @@ before_script:
   # verbosely complaining about a security problem; see issue 54
   - mkdir $HOME/.python-eggs
   - chmod og-w $HOME/.python-eggs
-  - pep8 ga4gh tests
 
-script: nosetests --with-coverage --cover-package ga4gh
-                  --cover-inclusive --cover-min-percentage 80
+script: 
+  - pep8 ga4gh tests scripts 
+  - nosetests --with-coverage --cover-package ga4gh
+              --cover-inclusive --cover-min-percentage 80
--- a/scripts/run_tests.py
+++ b/scripts/run_tests.py
@@ -1,0 +1,47 @@
+"""
+Script to imitate a Travis CI run
+"""
+
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import subprocess
+
+import yaml
+
+
+class TravisSimulator(object):
+
+    logStrPrefix = '***'
+
+    yamlFileLocation = '.travis.yml'
+
+    def parseTestCommands(self):
+        yamlFile = file(self.yamlFileLocation)
+        yamlData = yaml.load(yamlFile)
+        return yamlData['script']
+
+    def runTests(self):
+        testCommands = self.parseTestCommands()
+        for command in testCommands:
+            returnCode = self.runCommand(command)
+            if returnCode != 0:
+                self.log('ERROR')
+                return
+        self.log('SUCCESS')
+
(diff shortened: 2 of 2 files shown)


dcolligan:
cc'ing everyone who was chatting in #95 

@pashields @jeromekelleher @Naburimannu

jeromekelleher:
This is pretty neat! So, the idea is that we can run this script before pushing to save the trouble of running the travis build? Seems like a useful tool to me.

Do we need the extra `dev_scripts` dir? It seems to me that they're all dev scripts really - this one and `generate_schemas.py` don't need to be in the root as they don't import  `ga4gh`. Maybe move this one into `scripts/`? Or rename `scripts/` to `dev_scripts` if you prefer?

dcolligan:
@jeromekelleher The idea is Travis will likely not run on the commit we want it to, when we want it to.  When one pushes a PR, Travis will do a merge into master and then run CI on that.  That CI result is different than the one that we really want, which is the merge into master *at the time that the merge is going to take place.*  That is, master may have changed between when Travis runs CI and when we want to run CI.  This is why we couldn't catch the test breakage before.

I'm not sure if re-submitting a Travis job will merge onto the current master, or the master that it previously ran on.  If it's the second, then we definitely need my change, and should incorporate it into our merging flow.  If it's the first, then the importance of my change reduces to 1) a convenient way for devs to see if travis will reject their change before submitting the PR 2) a faster way to get travis results than through travis itself.  In either case, I think it's warranted.

jeromekelleher:
>I'm not sure if re-submitting a Travis job will merge onto the current master, or the master that it previously ran on. If it's the second, then we definitely need my change, and should incorporate it into our merging flow. If it's the first, then the importance of my change reduces to 1) a convenient way for devs to see if travis will reject their change before submitting the PR 2) a faster way to get travis results than through travis itself. In either case, I think it's warranted.

I think it's (1), as the documentation that @pashields shows. I've kicked off a bunch of previously failing PRs, which are fine now, so I think this is fairly conclusive. I still like this PR though, it's a useful tool. It's tedious having to remember to check coverage and pep8 each time, and travis builds takes ages, so this is a neat shortcut.

pashields:
I'm good with this change, but it does not replace a travis run, particularly when we start to talk about supporting more than one version of python. But travis also provides a clean repeatable env. "run_tests.py worked on my box" is not a valid excuse.

In the long run, we may wish to actually parse the travis yaml file rather than manually sync these. If this is a real blocker for work right now, I'm fine to have a TODO in there.

Also, to be clear, I'd view it as the responsibility of the person running the merge to re-run travis.

dcolligan:
Ok, submitted a change where I'm now parsing the yaml, but I needed to move some things around in the `.travis.yml` file.  It would seem to make sense to me to keep everything that we want to run as an actual test (vs setup) under the `script:` entry.

pashields:
awesome. Assuming tests pass, I can happily +1 this.

Naburimannu:
+1

jeromekelleher:
This really is very nice! One thing: is yaml a third party lib? If so, I think there should be a blank line between the imports.

I'm not suggesting we do it here, but I think we should try to keep track of all the third party libs we're using for testing and dev. It'd be nice to be able to install all these from a `requirements.txt` or whatever when testing in a virtualenv.

dcolligan:
@jeromekelleher pushed a change with the extra newline you requested

File an issue for a requirements.txt, that sounds like a good idea

PR #98, opened 2015-01-27
Style guide

pashields:
This is a starting point for a style guide. It is intentionally sparse for now. Let's first make sure that we are OK with this format and the principles outlined here. Any additions should be added in a separate PR afterwards.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -1,0 +1,16 @@
+# GA4GH Server Style Guide
+
+The code follows the guidelines of `PEP 8
+<http://legacy.python.org/dev/peps/pep-0008>`_ in most cases. The only notable
+difference is the use of camel case over underscore delimited identifiers; this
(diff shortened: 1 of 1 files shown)


dcolligan:
Why do variable name rules vary with the size of the enclosing function?  Shouldn't they always be descriptive?  Have there been other conversations about this?

Otherwise, +1

pashields:
There have been other discussions, I'd have to dig for them. Lambda functions or list comprehensions are the big examples here. So for instance, inverting a dictionary might be:
dict((v,k) for k,v in dict.iteritems())

I chose scope line length as a measure simply because it is universal across lambdas, comprehensions, etc.

jeromekelleher:
Great, thanks @pashields. +1.

As this is pretty uncontroversial up to this point, and people can open specific PRs to extend and refine this, I'm going to merge this now.

Issue #14, opened 2014-09-03
What is the correct mapping VCF GT values to `Call` objects?


jeromekelleher:
The mapping VCF genotype values to GA4GH `Call` objects  is currently incomplete/incorrect. Questions:

1. If the delimiter is `|` then the genotype is phased. What should the value of the `phaseset` field be? If the delimiter is `/` then I assume phaseset should be null.
2. What does ./. indicate in terms of `Call` objects? No call exists an so we do not return a `Call` for the current `CallSet`?

mbaudrey:
I'm working on this issue and I'm a bit unclear on some of the specifications.

1. For genotypes containing periods, are only `.` and `.\.` valid inputs? It seems to me that we should also allow genotypes that contain both integers and periods, such as `1/.` or `.|0`, for flexibility. But perhaps it isn't possible for a call to be made on one allele and not the other?

If this format of genotype should be allowed, how would we represent it internally? Currently a standard genotype such as `1` or `0/2` would be represented as a list of the integers it contains (`[1]` or `[0, 2]`) and any genotype containing a period would be represented as `[-1]` regardless of the presence of multiple periods, delimiters, or integers.

If we do allow mixed integers and periods, it the simplest solutions is that the `.`, rather than the entire genotype, should be replaced with a `-1`. So the genotype `1/.` would be represented as `[1, -1]`.

2. What do phase set values mean and how should they be stored? The specifications of VCF format only clarify that the PS field should contain a 32-bit integer and that genotypes with the same value are in the same phase set.

richarddurbin:
For VCF and the current Variant model I think we can safely rule out mixed integer and . genotypes.  Normally the total sum count is known/implicit, e.g. when diploid 1|. must be 1|1 because the allele counts must sum to 2.
So Cassie's rules that if there is any dot then call the genotype as missing seem OK to me.  We can add a note to this effect.

Richard

On 7 Jan 2015, at 19:52, mbaudrey <notifications@github.com> wrote:

> I'm working on this issue and I'm a bit unclear on some of the specifications.
> 
> For genotypes containing periods, are only . and .\. valid inputs? It seems to me that we should also allow genotypes that contain both integers and periods, such as 1/. or .|0, for flexibility. But perhaps it isn't possible for a call to be made on one allele and not the other?
[...]

PR #99, opened 2015-01-27
Improved genotype handling in wormtable variant sets

mbaudrey:
Issue #14

### DIFF
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -57,6 +57,7 @@ class WormtableVariantSet(VariantSet):
     # These must be bytes literals for integration with wormtable.
     GENOTYPE_LIKELIHOOD_NAME = b"GL"
     GENOTYPE_NAME = b"GT"
+    PHASESET_NAME = b"PS"
 
     def __init__(self, variantSetId, wtDir):
         """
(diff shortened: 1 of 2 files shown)


Naburimannu:
Even though it's trivial, it looks like you'll need to fix the PEP8 whitespace error before Travis will even try the new unit tests.

mbaudrey:
It's taken care of. Sorry about that.

jeromekelleher:
This is excellent, thank @mbaudrey! I've got two suggestions for improving the code layout, but generally I think it's great. 

The first suggestion would be that it would be good to change `convertGenotype` and `convertPhaseset` from static methods of `WormtableVariantSet` to standalone functions, which we might call `convertVCFGenotype` and `convertVCFPhaseset`. The functions are simple standalone bits of functionality, so I think they would be better like this.

For the tests, I think it would be better to group together all of the new test functions into one, which we might go like this:
```python
    def testGenotypeConversion(self):
        # Data is a list of tuples consisting of the VCF genotype and
        # phaseset values followed by the expected GA4GH converted
        # values.
        data = [
            # Genotype with no call
            ("./.", "0", [-1], None),
            # Half call where only the second half is called
            ("./0", "25", [-1], None),
            # ETC
        ]
        for vcfGenotype, vcfPhaseset, ga4ghGenotype, ga4ghPhaseset in data:
            result = variants.convertVCFGenotype(vcfGenotype, vcfPhaseset)
            convertedGenotype, convertedPhaseset = result
            self.assertEqual(ga4ghGenotype, convertedGenotype)
            self.assertEqual(ga4ghPhaseset, convertedPhaseset)
```
This just makes it a bit easier to see all of the test examples in one place, and saves on some repetitive code.

diekhans:
I am over two minds on this.  While the terseness of the coding
this as a loop is nice, it defeats the test cases are
independent goal.

I would be inclined to have a bunch of one-line test cases that
call a common function, would be roughly twice as many lines as
this.

jeromekelleher:
It's a good point, and I agree. The main goal of the above was to make it easier to add new test cases. What about something like this?
```python
    def testGenotypeUnphasedNoCall(self):
        self.verifyGenotypeConversion("./.", "0", [-1], None)
```
This makes it nice and quick to write new test cases without duplicating code.

BTW, this is (as far as I can think) the only example where I think it's better not to have docstring comments on a method. When nosetests runs these in verbose mode, it uses the docstring comments rather than the test method name in the output, which seems less useful. So, here I would definitely document the `verifyGenotypeConversion` method (as this isn't a test case, but a utility function), but leave the actual test cases without docstring comments. 

Does this make sense? I'll make a PR on the style guide writing this up soon.

diekhans:
+16

Agree on the doc string, I fix the behavior of the unittest with
doc strings less than ideal.

Meaningful tests names and then comments that go into more
detail if necessary are the way to go.

kerrydc:
+1

jeromekelleher:
I think we were hoping to update the unit tests here to follow the pattern outlined above before merging. It's not worth backing out of the merge though, as it's mostly right. @mbaudrey, would you mind making a follow-up PR with the changes outlined in my first comment above?

Naburimannu:
@jeromekelleher I thought we should be making the follow-up PR with the changes from @diekhans' comments / your second comment - one function per test case, even if they're just calling a common utility function with different arguments?

jeromekelleher:
Yeah, you're right, that was confusing! What I meant to say was, update to have tests like
```python
    def testGenotypeUnphasedNoCall(self):
        self.verifyGenotypeConversion("./.", "0", [-1], None)
```
I think it would also be good to change `convertVCFGenotype` to a function, since it doesn't depend on the class in any way. Does this make sense @mbaudrey?

mbaudrey:
Yes, I can make a follow up PR. To clarify, both convertVCFGenotype and
convertVCFPhaseset should be standalone functions? And should they still be
in variants.py or is there a better place to put them?

jeromekelleher:
Yes, both should be standalone, and variants.py is where they should live. Thank @mbaudrey!

diekhans:
It's nice to have classes/modules specific to functional areas
to place some standalone functions.  Agree it doesn't belong in
the worm table module, but don't know the code structure well enough
to suggest better.

kerrydc:
I'm doing this in my Tabix PR that should be submitted this afternoon.
Kerry

Issue #41, opened 2014-11-24
Add Beacon Server API support


hershman:
With the change over to Flask (see issue #38 and PR #40 for discussion and progress) and our support for plain VCF files using the `TabixBackend` we have the infrastructure to very easily make a Beacon  server. This could be trivial to deploy using the PyPI package (see issue #36) and would require only a path to a directory of VCF files to configure.

hershman:
I am working on PR for Beacon Spec (0.2)[https://docs.google.com/document/d/154GBOixuZxpoPykGKcPOyrYUcgEXVe2NvKx61P4Ybn4/edit#heading=h.234otr74rjti].

kerrydc:
Isn't this a bit premature? I mean, a Beacon is something that we don't
need until we have a real, deployable app. Outstanding API methods,
benchmarking, testing, security, logging, and so on are all higher
priority. It is likely the Beacon spec will continue to evolve while we
finish up implementing this basic functionality, rendering prior work
obsolete.
Kerry

jeromekelleher:
Not sure I agree @kerrydc - a Beacon could be something that we can turn into a real deployable app much more quickly than the full API. The sooner we ship working code, the better, and I think a lot of people would be interested in a simply deployable beacon. Some of us could concentrate on testing, logging etc for this simpler instance while the development for the larger API is also going on by others. 

Feel free to open an early PR once you're ready for comments @hershman.

maximilianh:
I think Beacons should not depend on any of the other GA4GHs API. Beacons
might often run on the external public legacy webserver, not necessarily on
an internal cloud. Getting security approval is easier with a simple
one-file webservice than a host of libraries, tools and dependencies.

hershman:
Hi @maximilianh, Thanks for joining the discussion. Your easily deployable beacon at https://github.com/maximilianh/ucscBeacon is really awesome!  

I don't it would silly for the reference server to not also be able to host a beacon in addition.  I have an implementation that will not be hard to clean up and catch up to the current Master branch and Beacon spec https://github.com/hershman/server/tree/beacon

It would be ideal if the same code base could be used, but the languages seem different enough and the amount of implantation seems small enough that I'm excited to forge ahead if the community is.

bartgrantham:
[...]

Issue #26, opened 2014-10-20
A beacon client would be useful


jeromekelleher:
Adding scriptable support for the beacon API would be very useful. It would be good to have a `beacon` function in the `ga4gh.client.HTTPClient` class, following the existing pattern. It would then be simple to add a CLI (either as part of the existing client application or as its own standalone program, if it was sufficiently useful) using this method. This would also require #25 to be solved, as Beacons will most often be at prefixed endpoints.

kozbo:
This effort is being tracked over [here](https://github.com/BD2KGenomics/ga4gh-integration/issues/17) I did not create a Beacon endpoint in the client lib, but rather created a stand alone service that would work as an add-on to any GA4GH server. I will post code soon, just working out consistency with the current beacon structures spec.

kozbo:
@jeromekelleher shall we close this issue and track progress in the issue in BD2KGenomics/ga4gh-integration?

jeromekelleher:
Good idea @kozbo, closing.

PR #103, opened 2015-01-29
Beacon 0.2

hershman:
This is an early stage PR that implements a Beacon 0.2 server (Issue #41) and a homepage for the client (Issue #26 ).  I still need to work on tests, but I wanted to release it to begin the discussion.

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -9,6 +9,7 @@ from __future__ import unicode_literals
 import os
 
 import ga4gh.protocol as protocol
+from ga4gh.frontend import app
 
 
 class Backend(object):
@@ -91,14 +92,12 @@ class Backend(object):
             self.variantsGenerator)
 
     def infoBeacon(self, request):
-        from frontend import app
         response = protocol.BeaconResource()
         response.id = app.config["BEACON_ID"]
         response.name = app.config["BEACON_NAME"]
         response.organization = app.config["BEACON_ORGANIZATION"]
         response.description = app.config["BEACON_DESCRIPTION"]
         response.api = "0.2"
-        import os
         hostName = os.environ.get("HTTP_HOST", "localhost")
         response.homepage = "http://%s/" % hostName
         return response.toJSONString()
@@ -107,17 +106,18 @@ class Backend(object):
         """
         Searches beacon but need to update to 0.2
         """
-        from frontend import app
-        def samevars(result, allele):
+
+        def sameAllele(variants, allele):
+            """Checks if any variaints have the alternative allele specified"""
             if allele is None:
                 return "True"
-            for variant in result:
+            for variant in variants:
                 if allele == "D":
                     if len(variant.referenceBases) > 1:
                         return "True"
                 else:
-                    for alt in variant.alternateBases:
-                        if allele in alt:
+                    for alternativeBase in variant.alternateBases:
+                        if allele in alternativeBase:
                             return "True"
             return "False"
 
@@ -131,22 +131,19 @@ class Backend(object):
         if 'dataset' in request.values:
             query.dataset = request.values["dataset"]
 
-        #Better but doesn't work
         result = []
-        variantSetIds = request.values["population"]
-        print(variantSetIds)
-        startPosition = int(request.values['coord'])
-        variantSetId = request.values["population"]
+        startPosition = query.position
+        variantSetId = query.dataset
         if variantSetId in self._variantSetIdMap:
             variantSet = self._variantSetIdMap[variantSetId]
             for variant in variantSet.getVariants(
-                str(request.values["chr"]), startPosition, startPosition,
+                query.chromosome, startPosition, startPosition,
                 None, []):
                 result.append(variant)
         else:
-            raise Exception, "%s missing" % variantSetId
+            raise Exception, "%s is not a valid variantSetId for this server" % variantSetId
 
-        exists = samevars(result, request.values['allele'])
+        exists = sameAllele(result, query.allele)
 
         request_response = protocol.ResponseResource()
         request_response.exists = exists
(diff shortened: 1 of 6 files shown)


ga4gh/frontend.py
```diff
@@ -12,13 +12,31 @@
...
 app = api.FlaskAPI(__name__)
 app.config.from_object('ga4gh.server.config:DefaultConfig')
 if os.environ.get('GA4GH_CONFIGURATION') is not None:
     app.config.from_envvar('GA4GH_CONFIGURATION')
 
+app.config['DEFAULT_RENDERERS'] = [
+    'flask.ext.api.renderers.JSONRenderer',
+    'flask.ext.api.renderers.HTMLRenderer',
+    'flask.ext.api.renderers.BrowsableAPIRenderer',
+]
```
hershman: I needed these changes to get the homepage to render, but I fear this may break the BrowsableAPIRenderer.  I tried using [@set_renderers(HTMLRenderer)](http://www.flaskapi.org/api-guide/renderers/#htmlrenderer), but am getting TypeError: 'type' object is not iterable

Naburimannu: Looks like that's undergone some recent revision in Flask, and may not be
settled: https://github.com/tomchristie/flask-api/issues/8

I *think* the resolution is that they want [HTMLRenderer]

ga4gh/backend.py
```diff
@@ -9,6 +9,7 @@
 import os
 
 import ga4gh.protocol as protocol
+from ga4gh.frontend import app
```
jeromekelleher: We're breaking the isolation between the frontend and backend here. The backend should definitely not depend on the frontend for all sorts of reasons.

ga4gh/frontend.py
```diff
@@ -40,11 +58,23 @@ def handleHTTPOptions():
...
 def index():
-    flask.abort(404)
+    if app.config["BEACON"]:
+        # TODO: Get list of variant_sets from backend
+        return flask.render_template('beaconClient.html', variant_sets=app.backend._variantSetIdMap)
+    else:
+        flask.abort(404)
+
+
+if app.config["BEACON"]:
```
jeromekelleher: This won't work in general, because the config won't be set at module load time.

jeromekelleher:
Thanks for this @hershman, this is a great start! I think if we step back a 
bit and try to add support for the Beacon protocol bit-by-bit then we 
can get this in nice and quickly. There's a few things I think we can 
do the make our lives easier. This will involve removing some of the 
things that you've done, which I hope you won't mind. We need to make 
sure that everything we commit from now on has a good level of test
coverage, and trimming down the size of changes is the best way to make
sure this happens. 

Firstly, I think we should separate out the problem of supporting the 
Beacon _protocol_ from providing a Beacon HTML frontend. These are not 
the same thing, and we need to support the protocol first before we 
should think about providing a HTML interface. So, I would simplify
here by removing all of the HTML rendering stuff and the templates.

Secondly, I this we should implement the Beacon protocol as it's 
defined in v0.5.1. We really want to stick with supporting a specific
version of the protocol as defined in a released tarball. If we start
editing _protocoldefinitions.py we're never going to know exactly 
what version of the protocol we're on, and we'll lose a lot of the
gains that we make by having machinery to automatically process the 
schema. Once the 0.2 version of Beacon is finished and added to the 
upstream schema, we can argue for a new release of the schema to be 
made. We can add support for this into the reference server then.

So, my suggestion would be to start with the Beacon as defined in 
v0.5.1 and follow exactly the same patterns as we are using for 
the rest of the protocol. We'll need to add test cases for the new 
queries and it would be useful to have support in the client 
CLI for running Beacon queries against the server. I wouldn't worry
too much about configuring the Beacon for now, let's just add
support for the queries and think about how to configure a Beacon
in standalone mode later.

ga4gh/backend.py
hershman: Beacon 0.2 includes a bunch of fields that are probably best set in a config file. I cringed when I wrote this line, but I needed the app in order to get to the config.  Is there a better an alternative way  to access to app.config or a better design I could use?

Naburimannu:
Agreed that v0.5.1 Beacon should be supported unless there's some strong reason not to; -1 to manual edits of `_protocol_definitions`.

hershman:
The edits in _protocol_definitions.py were generated by this modified version of the [generate schemas](https://gist.github.com/hershman/73fa968e0ce5277c6ecc) that uses a local directory instead of downloading the schemas.  I made a [PR in the schemas](https://github.com/ga4gh/schemas/pull/235) repository for Beacon 0.2 so hopefully it will become part of v0.5.2.

The beacon group is moving fast - I expect 0.2 to be the standard soon. Instead of working on a standard that's about to be obsolete, what do you think on holding off this PR until the schema is updated?

jeromekelleher:
Yeah, that sounds like a plan @hershman, let's pick this up again when the Beacon spec is finished. I'm going to close this PR for now, since things will have moved on a lot when it comes through, and we don't want to keep unmergeable PRs lying around for too long. I'm looking forward to seeing it again though!

ga4gh/backend.py
jeromekelleher: I guess we'd want to fill out the config dependent stuff in the frontend. We'll have to do a bit of thinking to see what the best approach is, but I'm sure there's a simple solution which preserves the frontend/backend split.

mfiume:
Thanks so much @jeromekelleher and @hershman. I anticipate that we'll be able to stamp the 0.2 Beacon API final soon and we can proceed. Just waiting on a last round of review.

jeromekelleher:
That's great @mfiume, we're really looking forward to baking Beacon support in here. Hopefully once the Beacon 0.2 spec is final we can roll a release of the API spec and start implementing it in the server.

Issue #92, opened 2015-01-23
Explore using __slots__ for protocol objects


Naburimannu:
Some of our memory burden (observed #10 ) comes from the dict allocated for every object; we can have a quarter million or more objects allocated on the server to answer a single query. Python classes can explicitly declare `__slots__` = [...] with a fixed immutable list of fields to prevent dict allocation.

This might save 15% of the peak memory burden (and perhaps provide speedup due to lack of memory churn?); @pashields points out that only a couple of lines of change in the codegen script will be needed. It would be cleaner, more pythonic, and more amenable to error handling than my previously prototyped ad-hoc string aggregation.

Naburimannu:
Waiting until we have server benchmark and heap memory profiling live to investigate.

Naburimannu:
Short of a formal investigation: https://github.com/Naburimannu/server/commit/8fbf3f526c14c907b5751d57d9a68494b9182376 spends  6.45-7.1 seconds in server_benchmark.py and never measures more than about 350MB on top; without that one patch, server_benchmark.py takes 10+ seconds and measures more than 650MB on top - considerably more savings than I expected.

We should be able to do this in a more principled way by using `__slots__` for all the protocol classes. The drawback is that our users won't be able to attach arbitrary new fields to protocol objects. There may well be a way that lets us cleanly use `__dict__` on the more heavyweight objects, and just switch to `__slots__` for GACalls (which server_benchmark is handling 250,000 of).

The designed way around that inflexibility in Python would be for them to create their own subclasses, which by default get a __dict__ and can be freely extended, but then (1) we have to provide them with some sort of factory access, and (2) when they do drop in a non-slots subclass of GACall they lose all this performance.

pashields:
This seems like a really good idea with big savings, unless I've missed something huge. @Naburimannu I'm moderately familiar with the code if you want a hand with this at any point.

jeromekelleher:
I've no objection to using `__slots__` if it doesn't make things more complicated. It seems like a pretty hefty performance gain, so I'd be very enthusiastic about this change. I don't think it's a huge loss that the protocol classes don't allow arbitrary fields being added - arguably, this is a good thing as it prevents common spelling errors.

Naburimannu:
[...]

PR #105, opened 2015-01-30
Heap profiling inside server_benchmark

Naburimannu:
If "--profile heap" is passed to server_benchmark.py, prints out
a heap snapshot inside Backend.runSearchRequest(), which currently
has high peak memory allocation.

Adds a dependency on guppy (currently 0.1.10).

Addresses https://github.com/ga4gh/server/issues/46.

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -57,6 +57,7 @@ class Backend(object):
         (object, nextPageToken) pairs, and be able to resume iteration from
         any point using the nextPageToken attribute of the request object.
         """
+        self.startProfile()
         # TODO change this to fromJSONDict and validate
         request = requestClass.fromJSONString(requestStr)
         pageList = []
@@ -68,6 +69,7 @@ class Backend(object):
         response = responseClass()
         response.nextPageToken = nextPageToken
         setattr(response, pageListName, pageList)
+        self.endProfile()
         return response.toJSONString()
 
     def searchVariantSets(self, request):
@@ -132,6 +134,12 @@ class Backend(object):
                     nextPageToken = "{0}:{1}".format(
                         variantSetIndex, variant.start + 1)
                     yield variant, nextPageToken
+
+    def startProfile(self):
+        pass
+
+    def endProfile(self):
+        pass
 
 
 class MockBackend(Backend):
--- a/server_benchmark.py
+++ b/server_benchmark.py
@@ -15,6 +15,20 @@ import ga4gh.backend
 import ga4gh.protocol as protocol
 import ga4gh.datamodel.variants as variants
 
+import guppy
+
+
+class HeapProfilerBackend(ga4gh.backend.Backend):
+    def __init__(self, dataDir, variantSetClass):
+        super(HeapProfilerBackend, self).__init__(dataDir, variantSetClass)
+        self.profiler = guppy.hpy()
+
+    def startProfile(self):
+        self.profiler.setrelheap()
+
+    def endProfile(self):
+        print(self.profiler.heap())
+
 
 def _heavyQuery():
     """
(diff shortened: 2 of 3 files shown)


Naburimannu:
The follow-on I'm envisioning to this would be to add "--profile cpu", a dependency on cProfile, and some encapsulation of the idea of a profiler, but I haven't prototyped that yet, and there may be some impedance mismatch with the way cProfile wants to be invoked. This CL by itself is enough to give us a repeatable measurement to consider https://github.com/ga4gh/server/issues/92, and the heap profile is a *lot* more stable than the time to run the query reported by server_benchmark.py.

ga4gh/backend.py
```diff
@@ -10,6 +10,7 @@
 
 import ga4gh.protocol as protocol
 
+from guppy import hpy
```
jeromekelleher: Do we need this import?

ga4gh/backend.py
```diff
@@ -57,6 +59,8 @@ def runSearchRequest(
         (object, nextPageToken) pairs, and be able to resume iteration from
         any point using the nextPageToken attribute of the request object.
         """
+        if self.profiler is not None:
```
jeromekelleher: What if we did this in a subclass rather than updating the code for the backend object? So, we just have
```python
class ProfilingBackend(backend.Backend):
      def runSearchRequest(stuff):
             profiler.setrelheap()
             super(ProfilingBackend, self).runSearchRequest(stuff)
             print(profiler.heap())
```
This way, we wouldn't need to touch backend.py at all.

server_benchmark.py
```diff
@@ -15,6 +15,8 @@
 import ga4gh.protocol as protocol
 import ga4gh.datamodel.variants as variants
 
+from guppy import hpy
```
jeromekelleher: I tend to prefer `import guppy` and use `guppy.hpy` (see the style PR). However, this might  be the pattern required for this to work, so feel free to ignore in this case.

jeromekelleher:
I like it a lot @Naburimannu. Will my `ProfilingBackend` idea work?

Naburimannu:
I love the idea of ProfilingBackend, but profiler.heap() needs to be called *before* runSearchRequest() returns, because otherwise all those objects go out of scope and disappear from the record. We can get a little bit more abstraction there by wrapping the profiler, but we need to intrude something into the backend.

Fixed import, minor PEP8 issues.

jeromekelleher:
Ah, OK. Is there any cost associated with importing guppy and not invoking the profiler?

server_benchmark.py
```diff
@@ -91,9 +93,26 @@ def benchmarkOneQuery(request, repeatLimit=3, pageLimit=3):
...
+                        choices=['none', 'heap'],
+                        help='"heap" runs a heap profiler '
+                             'once inside the backend')
+    parser.add_argument('--repeatLimit', type=int, default=3, metavar='N',
+                        help='how many times to run each test case '
+                             '(default: %(default)s)')
+    parser.add_argument('--pageLimit', type=int, default=3, metavar='N',
+                        help='how many pages (max) to load '
+                             'from each test case (default: %(default)s)')
+    args = vars(parser.parse_args())
```
jeromekelleher: Why the `vars` call here? The usual pattern is 
```python
args = parser.parse_args()
pageLimit = args.pageLimit # Or you might just use args.pageLimit directly
```

ga4gh/backend.py
```diff
@@ -10,6 +10,8 @@
 
 import ga4gh.protocol as protocol
 
+import guppy
```
jeromekelleher: Ah, I think I see how this works now. We shouldn't import guppy here as this creates an unnecessary dependency on the backend.

server_benchmark.py
Naburimannu: Good catch! I'd been misled by skimming documentation, and it's much cleaner without the vars() call.

jeromekelleher:
OK, what about this: We make `startProfile()` and `endProfile()` methods in the `Backend` class, which by default do nothing, and we put calls to these where you have the if conditions now. Then, in `ProfilingBackend` we make a heapy instance, which will do all the profiling for us. Will that work, or does heapy need to be called in the actual context?

Naburimannu:
@jeromekelleher: Yes, that's much cleaner. I still don't see guppy getting installed by Travis. @dcolligan, is there anything special I should have to do to get that working? Do I need to land the requirements.txt change before I land the change that has the new requirements?

jeromekelleher:
Looks great @Naburimannu, thanks. I think the Travis build is failing because we're not actually using requirements.txt there. I've made a PR with a fix at #109 --- might be easier to wait until this has gone through, and add your guppy requirement into the updated requirements.txt.

+1 to merge once the dependency issues have been resolved.

dcolligan:
@Naburimannu the master commit that is the merge-base of your branch and master just adds a requirements.txt file but doesn't do anything with it.  After @jeromekelleher's commit that drives the Travis build off of the requirements.txt file lands, you should rebase on top of master and add guppy to the requirements.txt file

(btw, is there any way in GitHub to view where off of master a given branch -- say, the one that this PR refers to -- diverges from master?  I did it the old fashioned way by adding a remote repo locally and then using the git command line tools, but it would be convenient if there were a faster way...)

jeromekelleher:
I use this quite a lot @dcolligan : https://github.com/ga4gh/server/network

dcolligan:
@jeromekelleher thanks, that's what I was looking for

Naburimannu:
Merged, squashed, Travis is happy now.

melaniedc:
+1

PR #108, opened 2015-01-30
76 client endpoints 2

### LINKED ISSUE #76: Extend client library to include full protocol

dcolligan:
The remaining client endpoints.

There were a lot more design decisions to be made here vs the last client endpoint PR.  Feedback appreciated.

Will close #76

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -76,13 +76,16 @@ def server_main(parser=None):
 ##############################################################################
 
 
-class AbstractSearchRunner(object):
+class AbstractQueryRunner(object):
     """
-    Abstract base class for search runner classes
+    Abstract base class for runner classes
     """
     def __init__(self, args):
         self._workarounds = set(args.workarounds.split(','))
         self._key = args.key
+        self._verbosity = args.verbose
+        self._httpClient = client.HttpClient(
+            args.baseUrl, args.verbose, self._workarounds, self._key)
 
     def usingWorkaroundsFor(self, workaround):
         """
@@ -90,24 +93,45 @@ class AbstractSearchRunner(object):
         """
         return workaround in self._workarounds
 
-    def setHttpClient(self, request, args):
-        """
-        Sets the _httpClient and other common attributes
-        """
-        request.pageSize = args.pageSize
-        self._minimalOutput = args.minimalOutput
-        self._request = request
-        self._verbosity = args.verbose
+
+class AbstractGetRunner(AbstractQueryRunner):
+    """
+    Abstract base class for get runner classes
+    """
+    def __init__(self, args):
+        super(AbstractGetRunner, self).__init__(args)
+        self._id = args.id
         self._httpClient = client.HttpClient(
             args.baseUrl, args.verbose, self._workarounds, self._key)
 
-    def _run(self, methodName, attrName=None):
+    def _run(self, method):
+        response = method(self._id)
+        print(response.id)
+
+
+class AbstractSearchRunner(AbstractQueryRunner):
+    """
+    Abstract base class for search runner classes
+    """
+    def __init__(self, args):
+        super(AbstractSearchRunner, self).__init__(args)
+
+    def _setRequest(self, request, args):
+        """
+        Sets the _httpClient and other common attributes
+        """
+        self._minimalOutput = args.minimalOutput
+        if 'pageSize' in args:
+            # GAListReferenceBasesRequest does not have a pageSize attr
+            request.pageSize = args.pageSize
+        self._request = request
+
+    def _run(self, method, attrName=None):
         """
         Runs the request given methodname and prints out
         the each result's attrName attribute if it is provided.
         If not, prints each entire result object.
         """
-        method = getattr(self._httpClient, methodName)
         results = method(self._request)
         for result in results:
             if attrName is None:
@@ -117,26 +141,26 @@ class AbstractSearchRunner(object):
                 print(attr)
 
 
-class VariantSetSearchRunner(AbstractSearchRunner):
+class SearchVariantSetsRunner(AbstractSearchRunner):
     """
     Runner class for the variantsets/search method.
     """
     def __init__(self, args):
-        super(VariantSetSearchRunner, self).__init__(args)
+        super(SearchVariantSetsRunner, self).__init__(args)
         request = protocol.GASearchVariantSetsRequest()
         setCommaSeparatedAttribute(request, args, 'datasetIds')
-        self.setHttpClient(request, args)
+        self._setRequest(request, args)
 
     def run(self):
-        self._run('searchVariantSets', 'datasetId')
+        self._run(self._httpClient.searchVariantSets, 'datasetId')
 
 
-class VariantSearchRunner(AbstractSearchRunner):
+class SearchVariantsRunner(AbstractSearchRunner):
     """
     Runner class for the variants/search method.
     """
     def __init__(self, args):
-        super(VariantSearchRunner, self).__init__(args)
+        super(SearchVariantsRunner, self).__init__(args)
         request = protocol.GASearchVariantsRequest()
         request.referenceName = args.referenceName
         request.variantName = args.variantName
@@ -151,11 +175,11 @@ class VariantSearchRunner(AbstractSearchRunner):
         else:
             request.callSetIds = args.callSetIds.split(",")
         setCommaSeparatedAttribute(request, args, 'variantSetIds')
-        self.setHttpClient(request, args)
+        self._setRequest(request, args)
 
     def run(self):
         if self._minimalOutput:
-            self._run('searchVariants', 'id')
+            self._run(self._httpClient.searchVariants, 'id')
         else:
             results = self._httpClient.searchVariants(self._request)
             for result in results:
@@ -180,91 +204,141 @@ class VariantSearchRunner(AbstractSearchRunner):
         print()
 
 
-class ReferenceSetSearchRunner(AbstractSearchRunner):
+class SearchReferenceSetsRunner(AbstractSearchRunner):
     """
     Runner class for the referencesets/search method.
     """
     def __init__(self, args):
-        super(ReferenceSetSearchRunner, self).__init__(args)
+        super(SearchReferenceSetsRunner, self).__init__(args)
         request = protocol.GASearchReferenceSetsRequest()
         setCommaSeparatedAttribute(request, args, 'accessions')
         setCommaSeparatedAttribute(request, args, 'md5checksums')
-        if self.usingWorkaroundsFor(client.HttpClient.workaroundGoogle):
-            # google says assemblyId not a valid field
-            del request.__dict__['assemblyId']
-        self.setHttpClient(request, args)
+        self._setRequest(request, args)
 
     def run(self):
-        self._run('searchReferenceSets', 'id')
+        self._run(self._httpClient.searchReferenceSets, 'id')
 
 
-class ReferencesSearchRunner(AbstractSearchRunner):
+class SearchReferencesRunner(AbstractSearchRunner):
     """
     Runner class for the references/search method
     """
     def __init__(self, args):
-        super(ReferencesSearchRunner, self).__init__(args)
+        super(SearchReferencesRunner, self).__init__(args)
         request = protocol.GASearchReferencesRequest()
         setCommaSeparatedAttribute(request, args, 'accessions')
         setCommaSeparatedAttribute(request, args, 'md5checksums')
-        self.setHttpClient(request, args)
+        self._setRequest(request, args)
 
     def run(self):
-        self._run('searchReferences', 'id')
+        self._run(self._httpClient.searchReferences, 'id')
 
 
-class ReadGroupSetsSearchRunner(AbstractSearchRunner):
+class SearchReadGroupSetsRunner(AbstractSearchRunner):
     """
     Runner class for the readgroupsets/search method
     """
     def __init__(self, args):
-        super(ReadGroupSetsSearchRunner, self).__init__(args)
+        super(SearchReadGroupSetsRunner, self).__init__(args)
         request = protocol.GASearchReadGroupSetsRequest()
         setCommaSeparatedAttribute(request, args, 'datasetIds')
         request.name = args.name
-        self.setHttpClient(request, args)
+        self._setRequest(request, args)
 
     def run(self):
-        self._run('searchReadGroupSets', 'id')
+        self._run(self._httpClient.searchReadGroupSets, 'id')
 
 
-class CallSetsSearchRunner(AbstractSearchRunner):
+class SearchCallSetsRunner(AbstractSearchRunner):
     """
     Runner class for the callsets/search method
     """
     def __init__(self, args):
-        super(CallSetsSearchRunner, self).__init__(args)
+        super(SearchCallSetsRunner, self).__init__(args)
         request = protocol.GASearchCallSetsRequest()
         setCommaSeparatedAttribute(request, args, 'variantSetIds')
         request.name = args.name
-        self.setHttpClient(request, args)
+        self._setRequest(request, args)
 
     def run(self):
-        self._run('searchCallSets', 'id')
+        self._run(self._httpClient.searchCallSets, 'id')
 
 
-class ReadsSearchRunner(AbstractSearchRunner):
+class SearchReadsRunner(AbstractSearchRunner):
     """
     Runner class for the reads/search method
     """
+    class GASearchReadsRequestGoogle(protocol.ProtocolElement):
+
+        __slots__ = ['end', 'pageSize', 'pageToken', 'readGroupIds',
+                     'referenceName', 'start']
+
+        def __init__(self):
+            self.end = None
+            self.pageSize = None
+            self.pageToken = None
+            self.readGroupIds = []
+            self.referenceName = None
+            self.start = 0
+
     def __init__(self, args):
-        super(ReadsSearchRunner, self).__init__(args)
+        super(SearchReadsRunner, self).__init__(args)
         request = protocol.GASearchReadsRequest()
+        if self.usingWorkaroundsFor(client.HttpClient.workaroundGoogle):
+            # google says referenceId not a valid field
+            request = self.GASearchReadsRequestGoogle()
         setCommaSeparatedAttribute(request, args, 'readGroupIds')
         request.start = args.start
         request.end = args.end
         request.referenceId = args.referenceId
         request.referenceName = args.referenceName
-        if self.usingWorkaroundsFor(client.HttpClient.workaroundGoogle):
-            # google says referenceId not a valid field
-            del request.__dict__['referenceId']
-        self.setHttpClient(request, args)
+        self._setRequest(request, args)
 
     def run(self):
-        self._run('searchReads', 'id')
+        self._run(self._httpClient.searchReads, 'id')
 
 
-class BenchmarkRunner(VariantSearchRunner):
+class ListReferenceBasesRunner(AbstractSearchRunner):
+    """
+    Runner class for the references/{id}/bases method
+    """
+    def __init__(self, args):
+        super(ListReferenceBasesRunner, self).__init__(args)
+        request = protocol.GAListReferenceBasesRequest()
+        request.start = args.start
+        request.end = args.end
+        self._id = args.id
+        self._setRequest(request, args)
+
+    def run(self):
+        method = self._httpClient.listReferenceBases
+        for base in method(self._request, self._id):
+            print(base.sequence)
+
+
+class GetReferenceSetRunner(AbstractGetRunner):
+    """
+    Runner class for the referencesets/{id} method
+    """
+    def __init__(self, args):
+        super(GetReferenceSetRunner, self).__init__(args)
+
+    def run(self):
+        self._run(self._httpClient.getReferenceSet)
+
+
+class GetReferenceRunner(AbstractGetRunner):
+    """
+    Runner class for the references/{id} method
+    """
+    def __init__(self, args):
+        super(GetReferenceRunner, self).__init__(args)
+
+    def run(self):
+        self._run(self._httpClient.getReference)
+
+
+class BenchmarkRunner(SearchVariantsRunner):
     """
     Runner class for the client side benchmarking. This is intended to give
     rough figures on protocol throughput on the server side over various
@@ -333,6 +407,15 @@ def addEndArgument(parser):
     parser.add_argument(
         "--end", "-e", default=1, type=int,
         help="The end of the search range (exclusive).")
+
+
+def addIdArgument(parser):
+    parser.add_argument("--id", default=None, help="The id of the object")
+
+
+def addGetArguments(parser):
+    addIdArgument(parser)
+    addUrlArgument(parser)
 
 
 def addUrlArgument(parser):
@@ -413,7 +496,7 @@ def client_main(parser=None):
         "variants-search",
         description="Search for variants",
         help="Search for variants.")
-    vsParser.set_defaults(runner=VariantSearchRunner)
+    vsParser.set_defaults(runner=SearchVariantsRunner)
     addUrlArgument(vsParser)
     addVariantSearchOptions(vsParser)
 
@@ -422,7 +505,7 @@ def client_main(parser=None):
         "variantsets-search",
         description="Search for variantSets",
         help="Search for variantSets.")
-    vssParser.set_defaults(runner=VariantSetSearchRunner)
+    vssParser.set_defaults(runner=SearchVariantSetsRunner)
     addUrlArgument(vssParser)
     addPageSizeArgument(vssParser)
     addDatasetIdsArgument(vssParser)
@@ -432,7 +515,7 @@ def client_main(parser=None):
         "referencesets-search",
         description="Search for referenceSets",
         help="Search for referenceSets")
-    rssParser.set_defaults(runner=ReferenceSetSearchRunner)
+    rssParser.set_defaults(runner=SearchReferenceSetsRunner)
     addUrlArgument(rssParser)
     addPageSizeArgument(rssParser)
     addAccessionsArgument(rssParser)
@@ -446,7 +529,7 @@ def client_main(parser=None):
         "references-search",
         description="Search for references",
         help="Search for references")
-    rsParser.set_defaults(runner=ReferencesSearchRunner)
+    rsParser.set_defaults(runner=SearchReferencesRunner)
     addUrlArgument(rsParser)
     addPageSizeArgument(rsParser)
     addAccessionsArgument(rsParser)
@@ -457,7 +540,7 @@ def client_main(parser=None):
         "readgroupsets-search",
         description="Search for readGroupSets",
         help="Search for readGroupSets")
-    rgsParser.set_defaults(runner=ReadGroupSetsSearchRunner)
+    rgsParser.set_defaults(runner=SearchReadGroupSetsRunner)
     addUrlArgument(rgsParser)
     addPageSizeArgument(rgsParser)
     addDatasetIdsArgument(rgsParser)
@@ -468,7 +551,7 @@ def client_main(parser=None):
         "callsets-search",
         description="Search for callSets",
         help="Search for callSets")
-    csParser.set_defaults(runner=CallSetsSearchRunner)
+    csParser.set_defaults(runner=SearchCallSetsRunner)
     addUrlArgument(csParser)
     addPageSizeArgument(csParser)
     addNameArgument(csParser)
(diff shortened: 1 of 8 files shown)


ga4gh/client.py
```diff
@@ -15,13 +15,26 @@
 requests.packages.urllib3.disable_warnings()
 
 
-class HTTPClient(object):
-
-    workaroundGoogle = 'google'
-
+class HttpClient(object):
```
jeromekelleher: PEP8 says "Note: When using abbreviations in CapWords, capitalize all the letters of the abbreviation. Thus HTTPServerError is better than HttpServerError."

Might as well stick to PEP8 for this, unless there's a compelling reason.

ga4gh/client.py
```diff
@@ -15,13 +15,26 @@
...
     """
-    Simple HTTP client for the GA4GH protocol.
+    GA4GH Http Client
     """
+    workaroundGoogle = 'google'
+
+    class RunRequestRequest(object):
```
jeromekelleher: This name seems confusing to me. `RunRequestContainer` maybe?

ga4gh/client.py
```diff
@@ -29,7 +42,88 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
...
+        return response
+
+    def _deserializeResponse(self, response, protocolClass):
+        jsonResponseString = response.text
+        self._updateBytesRead(jsonResponseString)
+        self._debugResponse(jsonResponseString)
+        responseObject = protocolClass.fromJSONString(jsonResponseString)
+        return responseObject
+
+    def runRequestSingle(self, rrr):
```
jeromekelleher: Ah, this is a `RunRequestRequest`, so `rrr`. Would `requestContainer` be clearer?

jeromekelleher: Or `requestWrapper` maybe?

jeromekelleher:
This looks good @dcolligan, some very useful abstractions here to simplify running all these different queries. I think we can make the naming a little bit better, but otherwise I like it a lot. I don't have time to play with this properly right now, but I'll try and get to it in the next couple of days and get back to you with some more comments.

ga4gh/client.py
dcolligan: I'd like to advocate disobeying this part of PEP8.  Noun boundaries are difficult to parse with all-caps acronyms, especially with a variable name that has multiple acronyms.  For instance:

HttpClient is easier to read than HTTPClient
HttpApiClient is easier to read than HTTPAPIClient

jeromekelleher: I tend to agree, but have tried to resist the urge to add my own preferences where they contradict PEP8. Want to make a PR against the STYLE.md with the argument, and update any existing code at the same time? Let's keep it as HTTPClient here for now to avoid confusing issues.

dcolligan: Added the STYLE.md change to this PR.

Naburimannu:
-1 on the latest commit, I'd like to consider the style guide change separately from this PR.

jeromekelleher:
I agree with @Naburimannu --- please take out the last commit and submit it as a standalone PR. We need to make sure that all style choices are made in a transparent manner.

dcolligan:
Alright, I moved the style change request to #123

jeromekelleher:
Excellent, thanks. Can we change back to HTTPClient for now also? We'll need to do a project wide search-and-replace for these strings if it's accepted anyway.

dcolligan:
Ok, changed the name back to HTTPClient

ga4gh/client.py
```diff
@@ -29,7 +54,89 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
...
+                json.loads(jsonString), sort_keys=True, indent=4)
+            print(pp)
+
+    def _checkStatus(self, response):
+        if response.status_code != 200:
+            print(response.text)
+            raise Exception("Url {0} had status_code {1}".format(
+                response.url, response.status_code))
+
+    def _constructUrl(self, url, id=None):
```
jeromekelleher: s/id/id_ here and below.

ga4gh/client.py
```diff
@@ -29,7 +54,89 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
...
+        else:
+            return fullUrl.format(id=id)
+
+    def _updateBytesRead(self, jsonString):
+        self._bytesRead += len(jsonString)
+
+    def _doRequest(self, request, url, requestMethod, id):
+        fullUrl = self._constructUrl(url, id=id)
+        params = self._addAuth()
+        method = getattr(requests, requestMethod)
```
jeromekelleher: method isn't used.

ga4gh/client.py
```diff
@@ -29,7 +54,89 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
...
+            print(pp)
+
+    def _checkStatus(self, response):
+        if response.status_code != 200:
+            print(response.text)
+            raise Exception("Url {0} had status_code {1}".format(
+                response.url, response.status_code))
+
+    def _constructUrl(self, url, id=None):
+        fullUrl = posixpath.join(self._urlPrefix, url).format(id=id)
```
jeromekelleher: Aren't we doing the same thing twice here? It would be clearer I think if we left out _construcURL entirely, as it's really just a one-liner.

ga4gh/client.py
```diff
@@ -29,7 +54,89 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
         self._workarounds = workarounds
         self._key = key
 
-    def runRequest(self, request, url, protocolClass, listAttr):
+    # TODO temporary auth solution
+    def _addAuth(self):
```
jeromekelleher: _getAuth() would be more accurate here since we're returning something and not updating state.

ga4gh/client.py
```diff
@@ -29,7 +54,89 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
...
+            headers = {"Content-type": "application/json"}
+            jsonString = request.toJSONString()
+            response = requests.post(
+                fullUrl, jsonString, headers=headers, verify=False,
+                params=params)
+        elif requestMethod == 'get':
+            if request is not None:
+                requestDict = request.toJSONDict()
+                params.update(requestDict)
+            response = requests.get(
```
jeromekelleher: Join these two lines.

ga4gh/client.py
```diff
@@ -29,7 +54,89 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
...
+        fullUrl = posixpath.join(self._urlPrefix, url).format(id=id)
+        if id is None:
+            return fullUrl
+        else:
+            return fullUrl.format(id=id)
+
+    def _updateBytesRead(self, jsonString):
+        self._bytesRead += len(jsonString)
+
+    def _doRequest(self, request, url, requestMethod, id):
```
jeromekelleher: Why not make _doRequest take a requestContainer as a parameter?

ga4gh/client.py
```diff
@@ -38,31 +145,60 @@ def runRequest(self, request, url, protocolClass, listAttr):
...
+            # yield the results
+            if requestContainer.listAttr is None:
+                yield responseObject
+            else:
+                obj = getattr(responseObject, requestContainer.listAttr)
+                for extract in obj:
+                    yield extract
+
+            # update the pageToken and notDone
+            if hasattr(responseObject, 'nextPageToken'):
```
jeromekelleher: This feels a bit wrong... I wonder if there's a different abstraction to RequestContainer we can use? Instead of just basically being a named tuple, it might have some operations which would simplify things a bit here? Having said that, it's a complicated mess you're trying to model here, and I don't have any better ideas, so there's nothing wrong with this if you're happy with it.

ga4gh/client.py
```diff
@@ -38,31 +145,60 @@ def runRequest(self, request, url, protocolClass, listAttr):
...
+                requestContainer.request.pageToken = \
+                    responseObject.nextPageToken
+                notDone = responseObject.nextPageToken is not None
+            else:
+                notDone = False
+
+    def findReferenceSet(self, id_):
```
jeromekelleher: Would `getReferenceSet` would be a little clearer here (and for the other GET methods)?

ga4gh/cli.py
```diff
@@ -223,23 +261,77 @@ class ReadsSearchRunner(AbstractSearchRunner):
...
-        if self.usingWorkaroundsFor(client.HTTPClient.workaroundGoogle):
-            # google says referenceId not a valid field
-            del request.__dict__['referenceId']
         self.setHttpClient(request, args)
 
     def run(self):
         self._run('searchReads', 'id')
 
 
+class ReferenceBasesSearchRunner(AbstractSearchRunner):
```
jeromekelleher: I think it would be clearer to use the patterns `SearchXRunner`, and `GetXRunner` to map to the search and get API methods.

ga4gh/cli.py
```diff
@@ -223,23 +261,77 @@ class ReadsSearchRunner(AbstractSearchRunner):
...
+
+
+class ReferenceSetFindRunner(AbstractFindRunner):
+    """
+    Runner class for the referencesets/{id} method
+    """
+    def __init__(self, args):
+        super(ReferenceSetFindRunner, self).__init__(args)
+
+    def run(self):
```
jeromekelleher: Do we need this indirection? Why not just have
```python
def run(self):
      self._httpClient.getReferenceSet(self._id)
```
If we do need to use the indirection, I'd rather use
```python
def run(self):
      self._run(self._httpClient.getReferenceSet)
```
and save on the strings and getattr calls.

jeromekelleher:
Phew, this is a mammoth patch @dcolligan! I haven't gone through all of it yet, but it's adding a huge chunk of functionality client-side. You've done a great job making a slightly messy protocol manageable. I wonder if we can simplify things a bit though. I've found myself chasing a lot of links in the chain to find out what is actually happening. It's probably worth going through things again and seeing if they can be streamlined a bit I think.

I'll try and get through the rest as soon as I can.

ga4gh/client.py
dcolligan: The only thing I was notably unhappy with in my refactor is the need to create runRequestSingle instead of just having a general runRequest method.  But python's insistence that methods that have a yield not have any kind of return X statement precludes that, AFAICT.  (Unless we want to return a generator that yields one item for get* calls, but that just feels like an abuse of semantics.)

dcolligan:
@jeromekelleher Pushed some more commits which addressed your concerns

ga4gh/cli.py
```diff
@@ -90,23 +90,48 @@ def usingWorkaroundsFor(self, workaround):
...
+
+
+class AbstractSearchRunner(AbstractQueryRunner):
+    """
+    Abstract base class for search runner classes
+    """
+    def __init__(self, args):
+        super(AbstractSearchRunner, self).__init__(args)
+
     def setHttpClient(self, request, args):
```
jeromekelleher: I think this should be refactored a bit:
1. Move the HTTPClient instantiation into the constructor of AbstractQueryRunner - this is shared by all the runners. Also, verbosity should be set in there.
2. Change `setHttpClient` to `_setRequest`.

ga4gh/cli.py
```diff
@@ -153,94 +178,159 @@ def __init__(self, args):
...
+            self.pageSize = None
+            self.pageToken = None
+
     def __init__(self, args):
-        super(ReferenceSetSearchRunner, self).__init__(args)
-        request = protocol.GASearchReferenceSetsRequest()
-        setCommaSeparatedAttribute(request, args, 'accessions')
-        setCommaSeparatedAttribute(request, args, 'md5checksums')
+        super(SearchReferenceSetsRunner, self).__init__(args)
         if self.usingWorkaroundsFor(client.HTTPClient.workaroundGoogle):
```
jeromekelleher: In general, for this type of situation I like the form
```python
request = protocol.GASearchReferenceSetsRequest()
if self.usingWorkaroundsFor(client.HTTPClient.workaroundGoogle):
    request = self.GASearchReferenceSetsRequestGoogle()
```
I think it communicates the default state a bit better, and it saves a line of code. OK, we allocate an object that we don't use, but this is immaterial in the larger scheme of things.

Feel completely free to ignore this, I'm just making a comment!

ga4gh/client.py
```diff
@@ -29,7 +54,89 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
...
+        if self._debugLevel > 1:
+            # TODO use a logging output and integrate with HTTP client more
+            # nicely... also, logging / error handling in general
+            print("json response:")
+            pp = json.dumps(
+                json.loads(jsonString), sort_keys=True, indent=4)
+            print(pp)
+
+    def _checkStatus(self, response):
+        if response.status_code != 200:
```
jeromekelleher: It would be nice to use `requests.codes.ok` here instead of 200. See http://docs.python-requests.org/en/latest/user/quickstart/#response-status-codes

ga4gh/client.py
```diff
@@ -29,7 +54,89 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
...
+            print(response.text)
```
jeromekelleher: Add a TODO here for making a custom exception. Or we could use `requests.raise_for_status()` and use their exceptions for HTTP events.

ga4gh/client.py
```diff
@@ -16,12 +16,37 @@
...
     """
+    workaroundGoogle = 'google'
+
+    class RunRequestContainer(object):
+        """
+        A container class for arguments to runRequest
+        """
+        def __init__(
+            self, request=None, url=None, protocolClass=None,
+                listAttr=None, requestMethod='post', id_=None):
```
jeromekelleher: I think the term `request` is being heavily overloaded here to mean several different things. I'd suggest

- requestMethod -> httpMethod
- protocolClass -> protocolResponseClass
- request -> protocolRequest

ga4gh/client.py
```diff
@@ -16,12 +16,37 @@
...
+            self.request = request
+            self.url = url
+            self.protocolClass = protocolClass
+            self.listAttr = listAttr
+            self.requestMethod = requestMethod
+            self.id_ = id_
```
jeromekelleher: There's a tricky style issue here. Should we have `self.id_` or `self.id`? Having an `id` attribute on an object is fine as far as I know (and we're doing it plenty of other places), so I'd probably prefer `self.id`. Definitely variables should be called `id_` though.

ga4gh/client.py
```diff
@@ -29,7 +54,89 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
...
+        Used for requests where a single object is the expected response.
+        """
+        # do the request
+        response = self._doRequest(requestContainer)
+
+        # deserialize json response
+        responseObject = self._deserializeResponse(
+            response, requestContainer.protocolClass)
+
+        # return the object
```
jeromekelleher: Is this test necessary? Shouldn't all the things we use `runRequestSingle` on just have a single object by definition?

ga4gh/client.py
```diff
@@ -29,7 +54,89 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
...
+            headers = {"Content-type": "application/json"}
+            jsonString = requestContainer.request.toJSONString()
+            response = requests.post(
+                fullUrl, jsonString, headers=headers, verify=False,
+                params=params)
+        elif requestContainer.requestMethod == 'get':
+            if requestContainer.request is not None:
+                requestDict = requestContainer.request.toJSONDict()
+                params.update(requestDict)
+            response = requests.get(fullUrl, params=params)
```
jeromekelleher: We should have a final `else` here with an assert(False) in it, in case httpMethod is neither 'post' or 'get'. I'm a bit worried about these hard-coded strings being used for control flow. Can we used some constants maybe?

ga4gh/client.py
```diff
@@ -38,31 +145,53 @@ def runRequest(self, request, url, protocolClass, listAttr):
...
+
+    def getReferenceSet(self, id_):
+        """
+        Returns a referenceSet from the server
+        """
+        requestContainer = self.RunRequestContainer(
+            url="referencesets/{id}",
+            protocolClass=protocol.GAReferenceSet,
+            requestMethod='get', id_=id_)
+        return self.runRequestSingle(requestContainer)
```
jeromekelleher: I think we're overcomplicating things here by pushing too much up into _doRequest, and deciding afterwards what we want to run based on what's in the container object. What about this:
```python
def getReferenceSet(self, id_):
     return self.runGetRequest("referencesets", id_, protocol.GAReferenceSet)
```
`runGetRequest` is then much simpler, and this avoids all the hard-coded strings.

ga4gh/client.py
```diff
@@ -71,57 +200,64 @@ def searchVariants(self, request):
         is handled transparently, so that several HTTP requests may be made
         while this method executes.
         """
-        return self.runRequest(
-            request, "variants/search", protocol.GASearchVariantsResponse,
-            "variants")
+        requestContainer = self.RunRequestContainer(
+            request, "variants/search",
+            protocol.GASearchVariantsResponse, "variants")
+        return self.runRequest(requestContainer)
```
jeromekelleher: So, here we'd want 
```python
def searchVariants(self, request):
     return self.runPostRequest(request, "variants/search", protocol.GASearchVariantsResponse, "variants")
```

jeromekelleher:
Ok, this is making more sense to me now. My comments at the end are sort-of contradicting the comments at the start, as I've changed my mind about the request container abstraction. I think the complexity that we're seeing and the difficulty in backtracing exactly what's happening in a given method is due to the container abstraction and trying to do all of the http requesting logic in a single method. I think a better option would be to write different `run` methods for the distinct patterns we see (which there are three I think? GET a single object, POST a list of objects and POST a single object?), and compose their functionality from the building blocks you already have lying around like `deserialise` and so on. I think this will make the code a lot simpler and easier to follow.

ga4gh/client.py
jeromekelleher: Or maybe
```
def searchVariants(self, request):
     return self.runSearchRequest(request, "variants", protocol.GASearchVariantsResponse)
```
We should exploit the common patterns in the various protocol methods as much as possible. We know we're running a search request on 'variants' so we can build the URL appropriately and also know that the list attribute should be `variants`.

dcolligan:
We could do that, but why not have a general method that can do everything (except for very well-defined and reoccurring cases like GET/single-object)?  Adding additional client methods will likely necessitate a refactor when they don't fall into one of our pre-defined buckets.  Plus it's another method(s) to test.

Before the list reference bases endpoint, I separated everything into a GET/single-object bucket and a POST/multiple-object bucket, but then that broke down since list reference bases is GET/multiple-object.

jeromekelleher:
I agree there's a tension between making it easy to test and making it easy to read, but I think that easy to read must win every time. I just can't tell easily what's going to happen in the big complicated _doRequest method for a given endpoint, and this is definitely an alarm bell. I should be able to start at the bottom of the call chain, press # or ctrl-] a few times (assuming vim) and know exactly what is going to happen, without having to remember exactly what parameters I'm carrying back up the chain.

ga4gh/client.py
dcolligan: Yeah, it's actually fine to have a obj.id.  I was just thinking it wasn't because my syntax highlighter highlights it.  It is not to be confused with python's builtin id(obj) function.

dcolligan: Let's just keep id_, since it keeps one from mindlessly writing `id = something` or `self.id = id` or some such and overriding the builtin id function.

diekhans: Danny Colligan <notifications@github.com> writes:
> Let's just keep id_, since it keeps one from mindlessly writing id = something
> or self.id = id or some such and overriding the builtin id function.

  self.id_ = id

;-)

ident ?

ga4gh/client.py
dcolligan: Python doesn't support constants

ga4gh/client.py
dcolligan: The point of the RunRequestContainer was to decouple the cli methods from the client methods by providing an object that any cli method could create and any client method could use.  I'm not sure tightly coupling them again would be a good idea.

ga4gh/client.py
dcolligan: Same decoupling comment as above.

jeromekelleher:
I disagree I'm afraid. Here's my (incomplete, untested and un-pep8 styled) version:
```python
    def runGetRequest(self, protocolRequest, objectName, responseClass, id_):
        url = "{objectName}/{id}"
        fullUrl = posixpath.join(
            self._urlPrefix, url).format(id=id_, objectName=objectName)
        params = self._getAuth()
        requestDict = protocolRequest.toJSONDict()
        params.update(requestDict)
        response = requests.get(fullUrl, params=params)
        return self._deserializeResponse(response, responseClass)

    def getReferenceSet(self, id_):
        """
        Returns a referenceSet from the server
        """
        return self.runGetRequest(protocolRequest, "referencesets",
                protocol.GAReferenceSet, id_)

    def getReference(self, id_):
        """
        Returns a reference from the server
        """
        return self.runGetRequest(protocolRequest, "references",
                protocol.GAReference, id_)

    def listReferenceBases(self, protocolRequest, id_):
        """
        Returns an iterator over the bases from the server
        """
        return self.runListRequest(protocolRequest, "references/{id}/bases",
                protocol.GAListReferenceBasesResponse, id_)

    def searchVariants(self, protocolRequest):
        """
        Sends the specified GASearchVariantsRequest to the server and returns
        an iterator over the returned set of GAVariant objects. Result paging
        is handled transparently, so that several HTTP requests may be made
        while this method executes.
        """
        return self.runSearchRequest(protocolRequest, "variants",
                protocol.GASearchVariantsResponse, "variants")

    def searchVariantSets(self, protocolRequest):
        """
        Returns an iterator over the VariantSets from the server.
        """
        return self.runSearchRequest(protocolRequest, "variantsets",
                protocol.GASearchVariantSetsResponse, "variantSets")

    def searchReferenceSets(self, protocolRequest):
        """
        Returns an iterator over the ReferenceSets from the server.
        """
        return self.runSearchRequest(protocolRequest, "referencesets",
                protocol.GASearchReferencesRequest, "referenceSets")

    def searchReferences(self, protocolRequest):
        """
        Returns an iterator over the References from the server
        """
        return self.runSearchRequest(protocolRequest, "references",
                protocol.GASearchReferencesRequest, "references")

    def searchCallSets(self, protocolRequest):
        """
        Returns an iterator over the CallSets from the server
        """
        return self.runSearchRequest(protocolRequest, "callsets",
                protocol.GASearchCallSetsRequest, "callSets")

    def searchReadGroupSets(self, protocolRequest):
        """
        Returns an iterator over the ReadGroupSets from the server
        """
        return self.runSearchRequest(protocolRequest, "readgroupssets",
                protocol.GASearchReadGroupSetsResponse, "readGroupSets")

    def searchCallSets(self, protocolRequest):
        """
        Returns an iterator over the CallSets from the server
        """
        return self.runSearchRequest(protocolRequest, "callsets",
                protocol.GASearchCallSetsRequest, "callSets")

    def searchReads(self, protocolRequest):
        """
        Returns an iterator over the Reads from the server
        """
        return self.runSearchRequest(protocolRequest, "reads",
                protocol.GASearchReadsRequest, "alignments")
```

This is a _lot_ less code, and seems a lot easier to read to me. OK, so runListRequest and runSearchRequest will share some code, but we can compose their functionality from smaller pieces (like deseraliseResponse), and then they are nice and simple.

pashields:
Wow, this is a big patch. I haven't been following this patch as it's been iterated on, so take what I have to add with a grain of salt. My only comments are on the client.py changes.

The greatest argument I have against the generic request container is the fact that it's not totally generic. At https://github.com/dcolligan/server/blob/76_client_endpoints_2/ga4gh/client.py#L91 I see an if statement depending on the type of the request. To me, that seems to say that we have not appropriately encapsulated all the behavior here. It's always preferable not to duplicate code, but I'm not convinced that we haven't shared code here simply by making multiple distinct types into one here. There is probably a more polymorphic solution here and I can imagine that being worthwhile to pursue. Put another way, if you are going to have containers, I would have them contain all behavior specific to them.

Another (more off the beaten path) way to think about this whole problem is to switch from thinking about a monolithic class based approach to a more middleware-oriented approach. Put another way, instead of gathering all the logic into a single function and some helpers, can we decompose the problem a bit more. We know we have to create a request -> execute that request -> return that response. Perhaps we can break that up even more to
```
[create basic request -> add auth -> add path -> add params -> ...] -> [execute request -> check status -> ...] -> [deserialize response from json -> deserialize to GA objects -> convert to final return format -> ...]
```
I think you've made great progress toward doing something like this just by breaking behavior out into functions, but you could go further and actually describe their execution flow as data rather than code. The biggest benefit of this is that adding new functionality typically does not increase complexity significantly. It can be difficult to do this while achieving the desired high cohesion/low coupling.

You may have already considered a lot of this and rejected it. If any of it sounds useful but really really obtuse, let me know and I can try to spend some more time with the patch and formulate some more concrete ideas.

dcolligan:
Alright, pushed a new change that implements Jerome's design that we talked about

jeromekelleher:
Excellent! Thanks @dcolligan, this looks great. I've been playing with it a bit, and I think we can simplify it down a bit and also get all of our http requests going through one place. Here's one way we might do it:
```python
    def _doRequest(self, httpMethod, url, protocolResponseClass,
                   httpParams={}, httpData=None):
        """
        TODO document this central method.
        """
        headers = {"Content-type": "application/json"}
        params = self._getAuth()
        params.update(httpParams)
        response = requests.request(
            httpMethod, url, params=params, data=httpData, headers=headers)
        self._checkStatus(response)
        return self._deserializeResponse(response, protocolResponseClass)

    def runSearchRequest(self, protocolRequest, objectName,
                         protocolResponseClass, listAttr):
        """
        Runs the specified request at the specified objectName and instantiates
        an object of the specified class. We yield each object in listAttr.
        If pages of results are present, repeat this process until the
        pageToken is null.
        """
        fullUrl = posixpath.join(self._urlPrefix, objectName + '/search')
        notDone = True
        while notDone:
            data = protocolRequest.toJSONString()
            responseObject = self._doRequest('POST', fullUrl,
                    protocolResponseClass, httpData=data)
            for extract in getattr(responseObject, listAttr):
                yield extract
            notDone = self._updateNotDone(responseObject, protocolRequest)

    def runListRequest(self, protocolRequest, url,
                       protocolResponseClass, id_):
        """
        Asks the server to list objects of type protocolResponseClass and
        returns an iterator over the results.
        """
        fullUrl = posixpath.join(self._urlPrefix, url).format(id=id_)
        notDone = True
        while notDone:
            requestDict = protocolRequest.toJSONDict()
            responseObject = self._doRequest('GET',
                    fullUrl, protocolResponseClass, requestDict)
            yield responseObject
            notDone = self._updateNotDone(responseObject, protocolRequest)

    def runGetRequest(self, objectName, protocolResponseClass, id_):
        """
        Requests an object from the server and returns the object of
        type protocolResponseClass that has id id_.
        Used for requests where a single object is the expected response.
        """
        url = "{objectName}/{id}"
        fullUrl = posixpath.join(
            self._urlPrefix, url).format(id=id_, objectName=objectName)
        return self._doRequest('GET', fullUrl, protocolResponseClass)
```
This needs some PEP8ing and more testing, but I've run it though your Google examples and it seems to work. What do you think?

dcolligan:
Looks good, except for that the headers should only be set if httpData is not None.  I'll incorporate and resubmit the branch.

dcolligan:
Alright, pushed a new (squashed) commit.

jeromekelleher:
Superb! This looks really great now, thanks for all the hard work @dcolligan. I think we're ready to merge.

First, we need to rebase and fix some conflicts (toJSONString etc). Also, you seem to have picked up a bunch of random commits on .travis.yml, STYLE.md and .gitignore which should be removed. After that, I'm +1.

dcolligan:
The `STYLE.md` change is unintentional but everything else is intentional.  The change in `.travis.yml` is just a comment to notify people about its use in `run_scripts.py`.  The `.gitignore` change deletes extraneous lines.

I just pushed a new commit rebased on top of master which rectified the `STYLE.md` deletion and also addresses #132

jeromekelleher:
Great, let's merge this beast! 

@pashields, @Naburimannu, would you mind having a quick look at this?

pashields:
+1 looks good to me.

ga4gh/cli.py
```diff
@@ -76,13 +76,16 @@ def server_main(parser=None):
 ##############################################################################
 
 
-class AbstractSearchRunner(object):
+class AbstractQueryRunner(object):
     """
-    Abstract base class for search runner classes
+    Abstract base class for runner classes
```
Naburimannu: In what sense do you mean "abstract" here? This isn't in the sense of https://www.python.org/dev/peps/pep-3119/, or I think in the C++/Java sense...

dcolligan: Sure it is, we just don't have a compiler to enforce it ;)

In any case, the class name isn't really part of the PR

jeromekelleher: I'm with @dcolligan here; this is Python, we don't need a compiler to tell us what's good and bad!

jeromekelleher:
Merged! Thanks @dcolligan, this is a huge step forward client side.

PR #109, opened 2015-01-30
Trimmed requirements.txt and updated travis conf.

jeromekelleher:
This cuts down the list of requirements to closer to what we really need, and also removes the needlessly specific version requirements. (If we need a >= version at some point we should add it then.)

Also uses this file to install our requirements on Travis.


dcolligan:
I'm +1 if this passes CI -- yay for removing cruft and for eating our own dogfood -- but, out of curiosity, how did you come up with this new requirements.txt list?

Naburimannu:
+1

jeromekelleher:
I just deleted stuff I either didn't recognise, or seemed completely mad (like numpy!). I've tried it in a new virtualenv, and it seems to work OK. I think Travis has a bunch of stuff installed by default which would take ages to install every time someone wanted it. Looks like numpy is a popular enough requirement for this.

dcolligan:
People are waiting on this, and it's a small enough change, so I'll merge.

Issue #107, opened 2015-01-30
variants/search client end point needs more detail


jeromekelleher:
With the changes in #102, we lost the detailed output for the variants/search function. We need to get this back.

In general, it would be good to have an option to provide either short or detailed output for each object. These do not need to be filled in for all objects to close this issue.

dcolligan:
Fixed in #110

PR #110, opened 2015-01-30
Adds optional extended outputFormat

dcolligan:
Issue #107

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -95,6 +95,7 @@ class AbstractSearchRunner(object):
         Sets the _httpClient and other common attributes
         """
         request.pageSize = args.pageSize
+        self._minimalOutput = args.minimalOutput
         self._request = request
         self._verbosity = args.verbose
         self._httpClient = client.HTTPClient(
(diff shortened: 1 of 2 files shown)


ga4gh/cli.py
```diff
@@ -364,7 +388,10 @@ def client_main(parser=None):
         "--workarounds", "-w", default='', help="The workarounds to use")
     parser.add_argument(
         "--key", "-k", help="The auth key to use")
-    subparsers = parser.add_subparsers(title='subcommands',)
+    parser.add_argument(
+        "--outputFormat", "-o", help="The output format to use",
+        default="minimal", choices=["minimal", "extended"])
```
jeromekelleher: A boolean flag would seem more natural here. What about -O, --minimalOutput, default=False? We keep the semantics we  have from before then and we can just add the --minimalOutput flag for your testing script.

dcolligan:

ga4gh/cli.py
```diff
@@ -153,7 +154,30 @@ def __init__(self, args):
         self.setHttpClient(request, args)
 
     def run(self):
-        self._run('searchVariants', 'id')
+        if self._minimalOutput is True:
```
jeromekelleher: `if self._minimalOutput: ` is the canonical way of testing a boolean.

scripts/google_examples.py
```diff
@@ -33,19 +34,24 @@ def runTests():
...
         }
         cmd = cmdStr.format(**cmdDict)
+        splits = cmd.split()
+        cleanCmd = ' '.join(splits)
         print(separator)
-        print(cmd)
+        print(cleanCmd)
         print(separator)
-        splits = cmd.split()
         proc = subprocess.Popen(splits)
```
jeromekelleher: `subprocess.check_call(splits)` would be a slightly cleaner way of doing this.

jeromekelleher:
+1 modulo nits. I think we should follow this up with something a bit more generic which gives useful output for all the other commands when #108 is done. Thanks @dcolligan.

scripts/google_examples.py
dcolligan: Yes, but that way suppresses all the output of the command, which we don't want.

dcolligan: Nevermind, my mistake.

dcolligan:
@jeromekelleher pushed commit that addressed nits

Issue #111, opened 2015-01-30
Frontend should display something


melaniedc:
Maybe have a basic input form and/or usage instructions on "/", and something to display (or allow download of) results for the implemented methods.

Thoughts? What would be useful to have here?

dcolligan:
+1 for the general idea, but I think maybe it would be more useful to have status/info about the server, analogous to a phpinfo page

jeromekelleher:
I'm +0 on the idea, but not sure what we could usefully display right now. I guess a phpinfo type display would be useful. I'm mildly negative about having any HTML output at all TBH, it seems like a blurring of the lines between the role of the web backend and HTML frontend. I'm happy to change my mind if other people feel strongly about this though!

jeromekelleher:
OK, @dcolligan's POC has changed my mind - we definitely want this! Some stuff we might show:

- List of URLs
- Summary of datasets being served 
- Traffic stats
- Uptime

These are just some ideas. I think the first two are probably the most useful for a user who has just deployed a server and wants to check that they're serving the data they expect.

dcolligan:
@jeromekelleher could you elaborate a little more about what you want re: "Summary of datasets being served" and "Traffic stats"?  AFAICT the backend does not now report what datasets are being served... what interface would you want to expose that information?  What traffic stats would you want to keep track of?

jeromekelleher:
Dunno, these are just vague ideas for discussion. For traffic stats, we could keep track of the number of bytes served/number of clients seen? This will probably cause trouble when we're dealing with multiple threads though, and the parent HTTP server will most likely do this anyway, so it's probably not worth the bother.

Re dataset information, I think it would be good to give a quick summary of the datasets and references that we're serving for user debugging purposes. We can just query the backend directly to determine this information. Of course we'll need to actually implement references and datasets first, so there's no point in doing anything specific with this for a while.

dcolligan:
Alright, I'll punt on the datasets and traffic stats for now.  We can file additional issues for the page at / as they come up.

In the meantime, I'd like to get #112 into master since making this change also forced me to update the exception handling code.

dcolligan:
Resolved in #112

@melaniedc or anyone else please file additional issues if you have ideas for adding onto the page

PR #112, opened 2015-01-30
Adds html info page served at /

dcolligan:
Displays the following info:
```
GA4GH reference server 0.5.1

Operations available

Method	Path
POST	/callsets/search
POST	/readgroupsets/search
POST	/reads/search
GET	/references/<id>
GET	/references/<id>/bases
POST	/references/search
GET	/referencesets/<id>
POST	/referencesets/search
POST	/variants/search
POST	/variantsets/search

Uptime

Running since 18 minutes ago (03:04:56 03 Feb 2015)
```

Issue #111

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -9,15 +9,54 @@ from __future__ import unicode_literals
 
 import os
 import traceback
+import datetime
 
 import flask
 import flask.ext.api as api
 import flask.ext.cors as cors
+import humanize
 
 import ga4gh.frontend_exceptions as frontendExceptions
 import ga4gh.protocol as protocol
 
-app = api.FlaskAPI(__name__)
+
+app = flask.Flask(__name__)
+
+
+class ServerStatus(object):
+
+    def __init__(self):
+        self.startupTime = datetime.datetime.now()
+
+    def getStatusInfo(self):
+        info = {
+            'urls': self._getUrls(),
+            'uptime': self._getUptime(),
+            'version': protocol.version,
+        }
+        return info
+
+    def _getUptime(self):
+        uptime = {
+            'natural': humanize.naturaltime(self.startupTime),
+            'datetime': self.startupTime.strftime("%H:%M:%S %d %b %Y")
+        }
+        return uptime
+
+    def _getUrls(self):
+        urls = []
+        rules = app.url_map.iter_rules()
+        excluded_methods = ('OPTIONS', 'HEAD')
+        excluded_rules = (
+            '/', '/flask-api/static/<path:filename>',
+            '/static/<path:filename>')
+        for rule in rules:
+            for method in rule.methods:
+                if (method not in excluded_methods and
+                        rule.rule not in excluded_rules):
+                        urls.append((rule.rule, method))
+        urls.sort()
+        return urls
 
 
 def configure(config="DefaultConfig", config_file=None):
@@ -28,6 +67,7 @@ def configure(config="DefaultConfig", config_file=None):
     if config_file is not None:
         app.config.from_pyfile(config_file)
     cors.CORS(app, allow_headers='Content-Type')
+    app.serverStatus = ServerStatus()
 
 
 def handleHTTPPost(request, endpoint):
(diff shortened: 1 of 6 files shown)


dcolligan:
Pushed new commit with exception handling improvements, tests

ga4gh/frontend.py
```diff
@@ -9,15 +9,19 @@
 
 import os
 import traceback
+import datetime
 
 import flask
 import flask.ext.api as api
```
jeromekelleher: Don't think we're using this now? I use [flake8](https://pypi.python.org/pypi/flake8) in my editor, it catches this sort of thing nicely.

ga4gh/frontend.py
```diff
@@ -66,9 +75,41 @@ def handle_error(exception):
     return response
```
jeromekelleher: There's a lot of functions here that seem like they should be grouped in a class. What about ```class StatusGenerator(object)``` which has a method `getInfoDict()` or something? We could instantiate this in configure() doing something like
```python
# ....
app.statusGenerator = StatusGenerator()
```
The startupTime can be allocated in the constructor of this. I'd like to avoid module level state where at all possible.

tests/test_exceptions.py
```diff
@@ -9,8 +9,52 @@
...
+
+    def run(self, *args, **kwargs):
+        # patching is required because flask.jsonify throws an exception
+        # if not being called in a running app context
+        dummyFlask = self.DummyFlask()
+        with mock.patch('flask.jsonify', dummyFlask):
+            super(TestExceptionHandler, self).run(*args, **kwargs)
+
+    def testMappedException(self):
+        originalExceptionClass, mappedExceptionClass = \
```
jeromekelleher: What are we testing for here? Why only the first element of `exceptionMap`?

tests/test_exceptions.py
```diff
@@ -34,9 +81,15 @@ def testCodeInvariants(self):
         codes = set()
         for clazz in self._getFrontendExceptionClasses():
```
jeromekelleher: I'd prefer either `cls` or `class_` here --- I don't think we use `clazz` anywhere else.

jeromekelleher:
Looks good @dcolligan, thanks. I've a few small refactoring requests, but I'm +1 after that.

ga4gh/frontend.py
```diff
@@ -66,9 +75,41 @@ def handle_error(exception):
...
 
+def getUrls():
+    urls = []
+    rules = app.url_map.iter_rules()
+    excluded_methods = ('OPTIONS', 'HEAD')
+    excluded_rules = (
+        '/', '/flask-api/static/<path:filename>', '/static/<path:filename>')
+    for rule in rules:
+        for method in rule.methods:
+            if method not in excluded_methods and \
```
jeromekelleher: Use parentheses in favour of line breaks here: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

ga4gh/frontend.py
dcolligan: This particular import is used, but if we want to catch them more generally we should incorporate flake8 or something else (pyilnt?) into the build.  File an issue for that if you wish.

jeromekelleher: Sorry, my bad. I don't think it's worth mandating flake8 checks, as they don't like some of the fancier things we're doing with _protocoldefinitions.py.

dcolligan: Each of these tools can be configured to include or exclude certain files, so I don't think that's a reason not to do it.

tests/test_exceptions.py
dcolligan: Just that the mapping behavior works.  [0] is an arbitrary element... I'll put a comment there.

dcolligan:
@jeromekelleher pushed a commit to address your concerns

ga4gh/frontend.py
jeromekelleher: Good point: see PR #124.

tests/test_exceptions.py
jeromekelleher: Why not loop over all of them? map doesn't return a list in py3, so this will break then.

Naburimannu:
+1

pashields:
+1

I'm happy to merge this but right now github says it won't apply cleanly even though travis passes. Any ideas @dcolligan?

dcolligan:
That just means there are merge conflicts with things that have been committed to master in the interim...  I'll rebase it and then you can merge it from there.

dcolligan:
Ok, just pushed a new commit... this one should pass the tests and then you can merge it after that @pashields

PR #121, opened 2015-02-02
Reference handling

jeromekelleher:
This is a very rough first pass at adding support for references. Any feedback would be great. This PR may-or-may not end up being what we use, so please don't wait on its inclusion to base your work on.

The main thing I'd like to get feedback on is the data input strategy. I've added a list REFERENCE_SETS to the Flask config, which will contain a list of directories containing our ReferenceSets. A reference set is basically a set of indexed FASTA files in a directory. During `frontend.configure()` we call `addReferenceSet(referenceDir)` on the backend. The idea then ultimately would be to do the same thing for datasets. The details of the different backends and file handlers should then all be isolated in backend and the datamodel modules, which is ultimately what we want I think. So, I think we should aim for something like this in configure:
```python
#....
cors.CORS(app, allow_headers='Content-Type')
# Create backend and setup the input data.
app.backend = backend.Backend()
for refDir in app.config["REFERENCE_SETS"]:
    app.backend.addReferenceSet(refDir)
for dataDir in app.config["DATA_SETS"]:
    app.backend.addDataset(dataDir)
```

I don't want to push all this through in this PR, I just wanted to explain the general idea and see if people agree with the strategy. Any thoughts?


ga4gh/backend.py
```diff
@@ -26,6 +27,19 @@ def __init__(self, dataDir, variantSetClass):
                 self._variantSetIdMap[variantSetId] = variantSetClass(
                     variantSetId, relativePath)
         self._variantSetIds = sorted(self._variantSetIdMap.keys())
+        self._nextReferenceSetId = 0
```
Naburimannu: http://ga4gh.org/documentation/api/v0.5.1/ga4gh_api.html#/schema/org.ga4gh.GAReferenceSet defines GAReferenceSet.id as a unique string; you're using sequential integers?

ga4gh/datamodel/references.py
```diff
@@ -6,8 +6,75 @@
...
 
 import ga4gh.protocol as protocol
 
-# TODO use something to translate reference data in FASTA format into
-# GA4GH native objects.
+
+class ReferenceSet(object):
+    """
+    Class representing a reference set in the GA4GH datamodel. A reference
+    set consists of a directory containing a set of indexed FASTA files.
```
Naburimannu: Your comment here describes the class in terms of an implementation, while your comment for Reference describes it in terms of the application domain. This makes it harder for me to understand the relationship.
A scientist shouldn't have to care about the directory full of indexed FASTA; what do *they* think the ReferenceSet is? Or is it just semantic noise we should be hiding from them?

jeromekelleher:
Closing this as #158 needs to be resolved first.

PR #122, opened 2015-02-02
Updated calls so that they are correctly added in Tabix.

kerrydc:
The tests demonstrate coverage and correct parsing of files. All pre-existing tests pass, and now, for the first time ever, the Tabix backend passes the compliance test suite!


ga4gh/datamodel/variants.py
```diff
@@ -298,11 +298,42 @@ def convertVariant(self, record):
...
+        variant.info = {}
+        infoFields = record[7].split(";")
+        for field in infoFields:
+            key, data = field.split("=")
+            dataList = data.split(",")
+            variant.info[key] = dataList
+        formatFields = record[8].split(":")
+        fieldIndicesMap = {}
+        index = 0
+        for field in formatFields:
```
jeromekelleher: ```for index, field in enumerate(formatFields)``` is the more Pythonic approach here.

ga4gh/datamodel/variants.py
```diff
@@ -314,16 +345,78 @@ def getVariants(self, referenceName, startPosition, endPosition,
...
-                yield self.convertVariant(record)
+                yield self.convertVariant(record, headerLine)
 
     def getMetadata(self):
-        # TODO: Implement this
-        ret = []
-        return ret
+        """
+        A VariantSetMetadata represents VCF header information. Since one
+        VairantSet can be comprised of many VCF files, with many duplicate
```
jeromekelleher: s/VairantSet/VariantSet

tests/test_backends.py
```diff
@@ -587,3 +587,73 @@ def testAcrossVariantSets(self):
...
+                                                       self._vcfHeaderLine)
+        self.assertTrue(protocol.GAVariant.validate(variant.toJSONDict()))
+        calls = variant.calls
+        for call in calls:
+            self.assertTrue(protocol.GACall.validate(call.toJSONDict()))
+
+    def testConvertVariant(self):
+        variant = self._tabixVariantSet.convertVariant(self._vcfLine,
+                                                       self._vcfHeaderLine)
+        print(variant)
```
jeromekelleher: Remove the print please.

jeromekelleher:
This is great, thanks @kerrydc! It's a big step forward. The main issue I have with this is that we're doing a lot of VCF parsing, which in my opinion we should avoid _at all costs_. VCF is a very difficult format to parse, and we would be a lot better off if we could get upstream tools to do it for us. Can pysam do any of this for us? I guess we'll be moving over to @bioinformed's interface to the htslib code soon, which will hopefully remove the need to do any real parsing, so this might be a good interim solution.

Also, I prefer @mbaudrey's code in which `convertVCFGenotype` is a function in the `variants` module.

jeromekelleher:
As an update to this, I've been playing with the pysam Tabix APIs. It turns out that there are two different interfaces, demoed here:
```python
"""
Code to demo working with pysam VCF.
"""

import sys

import pysam

def main():
    vcfFile = sys.argv[1]
    # Option 1, use the VCF interface
    # This is in https://github.com/pysam-developers/pysam/blob/master/pysam/cvcf.pyx
    vcf = pysam.VCF()
    vcf.connect(vcfFile)
    print("samples = ", vcf.getsamples())
    print("header = ")
    for k, v in vcf.getinfo().items():
        print("\t", k, "->", v)
    for record in vcf.fetch(b"1"):
        print(
            record.contig, record.pos, record.id, record.ref, record.alt,
            record.qual, record.filter, sep="\t")
        # This causes an error...
        # info = record.info
        for sample in record.samples:
            print("\t", sample, "\t", record[sample])

    # Option 2, use the asVCF interface.
    # This is in https://github.com/pysam-developers/pysam/blob/master/pysam/TabProxies.pyx
    tabixFile = pysam.TabixFile(vcfFile, parser=pysam.asVCF())
    for record in tabixFile.fetch(b"1"):
        print(
            record.contig, record.pos, record.id, record.ref, record.alt,
            record.qual, record.filter, record.info, sep="\t")
        for index, sample in enumerate(vcf.getsamples()):
            print("\t", sample, "\t", record[index])


if __name__ == "__main__":
    main()
```

The first approach is a very complete validating VCF parser, which doesn't seem to be documented or fully maintained. The second is a simple position based approach, which will at least save us the trouble of breaking the records.

As a pragmatic interim approach while @bioinformed's htslib code comes on line, I'd suggest we use a mixture of these two interfaces to the get the information we need. We can get the header information from the first interface, and use the second to get the rows. We'll need to parse the INFO and sample strings by ourselves, that shouldn't be too bad for now.

@bioinformed, @AndreasHeger, it would be great to get your input here. How does this look to you? Is this a reasonable summary of where VCF support is in pysam?

AndreasHeger:
Thanks, your summary is correct.

Option 2 (pysam.Tabixfile(fn).fetch(asVCF()) is the tabix interface. It 
provides random access via genomic coordinates through tabix and lazy 
parsing of columns within a row to avoid parsing samples 1-100 if only 
sample 101 is required. It is however unaware of the structure of the 
sample columns and ignores any information in the header of the VCF 
file. Both are returned as uninterpreted text. This is a fairly thin
layer on top of tabix and fully supported.

Option 1 (pysam.VCF() is a full parser of the VCF format contributed
by Gerton. It uses Option 2 for random access and lazy parsing, but
is fully aware of the header and sample structure. It is quite 
sophisticated in trying to work with VCF files of different versions
and those that are slightly malformatted. I believe it is quite feature 
complete. Adding it to pysam was experimental and it has not received
as much attention as the samtools/tabix components.

Kevin is adding a layer to the bcf/vcf part of htslib which will be the
main supported API to interact with bcf/vcf files. He mentioned it
would soon be ready. We were discussing whether to wait for htslib 1.2
before releasing the next pysam version (0.8.2).

I am not sure how deep htslib goes into the interpretation of the header
or sample column data. Is this something that pysam will need to keep?

Best wishes,
Andreas

AndreasHeger:
The new htslib VCF/BCF API support for pysam is currently hosted in this
branch <https://github.com/bioinformed/pysam/tree/bcf>.  I've adapted the
above code to the new API, which allows access to headers, records, indexed
queries, and parses subsets of samples (if requested).

from  __future__ import divisionfrom  __future__ import
print_functionfrom  __future__ import unicode_literals
import sysfrom pysam import BCFFile
def main():
    # This is https://github.com/bioinformed/pysam/blob/bcf/pysam/cbcf.pyx
    # Works on both VCF and BCF files
    vars = BCFFile(sys.argv[1])

    # Full header information is available
    print('samples = ', list(vars.header.samples))

    print("INFO headers =")
    for k,v in vars.header.info.iteritems():
        print('\tINFO {} -> {}'.format(k, v))

    # Return results only for the first sample
    # Must be called prior to fetching records
    vars.subset_samples([ vars.header.samples[0] ])

    for record in vars.fetch(b"1"):
        print(record.contig, record.pos, record.id or '.', record.ref,
record.alts,
              record.qual, ','.join(record.filter) or '.', sep="\t")

        for sample, geno in record.genos.iteritems():
            print('\t\tsample =', sample)
            for k, v in geno.iteritems():
                print('\t\t{} -> {}'.format(k, v))

if __name__ == '__main__':
    main()

jeromekelleher:
This htslib API looks perfect - we get all of the information we need in an easy to use manner. When can we expect this in a released version of pysam? Write support isn't critical, but it would be nice to have. (I'm not sure if it was Kevin or Andreas sent this message - github says it's Andreas, but I got the email from Kevin)

bioinformed:
I sent the last message with the adapted code sample.  I'm not sure why it
came out garbled.  Anyhow, I'm working on a minor API update and writing
some additional documentation in preparation for an alpha release.  Since
the code requires some fixes only present in htslib 1.2, Andreas and I are
discussing whether to merge the new API before or after the impending pysam
0.8.2 release.

-Kevin

pashields:
I work with @kerrydc and know that he's gone on vacation. I'm going to close this PR for now so it doesn't block any other requests. I think it's reasonable to reopen this in the future when kerry gets back and build this on top of whatever is on master at that point.

AndreasHeger:
Hi @jeromekelleher and all, we have just released pysam 0.8.2. This release includes @bioinformed 's bcf/vcf htslib API.

Please let us know of any comments and suggestions.

Best wishes,
Andreas

jeromekelleher:
This is great news, thanks @AndreasHeger!

PR #101, opened 2015-01-28
Refined the style imports section.

jeromekelleher:
Added guidelines on how to use imports.

PR #123, opened 2015-02-03
Change style guide abbreviations policy

dcolligan:
Breaking out this change from #108 

I'd like to advocate disobeying the part of PEP8 that says: "Note: When using abbreviations in CapWords, capitalize all the letters of the abbreviation. Thus HTTPServerError is better than HttpServerError." Noun boundaries are difficult to parse with all-caps acronyms, especially with a variable name that has multiple acronyms. For instance:

HttpClient is easier to read than HTTPClient
HttpApiClient is easier to read than HTTPAPIClient

PEP8 doesn't offer any rationale for the "Note" recommendation.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -11,6 +11,8 @@ Short variable names are OK when the scope is small (less than five lines)
 and the meaning is obvious. If the scope is longer or more complex,
 please use descriptive longer names.
 
+Contrary to PEP8, acronyms should be camel cased, not capitalized.  For instance, HTTPAPIClient should be written as HttpApiClient.
+
 ## Imports
 Imports should be structured into the following groups:
 


jeromekelleher:
I tend to agree, but am not convinced it's worth breaking from PEP8 for, so I'm +0 on this.

jeromekelleher:
Also, could you merge #101 and rebase please?

dcolligan:
I just merged #101 and rebased this PR

pashields:
+0 on this. I agree that it's more legible, but I'm not passionate about it.

diekhans:
Maybe a little late on this, but I agree with @dcolligan that things like HttpApiClient are MUCH easier to read.  I have seem some pretty horrible stuff trying to capitalize abbreviations.   Consistency with English is usually good, however it sometimes falls when it conflicts with other inconsistency with English like no spaces names.

Naburimannu:
+0 here likewise.

diekhans:
that is a +1 from me

PR #124, opened 2015-02-03
Added support for flake8 checks.

jeromekelleher:
Flake8 combines pep8 checks with some other lint checks. See http://flake8.readthedocs.org/en/2.2.3/ for details. These are the minimum changes required to get flake8 checks passing.

This also fixes a bug in the configure function in `frontend.py` which flake8 caught.


Naburimannu:
+1

PR #125, opened 2015-02-03
Fixes issues with workarounds for __slots__.

### LINKED ISSUE #117: Client no longer displays sent JSON in verbose mode
The client CLI provides a verbosity option to provide increasing levels of detail on the transactions between the client and server. This no longer shows the JSON sent to the server, which was a very useful debugging tool. Here is a possible behaviour for -v:

- -v: shows summary of HTTP messages sent to and received from server.
- -vv: shows full JSON messages sent and received.

jeromekelleher:
Workarounds for the Google API were broken after we changed over to using __slots__ for protocol elements. Also:

- Also fixes #117 as JSON output for requests is reinstated.
- Changed the Google variant example to return 1 record, as it was taking a long time to run.


Naburimannu:
+1
Sorry I missed that somehow when writing __slots__ (I think the patches might have came in in parallel?)

Do we understand the Google API variant thing? I should be meeting with that team today and can ask questions...

jeromekelleher:
No worries @Naburimannu, the patches came through in parallel so we didn't catch it. It's an easy fix, so no big deal.

We were just requesting a lot of variants in the query, so it naturally took some time. However, if you're in the mood for asking questions, it would be good to find out how long we're going to need to keep these workarounds...

dcolligan:
+1

this is a much better solution that I have in #108 

Also we can't really hold people accountable for unwittingly breaking `google_examples` if it's not in the CI, which presumably it will never be.

Naburimannu:
Actually, wait, -0: what happens when this gets tested with the workaroundGoogle active? It shouldn't be possible to remove an item from `__slots__`. ( I guess we focused on the drawback that you can't *add* anything to it when we discussed the pros & cons of the switch.)

jeromekelleher:
I think we're just removing an element from the `__slots__` list, not the actual underlying 'slots'. The `__slots__` list is  used in our JSON serialisation, so we get the effect we want, but if we called `object.attribute` it would still contain a value. We should probably check this though...

Naburimannu:
The Google workaround is a bug on their side; if we find any other mismatches between the spec and their implementation ping me or @calbach. Once they've deployed a fix I'll file an issue here, or just prep a PR.

@jeromekelleher: The `__slots__` list is directly tied to an in-memory layout, and I'd expect things to break if we were trying to mutate it at runtime - but I'd hope that the Python runtime treated it as const and disallowed what we're trying to do here.

jeromekelleher:
Actually, looking into this it's a bit more subtle than I might have thought, and I'm going to back out of it. I'll close the PR if everyone agrees.

How did you get around the problem in #108 @dcolligan?

dcolligan:
I just added new private classes in the respective methods in `cli.py` as alternatives to the protocol-defined classes.  See `GASearchReferenceSetsRequestGoogle` and `GASearchReadsRequestGoogle`

jeromekelleher:
OK, that's safer, let's go with your method. Closing.

calbach:
@Naburimannu fix just went live for the Google API issue of searching by `assemblyId`.

PR #126, opened 2015-02-04
83 conversion tools

dcolligan:
`ga2vcf.py` WIP

Wanted feedback about if I'm producing sane VCF headers before moving on.

Running the tool produces this file:

```
##fileformat=VCFv4.0
##INFO=<ID=LDAF,Number=1,Type=float,Description="MLE Allele Frequency Accounting for LD"
##INFO=<ID=AVGPOST,Number=1,Type=float,Description="Average posterior probability from MaCH/Thunder"
##INFO=<ID=RSQ,Number=1,Type=float,Description="Genotype imputation quality from MaCH/Thunder"
##INFO=<ID=ERATE,Number=1,Type=float,Description="Per-marker Mutation rate from MaCH/Thunder"
##INFO=<ID=THETA,Number=1,Type=float,Description="Per-marker Transition rate from MaCH/Thunder"
##INFO=<ID=CIEND,Number=2,Type=int,Description="Confidence interval around END for imprecise variants"
##INFO=<ID=CIPOS,Number=2,Type=int,Description="Confidence interval around POS for imprecise variants"
##INFO=<ID=END,Number=1,Type=uint,Description="End position of the variant described in this record"
##INFO=<ID=HOMLEN,Number=0,Type=uint,Description="Length of base pair identical micro-homology at event breakpoints"
##INFO=<ID=SVLEN,Number=1,Type=int,Description="Difference in length between REF and ALT alleles"
##INFO=<ID=SVTYPE,Number=0,Type=char,Description="Type of structural variant"
##INFO=<ID=AC,Number=0,Type=uint,Description="Alternate Allele Count"
##INFO=<ID=AN,Number=1,Type=uint,Description="Total Allele Count"
##INFO=<ID=AA,Number=0,Type=char,Description="Ancestral Allele, ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/pilot_data/technical/reference/ancestral_alignments/README"
##INFO=<ID=AF,Number=1,Type=float,Description="Global Allele Frequency based on AC/AN"
##INFO=<ID=AMR_AF,Number=1,Type=float,Description="Allele Frequency for samples from AMR based on AC/AN"
##INFO=<ID=ASN_AF,Number=1,Type=float,Description="Allele Frequency for samples from ASN based on AC/AN"
##INFO=<ID=AFR_AF,Number=1,Type=float,Description="Allele Frequency for samples from AFR based on AC/AN"
##INFO=<ID=EUR_AF,Number=1,Type=float,Description="Allele Frequency for samples from EUR based on AC/AN"
##INFO=<ID=VT,Number=0,Type=char,Description="indicates what type of variant the line represents"
##INFO=<ID=SNPSOURCE,Number=0,Type=char,Description="indicates if a snp was called when analysing the low coverage or exome alignment data"
##INFO=<ID=DS,Number=1,Type=float,Description="Genotype dosage from MaCH/Thunder"
##INFO=<ID=GL,Number=0,Type=float,Description="Genotype Likelihoods"
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT
```

### DIFF
--- a/ga4gh/converters.py
+++ b/ga4gh/converters.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import collections
+import re
 
 import pysam
 
@@ -223,4 +224,322 @@ class VcfConverter(AbstractConverter):
     """
     def __init__(self, httpClient, outputStream, searchVariantSetsRequest,
                  searchVariantsRequest):
-        raise NotImplementedError
+        super(VcfConverter, self).__init__(httpClient)
+        self.vcf = None
+        self.variantSet = None
+        self.outputStream = outputStream
+        self.searchVariantsRequest = searchVariantsRequest
+        self.searchVariantSetsRequest = searchVariantSetsRequest
+
+    def convert(self):
+        self.getVariantSet()
+        self.getCallSets()
+        self.getVariants()
+        self.createVcf()
+
+    def getVariantSet(self):
+        response = self._httpClient.searchVariantSets(
+            self.searchVariantSetsRequest)
+        variantSets = [variantSet for variantSet in response]
+        if len(variantSets) == 0:
+            raise VcfException("VariantSet not found")
+        if len(variantSets) > 1:
+            raise VcfException("More than one VariantSet returned")
+        self.variantSet = variantSets[0]
+        return variantSet
+
+    def getCallSets(self):
+        request = protocol.GASearchCallSetsRequest()
+        request.variantSetIds = [self.variantSet.id]
+        response = self._httpClient.searchCallSets(request)
+        callSets = [callSet for callSet in response]
+        self.callSets = callSets
+        return callSets
+
+    def getVariants(self):
+        response = self._httpClient.searchVariants(
+            self.searchVariantsRequest)
+        self.variantGenerator = response
+        return response
+
+    def createVcf(self):
+        self.vcf = Vcf(self.outputStream, self.variantSet, self.callSets)
+        self.vcf.writeHeader()
+        self.vcf.writeData(self.variantGenerator)
+
+
+class VcfMetadataInfoLine(object):
+    """
+    A line in the metadata of a VCF file
+    """
+    # see pysam source file cvcf.pyx
+    numberTypes = {
+        'NT_UNKNOWN': 0,
+        'NT_NUMBER': 1,
+        'NT_ALLELES': 2,
+        'NT_NR_ALLELES': 3,
+        'NT_GENOTYPES': 4,
+        'NT_PHASED_GENOTYPES': 5,
+    }
+
+    def __init__(self, infoDict):
+        self.infoDict = infoDict
+        for key, value in self.infoDict.items():
+            setattr(self, key, value)
+
+    def getNumbertype(self):
+        # The google api seems to return one of the following for number:
+        # - an integer
+        # - a period
+        # - None (only on custom header lines)
+        if re.match('[0-9]+', self.number) is not None:
+            return self.numberTypes['NT_NUMBER']
+        elif self.number == 'R':
+            return self.numberTypes['NT_ALLELES']
+        elif self.number == 'A':
+            return self.numberTypes['NT_NR_ALLELES']
+        elif self.number == 'G':
+            return self.numberTypes['NT_GENOTYPES']
+        else:
+            return self.numberTypes['NT_UNKNOWN']
+
+    def __getattr__(self, name):
+        # the pysam implementation looks up the following attributes
+        # on this object
+        if name == 'numbertype':
+            return self.getNumbertype()
+        elif name == "missingvalue":
+            return None
+        raise VcfException(
+            "VcfMetadataInfoLine has no attribute {0}".format(name))
+
+
+class VcfMetadata(object):
+    """
+    The metadata of a VCF file
+    """
+    def __init__(self):
+        self.info = {}
+        self.filter_ = {}
+        self.format_ = {}
+        self.header = []
+        self.switch = {
+            'INFO': self.info,
+            'FILTER': self.filter_,
+            'FORMAT': self.format_,
+        }
+
+    def processVariantSet(self, variantSet):
+        for metadata in variantSet.metadata:
+            infoLine = self._processMetadata(metadata)
+            if infoLine.key in self.switch:
+                self.switch[infoLine.key][infoLine.id] = infoLine
+            elif infoLine.key == 'fileformat':
+                # fileformat already written by pysam
+                continue
+            else:
+                keyVal = self._processHeader(infoLine)
+                self.header.append(keyVal)
+
+    def _processHeader(self, infoLine):
+        if infoLine.key == 'contig':
+            valueStr = []
+            for ilKey, ilVal in infoLine.info.items():
+                ilVal = ilVal[0]
+                if ilVal.find(' ') != -1:
+                    ilVal = '"{0}"'.format(ilVal)
+                fragment = '{0}={1}'.format(ilKey, ilVal)
+                valueStr.append(fragment)
+            joinedValueStr = ','.join(valueStr)
+            value = "<{0}>".format(joinedValueStr)
+        elif infoLine.key == 'ALT':
+            value = "<ID={id}, Description={description}>".format(
+                **infoLine.infoDict)
+        # TODO special cases for other fields: assembly, etc...
+        # see VCF doc section 1.2
+        else:
+            value = infoLine.value
+        return (infoLine.key, value)
+
+    def _processMetadata(self, metadata):
+        infoDict = metadata.toJsonDict()
+        infoLine = VcfMetadataInfoLine(infoDict)
+        return infoLine
+
+
+class VcfRow(object):
+    """
+    Represents a row in a VCF file
+    """
+    defaultChrom = 'defaultChrom'
+    defaultQuality = '50'
+    defaultFilter = []
+    defaultFormat = []
+    defaultNoSample = []
+
+    def __init__(self, variant, samples):
+        self.variantDict = {
+            'chrom': self.defaultChrom,
+            # pysam increments pos, so it needs to be an int
+            'pos': int(variant.start),
+            'id': variant.id,
+            'ref': variant.referenceBases,
+            'alt': variant.alternateBases,
+            'qual': self.defaultQuality,
+            'filter': self.defaultFilter,
+            'info': variant.info,
+            'format': self.defaultFormat,
+        }
+        for call in variant.calls:
+            self.variantDict[call.callSetName] = call.info
+        for sample in samples:
+            if sample not in self.variantDict:
+                self.variantDict[sample] = self.defaultNoSample
+
+    def toDict(self):
+        return self.variantDict
+
+
+class Vcf(object):
+    """
+    A convenience wrapper around pysam.VCF functionality
+    providing methods to write a VCF file.
+    """
+    def __init__(self, stream, variantSet, callSets):
+        self.stream = stream
+        self.vcf = pysam.VCF()
+        self.variantSet = variantSet
+        self.callSets = callSets
+        self.data = []
+        self.headerWritten = False
+
+    def writeHeader(self):
+        self.vcfMetadata = VcfMetadata()
+        self.vcfMetadata.processVariantSet(self.variantSet)
+        self.vcf.setinfo(self.vcfMetadata.info)
+        self.vcf.setfilter(self.vcfMetadata.filter_)
+        self.vcf.setformat(self.vcfMetadata.format_)
+        self.vcf.setheader(self.vcfMetadata.header)
+        self.samples = [callSet.sampleId for callSet in self.callSets]
+        self.vcf.setsamples(self.samples)
+        self.vcf.writeheader(self.stream)
+        self.headerWritten = True
+
+    def writeData(self, variantGenerator):
+        if self.headerWritten is False:
+            raise VcfException("Need to write header before writing data")
+        requestLimit = 1  # TODO find beter way to limit data
+        requestNumber = 0
+        for variant in variantGenerator:
+            if requestNumber >= requestLimit:
+                break
+            row = VcfRow(variant, self.samples)
+            # for debugging pysam, uncomment these lines
+            # and comment out the write_data call
+            # fakePysam = FakePysam(
+            #    self.samples, self.vcfMetadata.format_, self.vcfMetadata.info)
+            # fakePysam.write_data(self.stream, row.toDict())
+            self.vcf.write_data(self.stream, row.toDict())
+            requestNumber += 1
+
+
+#############################
+# TODO take out below
+# this is copy-and-pasted from pysam's implementation to help with debugging
+#############################
+
+
+class FakePysam(object):
+
+    NT_UNKNOWN = 0
+    NT_NUMBER = 1
+    NT_ALLELES = 2
+    NT_NR_ALLELES = 3
+    NT_GENOTYPES = 4
+    NT_PHASED_GENOTYPES = 5
+
+    def __init__(self, samples, format_, info):
+        self._version = 40
+        self._samples = samples
+        self._format = format_
+        self._info = info
+
+    def write_data(self, stream, data):
+        required = ['chrom','pos','id','ref','alt','qual','filter','info','format'] + self._samples
+        for k in required:
+            if k not in data: raise ValueError("Required key %s not found in data" % str(k))
+        if data['alt'] == []: alt = "."
+        else: alt = ",".join(data['alt'])
+        if data['filter'] == None: filter = "."
+        elif data['filter'] == []: 
+            if self._version == 33: filter = "0"
+            else: filter = "PASS"
+        else: filter = ';'.join(data['filter'])
+        if data['qual'] == -1: qual = "."
+        else: qual = str(data['qual'])
+
+        output = [data['chrom'], 
+                  str(data['pos']+1),   # change to 1-based position
+                  data['id'],
+                  data['ref'],
+                  alt,
+                  qual,
+                  filter,
+                  self.format_formatdata(
+                      data['info'], self._info, separator=";"),
+                  self.format_formatdata(
+                      data['format'], self._format, value=False)]
+        
+        for s in self._samples:
+            output.append(self.format_formatdata(
+                data[s], self._format, key=False))
+        
+        stream.write( "\t".join(output) + "\n" )
+
+    def format_formatdata( self, data, format, key=True, value=True, separator=":" ):
+        output, sdata = [], []
+        if type(data) == type([]): # for FORMAT field, make data with dummy values
+            d = {}
+            for k in data: d[k] = []
+            data = d
+        # convert missing values; and silently add definitions if required
+        for k in data:
+            self._add_definition( format, k, data[k], "(output)" )
+            for idx,v in enumerate(data[k]):
+                if v == format[k].missingvalue: data[k][idx] = "."
+        # make sure GT comes first; and ensure fixed ordering; also convert GT data back to string
+        for k in data: 
+            if k != 'GT': sdata.append( (k,data[k]) )
+        sdata.sort()
+        if 'GT' in data:
+            sdata = [('GT',map(self.convertGTback,data['GT']))] + sdata
+        for k,v in sdata:
+            if v == []: v = None
+            if key and value:
+                if v != None: output.append( k+"="+','.join(map(str,v)) )
+                else: output.append( k )
+            elif key: output.append(k)
+            elif value:
+                if v != None: output.append( ','.join(map(str,v)) )
+                else: output.append( "." )                    # should not happen
+        # snip off trailing missing data
+        while len(output) > 1:
+            last = output[-1].replace(',','').replace('.','')
+            if len(last)>0: break
+            output = output[:-1]
+        return separator.join(output)
+
+    def _add_definition(self, formatdict, key, data, line ):
+        if key in formatdict: return
+        self.error(line,self.ERROR_UNKNOWN_KEY,key)
+        if data == None:
+            formatdict[key] = FORMAT(key,self.NT_NUMBER,0,"Flag","(Undefined tag)",".")
+            return
+        if data == []: data = [""]             # unsure what type -- say string
+        if type(data[0]) == type(0.0):
+            formatdict[key] = FORMAT(key,self.NT_UNKNOWN,-1,"Float","(Undefined tag)",None)
+            return
+        if type(data[0]) == type(0):
+            formatdict[key] = FORMAT(key,self.NT_UNKNOWN,-1,"Integer","(Undefined tag)",None)
+            return
+        formatdict[key] = FORMAT(key,self.NT_UNKNOWN,-1,"String","(Undefined tag)",".")
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -12,6 +12,7 @@ import mock
 import ga4gh.cli as cli
 import ga4gh.protocol as protocol
 import ga4gh.client as client
+import ga4gh.converters as converters  # NOQA
 
 
 class TestNoInput(unittest.TestCase):


ga2vcf.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def _formatVariantSetToVcfHeader(self):
+        header = []
+        for metadata in self.variantSet.metadata:
+            infoLine = self._formatMetadataToMetainfoLine(metadata)
+            header.append(infoLine)
+        return header
+
+    def _formatMetadataToMetainfoLine(self, metadata):
+        key = "INFO"
```
dcolligan: How to distinguish INFO metainfo line from others?  (e.g. FORMAT, etc.)

bioinformed: At first glance:

   1. Better to create VCF 4.1 or 4.2 files.  The 4.0 spec is very dated.
   2. types are not right.  Based on the spec, they are case sensitive and
   must be one of: Integer, Float, Character, String or Flag (for INFO only)
   3. "Number=0" is nonsensical. If Number is 0 then the Type should be
   Flag. If Number is not known, then it should be ".".

I'll look again and more deeply in a bit.

jeromekelleher:
Looks like a good start @dcolligan, thanks. Some comments:

1. I think it would be good to make this a first-class program like ga4gh_server or ga4gh_client that is installed in the users path by setuptools. So, we should put the code for it into `cli.py` and edit `setup.py` to add ga2vcf to the list of entry points. Then, I think you should be able to reuse the majority of the code you've written for the variants-search client command.
2. It's a pity that the pysam API doesn't output more modern VCF and take care of writing out the INFO types for us. Since we're hoping to move over to the new htslib interface that @bioinformed is working on soon, I wouldn't worry too much about moving to VCF version 4.2 for now. Let's get something that works reasonably well using the existing pysam API and transition to the new API when it's ready.

@bioinformed, what's your take on this? Will the htslib based API support outputting BCF and VCF transparently? Will we need to create the headers, or will it be possible to generate the VCF in a fully programmatic way? Thanks for looking into this by the way!

bioinformed:
@jeromekelleher, the pysam VCF/BCF API _will_ allow fully programmatic creation of headers and records.  This functionality is not yet implemented, since I'm hoping to get feedback and some real world feedback on the immutable interface before making everything mutable (and tackling some thorny issues with object lifetimes and keeping C/Python data structures in sync so that Python scripts cannot _easily_ crash the interpreter).  Thus I think it is reasonable to keep hacking on a ga2vcf that writes straight VCF, since neither approach is rocket science and it will be a good template for when the mutable API is ready.

dcolligan:
@jeromekelleher Good idea on the `setup.py` refactor.

I think my questions about this are motivated more by the mismatch between what the GA4GH API currently provides (e.g. no FILTER or FORMAT headers; having to make a bunch of other requests to build up the rest of the metadata like the sample column names) and the VCF format.

It seems like the latest (and default) version pysam supports is 4.0: https://github.com/pysam-developers/pysam/blob/0.8.1/pysam/cvcf.pyx#L292

@bioinformed thanks for the good feedback

dcolligan:
@jeromekelleher Also, maybe our sample data isn't what the VCF API/format is expecting (e.g. the type strings not lining up)

jeromekelleher:
This is true @dcolligan --- I've filed an issue for this in #129. What does the Google API return for these values?

dcolligan:
Example Google API return value (part of a response from https://cloud.google.com/genomics/v1beta2/reference/variantsets/search ): https://gist.github.com/dcolligan/e122f4381be9e0e4bbb8

dcolligan:
Ok this file header output by the new version of the code (pulling from google) looks more reasonable by @bioinformed 's standards (but see the comment in the code about pysam not supporting VCF 4.2)
```
##fileformat=VCFv4.0
##INFO=<ID=FORMAT,Number=1,Type=FLOAT,Description="Genotype dosage from MaCH/Thunder">
##INFO=<ID=reference,Number=None,Type=UNKNOWN_TYPE,Description="None">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="GL based estimate of reference allele frequency">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Average posterior probability from MaCH/Thunder">
##INFO=<ID=INFO,Number=1,Type=INTEGER,Description="Raw read depth">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Genotyping depth model parameter M2">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Allele Frequency for samples from AFR based on AC/AN">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Posterior weighted chi^2 P-value for testing the association between group1 and group2 samples.">
##INFO=<ID=fileDate,Number=None,Type=UNKNOWN_TYPE,Description="None">
##INFO=<ID=INFO,Number=1,Type=INTEGER,Description="Difference in length between REF and ALT alleles">
##INFO=<ID=FORMAT,Number=1,Type=FLOAT,Description="Genotype quality">
##INFO=<ID=FORMAT,Number=1,Type=INTEGER,Description="Genotype Quality">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="For homozygous reference sites, the minimum genotype quality score">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Highest negative log10 concordance estimate with duplicate event">
##INFO=<ID=INFO,Number=.,Type=STRING,Description="List of samples">
##INFO=<ID=INFO,Number=4,Type=INTEGER,Description="# high-quality ref-forward bases, ref-reverse, alt-forward and alt-reverse bases">
##INFO=<ID=FORMAT,Number=1,Type=INTEGER,Description="# high-quality bases">
##INFO=<ID=reference,Number=None,Type=UNKNOWN_TYPE,Description="None">
##INFO=<ID=INFO,Number=1,Type=STRING,Description="The most probable unconstrained genotype configuration in the trio">
##INFO=<ID=INFO,Number=.,Type=STRING,Description="Assembly meta">
##INFO=<ID=INFO,Number=1,Type=INTEGER,Description="End position of the variant described in this record">
##INFO=<ID=FORMAT,Number=1,Type=INTEGER,Description="Phred-scaled strand bias P-value">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Max-likelihood estimate of the first ALT allele count (no HWE assumption)">
##INFO=<ID=INFO,Number=.,Type=INTEGER,Description="Allele count in genotypes">
##INFO=<ID=INFO,Number=.,Type=INTEGER,Description="Alternate Allele Count">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Allele Frequency for samples from EUR based on AC/AN">
##INFO=<ID=INFO,Number=1,Type=INTEGER,Description="Total number of alleles in called genotypes">
##INFO=<ID=INFO,Number=1,Type=INTEGER,Description="Total Allele Count">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Phred probability of all samples being the same">
##INFO=<ID=contig,Number=None,Type=UNKNOWN_TYPE,Description="None">
##INFO=<ID=FORMAT,Number=.,Type=INTEGER,Description="List of Phred-scaled genotype likelihoods, number of values is (#ALT+1)*(#ALT+2)/2">
##INFO=<ID=INFO,Number=1,Type=INTEGER,Description="# permutations yielding a smaller PCHI2.">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Genotype imputation quality from MaCH/Thunder">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="GL based estimate of alternate allele frequency">
##INFO=<ID=INFO,Number=.,Type=STRING,Description="indicates if a snp was called when analysing the low coverage or exome alignment data">
##INFO=<ID=INFO,Number=4,Type=FLOAT,Description="P-values for strand bias, baseQ bias, mapQ bias and tail distance bias">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Per-marker Mutation rate from MaCH/Thunder">
##INFO=<ID=INFO,Number=0,Type=FLAG,Description="Validation flag">
##INFO=<ID=FORMAT,Number=1,Type=STRING,Description="Genotype">
##INFO=<ID=FORMAT,Number=1,Type=FLOAT,Description="Copy number genotype quality for imprecise events">
##INFO=<ID=INFO,Number=2,Type=INTEGER,Description="Phred probability of the nonRef allele frequency in group1 samples being larger (,smaller) than in group2.">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Sum of homozygous reference likelihoods">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Sum of homozygous alternate allele likelihoods">
##INFO=<ID=FORMAT,Number=1,Type=INTEGER,Description="Copy number genotype for imprecise events">
##INFO=<ID=INFO,Number=2,Type=INTEGER,Description="Confidence interval around POS for imprecise variants">
##INFO=<ID=INFO,Number=1,Type=INTEGER,Description="Phred scaled PCHI2.">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Genotyping depth model parameter M1">
##INFO=<ID=INFO,Number=2,Type=INTEGER,Description="Inner confidence interval around END for imprecise variants">
##INFO=<ID=FORMAT,Number=.,Type=FLOAT,Description="Genotype likelihoods">
##INFO=<ID=FORMAT,Number=.,Type=FLOAT,Description="Genotype Likelihoods">
##INFO=<ID=FORMAT,Number=3,Type=FLOAT,Description="Likelihoods for RR,RA,AA genotypes (R=ref,A=alt)">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Allele Frequency for samples from ASN based on AC/AN">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Global Allele Frequency based on AC/AN">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Chi^2 based HWE test P-value based on G3">
##INFO=<ID=INFO,Number=2,Type=INTEGER,Description="Confidence interval around END for imprecise variants">
##INFO=<ID=INFO,Number=.,Type=STRING,Description="List of duplicate events preferred to this one">
##INFO=<ID=source,Number=None,Type=UNKNOWN_TYPE,Description="None">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="MLE Allele Frequency Accounting for LD">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Max-likelihood estimate of the first ALT allele frequency (assuming HWE)">
##INFO=<ID=INFO,Number=.,Type=INTEGER,Description="Length of base pair identical micro-homology at event breakpoints">
##INFO=<ID=INFO,Number=.,Type=STRING,Description="Assembly contig">
##INFO=<ID=reference,Number=None,Type=UNKNOWN_TYPE,Description="None">
##INFO=<ID=INFO,Number=1,Type=STRING,Description="indicates what type of variant the line represents">
##INFO=<ID=fileformat,Number=None,Type=UNKNOWN_TYPE,Description="None">
##INFO=<ID=INFO,Number=.,Type=STRING,Description="Validation method">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Inbreeding coefficient">
##INFO=<ID=ALT,Number=None,Type=UNKNOWN_TYPE,Description="Deletion">
##INFO=<ID=FORMAT,Number=.,Type=FLOAT,Description="Copy number likelihoods">
##INFO=<ID=INFO,Number=1,Type=STRING,Description="Type of structural variant">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Allele Frequency for samples from AMR based on AC/AN">
##INFO=<ID=INFO,Number=1,Type=STRING,Description="The most probable constrained genotype configuration in the trio">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Sum of heterozygous likelihoods">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Highest reciprocal overlap with duplicate event">
##INFO=<ID=INFO,Number=1,Type=INTEGER,Description="Root-mean-square mapping quality of covering reads">
##INFO=<ID=samtoolsVersion,Number=None,Type=UNKNOWN_TYPE,Description="None">
##INFO=<ID=INFO,Number=1,Type=INTEGER,Description="Log ratio of genotype likelihoods with and without the constraint">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Per-marker Transition rate from MaCH/Thunder">
##INFO=<ID=INFO,Number=3,Type=FLOAT,Description="ML estimate of genotype frequencies">
##INFO=<ID=INFO,Number=.,Type=STRING,Description="Sequence of base pair identical micro-homology at event breakpoints">
##INFO=<ID=INFO,Number=.,Type=STRING,Description="Information about merged calls: SOURCE, POS, and END">
##INFO=<ID=INFO,Number=1,Type=FLOAT,Description="Cluster separation in depth-based genotyping model">
##INFO=<ID=INFO,Number=1,Type=STRING,Description="Ancestral Allele, ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/pilot_data/technical/reference/ancestral_alignments/README">
##INFO=<ID=INFO,Number=1,Type=STRING,Description="Ancestral Allele, equal chimp reference allele">
##INFO=<ID=reference,Number=None,Type=UNKNOWN_TYPE,Description="None">
##INFO=<ID=INFO,Number=2,Type=INTEGER,Description="Inner confidence interval around POS for imprecise variants">
##INFO=<ID=fileformat,Number=None,Type=UNKNOWN_TYPE,Description="None">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT
```

bioinformed:
Those headers still look pretty garbled.


   1. Based on the spec, types are case sensitive and must be one of:
   Integer, Float, Character, String or Flag (for INFO only).  The current
   upper case values are illegal, as is UNKNOWN_TYPE.
   2. Within INFO, FILTER, and FORMAT metadata, the ID values must be
   unique.
   3. All metadata types in your example are output as ##INFO when many
   records are intended to be FORMAT or unstructured (key/value) metadata
      - e.g. ##INFO=<ID=source,Number=None,Type=UNKNOWN_TYPE,Description=
      "None"> should be ##source=somevalue
      - e.g. ##INFO=<ID=FORMAT,Number=.,Type=FLOAT,Description="Genotype
      likelihoods"> should be
##FORMAT=<ID=GL,Number=.,Type=Float,Description="Genotype
      likelihoods">

Here is a header from a random example file (using some VCF 4.1 features)
that shows how these metadata are intended to look:

##fileformat=VCFv4.1
##FORMAT=<ID=AD,Number=.,Type=Integer,Description="Allelic depths for the
ref and alt alleles in the order listed">
##FORMAT=<ID=DP,Number=1,Type=Integer,Description="Approximate read depth
(reads with MQ=255 or with bad mates are filtered)">
##FORMAT=<ID=GQ,Number=1,Type=Float,Description="Genotype Quality">
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
##FORMAT=<ID=PL,Number=G,Type=Integer,Description="Normalized, Phred-scaled
likelihoods for genotypes as defined in the VCF specification">
##INFO=<ID=AC,Number=A,Type=Integer,Description="Allele count in genotypes,
for each ALT allele, in the same order as listed">
##INFO=<ID=AF,Number=A,Type=Float,Description="Allele Frequency, for each
ALT allele, in the same order as listed">
##INFO=<ID=AN,Number=1,Type=Integer,Description="Total number of alleles in
called genotypes">
##INFO=<ID=BaseQRankSum,Number=1,Type=Float,Description="Z-score from
Wilcoxon rank sum test of Alt Vs. Ref base qualities">
##INFO=<ID=DB,Number=0,Type=Flag,Description="dbSNP Membership">
##INFO=<ID=DP,Number=1,Type=Integer,Description="Approximate read depth;
some reads may have been filtered">
##INFO=<ID=DS,Number=0,Type=Flag,Description="Were any of the samples
downsampled?"> ##INFO=<ID=Dels,Number=1,Type=Float,Description="Fraction of
Reads Containing Spanning Deletions">
##INFO=<ID=FS,Number=1,Type=Float,Description="Phred-scaled p-value using
Fisher's exact test to detect strand bias">
##INFO=<ID=HRun,Number=1,Type=Integer,Description="Largest Contiguous
Homopolymer Run of Variant Allele In Either Direction">
##INFO=<ID=HaplotypeScore,Number=1,Type=Float,Description="Consistency of
the site with at most two segregating haplotypes">
##INFO=<ID=InbreedingCoeff,Number=1,Type=Float,Description="Inbreeding
coefficient as estimated from the genotype likelihoods per-sample when
compared against the Hardy-Weinberg expectation">
##INFO=<ID=MQ,Number=1,Type=Float,Description="RMS Mapping Quality">
##INFO=<ID=MQ0,Number=1,Type=Integer,Description="Total Mapping Quality
Zero Reads"> ##INFO=<ID=MQRankSum,Number=1,Type=Float,Description="Z-score
From Wilcoxon rank sum test of Alt vs. Ref read mapping qualities">
##INFO=<ID=QD,Number=1,Type=Float,Description="Variant Confidence/Quality
by Depth">
##INFO=<ID=ReadPosRankSum,Number=1,Type=Float,Description="Z-score from
Wilcoxon rank sum test of Alt vs. Ref read position bias">
##UnifiedGenotyper="analysis_type=UnifiedGenotyper input_file=[reads.bam]
read_buffer_size=null phone_home=NO_ET read_filter=[]
intervals=[chr22:42020321-42527953] excludeIntervals=null
interval_set_rule=UNION interval_merging=ALL reference_sequence=
##contig=<ID=chr1,length=249250621,assembly=hg19>
##contig=<ID=chr10,length=135534747,assembly=hg19>
##contig=<ID=chr11,length=135006516,assembly=hg19>
##contig=<ID=chr11_gl000202_random,length=40103,assembly=hg19>
##contig=<ID=chr12,length=133851895,assembly=hg19>
##contig=<ID=chr13,length=115169878,assembly=hg19>
##contig=<ID=chr14,length=107349540,assembly=hg19>
##contig=<ID=chr15,length=102531392,assembly=hg19>
##contig=<ID=chr16,length=90354753,assembly=hg19>
##contig=<ID=chr17,length=81195210,assembly=hg19>
##contig=<ID=chr17_ctg5_hap1,length=1680828,assembly=hg19>
##contig=<ID=chr17_gl000203_random,length=37498,assembly=hg19>
##contig=<ID=chr17_gl000204_random,length=81310,assembly=hg19>
##contig=<ID=chr17_gl000205_random,length=174588,assembly=hg19>
##contig=<ID=chr17_gl000206_random,length=41001,assembly=hg19>
##contig=<ID=chr18,length=78077248,assembly=hg19>
##contig=<ID=chr18_gl000207_random,length=4262,assembly=hg19>
##contig=<ID=chr19,length=59128983,assembly=hg19>
##contig=<ID=chr19_gl000208_random,length=92689,assembly=hg19>
##contig=<ID=chrX,length=155270560,assembly=hg19>
##contig=<ID=chrY,length=59373566,assembly=hg19>
... ##reference=file:///data/reference/ucsc/hg19/ucsc.hg19.fasta

dcolligan:
@bioinformed Thanks for the feedback.  Question for you about the pysam source: what is a "numbertype" https://github.com/pysam-developers/pysam/blob/50698a6e2b8c39dc0d59172554a4e902681014b0/pysam/cvcf.pyx#L245 and how can they be derived?  Most objects that are passed to pysam's format_format need this attribute: https://github.com/pysam-developers/pysam/blob/50698a6e2b8c39dc0d59172554a4e902681014b0/pysam/cvcf.pyx#L424

bioinformed:
@dcolligan: numbertype is to accommodate the non-numeric cardinality specifiers allowed by the VCF spec (e.g. see https://samtools.github.io/hts-specs/VCFv4.2.pdf section 1.2.2). 

    NT_UNKNOWN: an unknown number of values (Number=".")
    NT_NUMBER: a fixed integral number of values (Number=[0-9]+)
    NT_ALLELES: the number of values is equal to the number of alleles (Number=R)
    NT_NR_ALLELES: the number of values is equal to the number of alt alleles (Number=A)
    NT_GENOTYPES: [..] of unphased genotypes (Number=G)
    NT_PHASED_GENOTYPES: intended to be the number of phased genotypes, but isn't supported by VCF 4.0 - 4.2

I've never used this pysam API before, so there may be nuances that I'm missing.

dcolligan:
@jeromekelleher can you review the first 250 lines of ga4gh/ga2vcf.py ?  It produces this file on a sample run: https://gist.github.com/dcolligan/580d430c37dc107cfd3b

jeromekelleher:
This is good progress, thanks @dcolligan. It's a pity we've got so much work to do here, but I guess this is how we figure out the problems with the protocol.

At a high level, I wonder if we can simplify things a bit. I think it would be useful split the work into two things: (1) a UI; and (2) a class that takes some arguments and writes a VCF (or SAM) stream to an opened file handle. All of the UI functionality should be isolated into cli.py.  We should be able to reuse all of the functionality for `variants/search` here, as this is almost exactly what we're doing. So, use the the RequestFactory to create a `SearchVariantsRequest` object, which is populated from the command line args. We then create a VcfConverter object and pass it this request and a file handle (sys.stdout for now).

I think it would also be good to group the conversion tools into a single module. So, we create a module called `ga4gh.converters` or something, and this has two classes `VcfConverter` and `SamConverter` that present a consistent interface. Something like
```python
class VcfConverter(AbstractConverter):
    def __init__(self, searchVariantsRequest, outputFile):
         # init
    def convert(self):
         # Do the actual conversion
```
Creating the `SearchVariantsSetsRequest` object should be as simple as 
```python
request = protocol.GASearchVariantSetsRequest()
request.datasetIds = self._searchVariantsRequest.datasetIds
```
which isn't worth the bother of using a factory for. Making the callsets request should be similarly simple. This avoids the circular import problem you mentioned in #144. 

With respect to the specific conversion code, I'm finding it tricky to figure out what type of objects you're working with in the various places. Maybe you could put a `ga` prefix on some of the variables to make this a bit more obvious (e.g., `gaMetadata`)? 

Having a quick scan of the VCF output, it looks roughly right to me (but I'm absolutely not a VCF expert). You seem to be missing the FORMAT field in the data line though, and the sample fields need genotype information and so on.

I think you're making great progress on a tricky problem here; hopefully with these simplifications we'll be able to iterate to a rough POC converter soon. We're really aiming at @bioinformed's BCF interface, so it's not worth putting  a lot of effort polishing this right now.

ga4gh/ga2vcf.py
```diff
@@ -0,0 +1,345 @@
...
+import ga4gh.ga2utils as ga2utils
+import ga4gh.request_factory as requestFactory
+
+
+class VcfException(Exception):
+    pass
+
+
+class VcfMetadataInfoLine(object):
+
```
jeromekelleher: I'm a bit confused about what this class actually does. Why the `__getattr__` trick?

ga4gh/ga2vcf.py
```diff
@@ -0,0 +1,345 @@
...
+            value = infoLine.value
+        return (infoLine.key, value)
+
+    def _processMetadata(self, metadata):
+        infoDict = metadata.toJsonDict()
+        infoLine = VcfMetadataInfoLine(infoDict)
+        return infoLine
+
+
+class VcfRow(object):
```
jeromekelleher: Do we really need a whole class for this? Isn't this just converting from a GA object to a dict of the form that pysam is expecting? Wouldn't a function do the same thing and avoid the overhead of creating an object for each VCF row?

ga4gh/ga2vcf.py
dcolligan: Deep in the pysam code it calls a numbertype attribute, which is why this needs data needs to be wrapped in an object and __getattr__ overridden 

https://github.com/pysam-developers/pysam/blob/50698a6e2b8c39dc0d59172554a4e902681014b0/pysam/cvcf.pyx#L427

ga4gh/ga2vcf.py
dcolligan: Perhaps, but I'll worry about performance later... conversion to a function is an easy change.  And (see the other comment about getattr) there might be nasty surprises in pysam that mandate this... we'll see.  Plus, I like the encapsulation.

jeromekelleher:
Pysam 0.8.2 has been [released](https://github.com/pysam-developers/pysam/releases) with BCF support. We should move over to this interface immediately, as the older VCF code isn't helping all that much.

The [source](https://github.com/pysam-developers/pysam/blob/master/pysam/cbcf.pyx) is the best documentation.

bioinformed:
The VCF/BCF API in pysam 0.8.2 doesn't allow for creation of files and
records from scratch.  So it is currently feasible to use the new code as a
backend, it isn't yet possible to use it for ga2vcf.  I'm working on adding
the mutable API right now, so it will be a few more weeks before alpha code
is ready and a few more before it part of the next pysam release.

jeromekelleher:
Thanks for the clarification @bioinformed, this is good to know.

dcolligan:
Been working on the SAM tool a bit... Would like some feedback on the TODOs in `ga4gh/ga2sam.py`, if anyone has any.  `python ga2sam_dev.py --key MY_KEY --workarounds google reads-search --start 51005353 --end 51005354 --readGroupIds ChhDTXZuaHBLVEZoQ2JpT2J1enZtN25Pb0IQAA --referenceName 22 --pageSize 1 https://www.googleapis.com/genomics/v1beta2/` produces this file: https://gist.github.com/dcolligan/27e4d41c4730879e6939

dcolligan:
Closing this PR for now... will file a separate one once we decide to implement ga2vcf

PR #127, opened 2015-02-04
verbose client display

MollyZhang:
#117

### DIFF
--- a/ga4gh/client.py
+++ b/ga4gh/client.py
@@ -39,6 +39,8 @@ class HTTPClient(object):
         notDone = True
         while notDone:
             jsonString = request.toJSONString()
+            if self._debugLevel > 1:
+                self.printJSONMessage("request:", jsonString)
             headers = {"Content-type": "application/json"}
             # make sure we correctly join with/out trailing slashes
             fullUrl = posixpath.join(self._urlPrefix, url)
@@ -53,16 +55,22 @@ class HTTPClient(object):
             if self._debugLevel > 1:
                 # TODO use a logging output and integrate with HTTP client more
                 # nicely.
-                print("json response:")
-                pp = json.dumps(
-                    json.loads(jsonString), sort_keys=True, indent=4)
-                print(pp)
+                self.printJSONMessage("response:", jsonString)
             resp = protocolClass.fromJSONString(jsonString)
             # TODO handle HTTP errors from requests and display.
             for extract in getattr(resp, listAttr):
                 yield extract
             request.pageToken = resp.nextPageToken
             notDone = resp.nextPageToken is not None
+
+    def printJSONMessage(self, header, jsonString):
+        """
+        Prints the specified jsonString to stdout for debugging purposes.
+        """
+        print("json", header)
+        pp = json.dumps(
+            json.loads(jsonString), sort_keys=True, indent=4)
+        print(pp)
 
     def searchVariants(self, request):
         """
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,6 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# Pycharm
+.idea/


ga4gh/cli.py
```diff
@@ -98,7 +98,7 @@ def setHttpClient(self, request, args):
         self._request = request
         self._verbosity = args.verbose
         self._httpClient = client.HTTPClient(
-            args.baseUrl, args.verbose, self._workarounds, self._key)
```
dcolligan: I bet this is going to break a lot of tests, etc.  I actually am modifying the interface in #108 so it might be worth waiting until that change lands to revisit this issue

ga4gh/cli.py
```diff
@@ -295,7 +306,7 @@ def addVariantSearchOptions(parser):
 
 def addVariantSetIdsArgument(parser):
     parser.add_argument(
-        "--variantSetIds", "-v",
+        "--variantSetIds", "-V",
```
dcolligan: I know this is just a one-character change, but I think because of what I mentioned in the above comment it might be worth breaking this change out into a separate PR (there also might be broken tests associated with this change)

ga4gh/cli.py
```diff
@@ -109,12 +109,23 @@ def _run(self, methodName, attrName=None):
         method = getattr(self._httpClient, methodName)
         results = method(self._request)
         for result in results:
-            if attrName is None:
-                print(result)
+            if self._verbosity == 1:
+                print(methodName, "request: \n", self._request)
```
dcolligan: It would probably be better to prefer using string formatting (using str's `format` method) over calling print with varargs

ga4gh/client.py
```diff
@@ -42,21 +41,12 @@ def runRequest(self, request, url, protocolClass, listAttr):
...
             fullUrl = posixpath.join(self._urlPrefix, url)
-            # TODO Can we get requests to output debugging information when
-            # debugLevel > 0?
             authUrl = self._addAuth(fullUrl)
             response = requests.post(
                 authUrl, jsonString, headers=headers, verify=False)
             response.raise_for_status()
             jsonString = response.text
             self._bytesRead += len(jsonString)
-            if self._debugLevel > 1:
```
dcolligan: I'm not sure @jeromekelleher wanted this ripped out; I think #117 was about adding functionality in addition to this... maybe he can comment

ga4gh/cli.py
```diff
@@ -15,7 +15,7 @@
 import ga4gh.backend as backend
 import ga4gh.protocol as protocol
 import ga4gh.datamodel.variants as variants
-
+import json
```
dcolligan: Rearrange these imports according to the `STYLE.md` section on imports

ga4gh/client.py
MollyZhang: the debugLevel in fact does the same thing as #117, it prints out the response in detail. there is some overlap

ga4gh/cli.py
```diff
@@ -109,11 +109,24 @@ def _run(self, methodName, attrName=None):
...
-                print(result)
+            if self._verbosity == 1:
+                print(methodName, "request: \n", self._request)
+                print(methodName, "response: \n", result)
+            elif self._verbosity > 1:
+                requestJson = json.dumps(
+                    json.loads(self._request.toJSONString()),
+                    sort_keys=True, indent=4)
+                print(methodName, "request json string: \n", requestJson)
+                responseJson = json.dumps(
```
dcolligan: I think these dump/load calls are probably redundant if you have access to the string to begin with... why can't you just print `response.toJSONString()`?

ga4gh/client.py
jeromekelleher: Yes, I think I'd rather just leave this as it is until we do proper logging. To fix this issue, just add a quick print of the same form for the request when `self._debugLevel > 1`.

jeromekelleher:
Thanks for this @MollyZhang, I like the use of pprint here. I think this is slightly misunderstanding the purpose of output verbosity though: increasing the verbosity level always adds to the output, it never changes the primary output. Try playing with ssh -v, ssh -vv for example to get a feel for how it's usually done. I've added a new issue on adding client logging in #130, which should make this sort of thing very easy to do.

MollyZhang:
Thanks @jeromekelleher for the input, I removed the changes I made previously, and just added a quick print for the request when self._debugLevel > 1., the same thing you did in PR #125. 
Now when verbosity or debug level > 1, the client side will print out both the the request and response JSON message. 
As to the different level of -v and -vv for the variant output, should we just leave it as is until #130 for adding client logging is solved?

jeromekelleher:
Thanks @MollyZhang, this is perfect. I think we should leave the more sophisticated logging out until #130 is addressed. Would you mind squashing these commits down please?

+1 (post squash)

pashields:
+1 after squash as well

MollyZhang:
commits squashed to one.

PR #133, opened 2015-02-06
Add a suite of generators for avro schemas.

pashields:
These can be used by tests to generate random instances of classes, both valid and invalid. This mostly pulls out some work started by @jeromekelleher and tries to make it a little less hairy looking.

There are lots of directions to take this in (mostly adding more tweakability and an actual test script for longer scale generative testing), but I think it's useful enough as it is here to get it in.

Let me know if y'all think this is useful.


pashields:
Fixing flake8 issues

jeromekelleher:
This is _very_ tidy @pashields, I like it a lot! A couple of minor points:

1. Do we need a whole sub-package for this? I think it's probably enough to make a module called `test.utils` and make an `InstanceGenerator` class in there or something similar. The instance generator class can then be allocated with the parameters you want, and we can then just sample from this for as long as we like.
2. I think a floating point value between 0 and 1 is better for expressing probabilities --- I assume this is what chanceIncorrect=25 is about? (Is this getting called?  I can't see the definition for it).

pashields:
Thanks @jeromekelleher 

1. I'm good with all that. So would you put the generators and their associated functions in ```test.utils``` or would you leave them in their own module?
2. Ah, sorry, that was cruft. I added support for generating potentially incorrect models, but it was not a good API all around. Removed now.

jeromekelleher:
I'd probably put all the functions into test.utils - you could either make them methods of the `InstanceGenerator` class (or whatever you decide to call it) or make them module private functions. I think it's a good idea to keep the number of files down where possible, and only break out new module hierarchies where we really need it.

pashields:
Ok @jeromekelleher did a little refactoring.

I've left the generators public so that users can eventually their own versions with tweaked options. Lmk if this works for you.

tests/test_schemas.py
```diff
@@ -84,61 +71,10 @@ def getInvalidValue(self, cls, fieldName):
...
-            embeddedClass = cls.getEmbeddedType(fieldName)
-            ret = self.getRandomInstance(embeddedClass)
-        elif typ.type in self.randomValueMap:
-            ret = self.randomValueMap[typ.type]()
-        else:
-            raise Exception("schema assumptions violated")
-        return ret
+        t = field.type
+        if isinstance(t, avro.schema.UnionSchema):
+            t = t.schemas[1]
```
Naburimannu: Nit: this looks like it could be accidentally correct, but is probably good enough?
(e.g. starting with a union (Int, String), we'll try to generate an incorrect int, so we generate random other type, and our RNG selects a string?)

Naburimannu:
+1

tests/test_schemas.py
pashields: yeah, you are right. that's bogus. Let me fix that.

tests/utils.py
```diff
@@ -0,0 +1,133 @@
...
+        "boolean": booleanGenerator(),
+        "int": integerGenerator(),
+        "long": longGenerator(),
+        "string": stringGenerator(),
+        "float": floatGenerator(),
+        "double": doubleGenerator()
+    }
+
+    def __init__(self):
+        self.typeToGenerator = dict(InstanceGenerator.typeToGenerator)
```
jeromekelleher: Why are we allocating a per-instance version of `typeToGenerator`? Will a class variable not work?

pashields: it’s overly defensive for now. As I said, I’d like to make it possible for the caller to supply generators, so I was making a defensive copy.

It’s not required right now though, I will remove it.

pashields:
thanks for all the feedback. should be fixed up. mind taking one more look @jeromekelleher  @Naburimannu?

tests/utils.py
```diff
@@ -0,0 +1,131 @@
...
+    def generateInvalidateTypeValue(self, *typeStrings):
+        """
+        Returns a value that is invalid for the supplied type.
+        """
+        if any(t in self.typeToGenerator for t in typeStrings):
+            remaining = dict((typeString, gen) for typeString, gen
+                             in self.typeToGenerator.iteritems()
+                             if typeString not in typeStrings)
+            return random.choice(remaining.values())
+        else:  # Not a primative type, can't be a bool, so just return boolean
```
Naburimannu: Nit: spell primitive

tests/utils.py
```diff
@@ -0,0 +1,131 @@
...
+        "int": integerGenerator(),
+        "long": longGenerator(),
+        "string": stringGenerator(),
+        "float": floatGenerator(),
+        "double": doubleGenerator()
+    }
+
+    def generateInstance(self, cls):
+        """
+        Generate an instance of the supplied class
```
Naburimannu: Nit: PEP257 is a little unclear but can we end with a period?

Naburimannu: Or else just get rid of the docstring here since it's pretty redundant with the name...

tests/test_schemas.py
```diff
@@ -14,6 +14,7 @@
 
 import tests
 import ga4gh.protocol as protocol
+import tests.utils
 
 # TODO We should add some tests here - see comment from @pashields on
 # https://github.com/ga4gh/server/pull/39
```
Naburimannu: Does your CL obsolete this comment?

Naburimannu:
+1 modulo nits still holds.

tests/utils.py
pashields: Done

tests/utils.py
pashields: Removed

tests/test_schemas.py
pashields: I think there is more to do here, but I'm not sure how valuable this comment is at this point. What say you @jeromekelleher ?

tests/utils.py
```diff
@@ -0,0 +1,128 @@
+import itertools
```
jeromekelleher: Please add a module level docstring followed by the stand `__future__` imports. We should document the public functions and classes also.

tests/utils.py
```diff
@@ -0,0 +1,128 @@
...
+import random
+import string
+
+import avro.schema
+
+
+def valueFromSetGenerator(possibilities=[]):
+    while True:
+        yield random.choice(possibilities)
```
jeromekelleher: My only (slight) concern here is the problem of reproducibility. When we use a global random generator, the values we use for a given test will depend on the initial random seed, and the order and number of tests that we run. If we use an `InstanceGenerator` which has a private instance of a random generator that we seed, then we can run these tests in any order or combination and be guaranteed to get the same results. 

Not sure if this is worth worrying about, but I thought I'd bring it up in any case.

tests/utils.py
pashields: Added.

tests/utils.py
pashields: That would work if you created a generator for each test yes. I don't think it would be a problem to add the ability to do this, but I'd prefer not to block this PR on it. Would you like a TODO or issue? LMK what you think.

pashields:
@dcolligan you have any thoughts?

tests/test_schemas.py
jeromekelleher: This comment should probably be removed all right - it's not giving us much. Your call @pashields.

jeromekelleher:
I think this is good as it is - it would be nice to add an issue raising the reproducability problem, but it should be treated separately.

+1.

dcolligan:
Looks great... +1 after squashing

tests/test_schemas.py
pashields: removed.

PR #134, opened 2015-02-08
Datasets

jeromekelleher:
This is a very rough work-in-progress for adding Dataset handling. This should probably go before #121, as there are some fairly fundamental and disruptive changes proposed here. This is just a sketch so we can discuss the ideas more concretely, so I'd really appreciate any feedback people might have here. Here's major changes:

1. We change the server CLI to take a single positional argument which is the config file. Within the config file we have a list DATASETS, which contains a list of directories. Each directory is assumed to correspond to exactly one dataset, and has a prescribed format. For now, we just have a 'variants' sub directory, but we will also have a 'reads' directory too.
2. Within the variants sub-directory, each directory corresponds to a VariantSet. For now, we inspect the directory to decide whether we should allocate a WormtableVariantSet or a TabixVariantSet. I've outlined what I think is a better idea in the comments based on 'Shards' of the variant set.
3. I've introduced the concept of a `DataObject` which should be the superclass of all objects that represent an aggregation of data in some way. So, Dataset, VariantSet, ReadSet, ReferenceSet, all of these should be DataObjects. Hopefully we can push a lot of the tedious housekeeping and file system walking code up here. Hopefully this will help make testing a bit easier also.  I think we should explicitly model this as a hierarchy, as this might make the tricky iteration/page token problems a bit simpler, as we can then view it as a tree-traversal problem and maybe generalise the code.

Feedback appreciated!


diekhans:
stupid question: has GA4GH even defined what a dataset actually is?

jeromekelleher:
The definition is in `reads.avdl`, and there are back references from a bunch of other objects. However, the inconsistency of `dataset` vs `callSet`, `variantSet`, `referenceSet`, readGroupSet` is irritating...

jeromekelleher:
Ooops, just clobbered my own changes with a git push -f from a stale branch. Will reinstate them later  when I get out my laptop.

diekhans:
Hmm, I was under the impression that each data type had it's own
*Set object at the root. and that they are not shared and maybe
fundamentally different.

I think open source development works better when code is being
written rather than discussed on conference calls..

jeromekelleher:
My understanding (and I could be completely wrong) is that the dataset is a top level container for read and variant data. So we'd have
```
dataset1/
       + readGroups
             + readGroupSet1
             + readGroupSet2
             + ....
       + variantSets
             + variantSet1
             + ....
dataset2/
       + ....
```
The dataset container is then used to determine access privileges --- is user X entitled to read from dataset Y?

diekhans:
Jerome Kelleher <notifications@github.com> writes:
> My understanding (and I could be completely wrong) is that the dataset is a top
> level container for read and variant data. So we'd have

and so little is actually written down.  I guess the reference
manual is the minutes from the conference calls.

This seems very problematic.  What if you have multiple variant
sets produce independently? 

> The dataset container is then used to determine access privileges --- is user X
> entitled to read from dataset Y?

This is very wrong ;-)  Access needs to be determined by
functions applied at multiple levels, not by a fixed container.
For instances, in TCGA germ-line variants are protected, somatic
variants are not.  Then they decided that whole genome somatic
show be considered protected, but not exome ...

sigh ....

jeromekelleher:
Well, it looks like this is what we have for now. I'd vote on implementing it (even if it's just a straw man) and use this to figure out exactly what is wrong with the model. At least then we have something concrete.

Would someone else like to weigh in here? @adamnovak you've done some work with datasets, what's your understanding of how they work/what they do?

diekhans:
YES.  implementing is the best way to validate!!

I will dig more and start a ticket.

lh3:
`Dataset` has not been formally defined. Google said it is used for permission control and added to the GA4GH APIs. Practically, Google uses a dataset for a project (e.g. 1000g, PGP, SGDP, etc). I have always wanted to define dataset in a more precise way.

> This is very wrong ;-)  Access needs to be determined by functions applied at multiple levels, not by a fixed container. For instances, in TCGA germ-line variants are protected, somatic variants are not. Then they decided that whole genome somatic show be considered protected, but not exome ...

Under this dataset model, TCGA could be divided into multiple datasets: "TCGA germline", "TCGA somatic", etc.

> The definition is in reads.avdl, and there are back references from a bunch of other objects. However, the inconsistency of dataset vs callSet, variantSet, referenceSet, readGroupSet, etc. is irritating...

A few people at Broad have concerns over the naming `callSet`. We rarely call a sample as a callSet. I personally do not like `readGroupSet`.

> I think open source development works better when code is being written rather than discussed on conference calls..

Couldn't agree more.

> YES.  implementing is the best way to validate!!

I also vote to implement dataset.The way @jeromekelleher has described is proper as I understand.

pgrosu:
I think on [line 81 in variants.py](https://github.com/jeromekelleher/server/blob/datasets/ga4gh/datamodel/variants.py#L81), it probably should be `variantSetId` instead of `ariantSetId`.

Regarding dataset, I would go as far as the following post: 

https://github.com/ga4gh/schemas/pull/74

Regarding enabling a more modularized and parallel approach, the following would should provide that flexibility:

https://github.com/ga4gh/schemas/issues/143

Basically a dataset and/or project can be formed ad hoc, while others can be become gold standards.  The above I think should provide that flexibility and granularity.

Paul

diekhans:
Heng Li <notifications@github.com> writes:
> Dataset has not been formally defined.

nor has id ;-)

> Google said it is used for permission control and added to the
> GA4GH APIs. Practically, Google uses a dataset for a project
> (e.g. 1000g, PGP, SGDP, etc).

This is a simplification beyond the reality of the ugliness
of the data and the projects.

> I have always wanted to define dataset in a more precise way.

I am old fashioned, I want to define what things before you
specify the data structures.

It's seems problematic to have back pointers to `dataset' in
contained objects.  It reduce flexibility.  These things are
better answerable via queries, not stored in object.

> Under this dataset model, TCGA could be divided into multiple datasets: "TCGA
> germline", "TCGA somatic", etc.

This is a very serious flaw.  When policy changes, the entire
data collection has to be reorganized.  Access control needs to
be a function that takes user and the desired objects and
returns yes or no.  This leaves actually defining access control
policy outside of GA4GH.

>     YES. implementing is the best way to validate!!
> 
> I also vote to implement dataset.The way @jeromekelleher has described is
> proper as I understand.

and time for a PR.

jeromekelleher:
Thanks very much for the input on this @diekhans and @lh3. I appreciate that the `dataset` abstraction has many problems, especially with respect to access control. However, we need to implement some sort of access control,  `datasetId`s are already in the schema, and assigning access permissions to users based on `datasetId`s is simple and at least minimally useful. I hope we can implement this simple model first, and discuss how we can improve on it as we gain experience with its deficiencies.

diekhans:
As long as the access control isn't defined in the dataset
record, only the dataset id is a parameter to it.

But at the moment, I would be happy to have zero access control  and
more API implemented ;-)

jeromekelleher:
There are folks working on implementing OAuth in the reference server @diekhans --- hopefully, we can have access control and more API!

diekhans:
Jerome Kelleher <notifications@github.com> writes:
> There are folks working on implementing OAuth in the reference server @diekhans
> --- hopefully, we can have access control and more API!

Excellent!

My two-cents on how authorization should be implemented is by
functions that take a user and an object and return yes/no.
While something at the level of data set might work for many
cases, one can imagine all kinds of interesting filter.  For
instance, one could only allow access to certain regions of the
genome.   Given a man/woman a data structure, and they will ask for
enhancement, give the a function and they can build all kinds of
cool things.

jeromekelleher:
Agreed @diekhans, this is a great way to do it, but it needs fleshing out at the schema/security policy level. Let's keep it simple in terms of the implementation for now.

Anyone have any feedback on my code here? @pashields, @Naburimannu and @dcolligan this is a sort of shell of the structure I was talking about the other day. Any thoughts on how it's looking so far?

diekhans:
Jerome Kelleher <notifications@github.com> writes:
> Agreed @diekhans, this is a great way to do it, but it needs fleshing out at
> the schema/security policy level. Let's keep it simple in terms of the
> implementation for now.

My argument would be that access control function vs declarative
data structures are far simpler (although I wouldn't go near as
far as call them on anything lower than at *Set type of object
with many clean use cases).

> Anyone have any feedback on my code here? @pashields, @Naburimannu and
> @dcolligan this is a sort of shell of the structure I was talking about the
> other day. Any thoughts on how it's looking so far?

dcolligan:
As others have pointed out, enforcing access control at the dataset level would be simple, but coarse.  (Which is certainly a good characteristic of a POC implementation.)

Is it true that no object within a dataset can be shared across dataset?  That is, could a, say, variantSet be in two of the same datasets?  If so, then this model will duplicate potentially a lot of data.  (An alternative would be a model where every object had a top-level directory, and then instances named by id within it, i.e.:)
```
variantSets/
    variantSetId1/
    variantSetId2/
callSets/
...
```
But you guys have a better handle on the data model so take my comments with a salt shaker.

If we do end up going with this model, be sure to add what datasets we are using to our brand new html page!

diekhans:
Dataset is very ill-defined at this point.  We can help shape it!

jeromekelleher:
Closing this as #158 needs to be resolved first.

PR #136, opened 2015-02-09
Fix new flake8 errors

pashields:
Flake8 added some new errors, here's quick fixes. All travis builds will fail for now, so let's merge this quickly.

PR #137, opened 2015-02-09
Removed unnecessary flake8 exception.

jeromekelleher:
Since #136 we don't need to make an exception for flake8 import checks (using # NOQA instead, which is much better).


dcolligan:
+1

pashields:
+1

Issue #13, opened 2014-09-03
Need to implement callsets/search method


jeromekelleher:
We need to implement the `callsets/search` method. Currently, the concept of a `CallSet` does not exist; we just generate `Calls` and map  the  `callSetId` and `callSetName` (do we _really_ need both of these?) to the sample name in VCF.  Can we simply map `CallSet.id`, `CallSet.name` and `CallSet.sampleName` to the VCF sample name?

dcolligan:
Fixed in #140

PR #140, opened 2015-02-10
added callSet Search backend and unit tests

MollyZhang:
#13

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -31,6 +31,7 @@ class Backend(object):
                 self._variantSetIdMap[variantSetId] = variantSetClass(
                     variantSetId, relativePath)
         self._variantSetIds = sorted(self._variantSetIdMap.keys())
+
         # References
         self._referenceSetIdMap = {}
         referenceSetDir = os.path.join(self._dataDir, "references")
@@ -50,6 +51,22 @@ class Backend(object):
                 readGroupSet = reads.ReadGroupSet(readGroupSetId, relativePath)
                 self._readGroupSetIdMap[readGroupSetId] = readGroupSet
         self._readGroupSetIds = sorted(self._readGroupSetIdMap.keys())
+
+        # callSets
+        self._callSetIdMap = {}
+        for variantSet in self._variantSetIdMap.values():
+            sampleNames = variantSet.getSampleNames()
+            variantSetId = variantSet.getVariantSetId()
+            for name in sampleNames:
+                if name not in self._callSetIdMap:
+                    self._callSetIdMap[name] = []
+                self._callSetIdMap[name].append(variantSetId)
+
+    def getVariantSetIdMap(self):
+        return self._variantSetIdMap
+
+    def getCallSetIdMap(self):
+        return self._callSetIdMap
 
     def parsePageToken(self, pageToken, numValues):
         """
@@ -156,6 +173,16 @@ class Backend(object):
             protocol.GASearchVariantsResponse, "variants",
             self.variantsGenerator)
 
+    def searchCallSets(self, request):
+        """
+        Returns a GASearchCallSetsResponse for the specified
+        GASearchCallSetsRequest Object.
+        """
+        return self.runSearchRequest(
+            request, protocol.GASearchCallSetsRequest,
+            protocol.GASearchCallSetsResponse, "callSets",
+            self.callSetsGenerator)
+
     # Iterators over the data hieararchy
 
     def _topLevelObjectGenerator(self, request, idMap, idList):
@@ -197,7 +224,7 @@ class Backend(object):
         by the specified request.
         """
         return self._topLevelObjectGenerator(
-                request, self._variantSetIdMap, self._variantSetIds)
+            request, self._variantSetIdMap, self._variantSetIds)
 
     def variantsGenerator(self, request):
         """
@@ -221,6 +248,37 @@ class Backend(object):
                     nextPageToken = "{0}:{1}".format(
                         variantSetIndex, variant.start + 1)
                     yield variant, nextPageToken
+
+    def callSetsGenerator(self, request):
+        """
+        Returns a generator over the (callSet, nextPageToken) pairs defined by
+        the specified request.
+        """
+        # if no variantSetIds are input from client,
+        # then set variantSetIds to all variantSetIds
+        if request.variantSetIds == []:
+            variantSetIds = self._variantSetIds
+        else:
+            variantSetIds = request.variantSetIds
+        name = request.name
+        if request.pageToken is not None:
+            startVariantSetIndex, startCallSetPosition = self.parsePageToken(
+                request.pageToken, 2)
+        else:
+            startVariantSetIndex = 0
+            startCallSetPosition = 0
+
+        for variantSetIndex in range(startVariantSetIndex, len(variantSetIds)):
+            variantSetId = variantSetIds[variantSetIndex]
+            if variantSetId in self._variantSetIdMap:
+                variantSet = self._variantSetIdMap[variantSetId]
+                callSetIterator = variantSet.getCallSets(
+                    name, startCallSetPosition)
+                for callSet, callSetPosition in callSetIterator:
+                    callSet.variantSetIds = self._callSetIdMap[callSet.name]
+                    nextPageToken = "{0}:{1}".format(
+                        variantSetIndex, callSetPosition + 1)
+                    yield callSet, nextPageToken
 
     def startProfile(self):
         pass
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -139,6 +139,12 @@ class WormtableVariantSet(VariantSet):
                     self._sampleNames.append(sampleName)
                 self._sampleCols[sampleName].append((infoName, col))
 
+    def getSampleNames(self):
+        return self._sampleNames
+
+    def getVariantSetId(self):
+        return self._variantSetId
+
     def convertInfoField(self, value):
         """
         Converts the specified value into an appropriate format for a protocol
@@ -252,6 +258,32 @@ class WormtableVariantSet(VariantSet):
                     yield self.convertVariant(row, sampleRowPositions)
                 else:
                     break
+
+    def getCallSets(self, name, startPosition):
+        """
+        Returns an iterator over the specified callset name. The parameters
+        correspond to the attributes of a GASearchCallSetsReuqest object.
+        """
+        # TODO: implement name string search after semantics is clarified
+        if name is not None:
+            raise NotImplementedError()
+        else:
+            callSetIds = self._sampleNames[startPosition:]
+            for i in range(len(callSetIds)):
+                yield self.convertCallSet(callSetIds[i]), i+startPosition
+
+    def convertCallSet(self, callSetId):
+        """
+        Converts the specified wormtable column into a GACallSet object.
+        """
+        callSet = protocol.GACallSet()
+        callSet.created = self._creationTime
+        callSet.updated = self._updatedTime
+        # TODO clarifying between id, name and sampleID in appropriate dataset
+        callSet.id = "{0}.{1}".format(self._variantSetId, callSetId)
+        callSet.name = callSetId
+        callSet.sampleId = callSetId
+        return callSet
 
     def getMetadata(self):
         """
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -185,7 +185,8 @@ def getReferenceSet(version, id):
 
 @app.route('/<version>/callsets/search', methods=['POST'])
 def searchCallSets(version):
-    raise frontendExceptions.PathNotFoundException()
+    return handleFlaskPostRequest(
+        version, flask.request, app.backend.searchCallSets)
 
 
 @app.route('/<version>/readgroupsets/search', methods=['POST'])
--- a/tests/unit/test_backends.py
+++ b/tests/unit/test_backends.py
@@ -146,6 +146,12 @@ class TestWormtableBackend(unittest.TestCase):
             request, pageSize, self._backend.searchVariantSets,
             protocol.GASearchVariantSetsResponse, "variantSets")
 
+    def getVariantSetIds(self):
+        """
+        Return all variantSetIds.
+        """
+        return [variantSet.id for variantSet in self.getVariantSets()]
+
     def getVariants(
             self, variantSetIds, referenceName, start=0, end=2 ** 32,
             pageSize=100, callSetIds=[]):
@@ -162,6 +168,16 @@ class TestWormtableBackend(unittest.TestCase):
         return self.resultIterator(
             request, pageSize, self._backend.searchVariants,
             protocol.GASearchVariantsResponse, "variants")
+
+    def getCallSets(self, variantSetId, pageSize=100):
+        """
+        Returns an iterator over the callsets in a specified variant set.
+        """
+        request = protocol.GASearchCallSetsRequest()
+        request.variantSetIds = [variantSetId]
+        return self.resultIterator(
+            request, pageSize, self._backend.searchCallSets,
+            protocol.GASearchCallSetsResponse, "callSets")
 
     def getReferenceNames(self, variantSetId):
         """
@@ -220,16 +236,16 @@ class TestWormtableBackend(unittest.TestCase):
         last = index.max_key(referenceName)
         return first[1], last[1]
 
-    def getCallSetIds(self, variantSetId):
+    def getWormtableCallSetIds(self, variantSetId):
         """
         Returns the set of callSetIds derived from the wormtable header.
         """
         table = self._tables[variantSetId]
-        callSetIds = set()
+        callSetIds = list()
         for col in table.columns()[8:]:  # skip VCF fixed cols
-            name = col.get_name()
-            if not name.startswith("INFO"):
-                callSetIds.add(name.split(".")[0])
+            name = col.get_name().split(".")[0]
+            if not name.startswith("INFO") and name not in callSetIds:
+                callSetIds.append(name.split(".")[0])
         return callSetIds
 
 
@@ -463,7 +479,7 @@ class TestVariants(TestWormtableBackend):
 
     def testSearchByCallSetIds(self):
         for variantSet in self.getVariantSets():
-            callSetIds = self.getCallSetIds(variantSet.id)
+            callSetIds = self.getWormtableCallSetIds(variantSet.id)
             # limit the subsets we check over to some small value.
             for subset in utils.powerset(callSetIds, 20):
                 for referenceName in self.getReferenceNames(variantSet.id):
@@ -480,8 +496,7 @@ class TestVariants(TestWormtableBackend):
                     ids.add(variant.id)
 
     def testAcrossVariantSets(self):
-        allVariantSets = [
-            variantSet.id for variantSet in self.getVariantSets()]
+        allVariantSets = self.getVariantSetIds()
         for permLen in range(1, len(allVariantSets) + 1):
             for variantSets in itertools.permutations(allVariantSets, permLen):
                 for referenceName in self.getCommonRefNames(variantSets):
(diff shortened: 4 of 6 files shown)


ga4gh/backend.py
```diff
@@ -26,6 +27,16 @@ def __init__(self, dataDir, variantSetClass):
                 self._variantSetIdMap[variantSetId] = variantSetClass(
                     variantSetId, relativePath)
         self._variantSetIds = sorted(self._variantSetIdMap.keys())
+        # self._callSetIdMap map each callSetId to all variantSets
+        # which this callSetId has calls in
+        for variantSetId in self._variantSetIds:
```
jeromekelleher: It would be more natural to use
```python
for variantSet in self._variantSetIdMap.values():
     sampleNames = variantSet.getSampleNames()
     # etc.
```
We should add the method `getSampleNames()` to VariantSet because accessing 'protected' instance variables is a bad idea. (In Python, the convention is that `__instanceVariable` is private to a class and `_instanceVariable` is private to a class and all of its subclasses.)

ga4gh/backend.py
```diff
@@ -26,6 +27,16 @@ def __init__(self, dataDir, variantSetClass):
...
+            sampleNames = self._variantSetIdMap[variantSetId]._sampleNames
+            for name in sampleNames:
+                if name not in self._callSetIdMap:
```
jeromekelleher: A neater approach to this is:
```python
if name not in self._callSetIdMap:
     self._callSetIdMap[name] = []
self._callSetIdMap[name].append(variantSetId)
```

ga4gh/backend.py
```diff
@@ -141,6 +162,38 @@ def startProfile(self):
...
+        the specified request.
+        """
+        # if no variantSetIds are input from client,
+        # then set variantSetIds to all variantSetIds
+        if request.variantSetIds == []:
+            variantSetIds = self._variantSetIds
+        else:
+            variantSetIds = request.variantSetIds
+        name = request.name
+
```
jeromekelleher: I don't think we need the blank line here -- we only need extra whitespace in a function when we really need to emphasise two distinct parts of the function. It's pretty clear what's happening here. Likewise below.

ga4gh/datamodel/variants.py
```diff
@@ -238,6 +238,37 @@ def getVariants(self, referenceName, startPosition, endPosition,
                 else:
                     break
 
+    def getCallSets(self, name, startCallSetPosition):
+        """
+        Returns an iterator over the specified callset name. The parameters
+        correspond to the attributes of a GASearchCallSetsReuqest object.
+        """
+        assert name is not None
+
```
jeromekelleher: delete extra whitespace in this function too.

ga4gh/datamodel/variants.py
```diff
@@ -238,6 +238,37 @@ def getVariants(self, referenceName, startPosition, endPosition,
...
+        callSetIds = self._sampleNames[startCallSetPosition:]
+        # get all substrings of name in a list nameStrings
+        l = len(name)
+        nameStrings = ([name[i:j + 1]
+                        for i in xrange(l) for j in xrange(i, l)])
```
jeromekelleher: Use `range` instead` of `xrange`. xrange is removed in Python 3.

tests/test_backends.py
```diff
@@ -138,6 +138,15 @@ def getVariantSets(self, pageSize=100):
             request, pageSize, self._backend.searchVariantSets,
             protocol.GASearchVariantSetsResponse, "variantSets")
 
+    def getVariantSetIds(self):
+        """
+        return all variantSetIds
+        """
+        variantSetIds = []
```
jeromekelleher: A list comprehension would probably be cleaner here:
```python
return [variantSet.id for variantSet in self.getVariantSets()]
```

ga4gh/frontend.py
```diff
@@ -134,7 +134,10 @@ def getReferenceSet(id):
 
 @app.route('/callsets/search', methods=['POST'])
 def searchCallSets():
-    raise frontendExceptions.PathNotFoundException()
+    if flask.request.method == "POST":
```
dcolligan: It might be good to do some refactoring here if we're going to have this same if POST then handleHttpPost else handleHttpOptions pattern in all of our endpoints.  I'm not sure if we want to do that in this PR or another one.

tests/test_end_to_end.py
```diff
@@ -101,4 +101,54 @@ def testVariantsSearch(self):
 
         # TODO: Add test cases for other methods when they are implemented.
 
+    def testCallSetsSearch(self):
```
dcolligan: This method is way too long.  If each thing you're testing doesn't depend on the results of the previous test, it would be good to break each of them out into their own method.

tests/test_backends.py
```diff
@@ -587,3 +595,57 @@ def testAcrossVariantSets(self):
...
                             variantSets, referenceName, 0, 2 ** 32,
                             pageSize=pageSize)
+
+
+class TestCallSets(TestWormtableBackend):
+    """
+    Tests the searchCallSets end point
+    """
+    def testCallSetIds(self):
+        """
```
jeromekelleher: Unit test functions shouldn't have docstring comments --- I'm updating the style document now with some examples and the rationale for this.

jeromekelleher:
This is great @MollyZhang, thanks! There's a problem with method name changes that have happened since you made your branch (toJSONString is now toJsonString etc) so you'll need to rebase and fix the problems. Because of this I haven't been able to try it out, but it looks very nice to me. You've done a great job including all of the tests, so I don't think there'll be any problems merging this once you've updated and fixed the issues outlined above.

ga4gh/frontend.py
MollyZhang: I made a new issue for this

tests/test_end_to_end.py
MollyZhang: I refactored the code it's a lot shorter now.

MollyZhang:
Thanks @jeromekelleher  and @dcolligan, I updated the code based on your comments.

ga4gh/backend.py
```diff
@@ -141,6 +161,37 @@ def startProfile(self):
...
+                request.pageToken, 2)
+        else:
+            startVariantSetIndex = 0
+            startCallSetPosition = 0
+
+        for variantSetIndex in range(startVariantSetIndex, len(variantSetIds)):
+            variantSetId = variantSetIds[variantSetIndex]
+            if variantSetId in self._variantSetIdMap:
+                variantSet = self._variantSetIdMap[variantSetId]
+                callSetIterator = variantSet.getCallSets(name,
```
dcolligan: just as a matter of style I think this is easier to read like this (and easier to maintain if/when extra arguments are to be added):

```
                callSetIterator = variantSet.getCallSets(
                    name, startCallSetPosition)
```

ga4gh/datamodel/variants.py
```diff
@@ -238,6 +244,35 @@ def getVariants(self, referenceName, startPosition, endPosition,
                 else:
                     break
 
+    def getCallSets(self, name, startCallSetPosition):
+        """
+        Returns an iterator over the specified callset name. The parameters
+        correspond to the attributes of a GASearchCallSetsReuqest object.
+        """
+        assert name is not None
```
dcolligan: Take out this assert and throw an exception here instead.  You can just create an Exception("your msg here") object with a TODO comment that says to use a better-thought-out exception class later.

ga4gh/datamodel/variants.py
```diff
@@ -238,6 +244,35 @@ def getVariants(self, referenceName, startPosition, endPosition,
...
+        callSetIds = self._sampleNames[startCallSetPosition:]
+        # get all substrings of name in a list nameStrings
+        l = len(name)
```
dcolligan: we don't use single-letter variable names, so change this to something like nameLength

tests/test_end_to_end.py
```diff
@@ -27,64 +27,64 @@ def setUp(self):
...
         self._wtTestFixture.tearDown()
 
-    def sendJSONPostRequest(self, path, data):
-        return self.app.post(path,
-                             headers={'Content-type': 'application/json'},
-                             data=data)
+    def sendJsonPostRequest(self, path, request):
+        """send json post, assert status code to be 200,
+        and return response data"""
+        response = self.app.post(path,
```
dcolligan: Same comment about indentation style here

tests/test_end_to_end.py
```diff
@@ -27,64 +27,64 @@ def setUp(self):
...
+                                 headers={'Content-type': 'application/json'},
+                                 data=request.toJsonString())
+        self.assertEqual(200, response.status_code)
+        if path == "/callsets/search":
+            responseData = protocol.GASearchCallSetsResponse.\
```
dcolligan: Again, I'd rather see this formatted as:
```
responseData = protocol. GASearchCallSetsResponse.fromJsonString(
    response.data)
```
(unless it can't fit on that line, in which case what you had was fine)

tests/test_end_to_end.py
MollyZhang: if I change to the style you suggested, the flake8 test will fail, and say this:
tests/test_end_to_end.py:40:13: E128 continuation line under-indented for visual indent

dcolligan: You're not wrapping the lines correctly.  Try:
```
response = self.app.post(
    path, headers={'Content-type': 'application/json'}, data=request.toJsonString())
```

MollyZhang: got it, thanks

MollyZhang:
thanks @dcolligan I changed all the lines you commented on

dcolligan:
Cool, I'm +1 if @jeromekelleher is

ga4gh/datamodel/variants.py
```diff
@@ -238,6 +244,37 @@ def getVariants(self, referenceName, startPosition, endPosition,
...
                     break
 
+    def getCallSets(self, name, startCallSetPosition):
+        """
+        Returns an iterator over the specified callset name. The parameters
+        correspond to the attributes of a GASearchCallSetsReuqest object.
+        """
+        # TODO: use better thought-out exception class later
+        if name is None:
+            raise("callset name can not be None")
```
jeromekelleher: Please use raise `raise Exception("..")`.

ga4gh/backend.py
```diff
@@ -141,6 +161,37 @@ def startProfile(self):
     def endProfile(self):
         pass
 
+    def callSetsGenerator(self, request):
```
jeromekelleher: It would be a bit neater to keep the `XGenerator` methods group together, and not have the startProfile and endProfile method sandwiched in between.

ga4gh/backend.py
```diff
@@ -141,6 +161,37 @@ def startProfile(self):
...
+
+        for variantSetIndex in range(startVariantSetIndex, len(variantSetIds)):
+            variantSetId = variantSetIds[variantSetIndex]
+            if variantSetId in self._variantSetIdMap:
+                variantSet = self._variantSetIdMap[variantSetId]
+                callSetIterator = variantSet.getCallSets(
+                    name, startCallSetPosition)
+                for callSet, callSetPosition in callSetIterator:
+                    callSet.variantSetIds = self._callSetIdMap[callSet.id]
+                    nextPageToken = "{0}:{1}".format(
```
jeromekelleher: We return an empty page of results at the end here. We need to return `None` as the `nextPageToken` when we are at the last result.

jeromekelleher:
Thanks for the updates @MollyZhang. We're getting closer, but I don't think we've got the required semantics. We seem to only have exact matching on names right now, where as we should have
1. If name is null, return all callsets;
2. if name is non-null, "return call sets for which a substring of the name matches this string."

So, if I specify name=HG then I should return all call sets that have names containing "HG".

ga4gh/datamodel/variants.py
```diff
@@ -238,6 +244,37 @@ def getVariants(self, referenceName, startPosition, endPosition,
...
+
+    def convertCallSet(self, callSetId):
+        """
+        Converts the specified wormtable column into a GACallSet object
+        """
+        callSet = protocol.GACallSet()
+        callSet.created = self._creationTime
+        callSet.updated = self._updatedTime
+        # TODO clarifying between id, name and sampleID in appropriate dataset
+        callSet.id = callSetId
```
jeromekelleher: This should be unique within the repository.  See e.g. `variant.id` in convertVariant.

tests/test_backends.py
```diff
@@ -587,3 +592,45 @@ def testAcrossVariantSets(self):
...
+
+    def testUniqueCallSetIds(self):
+        # verify that all callSetIds in each variantSet are unique
+        for variantSetId in self.getVariantSetIds():
+            callSetIds = set()
+            for callSetId in self.getCallSetIds(variantSetId):
+                self.assertTrue(callSetId not in callSetIds)
+                callSetIds.add(callSetId)
+
+    def testUniqueCallSetIdsAcrossVariantSets(self):
```
jeromekelleher: These two tests should be combined, as IDs should be unique repository wide.

tests/test_backends.py
```diff
@@ -587,3 +592,45 @@ def testAcrossVariantSets(self):
...
+
+    def testVariantSetIds(self):
+        # test the VariantSetId field of GACallSet object is correct
+        allCallSetIdMap = dict()
+        for variantSetId in self.getVariantSetIds():
+            for callSetId in self.getCallSetIds(variantSetId):
+                if callSetId not in allCallSetIdMap:
+                    allCallSetIdMap[callSetId] = []
+                allCallSetIdMap[callSetId].append(variantSetId)
+        wormtableCallSetIdMap = self._backend._callSetIdMap
```
jeromekelleher: We're breaking isolation here: we should never access any variable starting with an `_` from outside its own class or one of its subclasses. Do we really need to build up a map of all the callSetIds in the system to check this in any case? Wouldn't it be simpler to get each callSet in turn and just verify that its variantSetId is correct?

tests/test_backends.py
```diff
@@ -587,3 +592,45 @@ def testAcrossVariantSets(self):
...
+
+class TestCallSets(TestWormtableBackend):
+    """
+    Tests the searchCallSets end point
+    """
+    def testCallSetIds(self):
+        # verify that callSetIds are equal to sample columns in wormtable
+        for variantSetId in self.getVariantSetIds():
+            callSetIds = self.getCallSetIds(variantSetId)
+            variantSets = self._backend._variantSetIdMap[variantSetId]
```
jeromekelleher: We're also breaking isolation here. If we need to get these values we should add a method to do it.

jeromekelleher:
We're not really testing the search semantics here either @MollyZhang --- we'd need to have a fairly thorough test coverage of the actual search process here. This will be a bit tricky since we're working with substring search.

ga4gh/backend.py
MollyZhang: @jeromekelleher, A empty response with `None` as the `nextPageToken` is always returned as the last result, it is at runSearchRequest() in ga4gh/backend.py,  Whenever an empty search result is returned, nextPageToken is default to be None, and that's the last response.

MollyZhang:
@jeromekelleher The API definition of a callsets/search is problematic (see line 142 of variantmethods.avdl). It says "Only return call sets for which a substring of the name matches this string". If we follow this instruction, basically if we search for `HG00167`, since `H`, `G`, `0` etc are all substrings of this name, then all callsets that contain letter `H` or `G` or `0`, etc will all be returned, like `HG9999` or `NAT0872` etc will all be returned. That cannot be what the user wanted when he/she typed `HG00167`. I think what the API meant is that `Only return call sets for which the name matches a substring of these callsets names`, in that case, if we search `HG` it will return all callsets with `HG` in it.

Naburimannu:
@MollyZhang I think the actual meaning of the words you've quoted from the spec is exactly your preferred meaning: "a substring of the *name* [of the callset] matches this string". There's some ambiguity between the name-of-the-callset and the name-parameter-to-the query, but "this string" can't be referring to the callset, it has to be referring to the parameter to the query.

MollyZhang:
@Naburimannu That makes sense but the original sentence does sound ambiguous, I think adding a few words like `Only return call sets for which a substring of the call set name matches query string` will make it total unambiguous.

pgrosu:
@MollyZhang, this is just a simple conjunction to add more filtering on the results.  If you have an array of `variantSetIds` and a substring in `name`, then it will perform basically the same functionality as this SQL query:

```
SELECT * from CallSet 
WHERE 
CallSet.variantSetIds IN variantSetIds 
AND 
CallSet.name LIKE CONCAT('%', name, '%'); 
```
~p

diekhans:
I just submitted a pull request to change the name searches to
be exact match.  Substring is both not very useful and very
error prone.  We will probably want a full regular expression
search filter at some point.

jeromekelleher:
I'm with @diekhans here, substring search is of limited utility and dangerous. I'd suggest for now that we only support the name=null branch, and raise a NotImplemented if a string is passed. We can then wait for the semantics to be clarified upstream and still correctly implement a subset of the protocol as it is defined.

MollyZhang:
@jeromekelleher  All right now I only implemented when name = null, in that case all callsets will be returned. An exception "not implemented" is raised if any string is passed. 
I changed all other comments you had on the code previously as well.

Naburimannu:
Need to defer to subject matter experts on substring search.

ga4gh/datamodel/variants.py
```diff
@@ -238,6 +244,33 @@ def getVariants(self, referenceName, startPosition, endPosition,
...
 
+    def getCallSets(self, name, startPosition):
+        """
+        Returns an iterator over the specified callset name. The parameters
+        correspond to the attributes of a GASearchCallSetsReuqest object.
+        """
+        # TODO: use better thought-out exception class later
+        # TODO: implement name string search after semantics is clarified
+        if name is not None:
+            raise Exception("Not Implemented")
```
jeromekelleher: Change this to NotImplementedError() please.

tests/test_end_to_end.py
Naburimannu: This is still two tests, not one.

pgrosu:
These types of searches should be cached out separately as inverted indices for faster retrieval.  It is part of the Information Retrieval domain and basically how search engines get designed - and can easily be parallelized.  Below are a few link to my posts regarding this approach:

https://github.com/ga4gh/schemas/issues/154#issuecomment-56716514

https://github.com/ga4gh/schemas/issues/142#issuecomment-55518571

https://github.com/ga4gh/schemas/issues/144#issuecomment-55753407

It will even work on substrings if they are suffix trees.

Paul

README.txt
```diff
@@ -97,6 +97,12 @@ has variant name ``rs75454623``::
 
     $ ga4gh_client variants-search http://localhost:8000 -V 1000g_2013 -r 1 -e 60000 -n rs75454623  | less -S
 
+we can also run the "callset/search" method over all the variant sets in ga4gh-example-data,
```
jeromekelleher: I think we can remove this paragraph --- it's not essential to get people started.

jeromekelleher:
Can you update this PR and fix the test failures please @MollyZhang?

MollyZhang:
@jeromekelleher I updated the PR addressing your comments, added more backend tests, fixed the test failures, and rebased it. Thanks!

ga4gh/datamodel/variants.py
```diff
@@ -253,6 +259,33 @@ def getVariants(self, referenceName, startPosition, endPosition,
                 else:
                     break
 
+    def getCallSets(self, name, startPosition):
+        """
+        Returns an iterator over the specified callset name. The parameters
+        correspond to the attributes of a GASearchCallSetsReuqest object.
+        """
+        # TODO: use better thought-out exception class later
```
jeromekelleher: You can delete this TODO now - this is the correct exception.

tests/unit/test_backends.py
```diff
@@ -220,16 +234,16 @@ def getStartBounds(self, variantSetId, referenceName):
...
         """
         table = self._tables[variantSetId]
-        callSetIds = set()
+        callSetIds = list()
         for col in table.columns()[8:]:  # skip VCF fixed cols
-            name = col.get_name()
-            if not name.startswith("INFO"):
-                callSetIds.add(name.split(".")[0])
+            name = col.get_name().split(".")[0]
+            if (not name.startswith("INFO")) and (name not in callSetIds):
```
jeromekelleher: These brackets aren't necessary.

tests/unit/test_backends.py
```diff
@@ -146,6 +146,12 @@ def getVariantSets(self, pageSize=100):
             request, pageSize, self._backend.searchVariantSets,
             protocol.GASearchVariantSetsResponse, "variantSets")
 
+    def getVariantSetIds(self):
+        """
+        return all variantSetIds
```
jeromekelleher: This is pedantic, I know, but could you capitalise the first word and end the sentence with a full stop please?

MollyZhang:
@jeromekelleher fixed them

jeromekelleher:
There's a few more documentation issues; can you check that all the docstrings are in the required format?

MollyZhang:
@jeromekelleher  I updated all the docstrings to the required format

jeromekelleher:
Excellent, thanks @MollyZhang. 

+1 (after the commits are squashed).

MollyZhang:
commits squashed

PR #141, opened 2015-02-10
Add CPU profiling to server_benchmark

Naburimannu:
Use cProfile to report top 25% of CPU-time-consuming python
functions.

Addresses https://github.com/ga4gh/server/issues/46

---

Just the diagnostic code without any of the ideas I explored in https://github.com/ga4gh/server/issues/46.
Snippet at the end suggests we might add Backend.reportProfile(), but since this is isolated to a development script rather than deployed production code I don't think that's necessary until we have a second profiler that would use it, and I don't know what other profilers we might add.

### DIFF
--- a/server_benchmark.py
+++ b/server_benchmark.py
@@ -9,7 +9,9 @@ from __future__ import unicode_literals
 
 import re
 import time
+import pstats
 import argparse
+import cProfile
 
 import ga4gh.backend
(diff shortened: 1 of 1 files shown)


jeromekelleher:
+1

diekhans:
+1

PR #144, opened 2015-02-10
Create requestFactory and refactor cli to use it

### LINKED ISSUE #83: Client side conversion tools required

dcolligan:
A refactor in preparation for there being multiple code paths that use command line arguments to instantiate GAWhateverRequests.

Does not close #83

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -17,6 +17,106 @@ import ga4gh.protocol as protocol
 import ga4gh.datamodel.variants as variants
 
 
+def setCommaSeparatedAttribute(request, args, attr):
+    attribute = getattr(args, attr)
+    if attribute is not None:
+        setattr(request, attr, attribute.split(","))
+
+
+class RequestFactory(object):
+    """
+    Provides methods for easy inititalization of request objects
+    """
+    class GASearchReadsRequestGoogle(protocol.ProtocolElement):
+
+        __slots__ = ['end', 'pageSize', 'pageToken', 'readGroupIds',
+                     'referenceName', 'start']
+
+        def __init__(self):
+            self.end = None
+            self.pageSize = None
+            self.pageToken = None
+            self.readGroupIds = []
+            self.referenceName = None
+            self.start = 0
+
+    def __init__(self, args):
+        self.args = args
+        self.workarounds = set(self.args.workarounds.split(','))
+
+    def usingWorkaroundsFor(self, workaround):
+        """
+        Returns true if we are using the passed-in workaround
+        """
+        return workaround in self.workarounds
+
+    def createGASearchVariantSetsRequest(self):
+        request = protocol.GASearchVariantSetsRequest()
+        setCommaSeparatedAttribute(request, self.args, 'datasetIds')
+        request.pageSize = self.args.pageSize
+        request.pageToken = None
+        return request
+
+    def createGASearchVariantsRequest(self):
+        request = protocol.GASearchVariantsRequest()
+        request.referenceName = self.args.referenceName
+        request.variantName = self.args.variantName
+        request.start = self.args.start
+        request.end = self.args.end
+        if self.usingWorkaroundsFor(client.HttpClient.workaroundGoogle):
+            request.maxCalls = self.args.maxCalls
+        if self.args.callSetIds == []:
+            request.callSetIds = []
+        elif self.args.callSetIds == '*':
+            request.callSetIds = None
+        else:
+            request.callSetIds = self.args.callSetIds.split(",")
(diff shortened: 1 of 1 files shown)


Naburimannu:
This is adding a bit of indirection that isn't obvious. Can you give us an example of what the new use case is that drives it? (Admittedly I haven't reviewed all the outstanding PRs, but if it's in one of those then a reference to it in this PR would be fitting.)

jeromekelleher:
I think this is a good addition, but I'm not sure we need a whole new module for it. Also, `ProtocolFactory` seems overly general, since we're really just creating `XRequest` objects. I think it should be renamed `RequestFactory` and moved into the `ga4gh.cli` module, as really it's just about making turning command line arguments into `XRequest` objects.

Other than that, it looks good.

dcolligan:
@jeromekelleher pushed a commit rebased on top of master and addressing your concerns

@Naburimannu sure, this is in preparation for when other modules (I have ga2sam and ga2vcf in mind) need to create server requests... it just makes it so that we don't have to do the initialization logic in two places, once in the WhateverSearchRunner and once in whatever other module wants to create a GAWhateverRequest from the same cmd line args

jeromekelleher:
Looks great, thanks @dcolligan.

+1.

dcolligan:
I should note: once any other module (say, ga2vcf) makes use of RequestFactory, it should be put into its own module since, if it is not, ga2vcf will depend on cli and cli will depend on ga2vcf -- a circular module dependency (bad!).

jeromekelleher:
Actually, looking at this again in more detail, I'm not so sure about it. There's some nasty coupling between the calling objects and the factory.  These factory functions are useful, but should just be methods of `AbstractQueryRunner`.

Retracting my +1; sorry @dcolligan, I should have looked at this in more detail earlier.

ga4gh/cli.py
```diff
@@ -16,6 +16,109 @@
...
+        request.referenceId = args.referenceId
+        request.referenceName = args.referenceName
+        return request
+
+    @classmethod
+    def createGAListReferenceBasesRequest(cls, host, args):
+        request = protocol.GAListReferenceBasesRequest()
+        request.start = args.start
+        request.end = args.end
+        host._id = args.id
```
jeromekelleher: This is a violation of the isolation rules. If we need to set this `id` value, then this should be a method of `AbstractSearchRunner`, so that we can legitimately access `self._id`.

dcolligan: `host` may also be something else, not just a Runner class

dcolligan:
@jeromekelleher I disagree that they should be methods of anything in `cli.py` because that would create a circular module reference when classes outside `cli.py` need to use them... how do you propose resolving this?

jeromekelleher:
I'm not convinced we actually need them anywhere except the cli though --- see my comments about how we structure ga2vcf.

Naburimannu:
@dcolligan  Sorry for being dense, but for this (and for some of your other refactorings) I just don't have enough of an understanding to be able to evaluate the plans without really, really concrete examples.

dcolligan:
@Naburimannu ok, fair enough.   See #152 for where I'm going with this.

dcolligan:
Added a new commit with the following changes:
- puts requestFactory in its own module (because of the circular wiring problem)
- eliminates the `host` argument to all of requestFactory's methods (to remove coupling with caller)
- changes the requestFactory's classmethods to methods (to accomodate workarounds)

jeromekelleher:
Where is the circular import problem? I thought we were only using this in `cli.py`. We don't need a factory in `VcfConverter` as we can very simply derive the extra requests (`SearchVariantSets` and `SearchCallSets`)  we need from the `SearchVariantsRequest` used as a parameter. Can we not do something analogous for reads?

dcolligan:
Whoops, forgot about that decision.  Pushed an updated commit which puts the RequestFactory in the cli module.

dcolligan:
It looks like the build was failing because the newset version of flake added warnings that were not present in the previous version.  I just pushed a commit addressing these.

It does suck a bit that the libraries we depend upon can change out form under us during testing; on the plus side we always get the latest and greatest packages.  If this becomes too much of a problem, we can always fix versions in our `requirements.txt` file.

To get the latest version of flake, run:

`pip install --upgrade flake`

jeromekelleher:
Thanks @dcolligan, this looks good. This is definitely an improvement over what we have, as we're collecting all the code for creating these requests into one place, and simplifying the runner classes. 

+1.

Naburimannu:
+1. (My preference would be to +0, but since we're autogenerating the protocol implementation classes this seems like this is the best implementation we can come up with.)

PR #145, opened 2015-02-10
Refactor cli client methods

### LINKED ISSUE #83: Client side conversion tools required

dcolligan:
Refactor in preparation for reuse of a lot of the client cli functionality by the conversion tools

Does not close #83

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -16,6 +16,7 @@ import ga4gh.backend as backend
 import ga4gh.protocol as protocol
 import ga4gh.datamodel.variants as variants
 
+
 ##############################################################################
 # Server
 ##############################################################################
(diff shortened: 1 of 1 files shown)


pashields:
:star2: :+1: +1 :+1:  :star2: 

This is much easier to follow.

ga4gh/cli.py
```diff
@@ -416,8 +417,10 @@ def client_main(parser=None):
     vsParser.set_defaults(runner=VariantSearchRunner)
     addUrlArgument(vsParser)
     addVariantSearchOptions(vsParser)
+    return vsParser
+
 
-    # variantsets/search
+def addVariantSetsSearchParser(subparsers):
     vssParser = subparsers.add_parser(
```
jeromekelleher: I think we should use `parser` here as the variable name. There's no reason to come up with different variable names now, and they don't really help readability.

jeromekelleher:
This is great, much more readable. I'm +1 after the variable name change.

Naburimannu:
+0

dcolligan:
Okay, just pushed a rebased change with @jeromekelleher 's suggestions

Issue #139, opened 2015-02-09
Installed server can't find html files


dcolligan:
Opening the web page at / using `server_dev.py` works fine, but opening it with the installed version of the server does not since it's searching for the html files in a place where they haven't been installed.

```
$ ga4gh_server wormtable ga4gh-example-data
 * Running on http://0.0.0.0:8000/ (Press CTRL+C to quit)
 * Restarting with stat
Traceback (most recent call last):
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "build/bdist.macosx-10.10-intel/egg/ga4gh/frontend.py", line 117, in index
    'index.html', info=app.serverStatus.getStatusInfo())
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/flask/templating.py", line 127, in render_template
    return _render(ctx.app.jinja_env.get_or_select_template(template_name_or_list),
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/jinja2/environment.py", line 830, in get_or_select_template
    return self.get_template(template_name_or_list, parent, globals)
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/jinja2/environment.py", line 791, in get_template
    return self._load_template(name, self.make_globals(globals))
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/jinja2/environment.py", line 765, in _load_template
    template = self.loader.load(self, name, globals)
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/jinja2/loaders.py", line 113, in load
    source, filename, uptodate = self.get_source(environment, name)
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/flask/templating.py", line 60, in get_source
    return loader.get_source(environment, local_name)
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/jinja2/loaders.py", line 163, in get_source
    f = open_if_exists(filename)
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/jinja2/utils.py", line 150, in open_if_exists
    return open(filename, mode)
[...]

PR #149, opened 2015-02-12
Set zip_safe in setup.py to False

dcolligan:
This allows the server to serve html files from the installed
ga4gh package.

Issue #139

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,7 @@ setup(
     license='Apache License 2.0',
     long_description=ga4ghReadme,
     packages=["ga4gh", "ga4gh.datamodel"],
+    zip_safe=False,
     author="Global Alliance for Genomics and Health",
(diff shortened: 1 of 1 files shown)


jeromekelleher:
+1

Naburimannu:
Mostly +1, but @dcolligan 's comment doesn't match the description of the flag at http://flask.pocoo.org/docs/0.10/patterns/distribute/. What am I missing?

(The only mention I can find of setting it to false is on http://flask.pocoo.org/docs/0.10/extensiondev/, which shouldn't be relevant.)

Is this actually a fix, or just a workaround?

dcolligan:
@Naburimannu it might help to read the setuptools docs which is the relevant package here: https://pythonhosted.org/setuptools/setuptools.html#setting-the-zip-safe-flag

Naburimannu:
+1

Issue #130, opened 2015-02-05
Client library needs logging/debug output


jeromekelleher:
The client library currently does not have any facility for logging or debugging events that occur. We should include log prints for HTTP events at a minimum to allow us to debug network issues.

diekhans:
FWIW, attached [we will see if github can handle attachments] is
code I use to parse command line for logging level and facility,
setup loggers to syslog or stderr.

General gripe is that python doesn't split up messages so that
they don't get truncated by syslog.

dcolligan:
Doesn't look like GitHub supports anything but picture attachments

https://help.github.com/articles/issue-attachments/

Because we all know how useful uploading pictures of squirrels saying 'ship it' is vs txt uploads...

dcolligan:
Ok, you can upload a txt file using gist but you need to drag-and-drop it into the UI

diekhans:
ohhh modern technology
![adm3a](https://cloud.githubusercontent.com/assets/110053/6062972/a80e8fc8-ad08-11e4-89eb-9b11d421b8c4.jpg)

dcolligan:
Fixed in #150

PR #150, opened 2015-02-13
Added logger to client

dcolligan:
Produces these log messages:

_debugLevel 0:
nothing

_debugLevel 1:
```
ga4gh.client: POST https://www.googleapis.com/genomics/v1beta2/variants/search
```

_debugLevel 2:
```
ga4gh.client: POST https://www.googleapis.com/genomics/v1beta2/variants/search
[dump of Json response]
```

Issue #130

### DIFF
--- a/ga4gh/client.py
+++ b/ga4gh/client.py
@@ -8,6 +8,7 @@ from __future__ import unicode_literals
 import json
 import requests
 import posixpath
+import logging
 
 import ga4gh.protocol as protocol
 
@@ -28,6 +29,20 @@ class HttpClient(object):
         self._workarounds = workarounds
         self._key = key
 
+        # logging config
+        self._logger = logging.getLogger(__name__)
+        if self._debugLevel == 0:
+            logLevel = logging.WARN
+        elif self._debugLevel == 1:
+            logLevel = logging.INFO
+        else:
+            logLevel = logging.DEBUG
+        self._logger.setLevel(logLevel)
+        handler = logging.StreamHandler()
+        formatter = logging.Formatter('%(name)s: %(message)s')
+        handler.setFormatter(formatter)
+        self._logger.addHandler(handler)
+
     def getBytesRead(self):
         """
         Returns the total number of (non HTTP) bytes read from the server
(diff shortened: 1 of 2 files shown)


ga4gh/client.py
```diff
@@ -28,6 +29,20 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
         self._workarounds = workarounds
         self._key = key
 
+        # logging config
+        self.logger = logging.getLogger(__name__)
```
jeromekelleher: Should logger be a public attribute or protected (i.e., `self._logger`)? I'd usually default to protected.

ga4gh/client.py
```diff
@@ -46,17 +61,14 @@ def _usingWorkaroundsFor(self, workaround):
...
     def _debugResponse(self, jsonString):
-        if self._debugLevel > 1:
-            # TODO use a logging output and integrate with HTTP client more
-            # nicely... also, logging / error handling in general
-            print("json response:")
-            pp = json.dumps(
-                json.loads(jsonString), sort_keys=True, indent=4)
-            print(pp)
+        self.logger.debug("json response:")
+        pp = json.dumps(
```
jeromekelleher: The trouble with this is we're doing a lot of JSON manipulation even when _debugLevel = 0, which would probably slow things down a lot. Maybe we should remove the JSON formatting? Is there a clean way to keep pretty printing (which is handy) but without having to construct the string every time?

jeromekelleher:
Looks good, thanks @dcolligan. I've a few minor issues but I think this is generally what we want all right.

ga4gh/client.py
dcolligan: Hm, you're right.  I would have liked to just send everything to the logger and let its internal configuration determine what and what not to emit, but it looks like we will need to add in some of our own logic for performance reasons.

@jeromekelleher do you think it makes more sense to drive the logic off of `_debugLevel` or off of logger's log levels e.g. `if self.logLevel < logger.WARN` ?  If we do the former than _debugLevel only is used in __init__ and nowhere else, which is nice... on the other hand, doing the later forces us to understand more about how logger works...

jeromekelleher: Your call @dcolligan --- both have benefits and drawbacks...

dcolligan:
@jeromekelleher new commit addresses your concerns

jeromekelleher:
Nice. +1.

PR #151, opened 2015-02-13
Change the README so that the example usage works

dcolligan:
We changed the variantId option from -v to -V previously to avoid conflicts with the verbose flag, but didn't update the README to reflect that.


jeromekelleher:
+1

Naburimannu:
+1

PR #152, opened 2015-02-13
Add infrastructure for ga2vcf and ga2sam tools

### LINKED ISSUE #83: Client side conversion tools required

dcolligan:
Does not close #83

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -9,12 +9,68 @@ from __future__ import unicode_literals
 
 import time
 import argparse
+import sys
 
 import ga4gh.frontend as frontend
 import ga4gh.client as client
 import ga4gh.backend as backend
 import ga4gh.protocol as protocol
 import ga4gh.datamodel.variants as variants
+import ga4gh.converters as converters
+
+
+##############################################################################
+# ga2vcf
+##############################################################################
+
+
+def ga2vcf_main(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser(
+            description="GA4GH VCF conversion tool")
+    addClientGlobalOptions(parser)
+    subparsers = parser.add_subparsers(title='subcommands',)
+    addVariantsSearchParser(subparsers)
+    args = parser.parse_args()
+    request = protocol.GASearchVariantsRequest()
+    setCommaSeparatedAttribute(request, args, 'datasetIds')
+    # TODO add outputFile cli argument
+    outputFile = sys.stdout
+    vcfConverter = converters.VcfConverter(request, outputFile)
+    vcfConverter.convert()
+
+
+##############################################################################
+# ga2sam
+##############################################################################
+
+
+def ga2sam_main(parser=None):
+
+    def usingWorkaroundsFor(workaround):
+        return workaround in workarounds
+
+    if parser is None:
+        parser = argparse.ArgumentParser(
+            description="GA4GH SAM conversion tool")
+    addClientGlobalOptions(parser)
+    subparsers = parser.add_subparsers(title='subcommands',)
+    addReadsSearchParser(subparsers)
+    args = parser.parse_args()
+    workarounds = set(args.workarounds.split(','))
+    request = protocol.GASearchReadsRequest()
+    if usingWorkaroundsFor(client.HttpClient.workaroundGoogle):
+        # google says referenceId not a valid field
+        request = SearchReadsRunner.GASearchReadsRequestGoogle()
+    setCommaSeparatedAttribute(request, args, 'readGroupIds')
+    request.start = args.start
+    request.end = args.end
+    request.referenceId = args.referenceId
+    request.referenceName = args.referenceName
+    # TODO add outputFile cli argument
+    outputFile = sys.stdout
+    samConverter = converters.SamConverter(request, outputFile)
+    samConverter.convert()
 
 
 def setCommaSeparatedAttribute(request, args, attr):
--- a/ga4gh/converters.py
+++ b/ga4gh/converters.py
@@ -1,0 +1,40 @@
+"""
+Provides classes that take protocol requests, send that request to
+the server, and write a particular genomics file type with the results.
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+class AbstractConverter(object):
+    """
+    Abstract base class for converter classes
+    """
+    def __init__(self, request, outputStream):
+        self._request = request
+        self._outputStream = outputStream
+
+
+class SamConverter(AbstractConverter):
+    """
+    Converts a request to a SAM file
+    """
+    def __init__(self, searchReadsRequest, outputStream):
+        super(SamConverter, self).__init__(
+            searchReadsRequest, outputStream)
+
+    def convert(self):
+        raise NotImplementedError()
+
+
+class VcfConverter(AbstractConverter):
+    """
+    Converts a request to a VCF file
+    """
+    def __init__(self, searchVariantsRequest, outputStream):
+        super(VcfConverter, self).__init__(
+            searchVariantsRequest, outputStream)
+
+    def convert(self):
+        raise NotImplementedError()
--- a/ga2sam_dev.py
+++ b/ga2sam_dev.py
@@ -1,0 +1,11 @@
+"""
+Shim for running the ga2sam tool during development
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import ga4gh.cli
+
+if __name__ == "__main__":
+    ga4gh.cli.ga2sam_main()
--- a/ga2vcf_dev.py
+++ b/ga2vcf_dev.py
@@ -1,0 +1,11 @@
+"""
+Shim for running the ga2vcf tool during development
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import ga4gh.cli
+
+if __name__ == "__main__":
+    ga4gh.cli.ga2vcf_main()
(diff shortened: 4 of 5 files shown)


jeromekelleher:
I'd like to keep all the command line argument parsing and user interface stuff in cli.py and farm out the actual conversion to classes in a dedicated module --- see my comments on #126.

dcolligan:
@jeromekelleher ok, I read the comments in #126 and tried to implement your suggestions here.

A couple problems I still have:
- the lack of a request factory -- or some other way to abstract the details of protocol request instantiation -- leads to a lot of copy and pasting.  This is bad.
- the lack of a unified way to determine workarounds leads to copy-and-paste-and-slightly-modify, which is even worse.

What do you suggest to resolve these issues?  I still like my request factory, but it seems like you don't.

ga4gh/cli.py
```diff
@@ -9,12 +9,68 @@
...
+
+def ga2vcf_main(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser(
+            description="GA4GH VCF conversion tool")
+    addClientGlobalOptions(parser)
+    subparsers = parser.add_subparsers(title='subcommands',)
+    addVariantSetsSearchParser(subparsers)
+    args = parser.parse_args()
+    request = protocol.GASearchVariantSetsRequest()
```
jeromekelleher: This should be `SearchVariants` not `SearchVariantSets` because we want to be able to specify all of the range options available for `SearchVariants`. We will probably never want to convert a whole variant set, we'll just want to convert some variants for some small range we're interested in.

jeromekelleher:
This looks good, thanks for trying this out @dcolligan. I agree that we need some sort of request factory method, and we definitely don't want to copy-paste code. I've been thinking about this a bit, and I wonder if the right design is to extend argparse. We're doing a lot of stuff like 
```python
def addStartArgument(parser):
    parser.add_argument(
        "--start", "-s", default=0, type=int,
        help="The start of the search range (inclusive).")
```
which might be more natural to express as
```python
def addStartArgument(self):
    self.add_argument(
        "--start", "-s", default=0, type=int,
        help="The start of the search range (inclusive).")
```
So, the vague idea would be to make a class hierarchy that extends `argparse.ArgumentParser` and returns a fully instantiated `XRequest` object when you ask it.  For example, we might have
```python
def ga2vcf_main(parser=None): # We really should sort out the camelCasing here!
    if parser is not None:
        parser = SearchVariantsParser()
    parser.addOutputFileArgument()
    args, request = parser.parse_args()
    converter = converters.VcfConverter(request) 
    outputHandle = sys.stdout
    if args.outputFile is not None:
        outputHandle = open(args.outputFile, "w")
    converter = converters.VcfConverter(request, outputHandle)
    try:
        converter.convert()
    finally:
        if args.outputFile is not None:
            outputHandle.close()
```
What do you think, will this work?

dcolligan:
That would work, but it feels like it violates a separation of concerns between 1) argument parsing and 2) instantiating protocol objects.  I think it would be wrong to turn ArgumentParser into a DoEverythingBeforeARequestIsRunObject.  Furthermore, I think preferring composition to inheritance is always a good rule of thumb to follow.

You know, I think your only real objection in #144 was that RequestFactory took a `host` argument to all of its class methods.  I think I can refactor the code so that is no longer the case.

jeromekelleher:
I don't think this would lead to a `DoEverythingBeforeARequestIsRunObject` --- really, all your `RequestFactory` is doing is taking a bundle of `args` from argparse and returning an instantiated `Request` object. We could legitimately view this as parsing command line arguments into a `Request` object; that is, extending the functionality of `argparse.ArgumentParser`. Inheritance is a very powerful design principle, and I don't think there are any hard and fast rules for when it should and shouldn't be used. We should use whatever design solves the problem in hand in the simplest and clearest way (modulo a reasonable level of performance awareness).

Anyone else want to weigh in on this? @Naburimannu, @pashields any thoughts?

Having said all that, it doesn't really matter that much once we keep all of the command line interface stuff in `cli.py` and use the instantiated `Request` object as our interface to the converter classes. You're right that the only real objection I have to your `RequestFactory` are the side effects, so if you remove  these I'm happy to move on with this.

dcolligan:
Ok, I pushed a new commit in #144.  Let me know what you think.

Naburimannu:
My kibbitzing: I don't think inheritance is "more natural" than composition as @jeromekelleher suggested.

I *like* the idea of a class that processes arguments into a FooRequest, even though I'm not sure whether it'll smell - as with many things, my reaction is to say "Let's try it, and we can always introduce another intermediate layer in the future if it's really necessary." It seems without that, we have class A: arguments -> internal representation, and factory F: internal representation -> class instance. Without a principled reason for the internal representation, if it doesn't add complexity, let's collapse the two.

Here's a possible principled reason: would DoEverythingBeforeARequestIsRun need to know about all our internal classes, while ArgParse and a set of RequestFactories could have more limited coupling?

ga4gh/converters.py
```diff
@@ -0,0 +1,42 @@
+"""
+Provides classes that convert protocol requests to genomics data files
```
Naburimannu: It's not clear to me what it means to convert a request (in)to a file? Is this just the wrong preposition (or one of the transatlantic preposition shift issue), or do we need to be more explicit?

Naburimannu: (or just "file formats"? Although there again, is it to a file format, or some live protocol?)

Naburimannu:
Do we have a PR somewhere for ga2sam or ga2vcf so that we know this is the correct infrastructure to be building? Examples of what the output of this process will be?

dcolligan:
Okay, just pushed a new commit addressing some concerns.

Hopefully most of @Naburimannu 's concerns were resolved in the phone call.

ga4gh/converters.py
```diff
@@ -0,0 +1,43 @@
...
+
+class AbstractConverter(object):
+    """
+    Abstract base class for converter classes
+    """
+    def __init__(self, request, outputStream):
+        self.request = request
+        self.outputStream = outputStream
+
+    def close(self):
```
jeromekelleher: I don't think we're ever going to close a stream, since we've passed it in open. The canonical usage would be
```python
with open(fileName, "w") as outputFile:
    converter = VcfConverter(request, outputFile)
    converter.convert()
```

ga4gh/converters.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+class AbstractConverter(object):
+    """
+    Abstract base class for converter classes
+    """
+    def __init__(self, request, outputStream):
+        self.request = request
```
jeromekelleher: Should these be public or protected? I would default to protected for a substantial class like this.

jeromekelleher:
+1 modulo small nits.

dcolligan:

jeromekelleher:
Thanks @dcolligan --- I'll merge this tomorrow if there are no objections.

dcolligan:
@jeromekelleher this PR made good progress, but there are a few problems with this model that I've discovered:

1) The one request object passed to the Converter class may not be sufficient if the converter needs to make multiple kinds of requests to the server.  For instance, the `VcfConverter` can not infer the `datasetIds` argument needed in a `GAVariantSetsSearchRequest` from the `GAVariantsSearchRequest` that is passed to it.

2) The `httpClient` that the converters need to make requests needs to be initialized from the command line args (for instance `baseUrl`).  Either a) the client is initialized in the cli module and passed to the converter in another argument or b) we also pass the command line args to the converter in another argument.

3) The pysam API actually doesn't support writing to file streams.  The only thing it does support for SAM files is writing to stdout, but even that requires instantiation through a pysam object.  http://pysam.readthedocs.org/en/latest/usage.html#using-streams  So anything that the cli module could pass it would be pretty useless, thus calling into question whether this is the right converter interface at all...

How do you think we should resolve these issues?

jeromekelleher:
Hmmm, all good points @dcolligan.

1. This seems like a problem with the protocol to me. We should be able to get the metadata for a variant set without having to make a request for all the variant sets and filter them out.  Can you look into this and open an issue against the schemas repo, if you agree? In the mean time, here I would just pass in the extra arguments that are required one-by-one. I don't like the idea of passing an `argparse` args bundle around between modules are we are creating an implicit dependency on argparse.
2. I would initialise the httpClient in the CLI module, and pass it in as an argument to the Converter. We'd have something like `converter = VcfConverter(httpClient, request, dataSetIds)`.
3. OK, just adopt the pysam convention here and pass in "-" as the filename when writing to stdout.

dcolligan:
Wrt (1), I don't really understand this: "We should be able to get the metadata for a variant set without having to make a request for all the variant sets and filter them out."

That doesn't seem to be the problem to me.  The problem is that we have an interface that is expecting one kind of request object (`GAVariantsRequest`) but we need two request objects (`GAVariantSetsRequest` and `GAVariantsRequest`) in order to retrieve the data needed to write a VCF file.  It would seem inappropriate to me to have the data in one response object contain the data of the other -- it just seems like they are plainly separate objects.

dcolligan:
Basically I think it makes sense to toss out the converters interface design since it seems to work in neither the SAM case (since the pysam api doesn't support writing to arbitrary file streams) nor the VCF case (since more than one type of request object is needed to get the data to write the VCF file).

jeromekelleher:
We don't have to keep this precise interface. All I'm saying essentially is that we should keep the interface for converters completely separate from the command line interface. This means we can't pass argparse args bundles around as parameters, and we can't use infrastructure designed to turn argparse args bundles into request objects within the converter module. There's far too much coupling between them if we do.

PR #153, opened 2015-02-13
Add note about using a pre push hook

pashields:
We recently had an issue with a force push to master. This commit encourages contributors to use a pre-push hook that prevents such behaviors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,8 @@ Some general rules to follow:
   followed by an empty line and then any details that you want to share about 
   the commit.
 * Please try to follow the existing syntax style
+* We recommend using a ```pre push``` hook like the one found [here](https://gist.githubusercontent.com/pixelhandler/5718585/raw/1ee3948c6676b704fb9da424c65a767f2c88ae37/pre-push.sh)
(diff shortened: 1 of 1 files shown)


dcolligan:
I have a bunch of nits like

- the problem isn't so much that people are force pushing to master, but that they are pushing to upstream at all -- that should only be done through the GitHub UI; every other push should be to one's own repo
- force pushes to master branches are fine... so long as it is not the upstream master
- if someone is using a Git UI (all the more likely if someone is inexperienced with git, IMHO), will this script prevent them from pushing?  I doubt it... (but I may be wrong about this)

But I do recognize that if everyone installed this and only used the command line to push then future force pushes to upstream master would be prevented

+0

I would be more enthusiastic if this was a script that prevented pushes to a certain repository rather than (force pushes to) a certain branch.

jeromekelleher:
I'm with @dcolligan here - there's nothing wrong with force pushing to `origin/master`, we just don't want pushes to `upstream/master`. This won't do any harm though, so I'm also +0.

pashields:
Ok, no one seems super pumped about this, I'm just going to close it.

Issue #119, opened 2015-02-02
variants-search should with no arguments should return all variants on server


jeromekelleher:
The `variants-search` client command is a useful development tool, but tedious to use as it needs a lot of input to return anything. We should change the default behaviour to return variants from the server, if there are any. So:
- `--end` should default to some large number, so we always overlap with something. (Ideally, the `end` value should be nullable in the protocol so that we can specify this requirement more naturally);
- If no variantSetIds are specified, we should run a searchVariantSets request, and get the IDs for these from the server.

dcolligan:
Fixed in #154

PR #154, opened 2015-02-17
no arg variants-search call returns all variants

dcolligan:
Issue #119

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -17,6 +17,14 @@ import ga4gh.protocol as protocol
 import ga4gh.datamodel.variants as variants
 
 
+# the maximum value of a long type in avro = 2**63 - 1
+# (64 bit signed integer)
+# http://avro.apache.org/docs/1.7.7/spec.html#schema_primitive
+# AVRO_LONG_MAX = (1 << 63) - 1
+# TODO in the meantime, this is the max value wormtable can handle
+AVRO_LONG_MAX = (1 << 32) - 2
+
+
 def setCommaSeparatedAttribute(request, args, attr):
     attribute = getattr(args, attr)
     if attribute is not None:
@@ -122,11 +130,7 @@ class RequestFactory(object):
 ##############################################################################
 
 
-def server_main(parser=None):
-    if parser is None:
-        parser = argparse.ArgumentParser(
-            description="GA4GH reference server")
-    # Add global options
+def addGlobalOptions(parser):
     parser.add_argument(
         "--port", "-P", default=8000, type=int,
         help="The port to listen on")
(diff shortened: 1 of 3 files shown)


Naburimannu:
Looks like you committed more than intended, with parser/subparser architecture changes? The code that directly pertained to this issue looked fine...

dcolligan:
Yeah, I thought I might as well do a refactor analogous to #145 for the server cli while I was in there.  For the rationale behind the `scripts/` and `tests/` changes, see: https://github.com/ga4gh/server/pull/144#issuecomment-74724543

ga4gh/backend.py
```diff
@@ -118,6 +118,8 @@ def variantsGenerator(self, request):
         the specified request.
         """
         variantSetIds = request.variantSetIds
+        if len(variantSetIds) == 1 and variantSetIds[0] == '':
```
jeromekelleher: This is changing the  semantics of the protocol. We can't do this.

ga4gh/cli.py
```diff
@@ -406,7 +418,7 @@ def addStartArgument(parser):
 
 def addEndArgument(parser):
     parser.add_argument(
-        "--end", "-e", default=1, type=int,
+        "--end", "-e", default=60000, type=int,
```
jeromekelleher: Why 60000? I think this should be a very large value (i.e., the maximum possible value for the corresponding value in the Avro schema), so that we are guaranteed that this is always larger than the maximum in our data.

jeromekelleher:
This changes the semantics of the protocol to achieve its goals, which is absolutely forbidden. We *cannot* extend the protocol for any reason. Even if it's daft and clearly wrong, we must still implement it, as it is described. We can file bug reports against the protocol where we see problems (which we certainly *should* do), but we must implement a given version as it is stated.

The way to do this is to check if `args.variantSetIds` is None. If it is, we perform a `SearchVariantSets` request to find all variantSetIds.

dcolligan:
Pushed a new commit that addressed @jeromekelleher 's concerns.

Btw, if one runs a no arg variants-search request, the server will error, but I think that's a problem with our data.  The same error happens if one searches with the `-V 1000g_2011` flag or without it (but not with the `-V 1000g_2013` flag):

```
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/dannyc/ga4gh/server/ga4gh/frontend.py", line 174, in searchVariants
    return handleFlaskRequest(flask.request, app.backend.searchVariants)
  File "/Users/dannyc/ga4gh/server/ga4gh/frontend.py", line 116, in handleFlaskRequest
    return handleHttpPost(flaskRequest, endpoint)
  File "/Users/dannyc/ga4gh/server/ga4gh/frontend.py", line 81, in handleHttpPost
    responseStr = endpoint(request.get_data())
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 93, in searchVariants
    self.variantsGenerator)
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 65, in runSearchRequest
    for obj, nextPageToken in objectGenerator(request):
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 141, in variantsGenerator
    for variant in iterator:
  File "/Users/dannyc/ga4gh/server/ga4gh/datamodel/variants.py", line 211, in getVariants
    cols = [col for name, col in self._sampleCols[callSetId]]
KeyError: u'HG03279'
```

jeromekelleher:
This is still changing the semantics of the protocol I'm afraid. I  wasn't very clear earlier, but what I meant was we should do a `SearchVariantSets` on the *client* side. This PR shouldn't touch the backend at all, as it's purely a client side convenience.

dcolligan:
@jeromekelleher aha, that wasn't clear to me.  I've pushed another commit that has no backend changes.  I've filed #160 to deal with the backend exceptions.

ga4gh/cli.py
```diff
@@ -175,7 +197,16 @@ def __init__(self, args):
             request.callSetIds = None
         else:
             request.callSetIds = args.callSetIds.split(",")
-        setCommaSeparatedAttribute(request, args, 'variantSetIds')
+        # if no variantSets have been specified, send a request to
+        # the server to grab all variantSets and then continue
+        if args.variantSetIds is None:
+            variantSetsRequest = protocol.GASearchVariantSetsRequest()
+            response = self._httpClient.searchVariantSets(variantSetsRequest)
+            variantSets = [variantSet for variantSet in response]
```
Naburimannu: This line *reads* like a NOOP.

Naburimannu:
Client-side is +1 modulo my nitty question.

ga4gh/cli.py
jeromekelleher: I'm with @Naburimannu here - we don't need the intermediate `variantSets` list do we? I.e,
```python
variantSetIds = [variantSet.id for variantSet in response]
```

dcolligan:
Pushed a commit to address @Naburimannu nit

dcolligan:
@jeromekelleher I want your +1 on this before I label it mergeable

ga4gh/cli.py
```diff
@@ -521,7 +552,7 @@ def addNameArgument(parser):
 def addClientGlobalOptions(parser):
     parser.add_argument('--verbose', '-v', action='count', default=0)
     parser.add_argument(
-        "--workarounds", "-w", default='', help="The workarounds to use")
```
jeromekelleher: I think the old way is better actually - having a default of the empty makes creating the set of workarounds simpler and doesn't need a conditional.

jeromekelleher:
+1 modulo small nit.

ga4gh/cli.py
dcolligan: No, the old way would create a set of workarounds initialized to the set containing an empty string.  I think this is a better pattern to follow generally, since when sending arguments over the wire, sending an empty string instead of a null is usually not what you want.  (Although I agree it matters less in the specific workarounds case.)

jeromekelleher: Fair enough - I'll merge after a rebase.

dcolligan: Just pushed a rebase

PR #158, opened 2015-02-18
Easier paging

jeromekelleher:
This is a very rough first pass at making paging easier. This supersedes #134 and #121, as I think we need to resolve this problem before we make any progress on adding support for datasets or references.

Paging is currently too hard. At the moment, we're using different mechanisms to handle paging when we're dealing with `VariantSets` and `Variants`. This is very awkward, and makes it difficult and error prone to add support for more parts of the protocol. We need a single unified approach that makes it trivial to add support for new objects.

This is a very tentative first pass at doing this. Instead of having a sorted list of IDs that we use as our way of indexing into the variantSets, we abstract the details away into a `DataObjectCollection`, which is supposed to make this easy. The DataObjectCollection should handle searches in a generic way for high level objects (e.g., Datasets, VariantSets, ReferenceSets), and pass the search query down to more specialised classes when we are deeper in the tree.

In the generic case, we keep an ordered map of the IDs to the objects, stored here in an AVL tree. The pageToken we pass back is the ID of the next object, which is simple and straight forward to implement. See the definition of search here when `depth == 1`, which works well and I'm quite happy with. I've cobbled together something that sort-of works for depth == 2 here. It's very nasy, and shouldn't be taken too seriously. Ultimately, I hope we can have some elegant recursive approach here which makes the tree traversal logic completely transparent.

Questions:
1. Is this a good way to proceed? Is mapping the `nextPageToken` to the ID of the next object in the stream the right thing to do?
2. Is `DataObjectContainer` a good abstraction? Will this abstraction work when we try to support other parts of the protocol?

This is a crucial aspect of the design of the server in my opinion, so I'd really appreciate it if people (especially @Naburimannu, @pashields and @dcolligan) could think about it a bit and give me feedback. It's a critical blocker in the way of making progress on  backend support for the rest of the protocol, so I think it's very important we get our heads together to try and figure this one out.

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -8,7 +8,55 @@ from __future__ import unicode_literals
 
 import os
 
+import bintrees
+
 import ga4gh.protocol as protocol
+
+
+class DataObjectCollection(object):
+    """
+    A class representing a collection of data objects allowing for
+    simple and efficient paging using pageTokens.
+    """
+    def __init__(self):
+        self._idMap = bintrees.AVLTree()
+
+    def add(self, dataObject):
+        """
+        Adds the specified object to this collection.
+        """
+        self._idMap[dataObject.getId()] = dataObject
+
+    def parsePageToken(self, pageToken, numValues):
+        """
+        Parses the specified pageToken and returns a list of the specified
+        number of values. Page tokens are assumed to consist of a fixed
+        number of strings seperated by colons. If the page token does
+        not conform to this specification, raise a InvalidPageToken
+        exception.
+        """
+        tokens = pageToken.split(":")
+        # TODO define exceptions.InvalidPageToken and raise here.
+        if len(tokens) != numValues:
+            raise Exception("Invalid number of values in page token")
+        return tokens
+
+    def search(self, request):
+        """
+        Returns an iterator over the ProtocolElements defined by
+        the specified request.
+        """
+        maxKey = self._idMap.max_key()
+        if request.pageToken is None:
+            nextId = self._idMap.min_key()
+        else:
+            nextId, = self.parsePageToken(request.pageToken, 1)
+        for dataObject in self._idMap.value_slice(nextId, None):
+            protocolElement = dataObject.toProtocolElement()
+            nextId = None
+            if dataObject.getId() != maxKey:
+                nextId = self._idMap.succ_key(dataObject.getId())
+            yield protocolElement, nextId
 
 
 class Backend(object):
@@ -18,31 +66,14 @@ class Backend(object):
     """
     def __init__(self, dataDir, variantSetClass):
         self._dataDir = dataDir
-        self._variantSetIdMap = {}
+        self._variantSets = DataObjectCollection()
         # All directories in datadir are assumed to correspond to VariantSets.
         for variantSetId in os.listdir(self._dataDir):
             relativePath = os.path.join(self._dataDir, variantSetId)
             if os.path.isdir(relativePath):
-                self._variantSetIdMap[variantSetId] = variantSetClass(
-                    variantSetId, relativePath)
-        self._variantSetIds = sorted(self._variantSetIdMap.keys())
+                variantSet = variantSetClass(variantSetId, relativePath)
+                self._variantSets.add(variantSet)
 
-    def parsePageToken(self, pageToken, numValues):
-        """
-        Parses the specified pageToken and returns a list of the specified
-        number of values. Page tokens are assumed to consist of a fixed
-        number of integers seperated by colons. If the page token does
-        not conform to this specification, raise a InvalidPageToken
-        exception.
-        """
-        tokens = pageToken.split(":")
-        # TODO define exceptions.InvalidPageToken and raise here.
-        if len(tokens) != numValues:
-            raise Exception("Invalid number of values in page token")
-        # TODO catch a ValueError here when bad integers are passed and
-        # convert this into the appropriate InvalidPageToken exception.
-        values = map(int, tokens)
-        return values
 
     def runSearchRequest(
             self, requestStr, requestClass, responseClass, pageListName,
@@ -95,22 +126,10 @@ class Backend(object):
     def variantSetsGenerator(self, request):
         """
         Returns a generator over the (variantSet, nextPageToken) pairs defined
-        by the speficied request.
+        by the specified request.
         """
-        currentIndex = 0
-        if request.pageToken is not None:
-            currentIndex, = self.parsePageToken(request.pageToken, 1)
-        while currentIndex < len(self._variantSetIds):
-            variantSet = protocol.GAVariantSet()
-            variantSet.id = self._variantSetIds[currentIndex]
-            variantSet.datasetId = "NotImplemented"
-            variantSet.metadata = self._variantSetIdMap[
-                variantSet.id].getMetadata()
-            currentIndex += 1
-            nextPageToken = None
-            if currentIndex < len(self._variantSetIds):
-                nextPageToken = str(currentIndex)
-            yield variantSet, nextPageToken
+        for protocolElement in self._variantSets.search(request):
+            yield protocolElement
 
     def variantsGenerator(self, request):
         """
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -41,22 +41,56 @@ class DataObjectCollection(object):
             raise Exception("Invalid number of values in page token")
         return tokens
 
-    def search(self, request):
+    def search(self, request, depth):
         """
         Returns an iterator over the ProtocolElements defined by
         the specified request.
         """
-        maxKey = self._idMap.max_key()
-        if request.pageToken is None:
-            nextId = self._idMap.min_key()
+        if depth == 1:
+            maxKey = self._idMap.max_key()
+            if request.pageToken is None:
+                nextId = self._idMap.min_key()
+            else:
+                nextId, = self.parsePageToken(request.pageToken, 1)
+            for dataObject in self._idMap.value_slice(nextId, None):
+                protocolElement = dataObject.toProtocolElement()
+                nextId = None
+                if dataObject.getId() != maxKey:
+                    nextId = self._idMap.succ_key(dataObject.getId())
+                yield protocolElement, nextId
         else:
-            nextId, = self.parsePageToken(request.pageToken, 1)
-        for dataObject in self._idMap.value_slice(nextId, None):
-            protocolElement = dataObject.toProtocolElement()
-            nextId = None
-            if dataObject.getId() != maxKey:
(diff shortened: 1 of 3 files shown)


dcolligan:
In general, I agree that having a generic way of accessing objects on the backend / paging is a good idea.  I just worry about the resource constraints.  Your design means that we will be caching an AVL tree of all ids of each first class object, correct?  That seems like a lot of memory (not that we can't mitigate this later by writing the tree out to disk, or having some intermediate cache, or do something else fancy).

Also, I don't understand the meaning of `depth`.  What is it?

jeromekelleher:
I was thinking of the AVL tree as just for use in the objects we explicitly keep in memory, like `Dataset`s, `VariantSet`s, `ReferenceSet`s etc. For things like `Variant`s we'd use an implicit mapping to IDs using something like Wormtable or pysam/tabix. One possibility is to have a specific subclass of DataObjectCollection for a `VariantSet`, which over rides the functionality and doesn't have the AVL tree.

The depth idea here is that we use the same code to return the different objects. Suppose we have the following hierarchy:
```
   datasets
    dataset1     
        variantSet1
            variant1
            variant2
            variant3
        variantSet2
            variant1
            variant2
    dataset2
        variantSet1
            variant1
            variant2
```
If we want to list all datasets, we search this tree to depth 1 (returns dataset1, dataset2). If we want to list all variantSets, we search the tree to depth 2 (returns dataset1:variantSet1, dataset1:variantSet2, ..., dataset2:variantSet1). If we want to list all variants, we search to depth 3 (returns dataset1:variantSet1:variant1, dataset1:variantSet1:variant2, ..., dataset2:variantSet1:variant2). I'm not sure how useful this particular idea is, but I think we need to exploit this tree structure if we are going to make this generic.

dcolligan:
Hm, in that case wouldn't an AVL tree be a poor choice, since when it re-balances itself, you don't have any guarantee about which level each data type will be on?  It seems like we would want a non-balancing search tree here.

pgrosu:
Hi Jerome,

It think you will find the following two Google papers helpful in solving the issues you are experiencing:

[Storing and Querying Tree-Structured Records in Dremel](http://research.google.com/pubs/archive/43119.pdf)

[Dremel: Interactive Analysis of Web-Scale Datasets](http://research.google.com/pubs/archive/36632.pdf)

Of course there are always the option of a simple B-Tree or (Apache) Parquet implementation, but the above two from Google allow for implementing a distributed algorithm across multiple computers of trees with columnar storage for more optimized retrieval.

Paul

jeromekelleher:
The AVL tree isn't intended to store the full hierarchy, just the children at the internal nodes. It's going to be a very bushy tree (we might have say, 100 datasets and 10 variant sets in each dataset), so we need something to search the IDs of the objects. For example, suppose we're searching through all variantSets and we get a pageToken `dataset3:variantSet20`. (The IDs won't be of this form, but will be arbitrary strings --- I'm just using these as a concrete example.) We then need to position our iterator on `dataset3:variantSet20` and pick up the traversal from there. We do this using nested AVL tree lookups: we search the first AVL tree for dataSet3, and this gives us a reference to the corresponding object. We then search the AVL tree of variantSets in dataset3 for `variantSet20`, which positions us at the correct point in the AVL tree and we can then start iterating from there. Does this make sense?

I used an AVL tree because if gives us ordering semantics among the keys. We can't use dictionaries, as they have no concept of ordering, or at least not a useful one.  An AVL tree is probably overkill given the expected sizes we're dealing with, but I think it makes finding IDs simpler, so it's good for this reason. We could equally use a red-black tree or something similar, it wouldn't make any difference.

pgrosu:
So besides looking through long results, paging helps in case only parts of your query reach the server (the web is not always reliable) - usually these steps become distributed.  I understand the efficiency of an AVL tree having a O(log n) complexity for its operations and not wanting to use dictionaries, but even the Linux kernel had to hash their directory cache which can go down to O(1):

http://www.linuxjournal.com/article/7124

So if I want to do a matchmaker-type search for all samples containing a specific set of SNP calls - knowing the dangers with that approach - how efficient will that be? Usually there are several layers, as is currently the direction, as long as we don't have too many levels of indirection.  I'm not even including higher levels such as disease mapping at this point, but at some point it will become important.

So for now, the fastest approach would be some form of a localized hashtable at nodes that have a large collection of nodes/leaves as children - otherwise a master-hashtable is probably cleaner.

dcolligan:
@jeromekelleher yes, that makes sense -- you essentially have a nested AVL tree and `depth` refers to which level of nesting you are going to.  I don't have a strong opinion on AVL vs Red-Black vs whatever.  It looks like the `bintrees` packages supports both so it should be easy to change if we change our mind.

One thing that we should keep in mind is that all of our data is loaded on startup and then never written to again, correct?  So this WORM property might give us some advantages down the line.

lh3:
What are the keys of the AVL tree? @jeromekelleher seems to suggest the keys are IDs such that we can search for IDs, but wouldn't this discard the ordering we want to keep?

If I understand correctly, the right data structure could be a normal tree to keep the hierarchy and a hash table which maps (depth,ID) to the index of ID in the array it comes from. For example, suppose a dataset has two variantSets ["setFoo","setBar"]. We should have `hashTable[2,"setFoo"]=0`.

jeromekelleher:
The keys in the AVL trees are the object IDs as you say, but an AVL tree would maintain the sortedness requirement. The key property of an AVL tree that we want here is that after we've found a key (in log time) we can find its successor (in sort order) in constant time. 

However, I like your approach @lh3, as it is probably simpler to implement and doesn't require any external dependencies. Thanks for the input.

ga4gh/backend.py
```diff
@@ -39,10 +39,75 @@ def parsePageToken(self, pageToken, numValues):
         # TODO define exceptions.InvalidPageToken and raise here.
         if len(tokens) != numValues:
             raise Exception("Invalid number of values in page token")
-        # TODO catch a ValueError here when bad integers are passed and
-        # convert this into the appropriate InvalidPageToken exception.
-        values = map(int, tokens)
-        return values
+        return tokens
+
+    def search(self, request, depth):
```
Naburimannu: This is really two different functions with depth being used to choose between them.
Separate them instead.

lh3:
To construct an AVL tree, we need to define the "less than" operator between keys. How is this operator defined? Is it index(ID1)<index(ID2) or ID1<ID2 as a string comparison? In the former case, we would need a hash table to look up index(ID). In the latter case, we would discard the ordering.

jeromekelleher:
The comparison operator is just lexicographic ordering of strings I think. The order we return the (e.g.) `VariantSet`s is arbitrary, it just needs to be fixed so that we can be sure we pick up our iteration at the correct point when we resume from a `pageToken`. Does this make sense @lh3, or am I missing something important?

I am assuming static data here --- if the data changes mid query, then this ordering approach won't work. Lots of other things won't work either though, so I'm not sure if it's worth worrying about these sort of issues yet.

lh3:
Sorry. It seems that I have misunderstood your intention. I thought you wanted to keep an arbitrary ordering as is requested by the submitters. With my strategy, "variantSetFoo" could precede "variantSetBar", although "variantSetBar" is lexicographically smaller. I imagine that submitters *might* want to keep the submission ordering sometimes.

If the data is static, why not just use sorted arrays? AVL/RB/B-tree are only better when you want dynamic insertions and deletions.

Naburimannu:
@lh3: I think the PR needs work on naming and description. If I understood a conversation I had with @jeromekelleher on Friday, his **intent** in this is as much about representing the entire hierarchy of data types imported by GA4GH and handling queries that have to traverse that hierarchy as it is about paging.

From http://ga4gh.org/#/api:
Type hierarchy for Read data

We'll have multiple type hierarchies like that for Reads, and need to be able to page our responses to queries at any level in them, where queries may cross multiple levels of the hierarchy. In trying to add another data type to the Variants/VariantSets we started with (or seeing @MollyZhang's [PR to add CallSets] (https://github.com/ga4gh/server/pull/140)?) @jeromekelleher realized that the paging code was fragile.

In our discussion of this PR we're getting lost in data structure details and need a clearer picture of where we're going, but just editing the first paragraph of the PR might help readers understand better.

lh3:
I understand the broad picture about why we need ordering for paging. I am just curious about the details. More specifically, my questions are:

1. Do we want to keep the submission ordering? In theory, a submitter might submit an older batch of variant calls under ID "vsFoo" and a more recent batch under "vsBar". The submitter might expect to see "vsBar" is next to "vsFoo", but with an array/AVL tree, the opposite is true (because vsFoo is lexicographically larger than vsBar). However, in practice, lexicographical ID ordering probably works most of time. I don't mind too much if we discard the submission ordering.

2. Why do we need an AVL tree when a simple sorted array works? If we only need `search()` and `next()` operations, a sorted array is preferred. AVL/RB only works better when we want to perform `insert()` and `delete()`.

jeromekelleher:
Thanks for the summary @Naburimannu, this is a much better picture of where I'm trying to get to with this PR. 

To answer @lh3's questions:

1. I'd rather not get involved with submission semantics here, as this would require us to add some extra metadata to the VCF files etc. I think we should limit ourselves to a static snapshot of the data at a given time, with no knowledge of when the files arrived and so on.
2. Excellent question! You are right @lh3, I'll change this to use sorted arrays and binary search when I pick this up again.

pgrosu:
Even if it may seem everything looks clear now, we definitely need to ensure everyone's working from the same definitions in terms of interpretation and scope.  I feel it's time to step back for us to diagram out the code and it's scope in order to ensure we are all working from the same interpretation(s) before more code gets changed - otherwise we will keep solving the same problems repeatedly.  I can already foresee issues coming up, and for instance the diagram above (posted by @Naburimannu) has to do with reads and not variants (which the discussion was shifting towards), and reads usually are more static than are variants.  If I want to create variants for on-the-fly analysis from different ReadGroupSets, I'm not sure this approach might be flexible or optimal enough.

diekhans:
Heng Li <notifications@github.com> writes:
>  1. Do we want to keep the submission ordering? In theory, a submitter might
>     submit an older batch of variant calls under ID "vsFoo" and a more recent
>     batch under "vsBar".

I don't believe we should not encode any behavior that is not
explicitly specified in the API.  Submission order would be a
dangerous thing to rely on without an explicit coding of it
(submission date).

pgrosu:
@diekhans, I totally agree with you on this, but I think the original scope of the `server` project is to follow as closely the API implementation.  If we do too much of a scope creep, then it's easy to get into the situation where there would be quite a large divergence between the GA4GH API and the server implementation.  My preference is to have changes in `schemas` and then they would be propagated into `server`.  This way tracing issues would become more straightforward and the complexity of the API increases (i.e. functional analysis, etc.) .

~p

jeromekelleher:
Closing this PR as it's out of date.

Issue #160, opened 2015-02-19
Create backend exception framework


dcolligan:
The frontend has an exception hierarchy, but the backend doesn't really have any custom exception support at the moment.

Here's a good example of an exception we would like to catch (triggered by searching for a callSet inside a variantSet which doesn't have it: `python client_dev.py -O variants-search http://localhost:8000 -r 1 -c HG03279 -V 1000g_2011`):

```
Traceback (most recent call last):
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/dannyc/ga4gh/server/ga4gh/frontend.py", line 174, in searchVariants
    return handleFlaskRequest(flask.request, app.backend.searchVariants)
  File "/Users/dannyc/ga4gh/server/ga4gh/frontend.py", line 116, in handleFlaskRequest
    return handleHttpPost(flaskRequest, endpoint)
  File "/Users/dannyc/ga4gh/server/ga4gh/frontend.py", line 81, in handleHttpPost
    responseStr = endpoint(request.get_data())
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 93, in searchVariants
    self.variantsGenerator)
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 65, in runSearchRequest
    for obj, nextPageToken in objectGenerator(request):
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 141, in variantsGenerator
    for variant in iterator:
  File "/Users/dannyc/ga4gh/server/ga4gh/datamodel/variants.py", line 211, in getVariants
    cols = [col for name, col in self._sampleCols[callSetId]]
KeyError: u'HG03279'
```

Currently, this just bails and returns to the client:

```
{
  "errorCode": 7,
  "message": "Internal server error"
}
```

dcolligan:
Fixed in #162

PR #162, opened 2015-02-19
Add backend_exceptions

dcolligan:
The payload that comes back from an exception with a defined more-specific message is now something like:
```
{
  "errorCode": 4,
  "message": "callSet 'HG03279' not in variantSet '1000g_2011'"
}
```

Issue #160

### DIFF
--- a/ga4gh/backend_exceptions.py
+++ b/ga4gh/backend_exceptions.py
@@ -1,0 +1,30 @@
+"""
+Exceptions for the backend
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+class BackendException(Exception):
+    """
+    Base class of backend exceptions
+    """
+    def __init__(self):
+        super(BackendException, self).__init__()
+
+
+class CallSetNotInVariantSetException(BackendException):
+    """
+    Indicates a request was made for a callSet not in the actual variantSet
+    """
+    exceptionMessage = "callSet '{0}' not in variantSet '{1}'"
+
+    def __init__(self, callSetId, variantSetId):
+        super(CallSetNotInVariantSetException, self).__init__()
+        self.callSetId = callSetId
+        self.variantSetId = variantSetId
+
+    def toErrorMessage(self):
+        return self.exceptionMessage.format(
+            self.callSetId, self.variantSetId)
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -14,6 +14,7 @@ import pysam
 import wormtable as wt
 
 import ga4gh.protocol as protocol
+import ga4gh.backend_exceptions as backendExceptions
 
 
 def convertVCFPhaseset(vcfPhaseset):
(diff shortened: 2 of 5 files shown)


ga4gh/frontend.py
```diff
@@ -105,7 +109,10 @@ def handleException(exception):
 
     error = protocol.GAException()
     error.errorCode = exception.code
-    error.message = exception.message
+    if errorMessage is not None:
```
Naburimannu: It seems we could get rid of this conditional by either:
1. instead of saving errorMessage, overwriting exception.message (move lines 97-98 here)
2. in line 96, making the default value of errorMessage be exception.message

dcolligan: I don't think either of these would work because exception is potentially a different object before and after the two if statements in the middle of `handleException`.  That's why `errorMessage` is saved before it is potentially lost.

ga4gh/backend_exceptions.py
```diff
@@ -0,0 +1,30 @@
+"""
+Exceptions for the backend (i.e. that aren't returned to clients)
```
jeromekelleher: This isn't strictly true, is it? We do return a `CallSetNotInVariantSetException` to the client, so they know what happened.

ga4gh/datamodel/variants.py
```diff
@@ -207,9 +208,13 @@ def getVariants(self, referenceName, startPosition, endPosition,
             callSetIds = self._sampleNames
         else:
             readCols = self._table.columns()[:self._firstSamplePosition]
-            for callSetId in callSetIds:
-                cols = [col for name, col in self._sampleCols[callSetId]]
-                readCols.extend(cols)
+            try:
+                for callSetId in callSetIds:
```
jeromekelleher: I think the `try` should be as close to the source of the potential error as possible, e.g.
```python
for  callSetId in callSetIds:
    try:
          cols = [col for name, col in self._sampleCols[callSetId]]
    except KeyError:
          raise backendExceptions.CallSetNotInVariantSetException(...)
    readCols.extend(cols)
```

ga4gh/frontend.py
```diff
@@ -93,6 +93,10 @@ def handleHttpOptions():
 
 @app.errorhandler(Exception)
 def handleException(exception):
```
jeromekelleher: This method is quite tricky now. I think it's find, but I'd have to search about quite hard to figure out why things are the way they are. Can we get some comments to explain the reason for each of the separate chunks?

ga4gh/backend_exceptions.py
dcolligan: I'll take the parenthetical out; its truth is arguable, at the very least

ga4gh/datamodel/variants.py
dcolligan: yeah, I guess I fell for the evil-by-premature-optimization trap

ga4gh/frontend.py
dcolligan: Yeah, the trickiness may be in part because of the complexity of the model we have.  I'll add some comments.

dcolligan:

jeromekelleher:
Looks good, +1.

Naburimannu:
+1

Issue #161, opened 2015-02-19
Create api versioning framework


dcolligan:
Currently, the reference server does not have any API versioning.  We need to change this.  

Goals:
- easily be able to release future API versions -- meaning:
  - reusing previous API endpoints across versions where appropriate
  - easily being able to identify in the code which endpoints correspond to which api versions
  - adding additional endpoints should have minimal overhead
-  enforce some kind of api lifecycle: this could be as simple as off/on but preferably will be a bit more flexible (development / released / deprecated / deactivated)

Possible Implementation Steps:
- define lifecycle
- create programming model
  - create tags, data structures, etc.
  - override Flask default URL routing
- refactor code to adhere to programming model
  - add version information to URLs
  - tag (or whatever) each endpoint with appropriate version information
  - separate out endpoints into version-specific files as a convention 
- perhaps modify server response payload to include endpoint status information (this is actually probably prohibited by the GA4GH protocol spec, but of course that can be subject to change...)
- modify html page to provide info on endpoints
- create tests to enforce programming model
- create compliance suite for api to prevent regressions (probably another issue altogether)

dcolligan:
In today's conference call we downscoped this issue to just adding version information to URLs since we are likely not going to need to support multiple versions for a long time.

dcolligan:
Fixed in #163

PR #163, opened 2015-02-20
Add version info to URLs

dcolligan:
Issue #161

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -23,8 +23,35 @@ import ga4gh.protocol as protocol
 app = flask.Flask(__name__)
 
 
-class ServerStatus(object):
+class Version(object):
+    """
+    A major/minor/revision version tag
+    """
+    @classmethod
+    def parseString(cls, versionString):
+        versions = versionString.strip('vV').split('.')
+        return Version(*versions)
 
+    def __init__(self, major, minor, revision):
+        self.version = (major, minor, revision)
+
+    def __cmp__(self, other):
+        return cmp(self.version, other.version)
+
+    def __hash__(self):
+        return hash(self.version)
+
+    def __eq__(self, other):
+        return self.version == other.version
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+
+class ServerStatus(object):
+    """
+    Generates information about the status of the server for display
+    """
     def __init__(self):
         self.startupTime = datetime.datetime.now()
 
@@ -111,11 +138,16 @@ def handleException(exception):
     return response
 
 
-def handleFlaskRequest(flaskRequest, endpoint):
+def handleFlaskPostRequest(version, flaskRequest, endpoint):
+    if Version.parseString(version) != Version.parseString(protocol.version):
+        raise frontendExceptions.VersionNotSupportedException()
+
     if flaskRequest.method == "POST":
         return handleHttpPost(flaskRequest, endpoint)
-    else:
+    elif flaskRequest.method == "OPTIONS":
         return handleHttpOptions()
+    else:
+        raise frontendExceptions.MethodNotAllowedException()
 
 
 @app.route('/')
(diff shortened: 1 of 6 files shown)


ga4gh/versioning.py
```diff
@@ -0,0 +1,152 @@
...
+                self.minor == other.minor and
+                self.revision == other.revision)
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+
+class RouteGroup(object):
+    """
+    A group that a route belongs to
```
Naburimannu: This comment is tautological. A simple summarizing of the second sentence of the file doc would be better.

ga4gh/versioning.py
```diff
@@ -0,0 +1,152 @@
...
+            self.disabledVersion = None
+
+
+class VersioningManager(object):
+
+    # TODO externalize data fields to yaml file?
+    supportedVersionsData = ['0.5.1']
+
+    groupData = {
+        'V_0.5.1_RELEASE': {
```
Naburimannu: I *think* I can intuit how you intend this to be used, but could we have some examples on the bug (if not in commit message or comments) of how you'd meaninfully have multiple of these or how they'd change over time?

ga4gh/versioning.py
```diff
@@ -0,0 +1,152 @@
...
+from __future__ import unicode_literals
+
+import functools
+
+
+class Lifecycle(object):
+    """
+    A lifecycle stage for a route group
+    """
+    # TODO more lifecycle stages?
```
Naburimannu: Devil's Advocate:
(1) why do we need explicitly track Deactivated groups? why not just remove support from the server when we deactivate a group, and return some sort of "unsupported" error for invalid versions?
(2) why do we need Development? Is this going to be enforced in code in some way, rather than just through convention and out-of-band communication?

(If so, then one might eventually want to have a BETA state, but I think this is plenty for an initial implementation.)

ga4gh/frontend.py
```diff
@@ -169,6 +173,7 @@ def searchVariantSets():
     return handleFlaskRequest(flask.request, app.backend.searchVariantSets)
 
 
-@app.route('/variants/search', methods=['POST', 'OPTIONS'])
+@app.route('/<version>/variants/search', methods=['POST', 'OPTIONS'])
+@versioningManager.policy('V_0.5.1_RELEASE')
```
Naburimannu: Really tiny nit: if we're using VX.Y.Z for version strings, why add the extra underbar to make V_X.Y.Z here?

ga4gh/versioning.py
```diff
@@ -0,0 +1,152 @@
...
+            'enabledVersion': '0.5.1',
+        },
+    }
+
+    def __init__(self, app):
+        self.groups = self._initGroups()
+        self.supportedVersions = self._initSupportedVersions()
+        self.app = app
+
+    def _initSupportedVersions(self):
```
Naburimannu: This makes sense if we're going to externalize the data as proposed in line 95, but otherwise couldn't we just declare ```supportedVersionsData = set([Version.parseString('0.5.1')])```?

And does it even make sense to externalize the data? If we require code changes for new versions, we might as well make part of the process being updating this code. If we don't require code changes for new versions (other than a generated protocol schema update), it might make sense to externalize, but then we'll need a process for managing multiple schema definitions?

ga4gh/versioning.py
```diff
@@ -0,0 +1,152 @@
...
+        return groups
+
+    def policy(self, routeGroup):
+        def inner(func):
+            @functools.wraps(func)
+            def decoratedFunction(version):
+                if self.isAccessAllowed(routeGroup, version):
+                    return func()
+                else:
+                    raise Exception('Access denied')  # TODO custom exception
```
Naburimannu: If we have multiple route groups, is the first one that fails going to throw an exception back to the client? How does this fail over to the "correct" route group?

ga4gh/versioning.py
```diff
@@ -0,0 +1,152 @@
...
+            return decoratedFunction
+        return inner
+
+    def isAccessAllowed(self, routeGroup, versionString):
```
Naburimannu: The function name is a bit odd because it both checks to see if the version is part of this route group AND, if so, if access is allowed.

Naburimannu:
I think it's clear from a couple of my comments that I don't know well enough how we're proposing to support multiple versions in a single server to evaluate parts of this. (Some of which may be implicit in the parts of Flask I haven't grokked yet?)

If we're not going to support multiple versions simultaneously, this feels like overkill.

dcolligan:
@Naburimannu you're totally right that this effort is not properly spec'ed out, but writing up an implementation, no matter how buggy, does tend to stimulate conversation!  

@jeromekelleher is it the case that we will need to support multiple protocol versions simultaneously?

ga4gh/versioning.py
dcolligan: 1) We could do that.  But there's no way to return any information about the endpoint's lifecycle status to the end user via the protocol, however.  The user would need to read docs somewhere.

2) To differentiate it from ACTIVE, a status which communicates "do not modify these endpoints for any reason because clients now depend on them"

But yes, our lifecycle should be well defined, in any case.

ga4gh/versioning.py
dcolligan: The idea is that every route belongs to one, and only one, route group.  

One example of how route groups could change is flipping the route group's lifecycle from development to active when we want to declare a set of endpoints stable.

Another is to set a disabledVersion for the route group when they are at the end of their lives.  

Etc.

ga4gh/frontend.py
dcolligan: To make it a unique string, for easier grepping through the code.

A case could be made the other way, for consistency.

I suppose I prefer it the way I did it, to avoid entangling identifiers, but I don't have strong feelings on this decision.

ga4gh/versioning.py
dcolligan: There's only one routeGroup per route

ga4gh/versioning.py
dcolligan: If we need to manage multiple versions at once, it would make sense that we put this super-important, non-code information in a config file that people knew to look at super carefully when changes were made to it.

Yes, I agree supporting multiple versions is more difficult than just multiple endpoints.  It's multiple exceptions, multiple protocol objects, etc. ... not sure if we want to go down this road or not, but it's worth a discussion.

dcolligan:
Okay, updated this branch with the downscoped version of this issue as we discussed.

Naburimannu:
Thanks, Danny. Looks like +1.

ga4gh/frontend.py
```diff
@@ -23,8 +23,46 @@
...
+class Version(object):
+    """
+    A major/minor/revision version tag
+    """
+    @classmethod
+    def parseString(cls, versionString):
+        versions = versionString.strip('vV').split('.')
+        return Version(*versions)
+
+    def __init__(self, major, minor, revision):
```
jeromekelleher: Python uses a tuple for this of the form `(major, minor, revision)`. The advantage then is comparisons (<, ==, etc) are for free. Do we gain much from this class? Where do we use it?

ga4gh/frontend.py
```diff
@@ -111,11 +149,16 @@ def handleException(exception):
     return response
 
 
-def handleFlaskRequest(flaskRequest, endpoint):
+def handleFlaskPostRequest(version, flaskRequest, endpoint):
+    if Version.parseString(version) != Version.parseString(protocol.version):
```
jeromekelleher: Why not `version != protocol.version`? It's easy to make this more general when we need it.

tests/utils.py
```diff
@@ -11,6 +11,17 @@
 
 import avro.schema
 
+import ga4gh.protocol as protocol
+
+
+def applyVersion(routes):
```
jeromekelleher: Why the multiple input types here? This is generally considered an anti-pattern in Python. It would be simpler to just have a single string parameter, and handle the 'listification' in client code.

ga4gh/frontend.py
dcolligan: I think the class is useful, but you're right that replacing the major/minor/version attributes with a tuple would allow us to eliminate a lot of the methods.

ga4gh/frontend.py
dcolligan: Because `version` and `protocol.version` are different strings.  Specifically, `version` has a v in front of it.  This is considered a best practice for a url version string.

ga4gh/frontend.py
dcolligan: er, simplify, not eliminate

ga4gh/frontend.py
jeromekelleher: Oh yeah - that 'v' is very nasty. OK, let's keep the class and just store a tuple internally then.

dcolligan:

jeromekelleher:
Thanks @dcolligan, +1.

Issue #115, opened 2015-02-01
`generate_schemas.py` script should use textwrap instead of forking an `fmt` process


jeromekelleher:
When generating the embedded text representing the JSON schema of a protocol element, the `generate_schemas.py` script pipes the text through `fmt` to wrap the text. The `textwrap` module from the standard library provides a much better way to do this.

MollyZhang:
working on this

dcolligan:
Fixed in #164

PR #164, opened 2015-02-20
use textwrap module to replace fmt in process_schemas.py

MollyZhang:
issue #115 
When generating the embedded text representing the JSON schema of a protocol element, the process_schemas.py script pipes the text through fmt to wrap the text. The textwrap module from the standard library provides a much better way to do this.

### DIFF
--- a/scripts/process_schemas.py
+++ b/scripts/process_schemas.py
@@ -96,17 +96,7 @@ class SchemaClass(object):
                         if isinstance(dic, dict):
                             stack.append(dic)
         jsonData = json.dumps(schema)
-        with tempfile.TemporaryFile() as tmp:
-            tmp.write(jsonData)
-            tmp.seek(0)
-            # Filter the text through fmt to make it look like acceptable code.
-            subproc = subprocess.Popen(
-                ["fmt"], stdout=subprocess.PIPE, stdin=tmp)
-            (output, err) = subproc.communicate()
-            exitStatus = subproc.wait()
-        if exitStatus != 0:
-            msg = "Error occured running fmt: {0}: {1}".format(exitStatus, err)
-            raise Exception(msg)
+        output = "\n".join(textwrap.wrap(jsonData)) + "\n"
         return output
 
     def formatRequiredFields(self):
@@ -356,7 +346,7 @@ class SchemaProcessor(object):
 def main():
     parser = argparse.ArgumentParser(
         description="Script to process GA4GH Avro schemas. Requires "
-        "java and fmt external commands")
+        "java external command")
     parser.add_argument(
         "--outputFile", "-o", default="ga4gh/_protocol_definitions.py",
         help="The file to output the protocol definitions to.")
--- a/ga4gh/_protocol_definitions.py
+++ b/ga4gh/_protocol_definitions.py
@@ -21,8 +21,9 @@ No documentation
 "name": "populationId"}, {"default": null, "doc": "", "type": ["null",
 "string"], "name": "referenceVersion"}, {"default": null, "doc": "",
 "type": ["null", "string"], "name": "chromosome"}, {"default": null,
-"doc": "", "type": ["null", "long"], "name": "coordinate"}, {"default":
-null, "doc": "", "type": ["null", "string"], "name": "allele"}]}
+"doc": "", "type": ["null", "long"], "name": "coordinate"},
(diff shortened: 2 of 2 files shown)


Naburimannu:
Can you commit the new protocol definitions file simultaneously? I'm surprised that you don't have to set up indentation and wrap-width the way I had to for ```__slots__```; I'd expect PEP8/FLAKE errors on the output.

dcolligan:
I agree with @Naburimannu

MollyZhang:
@dcolligan and @Naburimannu , I added the new _protocol_definitions.py script. It seems that it does pass the travis test, but I am not sure about whether the text wrapping is stable.

dcolligan:
When you run it repeated times, does it produce the same file?

You could automate this by setting up some kind of script for yourself that ran the process 100 (or whatever times) and compares the contents of the file.

MollyZhang:
@dcolligan OK I am working on this

dcolligan:
It also may be the case that there's some kind of machine-dependent setting that is making this produce slightly different output.  Once you determine it's stable on your machine, you can let us know and we can pull down your branch and try on our machine before we agree to commit this to master.

MollyZhang:
I wrote this following script to test generating the output 100 times, and they are the same. Not sure different machine will give different output.  

import filecmp
import tempfile
import subprocess

REPEAT = 100

def main():
    tmpDir = tempfile.mkdtemp(prefix="process_schemas_output_test_")
    files = []
    all_files_same = True
    for i in range(REPEAT):
        files.append("{0}/{1}.py".format(tmpDir, i))
        args = ["python",
            "/Users/Molly/PycharmProjects/server/scripts/process_schemas.py",
            "-o", files[-1], "v0.5.1", "--avro-tools-jar",
            "/Users/Molly/Downloads/avro-tools-1.7.7.jar"]
        subprocess.check_call(args)
        if i > 0:
            print "comparing file {0}.py with file 0.py".format(i)
            if filecmp.cmp(files[0], files[i]) is not True:
                all_files_same = False
                print "{0} and {1} is different".format(files[0], files[1])
                break
            print "same"

    print "all files are same?: {0}".format(all_files_same)

if __name__ == "__main__":
    main()

dcolligan:
Your script is kind of hard to read... in the future use triple-backticks to 
```
make it look like this
           and preserve formatting
```
You can read more about that in the markdown documentation.

pgrosu:
@MollyZhang, can you please re-edit the code above with proper indentation?  To post as python code use the triple back-ticks at the beginning and at the end, and then post inside that section, below is a link describing this:

https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code

Thanks,
Paul

dcolligan:
Another thing you could do would be to upload the file to gist, if you didn't want to fiddle with markdown, and paste the link in the comments.

pgrosu:
@MollyZhang, can you also include the following file:

```
process_schemas.py
```
It would probably better to have a Readme to ensure what the requirements for a proper setup and successful running are.

~p

pgrosu:
@MollyZhang, found it :)

scripts/process_schemas.py
```diff
@@ -356,7 +346,7 @@ def run(self):
 def main():
     parser = argparse.ArgumentParser(
         description="Script to process GA4GH Avro schemas. Requires "
-        "java and fmt external commands")
+        "java external commands")
```
dcolligan: nit: this should be singular -- "java external command", not "commands"

dcolligan:
I checked out a branch with this branch's changes and ran the tool and I didn't see any difference in the generated file from the one committed to this branch so... @MollyZhang after you fix my nit and the tests pass:

+1

MollyZhang:
@pgrosu  I added some comments in the script and included the `process_schemas.py` file and avro tools, thanks for the feedback

MollyZhang:
@dcolligan  fixed the nit

pgrosu:
@MollyZhang, thank you for the added comments.  It also works for me, based on my output below:

```
comparing file 97.py with file 0.py
same
comparing file 98.py with file 0.py
same
comparing file 99.py with file 0.py
same
all files are same?: True
$
[process_schemas_output_test_t9lTFo]$ for file in ./*.py; do diff "$file" "./0.py"; done
[process_schemas_output_test_t9lTFo]$
```

Thanks,
Paul

Naburimannu:
+1

Issue #166, opened 2015-02-20
Have server return JSON for 404s


dcolligan:
Currently, the server (probably ever since the html page change) returns html instead of json for 404s.  Change it to return a 404 json error object.

dcolligan:
Fixed in #167

PR #167, opened 2015-02-20
404s and 405s now return json instead of html

dcolligan:
To ensure correctness, we will probably need to do the same with just about every 4xx error message (at least), but I think we can just add these as time goes on and people hit html erroneously.

Issue #166

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -142,12 +142,9 @@ def handleException(exception):
         exception = frontendExceptions.ServerException()
 
     # serialize the exception
-    error = protocol.GAException()
-    error.errorCode = exception.code
+    error = exception.toGaException()
     if errorMessage is not None:
         error.message = errorMessage
-    else:
-        error.message = exception.message
     response = flask.jsonify(error.toJsonDict())
     response.status_code = exception.httpStatus
     return response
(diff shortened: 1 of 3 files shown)


ga4gh/frontend.py
```diff
@@ -172,3 +170,25 @@ def searchVariantSets():
...
     return handleFlaskRequest(flask.request, app.backend.searchVariants)
+
+
+# The below methods ensure that JSON is returned for various errors
+# instead of the default, html
+
+
+@app.errorhandler(404)
+def pathNotFoundHandler(errorString):
+    return handleHttpError(frontendExceptions.PathNotFoundException)
```
jeromekelleher: Probably clearer to pass in an instance of the exception here rather than the class, i.e.,
```python
return handleHttpError(frontendExceptions.PathNotFoundException())
```

tests/unit/test_views.py
```diff
@@ -29,8 +29,12 @@ def sendVariantSetsSearch(self, data):
...
                              data=data)
 
-    def testServer(self):
-        self.assertEqual(404, self.app.get('/doesNotExist').status_code)
+    def test404sReturnJson(self):
+        response = self.app.get('/doesNotExist')
+        exception = protocol.GAException.fromJsonString(
+            response.get_data())
+        self.assertEqual(404, response.status_code)
+        self.assertEqual(5, exception.errorCode)
```
jeromekelleher: Can you add a TODO here to indicate we need to get rid of the hard coded GA error codes? HTTP error codes are fine as these are familiar.

jeromekelleher:
+1 -- nice catch @dcolligan.

dcolligan:
@jeromekelleher pushed new commit addressing your concerns

dcolligan:
rebased this on top of master... should be good to go

PR #171, opened 2015-02-23
Basic infrastructure for top-level objects.

jeromekelleher:
This PR is intended to add initial support for the top level objects in the reference and read hierarchies. It adds to the existing infrastructure in the simplest and least intrusive way, and should not be seen as 'finished work'.

The biggest difference here from a user perspective is that I've changed the required structure of the input `dataDir`. This was assumed to contain variantSets, and must now contain three directories: `variants`, `reads` and `references`. This is not the structure that we ultimately want (references should be independent of datasets), but it will get us going for now. If we agree on this structure I'll update the `ga4gh-example-data` set so that it follows this new layout.

There is no testing of the new code, since we don't have any example or test data to test it with. I'll be opening some issues for this. I hope that we can develop the code required to handle these top level objects independently of the changes we'll need to deal with the whole hierarchy in #158.

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -9,6 +9,8 @@ from __future__ import unicode_literals
 import os
 
 import ga4gh.protocol as protocol
+import ga4gh.datamodel.references as references
+import ga4gh.datamodel.reads as reads
 
 
 class Backend(object):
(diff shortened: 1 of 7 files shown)


ga4gh/backend.py
```diff
@@ -18,14 +20,34 @@ class Backend(object):
     """
     def __init__(self, dataDir, variantSetClass):
         self._dataDir = dataDir
+        # Variants
         self._variantSetIdMap = {}
-        # All directories in datadir are assumed to correspond to VariantSets.
-        for variantSetId in os.listdir(self._dataDir):
-            relativePath = os.path.join(self._dataDir, variantSetId)
+        variantSetDir = os.path.join(self._dataDir, "variants")
```
dcolligan: It seems like you're repeating yourself a lot here over the three object types... this could be refactored.

ga4gh/datamodel/reads.py
```diff
@@ -5,3 +5,57 @@
...
+    def __init__(self, id_, dataDir):
+        self._id = id_
+        self._dataDir = dataDir
+        self._readGroups = []
+        for fileType in ["sam", "bam"]:
+            pattern = "*.{}".format(fileType)
+            for path in glob.glob(os.path.join(self._dataDir, pattern)):
+                filename = os.path.split(path)[1]
+                localId = filename.split(".")[0]
+                readGroupId = "{}:{}".format(self._id, localId)
```
dcolligan: As a matter of style, should we prefer this format string syntax to (in this example, for instance) `"{0}:{1}"`

ga4gh/datamodel/references.py
```diff
@@ -5,3 +5,56 @@
...
+
+import ga4gh.protocol as protocol
+
+
+class ReferenceSet(object):
+    """
+    TODO document.
+    """
+    def __init__(self, id_, dataDir):
+        self._id = id_
```
dcolligan: Not sure if it would be worth deriving all of these objects from a common superclass.  They all seem to have an id, for instance... something to consider.

dcolligan:
Generally looks good, provided the TODOs are filled in.

ga4gh/datamodel/reads.py
jeromekelleher: I think so - we're not trying to support older versions of Python, and this form has been supported in 2.7 for a while now.

ga4gh/backend.py
jeromekelleher: I'm deliberately not trying to be tidy and refactor things nicely, as this code should be very short lived. Even a cleaned up and refactored version of this code is very ugly, and I'd rather it be obviously very ugly! I should add a TODO to this effect though.

ga4gh/datamodel/reads.py
dcolligan: https://docs.python.org/2/library/string.html#format-specification-mini-language

It looks like the only downside of using this syntax is that one can not reuse arguments in the string.  Usually that's fine, however.

ga4gh/datamodel/references.py
jeromekelleher: Definitely agree, we will certainly want a common superclass here. I'm not sure quite what it'll look like yet though, so I thought it best to just get something simple that works first.

jeromekelleher:
I've updated this and filled in the documentation TODOs @dcolligan. There are a lot of other TODOs, which I was hoping to turn into issues, as they are quite substantial pieces of work. How does it look to you?

dcolligan:
My opinion is unchanged.  Are you looking for a +1?

jeromekelleher:
Yeah - I was hoping to merge this and use the TODOs as issues so others could work on them. What would you like to see fixed before merging?

dcolligan:
Oh, nothing in particular.  You seem to be deliberately leaving this change ugly, which is fine, so I just wanted to clarify if this was the version you wanted to merge or not.

+1

Issue #116, opened 2015-02-01
`postMethods` should have a stable ordering in `generate_schemas.py`


jeromekelleher:
The postMethods should be sorted before writing, so that we have a stable ordering. See #114 for a discussion.

MollyZhang:
working on this

dcolligan:
Fixed in #174

PR #174, opened 2015-02-24
Solving unstable postMethods ordering in process_schemas.py

MollyZhang:
solving issue #116, which about that the postMethod is unstably ordered. 
The reason for the unstable ordering is because of line 207, the glob.glob() function returns an arbitrary order. Here is a discussion about glob.glob ordering
http://stackoverflow.com/questions/6773584/how-is-pythons-glob-glob-ordered


jeromekelleher:
+1

Naburimannu:
+1

dcolligan:
Great job, @MollyZhang

PR #177, opened 2015-02-24
Fixes import issue for tests utils.

jeromekelleher:
This fixes an issue in which `tests/test_backends.py` ERRORed when run in isolation using
```
nosetests -vs tests/test_backends.py
```

### DIFF
--- a/tests/unit/test_backends.py
+++ b/tests/unit/test_backends.py
@@ -17,7 +17,7 @@ import subprocess
 
 import wormtable as wt
 
-import tests
+import tests.utils as utils
 import ga4gh.backend as backend
 import ga4gh.protocol as protocol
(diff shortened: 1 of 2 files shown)


dcolligan:
That's weird... do you have an explanation of why this would break before this patch and how this patch fixes it?

dcolligan:
Oh, I see... nevermind.  Can you also fix this in other tests, too (even if it is a question of consistent style)?  Check out `test_schemas.py`

jeromekelleher:
Done - thanks for pointing this one out. To answer your question, I don't really have an explanation. Luckily, the method that works happens to follow our preferred style, so all is good.

Naburimannu:
+1

PR #179, opened 2015-02-24
Fixed workarounds bug.

jeromekelleher:
The recent change to the default value for `workarounds` left the client program in a broken state, as a lot of code assumed that the `workarounds` variable was always a string and not None. This changes the behaviour back, and fixes the bug.

Also changed the default value for referenceName to `1`, as `chrSim` was an ancient and useless relic.

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -258,9 +258,7 @@ class AbstractQueryRunner(object):
     Abstract base class for runner classes
     """
     def __init__(self, args):
-        self._workarounds = set()
-        if args.workarounds is not None:
-            self._workarounds = set(args.workarounds.split(','))
+        self._workarounds = set(args.workarounds.split(','))
         self._key = args.key
         self._verbosity = args.verbose
         self._httpClient = client.HttpClient(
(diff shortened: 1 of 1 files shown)


Naburimannu:
+1

I assume we didn't have any test coverage that hit workarounds?
Searching reads by refId is (I believe) still requiring workaround in the Google implementation.

dcolligan:
What code assumed `workarounds` was a string?

jeromekelleher:
The constructor of `RequestFactory`.

jeromekelleher:
@Naburimannu, this was actually in the nominal no-workarounds case. We should remedy this with some end-to-end tests with our client and server apps running.

dcolligan:
I'd rather change the default sink value of command line arguments to be none, rather than the empty string, even if it means changing the `RequestFactory` constructor.

As I've mentioned before, this avoids the failure mode of passing empty strings around in requests.

jeromekelleher:
OK, but do it in a follow up PR. The client is currently broken.

dcolligan:
Fair enough.  That would go well with an end-to-end test to prevent regressions.

dcolligan:
By the way, I remember some discussion of this on the conference call, but did we ever settle on the definition of unit vs integration vs end-to-end tests?

jeromekelleher:
I think we decided to leave them all as `unit` tests until we made some tests that take a _long_ time to run.

PR #182, opened 2015-02-24
Added install requires for humanize package.

jeromekelleher:
Fixes missing install requirement.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,8 +28,8 @@ f.close()
 ga4ghVersion = parseVersion("ga4gh/__init__.py")
 # Flask must come after all other requirements that have "flask" as a prefix
 # due to a setuptools bug.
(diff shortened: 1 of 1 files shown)


dcolligan:
Can't you just change that line to read in and parse the `requirements.txt` file, so we don't have to keep updating them in parallel?

pgrosu:
@dcolligan, I totally agree with this approach, but I could see an issue with automatizing the `wormtable` installation if it's not installed.

dcolligan:
A problem that isn't present by hardcoding it in the `setup.py` file?

pgrosu:
Good point :)

jeromekelleher:
They are for different things @dcolligan --- setup.py contains only what we need to deploy, whereas requirements.txt contains everything we need to develop.

dcolligan:
+1

PR #184, opened 2015-02-24
Sink values and end to end test

dcolligan:
change cli sink values from empty string to None

add end to end test

Questions: 
1) do want want to incorporate this new end-to-end test into the travis CI run?  If so, we have one of two choices: a) make a dummy backend for the ga4gh_server that doesn't require any genome data to be on the file system b) make sure data is on the file system when Travis runs somehow
2) what test suite should the test I've written be in?  or...
3) should we factor out the more general functionality I've written into a test framework for one of the suites?


ga4gh/cli.py
```diff
@@ -196,6 +202,9 @@ def addGlobalOptions(parser):
     parser.add_argument(
         "--config-file", "-F", type=str,
         help="The configuration file to use")
+    parser.add_argument(
```
jeromekelleher: Why do we want this? The reloader is generally pretty handy, and this server is just for development...

jeromekelleher:
I'm still a bit hazy on why `default=None` is better than `default=''`. Sorry for being slow, but could you explain this again?

I like the end-to-end test, but will need a bit of time to digest what's going on here.

ga4gh/cli.py
dcolligan: When starting the server within a test, we want one instance of the server to be launched, not two processes.  If the later, it makes it difficult to kill the server.

(Reloading is only useful to pick up changes in code when developing; when running a test this should not be happening.)

dcolligan:
Because `default=''` also requires special-case handling code, and is a less obvious sink value.  The likely result is that we send requests to the server that erroneously have `attribute: ''` instead of what we really want, `attribute: None`

dcolligan:
I've rearranged the code a bit and the Travis CI build is now failing in a way that I expected it to (because the data is not there) and reporting the error as such.

jeromekelleher:
I think we should defer this until after we've got simulated data working --- it's too difficult to do any real testing otherwise.

dcolligan:
Agreed, this change depends on a simulated backend.

dcolligan:
Closing for now

Issue #159, opened 2015-02-18
update README to include requirements.txt


dcolligan:
The setup described in the README is a bit out of date

macieksmuga:
Updating it based on my recent install experience

macieksmuga:
I wanted to get quick feedback from others before I modify the read, but here's my experience with installing the server from scratch (on a MacOS X 10.10 laptop with macports), and trying to run the nosetests:

The one external dependency not explicitly covered by the readme file is wormtable's dependency on Berkeley DB. While a link to the document that covers how to satisfy this dependency is eventually given, I was thinking of moving it up a couple of sentences in the readme, so that a first-time user following along with  the instructions doesn't get the inevitable error the first time they try to pip install wormtable.

The second thing I noticed, when running the nosetests, is that flask has presumably created a bunch of dependencies that aren't sufficiently covered by the existing requirements.txt file, namely:

mccabe, pyflakes, werkzeug, jinja2, itsdangerous, six

Unless I hear otherwise, I'll update the requirements.txt file with the above packages.

pgrosu:
@macieksmuga, absolutely agree and everything has to be documented clearly and in a step-by-step fashion, for different OS environments on both server or local implementation with logging for debugging.  This does not need to be complete in the first iteration, and can be streamlined over time.

It took me a couple of hours of setting everything up from scratch back in December on a cluster under only user-privilege level.  It definitely has to be a turn-key solution and we can't rely on the user to have significant expertise in setting things up, since working with the implementation and possibly updating the code would have a higher priority.

Paul

jeromekelleher:
It's probably a good idea to emphasise the Berkeley DB dependency for wormtable, as this would be a nasty stumbling block for a lot of users. We are planning on removing wormtable soon (hopefully!), so I wouldn't put too much work into this though.

For requirements.txt, I don't think there's any need to explicitly list the recursive requirements for Flask etc. Pip works these out, so we should just keep it to the list of packages that we use directly.

macieksmuga:
Thanks @jeromekelleher and @pgrosu for your feedback! 
Concerning the automatic dependency graph in pip, it didn't quite work out the first time in my case - which is why I had to perform the above listed pip installs manually after running the initial pip install -r requirements.txt 

I realized that the reason the process failed was due to wormtable install exiting on error. Once that was fixed, the flask dependencies were loaded automatically as expected.

PR #185, opened 2015-02-24
Updated README file with install instructions for first-time users.

macieksmuga:
The README was updated as follows:

*A new "installing the server" section was added near the top, specifying the steps for a successful first-time install, with examples.
*The documentation was updated to reflect running pip with the requirements.txt file
* the test section (running nosetests) originally located at the bottom of the file was moved up into the main install section, so that first-time users can easily verify their install.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -13,7 +13,7 @@ If you would like to help, please check out our list of
 
 Our aims for this implementation are:
 
-Simplicity/clarity
+**Simplicity/clarity**
     The main goal of this implementation is to provide an easy to understand
     and maintain implementation of the GA4GH API. Design choices
     are driven by the goal of making the code as easy to understand as
@@ -22,19 +22,73 @@ Simplicity/clarity
     that is useful in many cases where the extremes of scale are not
     important.
 
-Portability
+**Portability**
     The code is written in Python for maximum portability, and it
     should be possible to run on any modern computer/operating system (Windows
-    compatibility should be possible, although this has not been tested).  We use a
-    subset of Python 3 which is backwards compatible with Python 2 following the
-    current `best practices <http://python-future.org/compatible_idioms.html>`_.
-    In this way, we fully support both Python 2 and 3.
+    compatibility should be possible, although this has not been tested). Our coding
+    guidelines specify using a subset of Python 3 which is backwards compatible with Python 2 
+    following the current `best practices <http://python-future.org/compatible_idioms.html>`_.
+    The project currently does not yet support Python 3, as support for it is lacking in several
+    packages that we depend on. However, our eventual goal is to support both Python 2
+    and 3.
 
-Ease of use
+**Ease of use**
     The code follows the `Python Packaging User Guide
-    <http://python-packaging-user-guide.readthedocs.org/en/latest/>`_. This will
-    make installing the ``ga4gh`` reference code very easy across a range of
-    operating systems.
+    <http://python-packaging-user-guide.readthedocs.org/en/latest/>`_. 
+    Specifically, pip is used to handle python package dependencies (see below 
+    for details). This allows for easy installation of the ``ga4gh`` reference code 
+    across a range of operating systems.
+
+
+*******************************
+Installing the reference server
+*******************************
+
+Prerequisites:
+
+* Python 2.7,
+* Berkeley DB together with include and lib files (version 4.8 or higher),
+* Virtualenv (or another python sandboxing tool) is highly recommended.
+
+General installation procedure:
+
+* Install Berkeley DB (version 4.8 or higher) using your system's preferred 
+  package manager, see the `wormtable help page 
+  <https://pypi.python.org/pypi/wormtable>`_ for platform-specific details.
+
+* (On MacOS X, make sure the LDFLAGS and CFLAGS environment variables are set to
+  include the lib and include directories for the Berkeley DB install of your choice. 
+  The wormtable help page cited above provides more detailed instructions, or
+  see the `Appendix`_ to this README for an example install on that platform.)
+
+* Create a python sandbox directory using virtualenv, preferably 
+  *not* inside the ga4gh server directory. For an good introduction
+  to using virtualenv, see the `Python Guide page
+  <http://docs.python-guide.org/en/latest/dev/virtualenvs/>`_.
+  On some systems, you may need to specify the --no-site-packages
+  option to ensure a clean dependency install. For example, to 
+  create a sandbox named ``testenv``::
+
+  $ virtualenv --no-site-packages testenv
+
+* Make the virtualenv sandbox created above active::
+
+  $ source testenv/bin/activate
+
+* cd to the ga4gh server directory, and load the dependencies via pip::
+
+  $ cd [your ga4gh server directory]
+  $ pip install -r requirements.txt
+
+* Finally, run the install script, and run nosetests to confirm the install::
+
+  $ python setup.py install
+  $ nosetests
+
+A successfull install should result in a clean run of all the tests, 
+resulting in a line of dots followed by ``OK``. If this still isn't working,
+you may want to check the `Appendix`_ at the end of this README for 
+system-specific installation examples.
 
 ********************************
 Serving variants from a VCF file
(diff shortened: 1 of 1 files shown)


macieksmuga:
Note, this pull request is to resolve issue #159

pgrosu:
@macieksmuga This is great!  One small additional thing.  See if you can add this optional script that makes a local installation of `wormtable` automatic for non-root access - try it first since I patched this together a while ago.  Hopefully this will make it easier for non-Linux users - feel free to change it to what you think is best:

```
#!/bin/bash

# Download wormtable

wormtable_installation=`curl  https://pypi.python.org/pypi/wormtable | grep "tar.gz" | sed -e 's/\"/\n /g' | grep https | head -n1`

wget $wormtable_installation

wormtable_file=`echo $wormtable_installation | sed -e 's/\//\n /g' | tail -n1`

tar xzvf $wormtable_file

wormtable_directory=`tar tzvf $wormtable_file | head -n1 | sed -e 's/\///g' | sed -e 's/ /\n/g' | tail -n1`
wormtable_directory=`pwd`"/"$wormtable_directory

# Install berkley-db and wormtable

mkdir -p $HOME/.local
wget http://download.oracle.com/berkeley-db/db-4.8.30.tar.gz
tar -zxf db-4.8.30.tar.gz
cd db-4.8.30/build_unix/
../dist/configure --prefix=$HOME/.local
make install

cd $wormtable_directory

CFLAGS=-I$HOME/.local/include LDFLAGS=$HOME/.local/lib/ python setup.py install --user

export LD_LIBRARY_PATH=$HOME/.local/lib:$LD_LIBRARY_PATH
export PATH=$HOME/.local/bin:$PATH
```
Thanks,
Paul

README.txt
```diff
@@ -32,9 +32,70 @@ Portability
...
+    across a range of operating systems.
+
+
+*******************************
+Installing the reference server
+*******************************
+
+Prerequisites:
+
+- Python 2.7 or higher,
```
jeromekelleher: Delete "or higher" (we don't support 3.x)

jeromekelleher: s/48/4.8

README.txt
```diff
@@ -32,9 +32,70 @@ Portability
...
+- Virtualenv (or another python sandboxing tool) is highly recommended.
+
+General installation procedure:
+
+- Install Berkeley DB (version 48 or higher) using your system's preferred 
+  package manager, see the `wormtable help page 
+  <https://pypi.python.org/pypi/wormtable>`_ for platform-specific details.
+  In one developer's experience, installing db48 was preferable to db53 
+  on MacOS X via MacPorts. For example (on MacOS X with MacPorts):
+
```
jeromekelleher: I realise a lot of people use macs, but these are very mac specific instructions. For Debian/Ubuntu we can just `sudo apt-get install libdb-dev` and we're done. Can't we just provide a link to the wormtable documentation here?

README.txt
```diff
@@ -32,9 +32,70 @@ Portability
...
+Prerequisites:
+
+- Python 2.7 or higher,
+- Berkeley DB together with include and lib files (version 48 or higher),
+- Virtualenv (or another python sandboxing tool) is highly recommended.
+
+General installation procedure:
+
+- Install Berkeley DB (version 48 or higher) using your system's preferred
```
jeromekelleher: s/48/4.8

README.txt
```diff
@@ -32,9 +32,70 @@ Portability
...
+  $ export CFLAGS=-I/opt/local/include/db48/  
+  $ export LDFLAGS=-L/opt/local/lib/db48/
+
+- Create a python sandbox directory using virtualenv, preferably 
+  NOT inside the ga4gh server directory. For an good introduction
+  to using virtualenv, see the `Python Guide page
+  <http://docs.python-guide.org/en/latest/dev/virtualenvs/>`_.
+  On some systems, you may need to specify the --no-site-packages
+  option to ensure a clean dependency install. For example, to 
+  create a sandbox named ``testenv``:
```
jeromekelleher: This is RST, so we need two colons for these example sections to get the rendering right. i.e. " create a sandbox named ``testenv``::"

jeromekelleher:
This is a good update, thanks @macieksmuga. I'm conscious of the fact that we're trying to get rid of our dependency on wormtable and start working towards supporting VCF directly, so I'm not sure how much effort we should put in to making the Berkeley DB install experience as smooth as possible.

pgrosu:
Regarding making the optional script that makes the Berkley DB/Wormtable install experience as smooth as possible, I understand what you mean but currently it's still part of the requirements and we are posting it as a main part of the Readme.  Until we have the direct-VCF approach as the preferred deployment implementation, I think it would be helpful to make it as easy as possible to accommodate as many new users as possible to help on our project.  What do you think?

dcolligan:
Every *nix box should have a package manager that can install BDB.  If we provide any instructions for installing it, it should be about using apt, rpm, ports, whatever to install the required package, not a script for downloading a source install.

pgrosu:
@dcolligan, I agree but I just wanted to provide something as a last resort that will always work with minimal requirements under any environment. This is in case people just wanted to quickly get started without too much additional overhead.  Not all people are familiar with *nix boxes and installing packages - I've had my share of biologists that said "I just want to run it.  Give me one command."

dcolligan:
Can you verify your script works under any environment?  That's a trick question, of course.  Just to briefly enumerate some of the failure modes: no bash installed, no curl installed, no wget installed, oracle breaks their links, permissions problems... in short, with this script, I think, under many circumstances we are granting our unix-ignorant biologist even more latitude to screw up.

In any event, I don't think pure-and-simple biologists are going to find this server useful for a long time, and hopefully by that point BDB will be removed as a dependency.

pgrosu:
I was providing it as a starting point for a script with minimal requirements - again it's only a last resort.  We can all test it for our environments and write a disclaimer.  The only goal of this was so that the installation would not be the bottleneck, which was for me initially, but that might be because of some of the policies in the server environment I was running it under and I wanted the flexibility in debugging all the parts.

Regarding the package requirements, bash can be substituted by another shell, and wget/curl if not available can be temporarily locally compiled.

macieksmuga:
Would it make sense to have a couple of system-specific appendices at the end of the README file that people could refer to if they can't get the server running on their particular platform (because Wormtable/Berkeley DB)? Any platform-specific example code could then be relegated to this appendix.

dcolligan:
Sure, you can do that.

macieksmuga:
@pgrosu I added an appendix to the end of the README that can include more detailed install recipes for specific platforms. Caveat: I agree with @dcolligan about the desirability of keeping all dependency installs via a suitable platform-specific package manager, and not from source.

pgrosu:
@macieksmuga, Thank you for adding it.  I agree with both you and @dcolligan on this, but until we streamline the installation process for multiple environments, I feel that providing users with as many solutions that are as straightforward as possible, will help them and enable them to get excited to quickly start working with the code.

README.txt
```diff
@@ -30,11 +30,65 @@ Portability
...
+* Berkeley DB together with include and lib files (version 4.8 or higher),
+* Virtualenv (or another python sandboxing tool) is highly recommended.
+
+General installation procedure:
+
+* Install Berkeley DB (version 48 or higher) using your system's preferred 
+  package manager, see the `wormtable help page 
+  <https://pypi.python.org/pypi/wormtable>`_ for platform-specific details.
+
+* Make sure the LDPATH and CPATH environment variables include the lib
```
jeromekelleher: s/PATH/FLAGS

This is still OSX specific --- on Debian and Redhat, there's no need to worry about LDFLAGS and CFLAGS as they have software managers that actually work. However, it is a nasty stumbling block and lots of people do use macs, so  maybe just start this with "On OSX, make sure that LDFLAGS and CFLAGS...".

jeromekelleher:
This is great @macieksmuga --- the appendix is exactly the right thing to do. We should file an issue to fill in an example for Debian and Redhat derivatives also.

I'm +1 modulo the niggles above.

macieksmuga:
@jeromekelleher should I go ahead and squash the various micro-commits on this PR?

dcolligan:
If you're ready to submit you should always have the commit(s) in a form that makes sense to other developers to read.  Usually that means squashed down to one commit.

Also, please rebase your branch on top of master in the future rather than merging master into it.  If you do it this way, you'll get commits irrelevant to your changes included in your branch ("Added install requires for humanize package." here)

README.txt
```diff
@@ -30,11 +30,63 @@ Portability
...
+
+Prerequisites:
+
+* Python 2.7,
+* Berkeley DB together with include and lib files (version 4.8 or higher),
+* Virtualenv (or another python sandboxing tool) is highly recommended.
+
+General installation procedure:
+
+* Install Berkeley DB (version 48 or higher) using your system's preferred
```
dcolligan: Eliminate this parenthetical ... '48' is wrong, and you already stated the required version above

macieksmuga:
Since #191 can be fixed with a related edit to the README, I incorporated it into this pull request. Otherwise, all of @jeromekelleher and @dcolligan 's suggestions re #159 should be good to go.

dcolligan:
+1

README.txt
```diff
@@ -22,19 +22,73 @@ Simplicity/clarity
...
+* Berkeley DB together with include and lib files (version 4.8 or higher),
+* Virtualenv (or another python sandboxing tool) is highly recommended.
+
+General installation procedure:
+
+* Install Berkeley DB (version 4.8 or higher) using your system's preferred 
+  package manager, see the `wormtable help page 
+  <https://pypi.python.org/pypi/wormtable>`_ for platform-specific details.
+
+* (On MacOS X, make sure the LDPATH and CPATH environment variables are set to
```
jeromekelleher: LDFLAGS and CFLAGS.

jeromekelleher:
+1 once the CFLAGS and LDFLAGS has been fixed.

macieksmuga:
@jeromekelleher yikes - thanks for catching that!

Issue #189, opened 2015-02-25
Data-updating dev script


dcolligan:
Provide a script that, when run, correctly initializes the database.

jeromekelleher:
This is a bit vague @dcolligan --- we're just talking about the example data for development here, right?

I'm definitely +1 on the idea though.

pgrosu:
This is basically this:

https://github.com/ga4gh/server/issues/178#issuecomment-75970882

with the need to ensure things start from a clean environment.

And definitely +1.  If the data gets into the terabytes then we might need another solution, but for now we definitely want people to focus on the API more than worrying about having all the proper data/clean environment which should happen in the background - and this will basically solve it.

dcolligan:
Closed in #190

PR #190, opened 2015-02-25
Add update_data.py script

dcolligan:
Sample run:
```
$ python scripts/update_data.py
Archiving 'ga4gh-example-data' to 'ga4gh-example-data-archive/Wed_Feb_25_13:06:03_2015/ga4gh-example-data'
Archiving 'ga4gh-example-data.tar.gz' to 'ga4gh-example-data-archive/Wed_Feb_25_13:06:03_2015/ga4gh-example-data.tar.gz'
Archiving finished
Downloading 'http://www.well.ox.ac.uk/~jk/ga4gh-example-data.tar.gz' to 'ga4gh-example-data.tar.gz'
Downloading... 100.00% finished (  164M / 164M  )  # interactive ftw
Downloading finished
Extracting ga4gh-example-data.tar.gz
Extracting finished
Finished in 1 minute(s), 59 second(s)
```

Issue #189


dcolligan:
You know, we may just be able to include this in the `.travis.yml` if we wanted to run end-to-end tests on travis... not sure if that's either doable or desirable.

pgrosu:
+1, looks great!  I agree that having it in travis would help consistency and completeness, as long as the data is not too large. For now probably not having it in travis would speed things up, since I'm sure we might go through other iterations of how the data will get structured and thus downloaded.

Many thanks for creating this,
Paul

dcolligan:
Added an update... turns out the humanize module has got tons of functionality that I basically duplicated in a previous commit.

scripts/update_data.py
```diff
@@ -0,0 +1,138 @@
...
+            return result
+        return wrapper
+
+    def _report(self):
+        delta = self.end - self.start
+        timeString = humanize.time.naturaldelta(delta)
+        log("Finished in {} ({} seconds)".format(timeString, delta))
+
+
+class FileDownloader(object):
```
dcolligan: should this class be moved to its own module?  if so, where?

jeromekelleher: We do something very similar in `process_schemas.py`. Maybe we could have a `scripts/utils.py` module and reuse the FileDownloader here and in `process_schemas`? We'll need to start worrying about where these things are run, but it should be workable.

jeromekelleher:
This looks great @dcolligan, thanks. What do you think of the `scripts/utils.py` module? Perhaps this would also be a natural home for the decorators you proposed also in #192?

scripts/update_data.py
dcolligan: `scripts/utils.py` would be a good idea and I agree that `FileDownloader` should go in there.

I think we can safely assume that everything in `scripts` is run from the root server directory.

dcolligan:
The decorators should go in their own module under `ga4gh` since they are so general purpose and right now used mostly in `tests`.  I could see them being used in core server code as well.

Naburimannu:
+1, just holding off the merge because it isn't clear that Jerome & Danny have reached agreement?

dcolligan:
Yeah, hold off on the merge; I'm doing some refactoring.

jeromekelleher:
It's up to Danny - I'm happy to merge on this if he'd rather do the `scripts/utils` thing later (or not at all).

dcolligan:
Alright, pushed an update.  (Note the new launching syntax detailed in the commit message)

scripts/google_examples.py
```diff
@@ -11,6 +10,8 @@
 
 import yaml
 
+import scripts.utils as utils
```
jeromekelleher: If we express this as 
```python
import utils
```
we don't need to change the way we run the scripts. We also don't need to turn `scripts` into a package, so there's no need for the `__init__.py` file.

jeromekelleher:
This is really cool, thanks @dcolligan. Should we change the import structure? I think it'd be a bit simpler to use the `import utils` form, and would mean we can continue using `python scripts/x.py` method to run things, which means we can tab-complete in the shell.

Other than that, a resounding +1.

Naburimannu:
I certainly prefer doing it the way @jeromekelleher suggests, but I thought
@dcolligan said that wasn't working for him for some reason? Did we figure
out why?

dcolligan:
Alright, reverted to the old script launch method... should be good to merge now.

PR #192, opened 2015-02-26
add decorators module

dcolligan:
These are decorators that I have been adding in to scripts and tests in my PRs over the past couple days but are more generally useful, so I decided to put them into their own module.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -5,13 +5,96 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import functools
 import itertools
 import random
+import signal
 import string
+import time
 
 import avro.schema
 
 import ga4gh.protocol as protocol
+
+
+class TimeoutException(Exception):
+    """
+    A process has taken too long to execute
+    """
+
+
+class Repeat(object):
+    """
+    A decorator to use for repeating a tagged function.
+    The tagged function should return true if it wants to run again,
+    and false if it wants to stop repeating.
+    """
+    defaultSleepSeconds = 1
+
+    def __init__(self, sleepSeconds=defaultSleepSeconds):
+        self.sleepSeconds = sleepSeconds
+
+    def __call__(self, func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            while func(*args, **kwargs):
+                time.sleep(self.sleepSeconds)
+        return wrapper
+
+
+class Timeout(object):
+    """
+    A decorator to use for only allowing a function to run
+    for a limited amount of time
+    """
+    defaultTimeoutSeconds = 60
+
+    def __init__(self, timeoutSeconds=defaultTimeoutSeconds):
+        self.timeoutSeconds = timeoutSeconds
+
+    def __call__(self, func):
+
+        def _handle_timeout(signum, frame):
+            raise TimeoutException()
+
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            try:
+                # set the alarm and execute func
+                signal.signal(signal.SIGALRM, _handle_timeout)
+                signal.alarm(self.timeoutSeconds)
+                return func(*args, **kwargs)
+            finally:
+                # clear the alarm
+                signal.alarm(0)
+        return wrapper
+
+
+class AssertRaises(object):
+    """
+    Decorator that asserts a type of error is raised in the test.
+    Must be placed on a method of a class that derives from
+    unittest.TestCase.
+    """
+    def __init__(self, exceptionType):
+        self.exceptionType = exceptionType
+
+    def __call__(self, func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            exceptionRaised = False
+            testCaseSelf = args[0]
+            try:
+                func(*args, **kwargs)
+            except Exception as exception:
+                exceptionRaised = True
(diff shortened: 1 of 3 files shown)


tests/utils.py
```diff
@@ -14,6 +15,35 @@
...
 
 
+class AssertRaises(object):
+    """
+    Decorator that asserts a type of error is raised in the test.
+    Must be placed on a method of a class that derives from
+    unittest.TestCase.
+
+    unittest has an assertRaises method but it doesn't seem to work...
+    """
```
jeromekelleher: unittest.assertRaises works just fine in my experience - I don't see any need for this class.

jeromekelleher:
What are the use cases for these in the main code base? Until a decorator is used in more than one module it should be defined where it's used --- I don't see any benefit from having a separate module right now.

ga4gh/decorators.py
```diff
@@ -0,0 +1,92 @@
...
+
+class TimeoutException(Exception):
+    """
+    A process has taken too long to execute
+    """
+
+
+class Repeat(object):
+    """
+    A decorator to use for repeating a tagged function.
```
Naburimannu: This is the sort of might-as-well-be-code that I don't like to put in decorators, but I don't have a well-formed objection. Can you give an example of how it'd be useful? (github code search isn't finding it for me, if there is one already...)

I've skimmed PEP318, and the examples I see there are the more meta things I'd expected - adding typing, metadata, and perhaps patterns. Aspect-orientation also makes sense.

tests/utils.py
dcolligan: It didn't for me.  And even if it did, this class provides a convenient programming model that I don't think `unittest` offers.

ga4gh/decorators.py
dcolligan: Right now its only usage is in the test cases:
```
$ git grep Repeat
...
tests/unit/test_decorators.py:    @decorators.Repeat(sleepSeconds)
...
```
But the application I had in mind was waiting for a server to startup.  I have this in another branch:
```
    @utils.Timeout()
    @utils.Repeat()
    def _waitForServerStartup(self):
        self.server.poll()
        if self.server.returncode is not None:
            self._waitForServerErrLines()
            message = "Server process unexpectedly died; stderr: {0}"
            failMessage = message.format(self.getServerErrLines())
            self.fail(failMessage)
        return not self._isServerRunning()
```

jeromekelleher: Ah, I see, the end-to-end test-case. I've been meaning to think about this one a bit more, but perhaps we could avoid this trickiness by keeping things in the same process space? Is there some way we could wire the client and the server together without going over a socket explicitly? (Sorry, this should be on the other thread -- feel free to answer this question over there if you prefer)

dcolligan: That'd be a neat trick; I wouldn't be sure how to do it.  Even if we could, that might require a lot of hackery, and might not test all of the functionality we want by virtue of hacking around stuff.  What do you have against sockets?

jeromekelleher: It can be quite tricky in my experience have two processes and waiting for sockets to become ready. But you're right, it might be more hacking that it's worth to do it in a single process.

dcolligan: Not if you've got the right annotations ;)

In any event, I have some branches depending on this one, so I'd like to get it merged.

tests/utils.py
Naburimannu: @dcolligan Can you tell us how it didn't work for you?

dcolligan:
@Naburimannu to answer your comment that strangely got deleted:

"@dcolligan Can you tell us how it [`unittest.assertRaises`] didn't work for you?"

Yes, I was using it wrong.  I pushed an update deleting that comment.

dcolligan:
@Naburimannu @jeromekelleher is there anything preventing you guys from +1'ing this?  I'd like to get it in...

jeromekelleher:
I'm -1 on this until I see a pressing reason for needing to share these across modules. Just define them where they are used.

dcolligan:
Would you accept `Timeout` and `Repeat` in `tests/utils.py`, since they will be used in testing?

Also, I think both of those decorators should really be tested, so I want to include the tests somewhere.  I'm fine with not testing `Timed` and moving it to `scripts/utils.py` (where it already is).

jeromekelleher:
Sure, that sounds sensible.

Naburimannu:
My feedback is that I prefer to see Repeat as a code fragment, and
decorators used for metadata (which Timeout is marginally; Timed is
*really* a border case for me), but could be convinced otherwise by strong
argumentation or examples of the use of this style leading to readable and
easily-understood code. When I can't predict the behavior of a function at
its callsite without reading the definition, I'm unhappy, and as I said
previously my skimming of the relevant PEP seems to support that.

So I'm -0 on any of these as decorators, but won't stand in the way.

Tom

dcolligan:
Pushed a new commit that moves the newly defined decorators into `tests/utils.py`

jeromekelleher:
Can we shelve this until we have some immediate use cases? The discussion is too abstract otherwise: we should evaluate these decorators on whether they make solving problems that we are facing easier. My feeling is that they probably will come in handy when we're doing the end-to-end tests, but I don't want to make a judgement until we really dig into that problem.

dcolligan:
Alright.

Issue #186, opened 2015-02-25
Server should determine variant file handler type automatically


jeromekelleher:
At the moment, the server CLI takes an argument telling it what type of VariantSet to allocate. This is a poor design choice, and makes it difficult to try out new data. A simpler option would be to dynamically decide what type of VariantSet to allocate at run time. One possible outline would be

- Remove the variantSetClass from the `backend.Backend()` constructor, and call `variants.variantSetFactory(variantSetId, relativePath)` to construct the variantSet.
- Update the CLI to remove the redundant positional arg.
- In `variantSetFactory` we do a simple name based check: if the directory ends with ".wt" we allocate a `WormtableVariantSet` if not allocate a `TabixVariantSet`. (This will require us to update our example data.)

pgrosu:
+1, I'm always in favor of having the code adapt to the data :)

dcolligan:
@jeromekelleher can you update the directory structure in the example data to have this '.wt' convention?  I'm not exactly sure which level of the directory hierarchy you want it at.

jeromekelleher:
@dcolligan - done. I've also added the same data in tabix indexed VCF form, which should make it easy to compare the two approaches.

dcolligan:
Fixed in #204

PR #193, opened 2015-02-26
add fake backend

dcolligan:
The idea here is to be able to launch a data-free server with the command
```
ga4gh_server fake
```
In this way we can implement a future end-to-end test without needing to have any data on the file system.


ga4gh/backend.py
```diff
@@ -19,37 +20,46 @@ class Backend(object):
...
-        # Reads
-        self._readGroupSetIdMap = {}
-        readGroupSetDir = os.path.join(self._dataDir, "reads")
-        for readGroupSetId in os.listdir(readGroupSetDir):
-            relativePath = os.path.join(readGroupSetDir, readGroupSetId)
-            if os.path.isdir(relativePath):
-                readGroupSet = reads.ReadGroupSet(readGroupSetId, relativePath)
-                self._readGroupSetIdMap[readGroupSetId] = readGroupSet
-        self._readGroupSetIds = sorted(self._readGroupSetIdMap.keys())
+        if variantSetClass == variants.FakeVariantSet:
```
macieksmuga: Why is this done here, as opposed to using the `MockBackend(Backend)` class defined at the bottom of backend.py?

dcolligan: Because this PR is a POC.

I suppose the better thing to do would be to put Backend and MockBackend into some kind of class relationship and then instantiate the right one in the cli somehow.

I'm very open to feedback about the best way to do this.

macieksmuga: No worries, I'm just testing my understanding of the code by reading through some of the recent pull requests :)  I'll think a bit further about this, and try putting in a more constructive comment to this one next.

dcolligan:
@jeromekelleher @Naburimannu can I get some feedback on this?  end-to-end testing depends on having some way to launch a data-free server.

jeromekelleher:
I think we need to tackle these issues properly. It's no good making a no-data backend, as it won't have any data to serve, and won't be a very good test. We need a simulated backend, similar to what we started with back in the mists of time.  We should make a superclass `Backend` which does most of the work, and has `FileSystemBackend` and `SimulatedBackend` in which the specifics of instantiating the VariantSets, etc is taken care of. We should resolve #186 first, as the current backend interface really isn't very sensible.

So, I would recommend closing this PR and starting an issue for a SimulatedBackend.

Naburimannu:
And I think the right approach is DI, not subclassing (even though I often
don't like DI), but having not made time to turn my DI prototype into
something solid wouldn't object to a working subclassing implementation.

PR #194, opened 2015-02-26
Add pysam htslib VCF/BCF backend

bioinformed:


dcolligan:
This could use some tests... check out `tests/unit/test_backends.py` for an example of how we're doing it with wormtable currently.

bioinformed:
@dcolligan: I agree-- this new functionality does need proper testing.  However, the Wormtable tests are not abstract or easily convertible to a different backend.  In an ideal world, I'd have time to prepare proper unit tests and remedy the situation.  Unfortunately, I am focused on building the mutable VCF/BCF API in pysam and a variant comparison and benchmarking engine (in addition to my day job).  I'm sorry to leave work for others, but my upfront commitment was only to provide a backend implementation as a proof of principle due to lack of API documentation.

ga4gh/datamodel/variants.py
```diff
@@ -342,3 +343,118 @@ def getMetadata(self):
...
+    if isinstance(value, (list, tuple)):
+        return [str(v) for v in value]
+    else:
+        return [str(value)]
+
+
+_nothing = object()
+
+
+def is_empty_iter(it):
```
jeromekelleher: camelCase please.

ga4gh/datamodel/variants.py
```diff
@@ -342,3 +343,118 @@ def getMetadata(self):
...
+    """Return True iff the iterator is empty or exhausted"""
+    return next(it, _nothing) is _nothing
```
jeromekelleher: Why not ``return next(it, None) is None``?

ga4gh/datamodel/variants.py
```diff
@@ -342,3 +343,118 @@ def getMetadata(self):
...
+        self._created = protocol.convertDatetime(datetime.datetime.now())
+        self._chromFileMap = {}
+
+        for pattern in ['*.bcf', '*.vcf.gz']:
+            for filename in glob.glob(os.path.join(vcfPath, pattern)):
+                self._addFile(filename)
+
+    def _addFile(self, filename):
+        varFile = pysam.VariantFile(filename)
+        if not varFile.index:
```
jeromekelleher: We're using explicit comparison with None elsewhere for this sort of thing, i.e.
```python
if varFile.index is None:
    # do stuff
```

ga4gh/datamodel/variants.py
```diff
@@ -342,3 +343,118 @@ def getMetadata(self):
...
+            raise Exception("VCF/BCF files must be indexed")
+        for chrom in varFile.index:
+            # Unlike Tabix indices, CSI indices include all contigs defined
+            # in the BCF header.  Thus we must test each one to see if
+            # records exist or else they are likely to trigger spurious
+            # overlapping errors.
+            if is_empty_iter(varFile.fetch(chrom)):
```
jeromekelleher: The continue seems unnecessary here. Why not:
```
if not isEmpyIter(varFile.fetch(chrom)):
    if  # etc
```

ga4gh/datamodel/variants.py
```diff
@@ -342,3 +343,118 @@ def getMetadata(self):
...
+                continue
+            if chrom in self._chromFileMap:
+                raise Exception("cannot have overlapping VCF/BCF files.")
+            self._chromFileMap[chrom] = varFile
+
+    def convertVariant(self, record):
+        variant = protocol.GAVariant()
```
jeromekelleher: I don't think we need all the blank lines in this method --- these are all very simple statements, and don't need to be separated.

ga4gh/datamodel/variants.py
```diff
@@ -342,3 +343,118 @@ def getMetadata(self):
         # TODO: Implement this
         ret = []
         return ret
+
+
+def _encode_value(value):
```
jeromekelleher: camelCase this please.

ga4gh/datamodel/variants.py
```diff
@@ -342,3 +343,118 @@ def getMetadata(self):
...
+        variant.names = record.id.split(';') if record.id else None
+        variant.start = record.start          # 0-based inclusive
+        variant.end = record.stop             # 0-based exclusive
+        variant.referenceBases = record.ref
+        variant.alternateBases = list(record.alts)
+
+        # record.filter and record.qual are also available, when supported
+        # by GAVariant.
+
+        variant.info = {}
```
jeromekelleher: This should be set to the empty dict by default I think, so you can leave this statement out. Similarly for `variant.calls` below. (These extra list and dict allocations will add up when we're serving millions of variants!)

ga4gh/datamodel/variants.py
```diff
@@ -342,3 +343,118 @@ def getMetadata(self):
...
+
+        variant.calls = []
+        for name, call in record.samples.iteritems():
+            c = protocol.GACall()
+            c.genotype = _encode_value(call.allele_indices)
+
+            for key, value in record.info.iteritems():
+                if key == 'GT':
+                    c.genotype = list(call.allele_indices)
+                elif key == 'GL' and value:
```
jeromekelleher: Can you rewrite this without the implicit boolean conversion please? (i.e., `and value is not None`)

jeromekelleher:
This is great, thanks @bioinformed! I've got some minor style points to bring the code into line with the rest of the codebase, but otherwise it's perfect. I agree with you about the testing situation: we don't have any tests for the existing Tabix variant set, and this is something we will need to work on later.

+1 after the style issues are fixed and the commits squashed.

awz:
+1 - look forward to the merge +/- style updates & squash.

ga4gh/datamodel/variants.py
```diff
@@ -342,3 +343,118 @@ def getMetadata(self):
...
+                                          record.contig,
+                                          record.pos)
+
+        # TODO How should we populate these from VCF?
+        variant.created = self._created
+        variant.updated = self._created
+
+        variant.variantSetId = self._variantSetId
+        variant.referenceName = record.contig
+        variant.names = record.id.split(';') if record.id else None
```
jeromekelleher: Another implicit boolean conversion here: ```record.id.split(';') if record.id is not None else None```

jeromekelleher:
Would you mind making these minor mods please @bioinformed? We are very keen to merge this. (I thought I made a comment here a couple of days ago pinging you, but it must have gotten lost somewhere.)

bioinformed:
I'm *not* a fan of many of the project's stylistic decisions, but these are
easy enough.  I'll update the PR shortly.

ga4gh/datamodel/variants.py
bioinformed: The function is so general that it is likely to be copied into contexts where iteration may include None as a legitimate value.  Thus I create a unique object instance (_nothing) to use a sentinel.

jeromekelleher:
Thanks @bioinformed. Sorry to ask you to make them, but consistency is very important. We were sort of stuck with the camelCasing by the protocol, which I don't like at all (but I'd like a mixed camelCase and snake_case project even less). I would argue strongly about using non-booleans in conditionals though: `if x is not None` is a lot clearer than `if x` IMO.

ga4gh/datamodel/variants.py
bioinformed: I'll make these changes to match your preferred style.  I've been a Python programmer for over 15 years (since 1999 using Python 1.5) and have probably written millions of lines of Python code.  While we all have our personal quirks and preferences, I do find some of the GA4GH server style requirements a bit baffling and contrary to generally accepted styles like PEP-8.  That said, it is your party and therefore your rules.

ga4gh/datamodel/variants.py
bioinformed: I value breaks in my vertical space more than you?  Will change.

bioinformed:
Style changes are in.  Squashing commits destroyed the diffs in this pull request, but seems to have done the right thing overall.

ga4gh/datamodel/variants.py
jeromekelleher: We're following PEP8 as much as reasonably possible @bioinformed. The only real difference between our styles I think is the comparisons with None, which PEP8 has specific recommendations on: "Also, beware of writing if x when you really mean if x is not None -- e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!"

jeromekelleher:
Excellent, thanks!

+1

Issue #197, opened 2015-02-27
Calculate local test coverage numbers correctly


dcolligan:
A run of the coverage tool with `nosetests` will reuse cached coverage numbers, leading to discrepancies between what Travis reports and what `run_tests.py` reports.  Change this.

dcolligan:
Fixed in #198

PR #198, opened 2015-02-27
Delete cached coverage before computing coverage

dcolligan:
This won't change how coverage is computed on Travis (presumably because there aren't any cached coverage files lying around in a fresh install), but it will change local numbers to be accurate when using `run_tests.py`

Issue #197


jeromekelleher:
+1

Naburimannu:
+1

PR #199, opened 2015-03-02
modify process_schemas script to work on locally stored Avro schemas.

macieksmuga:
To allow for development of experimental server code using ga4gh schemas still in development (not tagged with an official version number), I modified the process_schemas.py script to allow generating the _protocol_definitions.py file from a local directory (for example, a local clone of a development fork of ga4gh/schemas). 

The script runs as before when not provided with an optional --output-file (-o) argument.

### DIFF
--- a/scripts/process_schemas.py
+++ b/scripts/process_schemas.py
@@ -282,6 +282,8 @@ class SchemaProcessor(object):
         string = "schemas-{0}".format(self.version[1:])
         self.schemaDir = os.path.join(self.tmpDir, string)
         self.avroJar = os.path.join(self.schemaDir, "avro-tools.jar")
+        self.sourceDir = args.inputSchemasDirectory
+        self.avroPath = "src/main/resources/avro"
 
     def cleanup(self):
         if self.verbosity > 1:
@@ -311,19 +313,44 @@ class SchemaProcessor(object):
             with open(os.devnull, 'w') as devnull:
                 subprocess.check_call(args, stdout=devnull, stderr=devnull)
 
-    def run(self):
-        url = "https://github.com/ga4gh/schemas/archive/{0}.tar.gz".format(
-            self.version)
-        self.download(url, self.sourceTar)
-        with tarfile.open(self.sourceTar, "r") as tarball:
-            tarball.extractall(self.tmpDir)
-        directory = os.path.join(self.schemaDir, "src/main/resources/avro")
+    def setupAvroJar(self):
         if self.avroJarPath is not None:
             self.avroJar = os.path.abspath(self.avroJarPath)
         else:
             url = "http://www.carfab.com/apachesoftware/avro/stable/java/"\
                 "avro-tools-1.7.7.jar"
             self.download(url, self.avroJar)
+
+    def getSchemaFromGitHub(self):
+        """
+        Downloads a tagged version of the schemas
+        from the official GitHub repo.
+        """
+        url = "https://github.com/ga4gh/schemas/archive/{0}.tar.gz".format(
+            self.version)
+        self.download(url, self.sourceTar)
+        with tarfile.open(self.sourceTar, "r") as tarball:
+            tarball.extractall(self.tmpDir)
+
+    def getSchemaFromLocal(self):
+        """
+        Copies schemas from a local directory.
+        """
+        destDir = os.path.join(self.schemaDir, self.avroPath)
+        if not os.path.exists(destDir):
+            os.makedirs(destDir)
+        avdlFiles = glob.iglob(os.path.join(self.sourceDir, "*.avdl"))
+        for avdlFile in avdlFiles:
+            if os.path.isfile(avdlFile):
+                shutil.copy2(avdlFile, destDir)
+
+    def run(self):
+        if self.sourceDir is None:
+            self.getSchemaFromGitHub()
+        else:
+            self.getSchemaFromLocal()
+        directory = os.path.join(self.schemaDir, self.avroPath)
+        self.setupAvroJar()
         cwd = os.getcwd()
         os.chdir(directory)
         for avdlFile in glob.glob("*.avdl"):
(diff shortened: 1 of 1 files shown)


scripts/process_schemas.py
```diff
@@ -271,7 +271,7 @@ class SchemaProcessor(object):
     """
     def __init__(self, args):
         self.version = args.version
-        self.destinationFile = args.outputFile
+        self.destinationFile = args.output_file
```
Naburimannu: I know underscore is PEP8 / Pythonic, but because the schema is camel-cased this project is also camel-cased.

scripts/process_schemas.py
```diff
@@ -280,6 +280,9 @@ def __init__(self, args):
         string = "schemas-{0}".format(self.version[1:])
         self.schemaDir = os.path.join(self.tmpDir, string)
         self.avroJar = os.path.join(self.schemaDir, "avro-tools.jar")
+        if args.input_schemas_directory is not None:
```
Naburimannu: Fix underscore -> casing.

scripts/process_schemas.py
```diff
@@ -343,20 +359,31 @@ def run(self):
...
+            self.runWithGitHub()
+
 
 def main():
     parser = argparse.ArgumentParser(
         description="Script to process GA4GH Avro schemas. Requires "
         "java external command")
     parser.add_argument(
-        "--outputFile", "-o", default="ga4gh/_protocol_definitions.py",
+        "--output-file", "-o", default="ga4gh/_protocol_definitions.py",
```
Naburimannu: Ugh - we're camel-cased internally, but I see some other command line flags are hyphenated.
Makes me want to reconsider the whole divergence-fromPEP8 thing.

jeromekelleher: There is inconsistency in some of the command line args all right @Naburimannu. Perhaps we should move to making all long-form arguments camelCase?

But yes, I very much support @Naburimannu's camelCasing comments here.

dcolligan: If we want to refactor cmd line flags to use camel casing, file an issue since that's going to be a sizable refactor, I predict.

jeromekelleher: Agreed. Let's remove the UI changes from here, and raise this as an issue to be discussed separately.

macieksmuga:
I followed the code generation convention as implemented by `argparse.ArgumentParser` which automatically converts `--command-line-option` to `parser.command_line_option` in compliance with PEP8. I agree we should move this discussion to a separate issue, and if I understand correctly, I need to camelCase the command line options to get this PR accepted.

macieksmuga:
I need to rebase the PR branch correctly, but I _think_ the camelCasing is now as was requested.

dcolligan:
+1

jeromekelleher:
I think we should rethink things a little bit here. The UI says that our first positional argument is the version number, and this is used in the output as  `_protocoldefinitions.version`. So, we never ignore the version number. Then, either we provide a input schema directory, or we don't. If we do, we use this to generate the schema and if we don't we download the schema from github at the tagged version.

These are minor differences, but it would be good to clarify the UI documentation on this.

scripts/process_schemas.py
```diff
@@ -334,6 +350,12 @@ def run(self):
         sg = SchemaGenerator(self.version, directory, self.destinationFile)
         sg.write()
 
+    def run(self):
+        if self.sourceTar is None:
+            self.runWithDir(self.schemaDir)  # local mode
```
jeromekelleher: This is confusing - we shouldn't pass an instance variable as a method parameter.

scripts/process_schemas.py
```diff
@@ -344,10 +366,15 @@ def main():
...
         "version",
-        help="The tagged git release to process, e.g., v0.5.1")
+        help="The tagged git release to process, e.g., v0.5.1. "
+        "Ignored if --inputSchemasDirectory is specified.")
     parser.add_argument(
         "--avro-tools-jar", "-j",
         help="The path to a local avro-tools.jar", default=None)
+    parser.add_argument(
+        "--inputSchemasDirectory", "-i",
+        help="Path to local schemas directory. If none, uses GitHub.",
```
jeromekelleher: We need more detail here. What directory are we assuming this points to?

scripts/process_schemas.py
```diff
@@ -282,6 +282,9 @@ def __init__(self, args):
         string = "schemas-{0}".format(self.version[1:])
         self.schemaDir = os.path.join(self.tmpDir, string)
         self.avroJar = os.path.join(self.schemaDir, "avro-tools.jar")
+        if args.inputSchemasDirectory is not None:
+            self.schemaDir = args.inputSchemasDirectory
```
jeromekelleher: I don't think we should work within the input directory tree, as this is potentially dangerous and can be confusing. I think we should leave self.schemaDir as it is, and instead copy the relevant files into the temporary directory. This way we're guaranteed to clean up properly after ourselves.

jeromekelleher:
This looks good @macieksmuga, I think we just need some minor changes and clarify the UI help a bit.

scripts/process_schemas.py
macieksmuga: Agree. I saw the problem with my initial approach – lots of generated intermediate files – as soon as I started playing with this code on my local clone of the schema repo.

macieksmuga:
@jeromekelleher would your concern about the two modes, github vs. local, be solved completely by better documentation, or do you think I should change around the control scheme – perhaps mandate either an -i option to feed in a local directory or -g to run with github-provided schemas? My preference is just to amend the documentation, and leave the control scheme as it is now: This script will likely only be run by somebody who's actively developing a server version against some experimental schemata (most of the time, "regular" users don't even need to notice this script, as its default output `_protocol_definitions.py` is already stored and downloaded directly from the repo).

scripts/process_schemas.py
macieksmuga: @jeromekelleher That's a good question. Currently, it's requiring the equivalent of the top-level directory of the cloned `ga4gh/schemas` project - so it assumes that there's a `src/main/resources/avro` subdirectory structure inside it. But on second thought, that's kind of a weird setup, which I guess is why you asked in the first place: Why shouldn't I just have this be the actual directory that contains the `.avdl` files (so in the case of a local clone of the `ga4gh/schemas` project located at `$LOCAL_GA4GH_SCHEMAS`, the argument to pass would then be `-i $LOCAL_GA4GH_SCHEMAS/src/main/resources/avro`.

macieksmuga:
@jeromekelleher Take a quick look at the revised usage and command-line help. I believe it addresses your concerns, and makes for an overall cleaner/simpler tool.

scripts/process_schemas.py
```diff
@@ -311,19 +313,44 @@ def convertAvro(self, avdlFile):
...
+        avdlFiles = glob.iglob(os.path.join(self.sourceDir, "*.avdl"))
+        for avdlFile in avdlFiles:
+            if os.path.isfile(avdlFile):
+                shutil.copy2(avdlFile, destDir)
+
+    def run(self):
+        if self.sourceDir is None:
+            self.schemaFromGitHub()
+        else:
+            self.schemaFromLocal()
```
jeromekelleher: It would be a little easier to understand if these were called `getSchemaFromGitHub` and `getSchemaFromLocal`.

jeromekelleher:
Excellent, this is exactly what I was hoping for;  thanks @macieksmuga. 

+1 modulo the tiny niggle above.

macieksmuga:
@jeromekelleher renamed schemaFrom* to getSchemaFrom* per your suggestion. I believe the code should (modulo Travis build) be good to go.

adamnovak:
+1, we definitely need this to do development targeting our moving-target schemas.

PR #200, opened 2015-03-02
Minimum patching up to make tabix backend work.

jeromekelleher:
The TabixVariantSet is currently broken (not surprisingly since we have no test coverage...). This PR is the minimum required to make it operational.

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -56,7 +56,7 @@ class Backend(object):
         self._callSetIdMap = {}
         for variantSet in self._variantSetIdMap.values():
             sampleNames = variantSet.getSampleNames()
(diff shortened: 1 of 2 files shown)


dcolligan:
How about adding a test to ensure we don't regress?

jeromekelleher:
It would be nice @dcolligan, but it's not trivial (otherwise it would already be done) and I don't have time to do it now. The Tabix backend is currently broken (so people can't use it) and this fixes it back to where it was.

jeromekelleher:
@dcolligan, any chance of a +1 and merge here? I've got a proposal for fixing testing on the TabixBackend in #208, which we'll need to discuss for a while. Can we merge this fix the breakage for the moment?

dcolligan:
+1 merging

PR #201, opened 2015-03-02
Add ga2sam tool

dcolligan:
Issue #83


ga4gh/cli.py
```diff
@@ -181,6 +127,109 @@ def createGAListReferenceBasesRequest(self):
...
+def ga2vcf_run(args):
+    # instantiate params
+    searchVariantsRequest = RequestFactory(
+        args).createGASearchVariantsRequest()
+    searchVariantSetsRequest = RequestFactory(
+        args).createGASearchVariantSetsRequest()
+    if args.outputFile is None:
+        outputStream = sys.stdout
+    else:
+        outputStream = file(args.outputFile, 'w')
```
jeromekelleher: Use `open` here rather than file().

ga4gh/converters.py
```diff
@@ -6,35 +6,236 @@
...
-        raise NotImplementedError()
+        header = self._getHeader()
+        targetIds = self._getTargetIds(header)
+        # pysam can't write to file streams (except for stdout)
+        # http://pysam.readthedocs.org/en/latest/usage.html#using-streams
+        if self._binaryOutput:
+            flags = "wb"
+        else:
+            flags = "wh"  # h for header
+        if self._outputFile:
```
jeromekelleher: `if self._outputFile is not None`

ga4gh/converters.py
```diff
@@ -6,35 +6,236 @@
...
+            fileString = self._outputFile
+        else:
+            fileString = "-"
```
jeromekelleher: It's purely a personal style thing, but I like the following pattern for these situations:
```python
fileString = "-"
if self._outputFile is not None:
     fileString = self._outputFile
```
This saves us a line of code. We could also have here
```python
fileString = "-" if self._outputFile is None else self._outputFile
```
which is a bit more concise too. It's a matter of taste and not something I feel strongly about, I just thought I'd bring it up.

ga4gh/converters.py
```diff
@@ -6,35 +6,236 @@
...
+        alignmentFile = pysam.AlignmentFile(
+            fileString, flags, header=header.toDict())
+        iterator = self._httpClient.searchReads(self._searchReadsRequest)
+        for read in iterator:
+            alignedSegment = SamLine.toAlignedSegment(read, targetIds)
+            alignmentFile.write(alignedSegment)
+        alignmentFile.close()
+
+    def _getHeader(self):
+        header = SamHeader()
```
jeromekelleher: I don't see much value in a `SamHeader` class here right now. Just make the dict directly here and return it. (If we need something more sophisticated in future then we should of course have a class)

ga4gh/converters.py
```diff
@@ -6,35 +6,236 @@
...
+
+    def __init__(self):
+        raise SamException("SamLine can't be instantiated")
+
+    @classmethod
+    def toAlignedSegment(cls, read, targetIds):
+        alignedSegment = pysam.AlignedSegment()
+        # QNAME
+        alignedSegment.query_name = \
+            read.fragmentName.encode('utf8')
```
jeromekelleher: Make a class variable `_encoding` and set this to 'utf8'.  I find these line breaks really ugly and difficult to read also. Can we use a short variable name here? What about `ret`? I.e.
```
ret = pysam.AlignedSegment()
ret.query_name = read.fragmentName.encode(self._encoding)
#etc
```
Since the method name is `toAlignedSegment` this should be abundantly clear?

ga4gh/cli.py
```diff
@@ -181,6 +127,109 @@ def createGAListReferenceBasesRequest(self):
...
+
+def ga2sam_main(parser=None):
+    # parse args
+    if parser is None:
+        parser = argparse.ArgumentParser(
+            description="GA4GH SAM conversion tool")
+    addClientGlobalOptions(parser)
+    addReadsSearchParserArguments(parser)
+    addBinaryOutputArgument(parser)
+    addOutputFileArgument(parser)
```
jeromekelleher: Not sure if it's defined here, but add a short option for --outputFile also. We should support both '-o file' and '--outputfile file'

ga4gh/cli.py
```diff
@@ -181,6 +127,109 @@ def createGAListReferenceBasesRequest(self):
...
+
+    # do conversion
+    samConverter = converters.SamConverter(
+        httpClient, searchReadsRequest, args.outputFile, args.binaryOutput)
+    samConverter.convert()
+
+
+def addBinaryOutputArgument(parser):
+    parser.add_argument(
+        "--binaryOutput", default=False,
```
jeromekelleher: Add a short option, '-b' here.

tests/unit/test_converters.py
```diff
@@ -0,0 +1,90 @@
...
+        httpClient = self.getHttpClient()
+        searchReadsRequest = self.getSearchReadsRequest()
+        binaryOutput = False
+        samConverter = converters.SamConverter(
+            httpClient, searchReadsRequest, filePath, binaryOutput)
+        samConverter.convert()
+
+        # read SAM file
+        samfile = pysam.AlignmentFile(filePath, "r")
+        reads = [_ for _ in samfile.fetch()]
```
jeromekelleher: Can we use `list(samfile.fetch())` here?

tests/unit/test_converters.py
```diff
@@ -0,0 +1,90 @@
...
+        self.assertEqual(reads[0].query_name, "SRR622461.77861202")
+        samfile.close()
+
+        # cleanup
+        os.remove(filePath)
+
+    def testPlainText(self):
+        self._testRoundTrip(False, 'sam.sam')
```
jeromekelleher: We should use a tempfile here, as we shouldn't ever write temporary files to CWD. Something like
```
with tempfile.NamedTemporaryFile() as fileHandle:
    self._testRoundTrip(False, fileHandle.name)
```

tests/unit/test_converters.py
```diff
@@ -0,0 +1,90 @@
...
+        searchReadsRequest = self.getSearchReadsRequest()
+        binaryOutput = False
+        samConverter = converters.SamConverter(
+            httpClient, searchReadsRequest, filePath, binaryOutput)
+        samConverter.convert()
+
+        # read SAM file
+        samfile = pysam.AlignmentFile(filePath, "r")
+        reads = [_ for _ in samfile.fetch()]
+        self.assertEqual(reads[0].query_name, "SRR622461.77861202")
```
jeromekelleher: Can we get a TODO for more indepth testing here please?

tests/unit/test_converters.py
```diff
@@ -0,0 +1,90 @@
...
+        samfile.close()
+
+        # cleanup
+        os.remove(filePath)
```
jeromekelleher: This isn't needed when we use the `with` construct below.

jeromekelleher:
This is superb, thanks @dcolligan! I have some minor style comments, but on the whole I like it a lot. I like the testing strategy: it's pragmatic, and it gets us moving. I can't comment on the accuracy of the SAM conversion, as I don't know very much about it. I'd be in favour of merging this PR, and asking some experts to review it afterwards and make some issues for any problems they find.

We are held back here by the lack of appropriate GET requests for `VariantSet`s, `ReadGroupSets` and `ReadGroup`s; we should make issues/open a PR to solve the problem on the schemas repo.

dcolligan:
@jeromekelleher pushed a commit to address your concerns

There seem to be no GET requests defined in the schema repo's HEAD for `/variantsets/{id}` nor `/readgroups/{id}`, but there is one for `/readgroupsets/{id}` in `readmethods.avdl`

dcolligan:
Filed this issue in the schemas repo: https://github.com/ga4gh/schemas/issues/245

tests/unit/test_cli.py
```diff
@@ -5,29 +5,84 @@
...
 
-import unittest
 import argparse
+import unittest
+import mock
 
 import ga4gh.cli as cli
 import ga4gh.protocol as protocol
 import ga4gh.client as client
+import ga4gh.converters as converters  # NOQA
```
jeromekelleher: Why NOQA? Shouldn't we just delete this line if it's not used?

tests/unit/test_converters.py
```diff
@@ -0,0 +1,90 @@
...
+class TestSamConverterRoundTrip(TestSamConverter):
+    """
+    Write a sam file and see if pysam can read it
+    """
+    def _testRoundTrip(self, binaryOutput):
+        with tempfile.NamedTemporaryFile() as fileHandle:
+            # write SAM file
+            httpClient = self.getHttpClient()
+            searchReadsRequest = self.getSearchReadsRequest()
+            binaryOutput = False
```
jeromekelleher: Why are you setting  binaryOutput = False?

ga4gh/cli.py
```diff
@@ -181,6 +127,109 @@ def createGAListReferenceBasesRequest(self):
...
+def ga2vcf_main(parser=None):
+    # parse args
+    if parser is None:
+        parser = argparse.ArgumentParser(
+            description="GA4GH VCF conversion tool")
+    addClientGlobalOptions(parser)
+    addOutputFileArgument(parser)
+    subparsers = parser.add_subparsers(title='subcommands',)
+    variantsSearchParser = addVariantsSearchParser(subparsers)
+    # TODO kinda ugly that we need to overload the parser with arguments
```
jeromekelleher: We can note here that this is a temporary workaround until we have `getVariantSet` in the protocol.

jeromekelleher:
Looks great @dcolligan, thanks. 

+1 modulo  minor comments above.

dcolligan:
@jeromekelleher pushed another commit fixing those problems

PR #204, opened 2015-03-03
Infer variant type instead of passing it via cli

dcolligan:
Issue #186

### DIFF
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -49,6 +49,13 @@ def convertVCFGenotype(vcfGenotype, vcfPhaseset):
     return genotype, phaseset
 
 
+def variantSetFactory(variantSetId, relativePath):
+    if variantSetId.endswith(".wt"):
+        return WormtableVariantSet(variantSetId, relativePath)
+    else:
+        return TabixVariantSet(variantSetId, relativePath)
+
+
 class VariantSet(object):
     """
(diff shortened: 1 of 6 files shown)


ga4gh/datamodel/variants.py
```diff
@@ -48,6 +48,13 @@ def convertVCFGenotype(vcfGenotype, vcfPhaseset):
     return genotype, phaseset
 
 
+def variantSetFactory(variantSetId, relativePath):
+    if True:  # TODO look for '.wt' in some part of directory path
```
dcolligan: @jeromekelleher can you provide some clarity here?  (see question on issue page)

jeromekelleher: Just use
```python
if variantSetId.endswith(".wt"): 
```
We're ripping out wormtable, so it doesn't need to be pretty.

dcolligan:
@jeromekelleher pushed update

jeromekelleher:
Nice, +1.

As an aside, the extra stuff you had to do here with the expected IDs makes me even more certain that hard coding strings in test cases is a bad idea...

jeromekelleher:
Oops, I forgot - we need to update the README also, as this will now tell people the wrong thing. Can you add this here please @dcolligan?

dcolligan:
@jeromekelleher pushed a new change with the README updated

Issue #175, opened 2015-02-24
Configuration options for server request/response validation required


jeromekelleher:
The `ProtocolElement.validate()` method allows us to check whether a given JSON dictionary is a valid instance of the protocol. We should have options to enable this checking for all incoming requests and all outgoing responses. We need both because we want them for different things: validation of input will be useful in productions settings as a security measure, and validation of output will be useful for development and testing.

One way to do this would be:

Give the `backend` object a reference to the Flask `config` object. We should add a method to backend called `setConfig` which can be called (for now) in `cli.server_main`, e.g. 
```python
backend = backend.Backend(...)
frontend.app.backend = backend
backend.setConfig(frontend.app.config)
```
This does create coupling between the frontend and the backend, but hopefully by just using the config object this isn't too bad.

It is then straightforward to update `Backend.runSearchRequest()` to optionally validate the request and response.

dcolligan:
I'm fine with this change, except for the coupling between frontend and backend.

It's probably a good idea to break out the flask config into its own module that can be then both imported by frontend and backend.

dcolligan:
Fixed in #205

PR #205, opened 2015-03-03
Add server request/response validation

dcolligan:
Issue #175

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -6,11 +6,13 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import json
 import os
 
 import ga4gh.protocol as protocol
 import ga4gh.datamodel.references as references
 import ga4gh.datamodel.reads as reads
+import ga4gh.backend_exceptions as backendExceptions
 
 
 class Backend(object):
@@ -20,6 +22,8 @@ class Backend(object):
     """
     def __init__(self, dataDir, variantSetClass):
         self._dataDir = dataDir
+        self._requestValidation = False
+        self._responseValidation = False
         # TODO this code is very ugly and should be regarded as a temporary
         # stop-gap until we deal with iterating over the data tree properly.
         # Variants
@@ -99,8 +103,9 @@ class Backend(object):
         any point using the nextPageToken attribute of the request object.
         """
         self.startProfile()
-        # TODO change this to fromJsonDict and validate
-        request = requestClass.fromJsonString(requestStr)
+        requestDict = json.loads(requestStr)
+        self.validateRequest(requestDict, requestClass)
+        request = requestClass.fromJsonDict(requestDict)
         pageList = []
         nextPageToken = None
         for obj, nextPageToken in objectGenerator(request):
@@ -110,6 +115,8 @@ class Backend(object):
         response = responseClass()
         response.nextPageToken = nextPageToken
         setattr(response, pageListName, pageList)
+        responseDict = response.toJsonDict()
+        self.validateResponse(responseDict, responseClass)
         self.endProfile()
         return response.toJsonString()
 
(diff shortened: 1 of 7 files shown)


ga4gh/backend.py
```diff
@@ -286,6 +292,18 @@ def startProfile(self):
     def endProfile(self):
         pass
 
+    def validateRequest(self, jsonDict, requestClass):
+        if ('REQUEST_VALIDATION' in flaskapp.app.config and
+                flaskapp.app.config['REQUEST_VALIDATION']):
+            if not requestClass.validate(jsonDict):
+                raise backendExceptions.RequestValidationFailureException
```
jeromekelleher: missing brackets here? ie, ` raise backendExceptions.RequestValidationFailureException()`

ga4gh/backend.py
```diff
@@ -286,6 +292,18 @@ def startProfile(self):
...
+
+    def validateResponse(self, jsonDict, responseClass):
+        if ('RESPONSE_VALIDATION' in flaskapp.app.config and
+                flaskapp.app.config['RESPONSE_VALIDATION']):
+            if not responseClass.validate(jsonDict):
+                raise backendExceptions.ResponseValidationFailureException
```
jeromekelleher: and here.

jeromekelleher:
I'm really not keen at all on the backend depending on flask. Is there any other way we can do this? We just need to do some configuration on the backend. Can't we just set some instance variable during configure? i.e.
```python
def configure(...):
    # All the stuff we already have...
    backend = backend.Backend(app.config["DATADIR"])
    backend.setRequestValidation(app.config["REQUEST_VALIDATION"])
    backend.setResponseValidation(app.config["REQUEST_VALIDATION"])
    app.backend = backend
```
This greatly simplifies things, and means we don't have to pass around an ugly config dictionary and do lots of testing on string constants. This will require replumbing things a bit, but solving #186 gets us most of the way there.

ga4gh/backend.py
dcolligan: Exceptions raised like this are implicitly instantiated.  Do we want to add parentheses as a matter of style?

Related: https://www.python.org/dev/peps/pep-0317/

jeromekelleher: Yeah, I think we should. It's just making an awkward special case out of no-parameter exceptions.

dcolligan:

ga4gh/backend.py
```diff
@@ -286,6 +293,22 @@ def startProfile(self):
     def endProfile(self):
         pass
 
+    def validateRequest(self, jsonDict, requestClass):
```
jeromekelleher: Document these methods please -- all public methods should have docstrings.

ga4gh/flaskapp.py
```diff
@@ -0,0 +1,73 @@
+"""
+Contains the actual flask startup code
+"""
```
jeromekelleher: I'm not really seeing the benefit of this new module. We still have lots of flasky stuff in frontend.py, so we're not breaking flask from everything else. I'd vote for keeping things are they are here.

dcolligan:
@jeromekelleher pushed a new change addressing your concerns

jeromekelleher:
Excellent, thanks @dcolligan. Why do we need to default to not validating requests are responses for testing? If we're failing validation at the moment, please open an issue on this once this PR has been merged.

+1.

dcolligan:
Because in some cases tests send data to the server that won't validate like `self.sendVariantsSearch('{"variantSetIds": [1, 2]}')` in `test_views.py`.  It's only for a small subset of tests that we default to non-validation.

Yes, I have verified that request-responses work (see, this is why we need an end to end test!).

dcolligan:
(And the fact that only the tests with malformed data failed testifies to this change working.)

jeromekelleher:
Thanks @dcolligan, just checking! (And yes, we _do_ need end-to-end tests, I completely agree).

+1.

Issue #173, opened 2015-02-23
Better test data required


jeromekelleher:
Currently our test data is limited to a handful of small and synthetic VCFs (see https://github.com/ga4gh/server/tree/master/tests/data). We need a wide spread of test data to exercise our APIs, across variants, reads and references. We should have a collection typical small data files spanning many anticipated use cases.

dcolligan:
This issue is too general.  Can you try and narrow the scope a bit?

jeromekelleher:
OK, fair enough. We need the following to close this issue:

1. Test reference data;
2. Test read data;
3. More test VCF data, covering different format versions, numbers of samples, ploidy levels, etc. How much more depends on how much we think it's worth covering exotic corner cases.

I'm not really informed enough to know what would be considered a representative spread of examples, so it would be good to get some wider opinions here. Anyone like to weigh in here?

pgrosu:
@jeromekelleher, I think you gave a great suggestion a while back in using the Illumina's Platinum Genomes dataset:

https://github.com/ga4gh/compliance/issues/13#issuecomment-59219513

~p

shajoezhu:
missing a test example such that has a variant site, where for the INFO and FORMAT field, NUMBER=R. 

For details, see http://samtools.github.io/hts-specs/VCFv4.2.pdf, section 1.2.2 and 1.2.4

shajoezhu:
NUMBER=R is new for vcf4.2, a related entry is found on http://sourceforge.net/p/samtools/mailman/message/32900784/

jeromekelleher:
Test data has been greatly improved since this issue.

PR #208, opened 2015-03-04
Framework for data driven tests.

jeromekelleher:
This PR adds the initial frameworks required for data driven tests, and some examples for variants and references. These tests work as follows:

1. We have a number of modules within the `tests/datadriven` module. Each module has a top level function that is a generator over the test cases that are run. This uses nose's ability to use generators that yield individual test cases (see http://nose.readthedocs.org/en/latest/writing_tests.html). Each test that we run is a effectively a pair (testMethod, dataDir).
2. In each module, we have one or more TestClasses, and we allocate an instance of this class for each data directory. The test class builds a model of the data in this directory and also allocates the corresponding object from `ga4gh.datamodel`. We then compare these two models, and ensure that they are consistent.
3. Each TestClass has a set of test methods, and each of these is called exactly once. Note that this does violate the idea that unit tests should be entirely independent of each other. However, with a little programming discipline I think we can make these tests effectively independent. The cost of building the alternative model of the data in question is not trivial, and it seems pointless to mandate that this is done before every single test.

Here is an example of what a run looks like:
```
$ nosetests -vs tests/datadriven/
tests.datadriven.test_references.ReferenceSetTest.testNumReferences:GRCh38 ... ok
tests.datadriven.ReferenceSetTest.testProtocolElementValid:GRCh38 ... ok
tests.datadriven.test_references.ReferenceSetTest.testNumReferences:GRCh38_copy ... ok
tests.datadriven.ReferenceSetTest.testProtocolElementValid:GRCh38_copy ... ok
tests.datadriven.test_references.ReferenceSetTest.testNumReferences:example_1 ... ok
tests.datadriven.ReferenceSetTest.testProtocolElementValid:example_1 ... ok
tests.datadriven.VariantSetTest.testProtocolElementValid:example_1 ... ok
tests.datadriven.test_variants.VariantSetTest.testSearchAllVariants:example_1 ... ok
tests.datadriven.VariantSetTest.testProtocolElementValid:example_2 ... ok
tests.datadriven.test_variants.VariantSetTest.testSearchAllVariants:example_2 ... ok
tests.datadriven.VariantSetTest.testProtocolElementValid:example_3 ... ok
tests.datadriven.test_variants.VariantSetTest.testSearchAllVariants:example_3 ... ok
tests.datadriven.VariantSetTest.testProtocolElementValid:example_4 ... ok
tests.datadriven.test_variants.VariantSetTest.testSearchAllVariants:example_4 ... ok

Ran 14 tests in 0.144s
```

Why not use standard unit tests? Unit tests are very valuable and we should continue to make lots of them. However, the principle thing that our server does is interact with data that it is provided with and serve this out to users. We will get data in all sorts of weird and wonderful formats with all sorts of quirks. We need to be able to deal with this, and to build in test cases to ensure we handle these quirks properly. When we find a data file that breaks our code, we add an example data file to our tests data, and the data driven tests automatically pick it up and run it.

The main disadvantage of this approach is that we can't extent `unittest.TestCase`, and so we lose our nice `self.assertX` methods. This is unfortunate, and is because of the way nose test generators work. We can possibly work around this with some nasty hackery, but I'm not sure if it's a good idea.

I'd hope to make this a fairly fundamental piece of infrastructure for how we move forward in terms of building out reads, variants and references support, so please review this PR and let me know what you think.


dcolligan:
A few things:
1) I agree that data-driven tests are needed for the project, generally.
2) I don't like numeric naming of just about anything, both because it's non-descriptive and it implicitly adds a sequencing constraint.  I would favor `example_whatever` names rather than `example_5` names.
3) So it looks like we're ok with adding binary files into the git repo now.  Does this mean we can add the latest ga4gh-example-data to the git repo?  That would be one way to unblock end-to-end testing.
4) 1 test per data file does seem pointless, but I agree it sucks to throw out unittest
5) The ability to add in a new data file and have everything magically work is a nice feature.
6) Maybe a README in the root of tests/datadriven about how to run the tests would be good.
7) Will these tests get picked up in a global nosetests run?  Do we want them to be?

jeromekelleher:
Good, glad you're happy with this. Some responses:
> 2) I don't like numeric naming of just about anything, both because it's non-descriptive and it implicitly adds a sequencing constraint. I would favor example_whatever names rather than example_5 names.

Agreed. I don't like this either, we should fix it for #173.

> 3) So it looks like we're ok with adding binary files into the git repo now. Does this mean we can add the latest ga4gh-example-data to the git repo? That would be one way to unblock end-to-end testing.

It's too big. We should keep our repo as small as we can.

> 4)  1 test per data file does seem pointless, but I agree it sucks to throw out unittest

It's worse than that with unitest: we cannot multiply test-cases by data. We could work around the setup costs using class and module level fixtures if this was the only problem. unittest assumes that every test function is executed exactly once, and this just doesn't work with a data driven approach.

> 6) Maybe a README in the root of tests/datadriven about how to run the tests would be good.

Good idea.

> 7) Will these tests get picked up in a global nosetests run? Do we want them to be?

Yes, and yes. If they get expensive we can run them less frequently, but right now it looks like they'll be plenty fast enough to run every time.

diekhans:
Yes, data driven tests as essential for this project.

I often do the intellectual cop-out of numbering test examples
and then regret it.  I do try to have example files that drive
multiple tests.  I do this with unittest all the time, just
cheat on the setup/tear down (cache results).

I am a fan of only storing the ASCII version of data (e.g. SAM
instead of BAM) and converting on the fly when running the
tests.  This does put a burden of having more tools installed to
run the test.  A fall bac option is to check both the ASCII and
binary versions into the tree.

jeromekelleher:
Can you give more details on how you do this with unittest @diekhans? This would be good to know - maybe I'm over-engineering things here and there's a simpler solution.

Agreed on only storing the ASCII versions. We are storing the tabix indexes at the moment, but maybe we should start doing this on the fly. Pysam will do this for us quite nicely.

macieksmuga:
@diekhans I'm thinking a little bit ahead to how we may be storing graph data for testing v0.6.x of the API (which I believe is essential for the development effort). A SQLite based dataset could always be represented by its creation script (CREATE TABLEs followed by INSERTs), and an AVRO based one can be expressed in JSON, so I think we'd be able to adhere to an ASCII-only policy if that's what ends up being mandated.

diekhans:
I would argue for a more human-editable format than the INSERTS
and then a program to load the SQLite database.

pgrosu:
@diekhans, basically if I understand correctly, you are looking for complete transparency in what happens to the data as it gets processed at each step and then compare to what it should be - while leaving optimizations to the data-format for a later time, after these steps are successful.

diekhans:
My main point is that keeping the master source for the test
data in git in the ASCII form make it easier to maintain and
understand.  Also, as you point out, if the binary format
changes, then it just requires changing the script that builds
the binary file.

pgrosu:
+1 @diekhans, Thank you for the clarification and I am in perfect agreement.

jeromekelleher:
I don't think we should strictly mandate text-only data (we might, for example, want to put in an old BAM file to make sure we read it correctly), but it is definitely the preferable option. I agree with @macieksmuga too --- we'll certainly want to get graph data in here.

jeromekelleher:
I've had another quick look at trying to hack support for unittest.TestCase in with this framework, and it looks very nasty. The idea would be to redefine the DataDrivenTest class at run time so that it inherits from unittest.TestCase: we'd then have all the nice assertX methods. (We have to do this at run time, because any compile-time subclasses of unittest.TestCase are collected by the unittest test discovery methods, and this breaks the apply-each-test-to-each-data-file pattern.) This is possible, I think, but not worth it in my opinion. It would lead to very mysterious looking code, obscure errors, and would probably be highly Python 2 specific.

Any thoughts on this: are we OK with using plain asserts for data driven tests?

diekhans:
Dumb question: couldn't the data driven tests be the same thing
as the conformance suite?

jeromekelleher:
It's not quite the same thing in my view @diekhans, but it does depend on how you define conformance suite. I would think that a conformance suite must be something that we can run against an arbitrary server, and operates only through the protocol. These data driven tests are highly specific to our internal structures. What have you in mind to a conformance suite?

diekhans:
See, I told you it was a dumb question.  I didn't get the
running at a lower level than the API.

Still, if conformance test cases could be used for other
these purposes, it might save development times.

BTW, I would say a conformance suites can run against an
arbitrary server that has the conformance suite data set loaded.

Cheers
Mark

jeromekelleher:
> BTW, I would say a conformance suites can run against an
> arbitrary server that has the conformance suite data set loaded.
Agreed!

dcolligan:
+1

jeromekelleher:
I think enough time has elapsed - can you merge this please @dcolligan?

PR #209, opened 2015-03-04
Module import tests (cycles, layering)

dcolligan:
Not sure if this is more trouble than it's worth at a project of our size, but...

And not even sure that this is the best way to do this, even if we do want it...

This is a test that ensures that no circular imports are present in the project.

Feedback welcome.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,4 +12,5 @@ pep8
 pysam
 pyyaml
 requests
+snakefood
 wormtable
--- a/tests/unit/test_imports.py
+++ b/tests/unit/test_imports.py
@@ -1,0 +1,554 @@
+"""
+Tests that the project's module import graph conforms to certain policies
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import copy
+import itertools
+import logging
+import operator
+import os
+import pprint
+import unittest
+import sys
+
+from snakefood.util import iter_pyfiles, setup_logging, is_python
+from snakefood.find import find_dependencies
+from snakefood.find import ERROR_IMPORT, ERROR_SYMBOL, ERROR_UNUSED
+from snakefood.fallback.collections import defaultdict
+from snakefood.roots import find_roots, relfile
+
+from tests import utils
+
+
+graph = None
+
+
+def globalSetUp():
+    # the import graph should only be generated once;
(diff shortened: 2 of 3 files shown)


dcolligan:
Apparently the best off-the-shelf dependency tool for python is snakefood, but its documentation is kind of thin: http://furius.ca/snakefood/

Probably worth using if we want to pursue enforcing any kind of policy wrt imports.

jeromekelleher:
Doesn't flake8 detect these for us?

dcolligan:
Nope.  We already do have one circular reference in the code (protocol.py and _protocol_definitions.py import each other).  Granted, this one is entirely expected.  (The inspiration for this change came from making the flakapp.py change, which of course is not going to make it in to master.)  If we do write a tool, it should be able to take a list of cycles to ignore.

So, questions:
1) Do we want this change (or, rather, a more polished form of it)?
2) If so, what other rules do we want to enforce besides no (unapproved) import cycles?  Possibly some layering rules (e.g. backend can't import frontend, etc.)?

jeromekelleher:
It it's not a lot of work, this sounds like a useful addition. Anyone else have any opinions?

dcolligan:
Alright, just checked in an update.  It's a lot of code, but basically runs two tests:

1) Checks for cycles in the module import graph.  We do have one cycle (see `ImportGraphCycleChecker.cycleExclusions`) presently, but we only fail on unexpected cycles.  If an unexpected cycle is detected, this test will fail with (one gets this error if one comments out the cycle exclusions):
```
PolicyException: Circular import reference: ga4gh/frontend_exceptions.py --> ga4gh/protocol.py --> ga4gh/_protocol_definitions.py --> ga4gh/protocol.py
```
2) Checks for unapproved imports that violate layering policy.  Layers are defined in `ImportGraphLayerChecker.layers`.  The rules are that modules in a layer can not import from layers equal to (provided the module is in a different moduleGroup) or above their own.  Unapproved imports will get a error like this (I had to sabotage the ga4gh code to emit this):
```
PolicyException: module 'ga4gh/serverconfig.py' in moduleGroup 'config' has dependency on module 'ga4gh/datamodel/reads.py' in moduleGroup 'datamodel'
```

jeromekelleher:
This _is_ a lot of code, but it does perform a useful task. 

+1

PR #211, opened 2015-03-06
Htslib data driven tests

jeromekelleher:
Initial POC for data driven tests in the HtslibVariantSet.  Adds the `self.assertEqual` method to the data driven tests, which might solve our issues with not using unittest.

This PR depends on #208.

### DIFF
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -459,7 +459,8 @@ class HtslibVariantSet(VariantSet):
         variant.start = record.start          # 0-based inclusive
         variant.end = record.stop             # 0-based exclusive
         variant.referenceBases = record.ref
-        variant.alternateBases = list(record.alts)
+        if record.alts is not None:
+            variant.alternateBases = list(record.alts)
         # record.filter and record.qual are also available, when supported
         # by GAVariant.
         for key, value in record.info.iteritems():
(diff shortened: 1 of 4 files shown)


ga4gh/datamodel/variants.py
```diff
@@ -450,9 +450,10 @@ def convertVariant(self, record):
         if record.id is not None:
             variant.names = record.id.split(';')
         variant.start = record.start          # 0-based inclusive
-        variant.end = record.stop             # 0-based exclusive
+        variant.end = record.start + len(record.ref)
```
bioinformed: This logic is incorrect for records with an END INFO field that is not equal to len(record.ref).  Better to use record.stop, since it handles this case properly.

lh3: +1 as Kevin said.

jeromekelleher: Ah, that makes sense. I was following the GA4GH documentation ("This is typically calculated by start + referenceBases.length") and wondering why my tests were failing. Thanks, I'll fix this.

bioinformed: The GA4GH docs are technically correct, but the qualifier "typically" hides
important details.  :)

jeromekelleher: Indeed! Great to have an expert eye on the VCF code, thanks for pointing this out.

jeromekelleher:
This is ready for review. I'm going to open a number of issues on filling out the gaps that have been identified here, rather than try to make this complete.

dcolligan:
+1

Issue #202, opened 2015-03-02
Add SimulatedBackend for testing


dcolligan:
Following up on the discussion in #193 

Create some kind of inheritance hierarchy (or do the equivalent with dependency injection?) for the backend.  Select amongst the different subclasses to use during server startup.

Blocker for merging #184

@jeromekelleher can you flesh out a little more what you had in mind here?

jeromekelleher:
Here are some thoughts:

1. We should add a SimulatedVariantSet. This should have a superclass shared with the other VariantSets, and we should try to share as much code as possible with HtslibVariantSet. (Not so much TabixVariantSet and WormtableVariantSet, since these will both soon be removed.) This can be based on the old version: https://github.com/ga4gh/server/blob/0d6522669fc3ad635deb6f293508c5a3b41ed156/ga4gh/server.py#L537

2. One way that we might use this is to create a class hierarchy for Backends. We have a root class, AbstractBackend (that does as much as possible of our work), FileSystemBackend and SimulatedBackend. FileSystemBackend corresponds to the backend that we already have, and SimulatedBackend differs by not taking a datadir as a parameter, but instead some parameters for how many variant sets it should instantiate, random seeds, etc.

3. We then figure out how to bring all this together --- through the config file in Flask seems like the obvious approach. In configure() if we see `DATADIR` is set, we allocate a `FileSystemBackend`. Otherwise, we should have a bunch of simulator parameters, and we then allocate a `SimulatedBackend`. This should give us the flexibility we need to use this both for development and for testing.

Does this make sense?

dcolligan:
Fixed in #212

PR #212, opened 2015-03-06
202 simulated backend

dcolligan:
First attempt; based off of PR #204 ; haven't modified tests yet; want @jeromekelleher 's feedback.

Issue #202


jeromekelleher:
This looks good @dcolligan, thanks. I think we should try to push the simulator aspects into the `datamodel.variants` module, and have simulated backend do as little as possible. I would imagine something like the following:
```python
class SimulatedBackend(AbstractBackend):
    def __init__(self, randomSeed, numVariantSets=1, numCalls=1, variantDensity=0.5):
        self._randomSeed = randomSeed
        self._randomGenerator = random.Random(randomSeed)
        for i in range(numVariantSets):
             variantSetId = "simVs{}".format(i)
             seed = self._randomGenerator().randint(0, 2**32)
             variantSet = variants.SimulatedVariantSet(seed, numCalls, variantDensity)
             self._variantSets.append(variantSet)
             self._variantSetIdMap[variantSetId] = variantSet
```
Then, we just inherit all of the search behaviour from the superclass. The `SimulatedVariantSet` class will need to implement `getVariants` like the existing variant sets, but it should be possible to do this easily enough in terms of the functionality that you have here.

In terms of configuration, what you have is perfect right now. We'll want to something a bit more hierarchical as we add more and more simulator parameters, but we can think about that later.

dcolligan:
@jeromekelleher addressed your comments in newest commit

jeromekelleher:
Perfect, thanks. This will make our lives a _lot_ easier for testing.

+1 after you've merged #214 and rebased this.

dcolligan:
@jeromekelleher merge me!

PR #214, opened 2015-03-09
Fixed README examples and queries on example data.

jeromekelleher:
Fixed the README so that example queries run. Also changed the default configuration to turn off validation by default so that queries run against tabix indexed data run.

(Note, there is some diff-noise in the README introduced by trailing whitespace being removed.)


dcolligan:
I would +1, but it looks like there's some version conflict issue with pep8 on Travis.  It looks like we might have to specify an older version of flake in the requirements.txt file.

jeromekelleher:
Yep, spot on @dcolligan. I've just pushed a commit that should fix this.

PR #218, opened 2015-03-09
Remove datadir arg

jeromekelleher:
This PR removes the mandatory DATADIR argument from the server CLI, and moves all configuration into the flask config space. This is an essential step towards deployability.

For development, we add a default configuration that we invoke from the CLI that makes the default data source ga4gh-example-data. This should make developers lives easy. We can override this with a config file specified on the command line though.

This will be updated and rebased once #212 has been merged. Feedback welcomed!

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -292,6 +292,12 @@ class AbstractBackend(object):
         self._responseValidation = responseValidation
 
 
+class EmptyBackend(AbstractBackend):
+    """
+    A GA4GH backend that contains no data.
+    """
+
+
 class SimulatedBackend(AbstractBackend):
     """
     A GA4GH backend backed by no data; used mostly for testing
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -11,11 +11,9 @@ import time
 import argparse
 import sys
 
-import ga4gh.backend as backend
 import ga4gh.client as client
 import ga4gh.protocol as protocol
 import ga4gh.converters as converters
-import ga4gh.datamodel.variants as variants
 import ga4gh.frontend as frontend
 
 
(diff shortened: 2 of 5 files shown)


ga4gh/cli.py
```diff
@@ -243,26 +241,11 @@ def addGlobalOptions(parser):
         "--port", "-P", default=8000, type=int,
         help="The port to listen on")
     parser.add_argument(
-        "--config", "-C", default='DefaultConfig', type=str,
+        "--config", "-c", default='DevelopmentConfig', type=str,
```
dcolligan: why the changes here and below of the short args to lowercase?

dcolligan:
Looks good, although I will want this to land after #212 since rebasing on top of that change will alter this one significantly.

ga4gh/cli.py
jeromekelleher: I think we usually fill up the lowercase letters first when making options. I'd imagine -f is a bit more common for filename input than -F. I'm happy to change it back if you don't like it, I'm not that bothered one way or the other.

jeromekelleher:
Good, glad you like it. We definitely want to land #212 first; I thought this was the best way to explain where I imagined our input was going after we merged the simulators!

ga4gh/cli.py
dcolligan: No, this is fine.  Just wondered what the rationale was.

jeromekelleher:
OK, this has been updated with the simulated backend and is ready for review.

One problem is the README is now out of date. I don't think there's much point in patching it up quickly here though, as we need to rework it substantially to start removing the wormtable vs tabix distinction, and to start documenting the config file format. This should be done separately I think - I'll open an issue if this PR is merged.

ga4gh/backend.py
```diff
@@ -292,6 +292,12 @@ def setResponseValidation(self, responseValidation):
         self._responseValidation = responseValidation
 
 
+class EmptyBackend(AbstractBackend):
```
dcolligan: What is the motivation behind EmptyBackend?

jeromekelleher: I thought it would be useful for testing - there are times when we want to assert some properties of the backend when we don't care about data at all (`test_views` uses it now I think). I didn't want to instantiate `AbstractBackend` directly as this feels a bit wrong, and we might change the behaviour of `AbstractBackend` later to now be directly instantiable.

dcolligan: Cool, I'm +1 on these.

macieksmuga:
Ah, flask config file. It's rather hard to figure out how to run the server now, for a casual user. README desperately needs a rewrite.

PR #219, opened 2015-03-09
End to end test

dcolligan:
change default sink values from empty strings to none

add end to end test

adds test annotations

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -246,6 +246,9 @@ def addGlobalOptions(parser):
     parser.add_argument(
         "--config-file", "-f", type=str, default=None,
         help="The configuration file to use")
+    parser.add_argument(
+        "--dont-use-reloader", default=False, action="store_true",
+        help="Don't use the flask reloader")
 
 
 def server_main(parser=None):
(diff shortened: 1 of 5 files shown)


jeromekelleher:
This looks good, but we need to change the way the server CLI is invoked. Probably the easiest thing to do is to write a configuration to a temp file, and pass this path to the server as a '-f' argument. This should make it easy to control the configuration of the server.

What are the start-up/shutdown times like here?

dcolligan:
```
$ nosetests tests/end_to_end/
...
Ran 1 test in 1.384s
```

dcolligan:
The time is mainly dependent on the utils.Repeat.sleepSeconds constant.  If we bump it down to 0.1 the execution time drops to 0.658s.  I'll make that change.

jeromekelleher:
OK, that's good, we can run these every time then. Excellent.

dcolligan:
@jeromekelleher why can't we just use the TestConfig for the server (as I do here)?

ga4gh/cli.py
```diff
@@ -246,6 +246,9 @@ def addGlobalOptions(parser):
     parser.add_argument(
         "--config-file", "-f", type=str, default=None,
         help="The configuration file to use")
+    parser.add_argument(
+        "--dont-use-reloader", default=False, action="store_true",
```
jeromekelleher: Maybe `--disable-reloader`?

jeromekelleher:
Yes, that's a better option actually. Can we put this on a different port please? It's a bit annoying to have to stop my dev server each time I want to run the unittests. Something large and random seems good.

Otherwise, +100. This is really tidy and runs really quickly and smoothly.

dcolligan:
Alright, 8001 it is...

ready to merge

Issue #71, opened 2015-01-13
Improve installation time/travis build time


jeromekelleher:
Installation of the `ga4gh` package in a virtualenv is very slow because we must compile [pysam](https://github.com/pysam-developers/pysam) each time. This also happens for our TravisCI builds, which is bad for both us and TravisCI.

Can we speed this up in any way?  Would a [binary wheel](http://lucumr.pocoo.org/2014/1/27/python-on-wheels/)  for pysam on Travis's infrastructure be feasible/desirable?

bioinformed:
+1 feasible (I use pysam wheels all the time)

jeromekelleher:
Excellent! Any quick pointers on how to go about this?

bioinformed:
The best solution is for Andreas and I to upload wheels to PyPI.  I'll ask
him if this is something he is willing to do.

Alternatively (and in the mean time), it isn't hard to set up a local wheel
cache to avoid rebuilding packages.  There are many sources of online
guides to setting up a local "wheelhouse", including
http://vincent.is/using-the-wheel/

bioinformed:
Here is a StackOverflow answer that gives a very detailed recipe for
TravisCI:
http://stackoverflow.com/questions/18067073/speedup-virtualenv-creation-with-numpy-and-pandas

dcolligan:
If I'm reading this correctly, it looks like we'd have to set up some remote server (not Travis) to host the wheels.  @jeromekelleher can we / do we want to do that?

Alternatively, if just anyone can upload arbitrary wheels to PyPI we could do that too... we could have a ga4gh-server-wheels package or something where we package pysam and whatever else we need.

jeromekelleher:
I think the bandwidth requirements would be fairly small @dcolligan, and it would be well worth it. If we get them built, I'm sure we'll find the required infrastructure to serve them. How much hassle is it to build the wheels? I've had a quick look a few times, and it never seems particularly trivial.

PyPI isn't an option as binary wheels for Linux aren't supported AFAIK.

dcolligan:
Here's some information on what travis is running:

http://docs.travis-ci.com/user/ci-environment/

"Travis CI virtual machines are based on Ubuntu 12.04 LTS Server Edition 64 bit"

dcolligan:
Fixed in #225

PR #225, opened 2015-03-09
wheels WIP

dcolligan:
On my local system installing into a new virtualenv off of this branch successfully skips the pysam compile step, which is the most timely step in our build.  These changes drop the build time on my system from around 5 minutes to under 30 seconds. 

Wheels are OS/arch dependent, to an extent.  The one that I built doesn't work on the Travis environment, which uses a 64 bit Ubuntu build.  We just need to get a wheel that the Travis environment finds acceptable, which might take a bit of guess-and-check (building the wheel on an identical Ubuntu environment would work).    The wheel we need can be created with the command `pip wheel pysam`.

We would also need to check in a new version of the wheel whenever pysam had changes we wanted to incorporate.

Questions: 
- is it acceptable to check in the wheelhouse to source control, as I've done here, or do we want to download it off of some remote server every time a branch is pushed to GitHub?
- do we want to use a wheelhouse for only pysam, all our required packages, or something in between?

Issue #71

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,11 @@ python:
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -qq libdb-dev
+  - wget 'http://www.well.ox.ac.uk/~jk/wheelhouse.tar'
+  - tar -xvf wheelhouse.tar
 
 install:
-  - pip install -r requirements.txt
+  - pip install --use-wheel --find-links=wheelhouse/ -r requirements.txt
   - python setup.py install
 
 before_script:


jeromekelleher:
Can you try out the pysam wheel stored [here](http://www.well.ox.ac.uk/~jk/wheelhouse/) please @dcolligan? I've made a Ubuntu 12.04 VM, and mirrored travis's Python setup as closely as I can. 

To answer your questions

- I wouldn't be keen on keeping the wheelhouse in git, and would rather keep it on a server. We can use the URL I've provided initially. If it becomes a problem we can look for better hosting.
- Just pysam would do I think. It's pretty tedious keeping these up to date!

dcolligan:
Here's the build for that wheel: https://travis-ci.org/dcolligan/server/builds/53819082

It looks like it accepted the wheel, but there's some symbol that it can't find (`PyUnicodeUCS2_DecodeUTF8`).  Apparently this has to do with some version mismatch of the interpreter and the library: http://stackoverflow.com/questions/11878666/python-undefined-symbol-pyunicodeucs2-decodeutf8

@jeromekelleher at this point we have three options:
1) continue guess-and-check: make another wheel with the above information and try again
2) build the wheel as part of the Travis build -- this would avoid version mismatches -- and then upload that to some target server.  We can then use that wheel in all Travis builds going forward
3) (I would not prefer this option) If pysam is going to be removed from our code soon, give up :(

jeromekelleher:
Thanks @dcolligan. (1) seems to be the only viable option here. I don't like (2) as we'd have to embed credentials for the uploading site into our Travis build and this seems dangerous. We're betting heavily on pysam doing lots of stuff for us, so it's hopefully going to be here for a long time.

I'll have another go at making the wheel.

jeromekelleher:
OK, I rebuilt my Python in the Ubuntu VM using the correct unicode size, and it all seems to work now. Seems like a result to me!

.travis.yml
```diff
@@ -5,9 +5,11 @@ python:
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -qq libdb-dev
+  - mkdir wheelhouse
+  - wget 'http://www.well.ox.ac.uk/~jk/wheelhouse/pysam-0.8.2.1-cp27-none-linux_x86_64.whl' -O 'wheelhouse/pysam-0.8.2.1-cp27-none-linux_x86_64.whl'
```
jeromekelleher: Is there any way to wget wheelhouse without embedding the filename? I can make a tar ball out of it if needs be.

dcolligan: Yeah, using a tarball would enable us to replace the mkdir and wget with a wget and tar -xzvf, which seems cleaner to me (plus enables us to add other wheels in the future without modifying `.travis.yml`, if we ever think that desirable).

dcolligan: @jeromekelleher or I could just squash and commit the existing changes since they seem to be working; your call.

jeromekelleher: It's available at http://www.well.ox.ac.uk/~jk/wheelhouse.tar now, give it a shot. It'll bite us later if we don't get rid of the version specific filenames in travis.yml.

pgrosu: How about just making a version-agnostic symbolic link at your server that points to the version you want and then the name will never change.  If you want you can have a script that makes it version-agnostic.

dcolligan:
@jeromekelleher alright, this latest version seems to work.

merge me!

jeromekelleher:
+1. Merging.

Issue #223, opened 2015-03-09
Unittest like assert* method required for datadriven.DataDrivenTest


jeromekelleher:
PR #211 includes a definition of `assertEqual` in `DataDrivenTest`, as this allows us to write test cases much more conveniently than raw asserts, and we cannot use unittest.TestCase in these tests for technical reasons. It would be useful to define a subset of the unittest `assert*` methods here so we can use them. Say:

- assertTrue, assertFalse
- assertIs, assertIsNot
- assertGreater, assertGreaterEqual
- assertLess, assertLessEqual
- assertRaises

These should be quick to implement, and will suffice for the vast majority of tests we will want to perform in the data driven scenario.

dcolligan:
Here are the methods unittest defines: https://docs.python.org/2/library/unittest.html
```
assertEqual(a, b)	a == b	 
assertNotEqual(a, b)	a != b	 
assertTrue(x)	bool(x) is True	 
assertFalse(x)	bool(x) is False	 
assertIs(a, b)	a is b	2.7
assertIsNot(a, b)	a is not b	2.7
assertIsNone(x)	x is None	2.7
assertIsNotNone(x)	x is not None	2.7
assertIn(a, b)	a in b	2.7
assertNotIn(a, b)	a not in b	2.7
assertIsInstance(a, b)	isinstance(a, b)	2.7
assertNotIsInstance(a, b)	not isinstance(a, b)	2.7

```

dcolligan:
And also (from the same page):
```
assertAlmostEqual(a, b)	round(a-b, 7) == 0	 
assertNotAlmostEqual(a, b)	round(a-b, 7) != 0	 
assertGreater(a, b)	a > b	2.7
assertGreaterEqual(a, b)	a >= b	2.7
assertLess(a, b)	a < b	2.7
assertLessEqual(a, b)	a <= b	2.7
assertRegexpMatches(s, r)	r.search(s)	2.7
assertNotRegexpMatches(s, r)	not r.search(s)	2.7
assertItemsEqual(a, b)	sorted(a) == sorted(b) and works with unhashable objs	2.7
assertDictContainsSubset(a, b)	all the key/value pairs in a exist in b	2.7
```
and
```
assertMultiLineEqual(a, b)	strings	2.7
assertSequenceEqual(a, b)	sequences	2.7
assertListEqual(a, b)	lists	2.7
assertTupleEqual(a, b)	tuples	2.7
[...]

PR #227, opened 2015-03-10
Add assert methods to data driven test suite

dcolligan:
Issue #223

### DIFF
--- a/tests/datadriven/__init__.py
+++ b/tests/datadriven/__init__.py
@@ -7,6 +7,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import contextlib
 import os
 import inspect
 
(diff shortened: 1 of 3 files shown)


tests/datadriven/__init__.py
```diff
@@ -52,11 +53,17 @@ class DataDrivenTest(object):
     corresponding to this, and then test that these objects have the
     properties that we expect.
     """
-    def __init__(self, setId, baseDir):
-        self._setId = setId
-        self._dataDir = os.path.join(baseDir, setId)
-        self._gaObject = self.getDataModelClass()(
-            self._setId, self._dataDir)
+    def __init__(self, *args):
+        if len(args) == 0:
```
jeromekelleher: Why are we doing this?

tests/datadriven/__init__.py
```diff
@@ -86,6 +93,130 @@ def assertEqual(self, a, b):
...
+        an assertion error.
+        """
+        if a == b:
+            raise AssertionError("{} == {}".format(a, b))
+
+    def assertTrue(self, x):
+        """
+        Tests that x is true.  If it is false, raise an assertion error.
+        """
+        if bool(x) is False:
```
jeromekelleher: `if not bool(x):`

tests/datadriven/__init__.py
```diff
@@ -86,6 +93,130 @@ def assertEqual(self, a, b):
...
+            raise AssertionError("{} is False".format(x))
+
+    def assertFalse(self, x):
+        """
+        Tests that x is false.  If it is true, raise an assertion error.
+        """
+        if bool(x) is True:
```
jeromekelleher: `if bool(x):`

tests/datadriven/__init__.py
```diff
@@ -86,6 +93,130 @@ def assertEqual(self, a, b):
...
+        """
+        Tests that a is less than or equal to b.
+        If not, raise an assertion error.
+        """
+        if a > b:
+            raise AssertionError(
+                "{} is not less than or equal to {}".format(a, b))
+
+    @contextlib.contextmanager
+    def assertRaises(self, exceptionType):
```
jeromekelleher: I don't quite follow this definition. Why do we yield, and why is it a contextmanager?

tests/datadriven/test_asserts.py
```diff
@@ -0,0 +1,105 @@
...
+import tests.datadriven as datadriven
+
+
+class TestAsserts(unittest.TestCase):
+
+    def setUp(self):
+        self.dataDrivenTest = datadriven.DataDrivenTest()
+
+    def testAssertEqual(self):
+        self.dataDrivenTest.assertEqual(0, 0)
```
jeromekelleher: Shouldn't this be:
```python
self.assertRaises(AssertionError, self.dataDrivenTest.assertEqual, 0, 1)
```

tests/datadriven/__init__.py
pgrosu: :)

tests/datadriven/test_asserts.py
```diff
@@ -0,0 +1,105 @@
...
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import unittest
+import tests.datadriven as datadriven
+
+
+class TestAsserts(unittest.TestCase):
+
+    def setUp(self):
```
jeromekelleher: Ah, I see, Can't we just have default values for the constructor of `DataDrivenTest` rather than unpacking the args tuple?

jeromekelleher:
Looks good @dcolligan, thanks. A few minor quibbles, but altogether exactly what we want.

tests/datadriven/__init__.py
dcolligan: So I can cleanly test the assert methods without doing unnecessary file operations, etc.

tests/datadriven/__init__.py
dcolligan: This is the syntax to allow us to use the `with` keyword when using `assertRaises`

tests/datadriven/test_asserts.py
dcolligan: Yeah, we could do that

tests/datadriven/test_asserts.py
dcolligan: I presume you mean the lines below the one you tagged.

But no, that's not how the assertRaises statement I created operates.  I find it more convenient to use the with syntax than separating out the arguments from the function.  It's a more general pattern.

dcolligan:

jeromekelleher:
Looks good, thanks @dcolligan. Two things:

1. Maybe it would be better to make a new class called `TestCase` in the datadriven module, and move all of the unittest-like stuff in there. The `DataDrivenTest` can then extend this, and we won't need to have funny constructor semantics to allow us test the assertion methods. We can also put in some documentation in the `TestCase` class explaining why we're recreating a subset of unittest.TestCase.
2. I feel pretty strongly that we should follow the standard Python 2.7 semantics for `assertRaises`. This should have the same signature and behaviour as the unittest method.

dcolligan:
Wrt (2), both semantics are supported in unittest: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises  

Like I said, I find the one that I wrote far more understandable and useful.

dcolligan:
Ok, I caved and implemented both versions of `assertRaises`

@jeromekelleher new commit pushed with your suggestions

jeromekelleher:
Thanks @dcolligan!

+1

Issue #222, opened 2015-03-09
Unit tests required for SimulatedBackend


jeromekelleher:
We need some basic tests for the simulator, to ensure that it works reasonably sensibly. We should make a new test module, `tests/test_simulator.py`, and create a test class in there, `TestVariantSet`. This class should test (e.g.)

1. The values passed into the constructor are stored correctly (after adding getter methods). For example, we'd want to check `self.assertEqual(seed, simulator.getRandomSeed()).
2. The variants we generate have the required properties (i.e., `start` and `end` are in the correct range).
3. The variants we generate are consistent. That is, if we allocate two simulators with the same seed and other parameters, they should produce identical variants when called on the same data.
4. The variants produced are `self-consistent`. This means that if we run the same query on a variantset more than once, we get the same result back.

These tests should be run at the VariantSet level. That is, we allocate instances of `datamodel.variants.SimulatedVariantSet`; we do not allocate any instances of `Backend`.

dcolligan:
Fixed in #229

Issue #195, opened 2015-02-26
Increase minimum required test coverage?


dcolligan:
We're at 95% test coverage right now, and in `.travis.yml` we're mandating 80% test coverage.  Do we want to bump that up to, say, 90 (or even higher?) so we don't regress?

Edit: we're actually at 85%... see below and #197

jeromekelleher:
+1 --- 90% seems like a good number.

dcolligan:
We're at 85% right now.  The `datamodel` and `cli` modules seem to be bringing us down.
```
Name                          Stmts   Miss  Cover   Missing
-----------------------------------------------------------
ga4gh                             1      0   100%
ga4gh._protocol_definitions     603     44    93%   38-39, 72-73, 111-112, 156-157, 251-252, 281-282, 311-312, 346-347, 432-433, 464-465, 502-503, 536-537, 821-822, 874-875, 916-917, 995-996, 1104-1105, 1216-1217, 1301-1302, 1379-1380, 1467-1468, 1697-1698
ga4gh.backend                   138     15    89%   39-43, 49-52, 82, 121, 131, 141, 151, 210, 218
ga4gh.backend_exceptions         14      0   100%
ga4gh.cli                       432    104    76%   28-40, 50-73, 100-105, 131, 133, 135, 169, 234, 245-248, 279-280, 306-312, 325, 338-341, 347-352, 358-370, 383, 396, 409, 422, 435-440, 448, 462-464, 475, 486, 496-512, 767, 788-789
ga4gh.client                    122     14    89%   39, 51, 56, 71, 75-77, 81-83, 87, 91-93, 112
ga4gh.converters                 17      6    65%   15-16, 24, 28, 36, 40
ga4gh.datamodel                   3      0   100%
ga4gh.datamodel.reads            33     20    39%   22-32, 38-43, 53-54, 60-63
ga4gh.datamodel.references       31     18    42%   24-35, 41-44, 55-56, 62-65
ga4gh.datamodel.variants        224     51    77%   195, 199, 221-224, 230-231, 249-260, 269, 320-328, 335-353, 361-371, 375-376
ga4gh.frontend                  123      9    93%   39, 42, 93, 95, 141, 162, 178, 183, 211
ga4gh.frontend_exceptions        74      0   100%
ga4gh.protocol                   70      7    90%   20-23, 34, 50, 120
ga4gh.serverconfig                7      0   100%
-----------------------------------------------------------
TOTAL                          1892    288    85%
```

[...]

PR #229, opened 2015-03-10
Add tests for simulated variant sets

dcolligan:
Increase required test coverage to 85%

Issue #222
Issue #195

### DIFF
--- a/tests/unit/test_simulator.py
+++ b/tests/unit/test_simulator.py
@@ -1,0 +1,103 @@
+"""
+Tests for simulated data
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import unittest
+
+import ga4gh.datamodel.variants as variants
+
+
+class TestSimulatedVariantSet(unittest.TestCase):
+    """
+    Test properties of the SimulatedVariantSet
+    """
+    def setUp(self):
+        self.randomSeed = 0
+        self.numCalls = 2
+        # ensure variantDensity is >= 1 so we get deterministic behavoir
+        self.variantDensity = 1
+        self.variantSetId = 3
+        self.simulatedVariantSet = self._getSimulatedVariantSet()
+        self.referenceName = 'ref'
+        self.startPosition = 100
+        self.endPosition = 103
+        self.variantName = 'unused'
+        self.callSetIds = ['unused']
+        self.bases = ["A", "C", "G", "T"]
+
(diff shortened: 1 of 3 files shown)


tests/unit/test_simulator.py
```diff
@@ -0,0 +1,93 @@
...
+
+class TestSimulatedVariantSet(unittest.TestCase):
+    """
+    Test properties of the SimulatedVariantSet
+    """
+    def setUp(self):
+        self.randomSeed = 0
+        self.numCalls = 1
+        # ensure variantDensity is >= 1 so we get deterministic behavoir
+        self.variantDensity = 2
```
jeromekelleher: Just set this to 1, it should be interpreted as a probability.

tests/unit/test_simulator.py
```diff
@@ -0,0 +1,93 @@
...
+        self.callSetIds = ['unused']
+        self.bases = ["A", "C", "G", "T"]
+
+    def _getSimulatedVariantSet(self):
+        simulatedVariantSet = variants.SimulatedVariantSet(
+            self.randomSeed, self.numCalls,
+            self.variantDensity, self.variantSetId)
+        return simulatedVariantSet
+
+    def _getSimulatedVariantsList(self, simulatedVariantSet=None):
```
jeromekelleher: Why do we need the parameter here? It looks to me like it's not actually used.

jeromekelleher:
+1 modulo niggles.

tests/unit/test_simulator.py
dcolligan: see `testConsistency`

dcolligan:
@jeromekelleher pushed commit with your suggestions

tests/unit/test_simulator.py
jeromekelleher: Nope, still not seeing where this parameter is used. It doesn't have any effect anyway, even if it is used: you don't use `simulatedVariantSet` anywhere in the function.

dcolligan: ah, well that's a bug in the code!

dcolligan:
@jeromekelleher pushed a commit to fix that bug

tests/unit/test_simulator.py
dcolligan: and check lines 84-85 ... that's where the parameter is passed

jeromekelleher:
Ah, now I get it.

+1.

Issue #226, opened 2015-03-10
tests/end_to_end/test_gestalt should use SimulatedBackend.


jeromekelleher:
The end-to-end tests should allocate a SimulatedBackend, and run some simple tests on it. For example, we should

1. Write a server config file to a tmp file, and pass this to the server using the '-f' argument. This allows us to set the simulator parameters, such as VARIANT_DENSITY and NUM_VARIANT_SETS.
2. We can then test the output of the client to see if it matches. For example, we can run a `search-variantsets`, and verify the number of lines written to stdout is correct. If we set VARIANT_DENSITY to 1, then we can also verify that the number of lines written to stdout for a variants-search is correct. For example, if we run
```
$ python client_dev.py variants-search http://localhost:8000/v0.5.1 -s0 -e2 
```
we should get 2*NUM_VARIANT_SETS lines in the output. This tests a lot of different aspects of the system and would be a very useful test.

dcolligan:
Running that client command doesn't produce any results.

With:
```
$ python server_dev.py -c TestConfig
```
We get:
```
$ python client_dev.py -vv variants-search http://localhost:8000/v0.5.1 -s0 -e2
ga4gh.client: POST http://localhost:8000/v0.5.1/variantsets/search
ga4gh.client: json request:
ga4gh.client: {
    "datasetIds": [],
    "pageSize": null,
    "pageToken": null
}
ga4gh.client: json response:
ga4gh.client: {
    "nextPageToken": null,
    "variantSets": []
}
ga4gh.client: POST http://localhost:8000/v0.5.1/variants/search
ga4gh.client: json request:
ga4gh.client: {
    "callSetIds": [],
    "end": 2,
    "pageSize": 100,
    "pageToken": null,
    "referenceName": "1",
    "start": 0,
    "variantName": null,
    "variantSetIds": []
}
ga4gh.client: json response:
ga4gh.client: {
    "nextPageToken": null,
    "variants": []
}
```

jeromekelleher:
We can't just use the bottled `TestConfig` here @dcolligan because it has an '__EMPTY__' variant set. We do need to set the config explicitly, or we can't test that the returned values are correct.

dcolligan:
Ah, forgot about the distinction between empty and simulated.

dcolligan:
Fixed in #230

PR #230, opened 2015-03-11
Make test_gestalt use SimulatedBackend

dcolligan:
We should test some kind of return values from the server before this is committed (as per the issue).

Issue #226

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -259,8 +259,8 @@ def server_main(parser=None):
     args = parser.parse_args()
     frontend.configure(args.config, args.config_file)
     frontend.app.run(
-            host="0.0.0.0", port=args.port, debug=True,
-            use_reloader=not args.dont_use_reloader)
+        host="0.0.0.0", port=args.port, debug=True,
+        use_reloader=not args.dont_use_reloader)
 
 
 ##############################################################################
--- a/tests/end_to_end/server.py
+++ b/tests/end_to_end/server.py
@@ -19,12 +19,6 @@ class ServerTest(unittest.TestCase):
     """
     Manages server bootup and shutdown for a test
     """
-    serverCmdLine = """
-        python server_dev.py
-        --dont-use-reloader
-        --config TestConfig
-        --port {} """
-
     def run(self, *args, **kwargs):
         self.setupServer()
         try:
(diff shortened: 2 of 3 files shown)


dcolligan:
@jeromekelleher this should work as specified in the issue now

jeromekelleher:
It would be nice to make the config parameterisable, so that we're not depending on the values embedded in test_server.py. I'm not sure this is a good name either -- do we run any tests in `test_server.py`? If not, we should just rename it `server.py`, or just move all of the functionality into `test_gestalt`, which seems to be the only place it's used.

dcolligan:
The config is parameterisable.  That's what `ServerTestConfigFile` is for.

I would advocate keeping the server startup stuff in a separate module, since I could definitely imagine this being used by other end to end tests in the future.

tests/end_to_end/test_server.py
```diff
@@ -112,3 +107,52 @@ def getServerErrLines(self):
...
+
+
+class ServerTestConfigFileDeterministic(ServerTestConfigFile):
+    """
+    Launch a server test with a custom configuration file
+    that ensures there is no pseudorandom behavior.
+    """
+    def setConfig(self):
+        self.config = """
+SIMULATED_BACKEND_NUM_VARIANT_SETS = 1
```
jeromekelleher: This doesn't look very configurable to me - what if I want 5 variantSets?

jeromekelleher:
OK, I won't argue about keeping the server stuff in a separate module, I think we should just rename it. It shouldn't be picked up and run by nosetests unless it contains actual test cases (which I think it doesn't?).

tests/end_to_end/test_server.py
dcolligan: Then you override `ServerTestConfigFile` and implement `setConfig` in your subclass with 5 variantsets and whatever else you want in the config.

dcolligan:
@jeromekelleher pushed a change with a rename

tests/end_to_end/test_server.py
jeromekelleher: What about just adding some parameters to setConfig here? numVariantSets=1 by default? I don't want to have to reproduce this exact code just to change stuff like that.

Also, we should have more than one variant set to test the client properly, and I'm dead against the tests depending on a value embedded within a string in a different module.

dcolligan: I'll move that code to the `test_gestalt.py` file then.

dcolligan:
@jeromekelleher pushed a change that moved some code around

jeromekelleher:
Excellent, thanks. +1.

PR #232, opened 2015-03-11
POC implementation of accessing remote data.

jeromekelleher:
Pysam and htslib provide the functionality to access files remotely over FTP/HTTP. We simply need to provide a URL as an argument, and the network accesses happen transparently. This seems like useful functionality, and might allow us to deploy easily on (e.g.) Amazon using S3 storage. This PR is a rough POC implementation to illustrate how we might do this.

Questions:
1. Is this something we want to support?
2. Will this be sufficient to access data stored in Amazon S3 (http://aws.amazon.com/public-data-sets/) like 1000G?

Issues:
1. How do we manage input data? (In this POC, I've just dumped a JSON file with the URLS into the variants dir. We should think hard about what the right way to do this is.)
2. How do we manage the index files? Currently, htslib downloads the index file and writes it to CWD. We'd need to be a lot more systematic about this for several reasons: (a) in a deployment under apache or another web server, the concept of CWD is a bit hazy, and we certainly don't want to make some random directories writeable by the `www-data` user; (b) when we're dealing with lots of variant sets, there is every chance that two index files from different variant sets will have the same file names.
3. How does this affect our latency? Will we get usable throughput via this approach? We need some profiling to answer these questions.

Here is an example `urls.json` file:
```json
{
    "urls": [ 
     "ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/release/20130502/supporting/bcf_files/ALL.chr1.phase3_shapeit2_mvncall_integrated_v5.20130502.genotypes.bcf", 
     "ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/release/20130502/supporting/bcf_files/ALL.chr10.phase3_shapeit2_mvncall_integrated_v5.20130502.genotypes.bcf"
    ]
}
```

### DIFF
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -6,11 +6,14 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import atexit
 import datetime
 import glob
+import json
 import os
 import random
-
+import shutil
+import tempfile
 import pysam
 
 import ga4gh.protocol as protocol
@@ -259,6 +262,15 @@ def isEmptyIter(it):
     return next(it, _nothing) is _nothing
 
 
+def _cleanupHtslibsMess(indexDir):
+    """
+    Cleanup the mess that htslib has left behind with the index files.
+    This is a temporary measure until we get a good interface for
+    dealing with indexes for remote files.
+    """
+    shutil.rmtree(indexDir)
+
+
 class HtslibVariantSet(AbstractVariantSet):
     """
     Class representing a single variant set backed by a directory of indexed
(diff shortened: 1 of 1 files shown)


bartgrantham:
Yay!  I've been waiting for this one for a while.  The hazy vision I had for this capability revolved around exploiting Amazon S3's ability to serve a bucket as a static website to scale the server horizontally.  If the server can be deployed in ElasticBeanstalk, this would provide a very easy path for scalable deployment for anyone with an Amazon account.  This, hopefully, provides a short circuit for anyone encountering resistance getting resources in their organizations.

Some thoughts from your questions:

1. I think having a JSON manifest in a standard location on the endpoint makes more sense than a local "urls.json" config.  If you consider a URL the same as a path to a DATADIR, you'd have something like `$ ga4gh_server https://genedata.example.com` as an analog.
2. I think getting this right is really important.  I'd suggest that in the absence of a configuration standard, there would be a convention to follow regarding the naming of data files and their associated index files, and the grouping of these files into sets.  IIRC, the server already does something like this?  My answer to the first question applies here as well, if you have an explicit manifest for a data set it should exist in a standard location in the DATADIR.
3. An HTTP Range query will always have more overhead than seek()'ing a file handle.  But I don't think the latency will be _that_ bad.  Reading large blocks of data may or may not be competitive overall depending on the local disk, network latency, etc.  The big gain is that the opportunity for parallelization is far greater if you're not bound by your local disk.  My experience and intuition tells me that accessing/serving the data from remote storage instead of the local filesystem will be slower in general for lower traffic but is essential for higher traffic.

So for 1 and 2, my suggestion is:

* the handling of data in local directories and the data at urls should be handled as similarly as possible
* the server is directed to the root path of the data via command line (we may outgrow this and need a conf file, but even then the conf should be limited to pointing to the data, not characterizing it)
* the configuration of the data is discovered via a JSON manifest in a standard location, falling back to a straightforward convention that is in harmony with people's common use of htslib in the case that there is no manifest.

jeromekelleher:
I've updated this to be slightly less nasty. Now we download the index files to a temporary directory, which we delete when we are finished. This is a lot better than dumping them into the CWD as we should now be able to run this under apache without hitting permissions problems. It would be nice to include this so that we can at least test out deployment on AWS to see how things work. 

What do we think?

ga4gh/datamodel/variants.py
```diff
@@ -6,11 +6,15 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import atexit
 import datetime
 import glob
+import json
 import os
+import os.path
```
dcolligan: this import isn't necessary, is it?

dcolligan:
+1 aside from nit

ga4gh/datamodel/variants.py
jeromekelleher: Indeed it isn't -- learn something new every day!

dcolligan: Speaking of which, check out this thread

http://stackoverflow.com/questions/101268/hidden-features-of-python

PR #233, opened 2015-03-11
Centralised and simplified exceptions.

jeromekelleher:
This commit removes the distinction between backend and frontend exceptions, which was proving unhelpful.

This should make exception handling and reporting more robust. Currently it is almost impossible to debug problems in a deployment setting.


ga4gh/exceptions.py
```diff
@@ -0,0 +1,179 @@
...
+        default. For more elaborate messages that must be contructed
+        at run time, we can override this method and use state held
+        in instance variables in the exception.
+        """
+        return self.message
+
+    @classmethod
+    def getErrorCode(cls):
+        """
+        Returns an integer that can be used to identif this class.
```
dcolligan: typo

ga4gh/frontend.py
```diff
@@ -140,46 +140,32 @@ def handleHttpOptions():
...
+    if app.config['DEBUG']:
+        print(traceback.format_exc(exception))
+    serverException = exception
+    if not isinstance(exception, exceptions.ServerException):
+        serverException = exceptions.convertException(exception)
+    error = serverException.toProtocolElement()
+    responseStr = error.toJsonString()
+    return flask.Response(
+        responseStr, status=serverException.httpStatus,
+        mimetype="application/json")
```
dcolligan: doesn't jsonify set the mimetype automatically?  I'd prefer those semantics

tests/unit/test_exceptions.py
```diff
@@ -38,86 +38,55 @@ def run(self, *args, **kwargs):
...
-            except TypeError:
-                # built-in python exceptions can't use inspect
-                originalException = originalExceptionClass()
-            mappedException = mappedExceptionClass()
-            response = frontend.handleException(originalException)
-            self.assertEquals(response.status_code, mappedException.httpStatus)
+    def getGa4ghException(self, data):
+        return protocol.GAException.fromJsonString(data)
 
     def testFrontendException(self):
```
dcolligan: change name of this test

tests/unit/test_exceptions.py
```diff
@@ -38,86 +38,55 @@ def run(self, *args, **kwargs):
...
-        exception = frontendExceptions.ObjectNotFoundException()
+        exception = exceptions.ObjectNotFoundException()
         response = frontend.handleException(exception)
         self.assertEquals(response.status_code, 404)
 
     def testBackendException(self):
```
dcolligan: change name of this test

ga4gh/frontend.py
jeromekelleher: I'm just following the pattern in the other methods for simplicity. We can change them all to jsonify if you prefer, I don't mind which we use once we're consistent. Your call.

ga4gh/exceptions.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        # We use the CRC32 digest of the class name as a unique code.
+        # We follow the recommendation of the Python docs to ensure
+        # that this value is signed 32 bit integer.
+        code = zlib.crc32(cls.__name__) & 0xffffffff
```
dcolligan: This could still potentially have collisions.  CRC does not have strong uniqueness guarantees.

ga4gh/exceptions.py
```diff
@@ -0,0 +1,179 @@
...
+    message = "The requested object was not found"
+
+
+class UnsupportedMediaTypeException(ServerException):
+    httpStatus = 415
+    message = "Unsupported media type"
+
+
+class VersionNotSupportedException(NotFoundException):
+
```
dcolligan: style: some of these classes have a newline between the class definition and the attributes, and some don't.  whatever your choice, be consistent (I prefer no newline, no comment if it's obvious what the class is for)

ga4gh/exceptions.py
```diff
@@ -0,0 +1,179 @@
...
+    def __init__(self, message):
+        self.message = message
+
+
+class RequestValidationFailureException(ServerException):
+    """
+    A validation of the request data failed
+    """
+
```
dcolligan: style: no newline between class doc comment and first method (this applies other places in the file as well)

ga4gh/exceptions.py
```diff
@@ -0,0 +1,179 @@
...
+
+class ServerError(ServerException):
+    """
+    Superclass of all exceptions that indicate a bug has occured.
+    """
+    httpStatus = 500
+
+
+class ResponseValidationFailureException(ServerError):
```
dcolligan: Did you forget to add RequestValidationFailureException ?

ga4gh/exceptions.py
```diff
@@ -0,0 +1,179 @@
...
+    Converts the specified exception that is not a subclass of the
+    ServerException into a ServerError so that it can be correctly
+    serialised and communicated back to the client.
+    """
+    serverException = ServerError()
+    serverException.message = str(exception)
+    return serverException
+
+
+class ServerException(Exception):
```
dcolligan: maybe rename this class to make the difference with ServerError more obvious?  Like BaseServerException or something

ga4gh/exceptions.py
dcolligan: Plus, this ties the code to the name of the class, so someone could break the protocol by renaming internal classes.  That strikes me as a bad trap to set.

Maybe this is fine for now, and we just put a big TODO on this to rethink before releasing v1 or whatever.

ga4gh/exceptions.py
```diff
@@ -0,0 +1,179 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import zlib
+
+import ga4gh.protocol as protocol
+
+
+def convertException(exception):
```
dcolligan: I don't like the name of this method; it's ambiguous.  Convert exception to what?

maybe convertToServerError ?

tests/unit/test_exceptions.py
```diff
@@ -38,86 +38,55 @@ def run(self, *args, **kwargs):
         with mock.patch('flask.jsonify', dummyFlask):
             super(TestExceptionHandler, self).run(*args, **kwargs)
 
-    def testMappedException(self):
```
dcolligan: The essence of this test would be good to keep around -- that every exception that we define can be successfully constructed.

ga4gh/frontend.py
dcolligan: Yeah, having mimetype strings in random methods definitely strikes me as something to avoid

dcolligan:
This looks generally good.

You'll need to modify the import graph layering test as well since you've added and deleted some modules.

ga4gh/frontend.py
jeromekelleher: We can't use jsonify in the other methods because they work with strings rather than dicts (and we definitely don't want to change this). I've changed the error handler to use jsonify again.

Working through the rest of the comments.

ga4gh/exceptions.py
jeromekelleher: We'll, we're testing to make sure all the codes are unique, so if we do have a collision it'll show up in CI. People can change the codes by renaming existing classes, but that's OK: any released version will be against a specific tag in git and we can then trace back to the actual class from there (we should write a little utility that takes a code and spits out the class name).

The reason I didn't use SHA1 or whatever is they output 16 byte digests, which correspond to very, very large numbers. Too big for the value in the protocol.

ga4gh/exceptions.py
jeromekelleher: No, this is a different class of error. It's not a ServerError if someone sends us unusable crud, but it certainly is if we _return_ unusable crud.

ga4gh/exceptions.py
dcolligan: Ok.  Tests make me sleep better at night.

jeromekelleher:
Thanks for the feedback @dcolligan, very useful. I've implemented the majority of your suggestions.

I backed out of the jsonify change because it actually makes it a lot easier to test the exceptions if we don't use it. Also, style-wise, pep8 mandates a blank line between the `__init__` and the class variable definitions (who knew?).

ga4gh/exceptions.py
```diff
@@ -0,0 +1,181 @@
...
+import ga4gh.protocol as protocol
+
+
+def getServerError(exception):
+    """
+    Converts the specified exception that is not a subclass of
+    BaseServerException into a ServerError so that it can be correctly
+    serialised and communicated back to the client.
+    """
+    assert not isinstance(exception, BaseServerException)
```
dcolligan: can we raise a specific error here?  I'm kind of allergic to the assert keyword

ga4gh/exceptions.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def __init__(self, message):
+        self.message = message
+
+
+class RequestValidationFailureException(BaseServerException):
+    """
+    A validation of the request data failed
+    """
+
```
dcolligan: delete this newline

ga4gh/frontend.py
```diff
@@ -117,69 +119,63 @@ def configure(config="DefaultConfig", configFile=None):
     app.backend = theBackend
 
 
+def getFlaskResponse(responseString, httpStatus=200):
```
dcolligan: excellent

dcolligan:
Cool, I'm +1 after you fix the above things.

ga4gh/exceptions.py
pgrosu: @dcolligan - Hehehe :)

jeromekelleher: What would you suggest? This is one of those "this should never happen, I'm just putting in an assert to be on the safe side" situations. It's a basic programming error if it does happen, so I think and assertion is appropriate.

ga4gh/exceptions.py
jeromekelleher: Done.

ga4gh/exceptions.py
dcolligan: AssertionError

dcolligan: I mean, I guess I don't feel strongly about it either way

dcolligan:
+1

jeromekelleher:
@dcolligan  - merge me please!

Issue #228, opened 2015-03-10
Set up travis to deploy to PyPi


dcolligan:
Right now we need to manually upload ga4gh source distributions through the PyPi web interface to release a version.  It would be much better if we had Travis do this for us.  Here's a HOWTO: http://docs.travis-ci.com/user/deployment/pypi/

jeromekelleher:
This seems like overkill and unnecessary work to me; once you have your PyPI credentials set up, you can use 
```
python setup.py sdist upload
```
and it's done.

dcolligan:
That's actually not the recommended way to do it since it goes over plaintext

https://python-packaging-user-guide.readthedocs.org/en/latest/distributing.html#uploading-your-project-to-pypi

jeromekelleher:
Didn't realise that, thanks. Still, twine looks pretty straightforward.

jeromekelleher:
Closing this due to inactivity. Twine is working fine for now I think.

PR #236, opened 2015-03-11
Add scripts/upload_source.py

dcolligan:
This is a script to easily upload a version of the server to pypi

Also do some refactoring

### DIFF
--- a/scripts/google_examples.py
+++ b/scripts/google_examples.py
@@ -6,18 +6,14 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import os
-import subprocess
-
-import yaml
+import shlex
 
 import utils
 
 
 def getKey():
-    filepath = 'scripts/auth.yml'
-    with open(filepath) as stream:
-        doc = yaml.load(stream)
-        key = doc['google']['key']
+    doc = utils.getAuthValues()
+    key = doc['google']['key']
     return key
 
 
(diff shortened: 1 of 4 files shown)


scripts/auth.yml.example
```diff
@@ -1,2 +1,5 @@
 google:
     key: YOUR_KEY
+pypi:
```
jeromekelleher: This makes me nervous (accidental commit of credentials to github) - does ~/.pypirc not work with twine?

jeromekelleher:
This seems like overkill to me also I'm afraid. It'll take a few seconds to run
```
python setup.py sdist
twine upload  dist/ga4gh-{version}.tar.gz
```
This will happen (say) once a month, and doesn't need to be automated. (Plus, I might not actually want to do the things that are in the script --- what if I want to keep a bunch of earlier versions in my dist directory for testing?)

The refactorings look good --- feel free to keep those. I'm -1 on the script itself though.

scripts/auth.yml.example
dcolligan: It looks like it does work with pypirc, fyi:

https://github.com/pypa/twine/blob/master/twine/utils.py

dcolligan:
@jeromekelleher  Alright, took out the upload script but kept the rest of the changes

dcolligan:
(although I would note a script like this might be useful to enforce some kind of pushing policy if/when we tackle #228 )

jeromekelleher:
Thanks @dcolligan, +1.

Issue #23, opened 2014-10-20
Simple WSGI deployment 


jeromekelleher:
Deploying the server on WSGI compliant HTTP servers is currently not straightforward. We need simple, well-documented recipies for deploying on common platforms such as Apache and Nginx. We may need to add some helper functions into `ga4gh.server` to make this as easy as possible.

dcolligan:
Fixed in #238

PR #238, opened 2015-03-12
Update readme and misc deployment fixes.

### LINKED ISSUE #224: README is out of date
The README has not been updated to reflect the change over to a fully config-file driven input system. We should also take this opportunity to simply the README, and start preparing the way for a wormtable-free, deployable server. Some notes:

- We should remove the distinction between wormtable and Tabix variants. Now, we should explain the structure of the data directory (reads and variants), and state we can support VCF and BCF files.
- Our worked examples should only include references to the raw VCF files, and not mention wormtable at all. We should explain how to create a custom config file, and run this using the '-f' argument to the server.
- Remove the section on `Converting 1000G data`.
- Update the Code Layout section, to reflect more recent developments.

jeromekelleher:
- Updated the README to restructure to be more in line with the current
  state of the project.
- Added instructions for deployment on Apache and Ubuntu.
- Updated the configuration interface to be more sensible in this
  context.
- Fixed deployment issue in which the index.html template was not being
  included in the tarball.
- Bumped version number to 0.1.0a2 to distinguish this version from the
  current PyPI package.

The README is a draft and will certainly need a lot more work before it is satisfactory. This process can either as part of this PR or can be a follow up, depending on how people feel.

I would suggest that we tag version 0.1.0a2 once we have verified that the deployment instructions work and we are happy with the README. (The deployment instructions don't quite work right now as as the PyPI package doesn't have some required fixes.) Once we have tagged 0.1.0a2 we should upload this version to PyPI.

This closes #224 and #23


dcolligan:
Looks good 

+1

PR #243, opened 2015-03-12
Add datadriven BAM files

dcolligan:
- add bam2bam
- rename datadriven SAM file dir


jeromekelleher:
Good, we want BAMs too. Come to think of it though, can't we just use samtools to do the jobs of the new scripts?

dcolligan:
Presumably, since pysam and samtools are different front ends to the same functionality.  Is that an argument against merging this?

jeromekelleher:
No, I want to keep the BAM files, so we'll definitely merge some version of this. What I'm wondering is if we need the scripts at all - if we can reproduce the files we have from the source and a samtools one-liner, then we should probably just do that (recording the one-liner in the README).

dcolligan:
I glanced over the samtools docs and didn't see anything obvious, but the terminology is all greek to me, so maybe I'm missing something.

pgrosu:
@dcolligan, what's unclear in samtools?  It's good to clear this up in order to start from the same interpretation before implementation.

diekhans:
Hannas can help with samtools..



... Sent from my computer phone

-----Original Message-----

dcolligan:
Looking over the samtools documentation again, it seems like there's no way to do precisely what we want, but I've asked Hannes to comment on this PR just to make sure.

It seems like it is possible to use samtools to:
1) using the view command, select a certain subset of alignments based on certain characteristics of those alignments
2) using the view command, use a region specifier to select a subset of alignments

Unfortunately, neither of these work for our case since we don't want to assume anything about the contents of the SAM/BAM input file (aside from it having more than 10 data lines).

dcolligan:
Yes, an alternative is to use the command line to set up a sequence of original BAM -> samtools convert to SAM -> use awk/head/whatever to get the required lines -> convert to the destination file format.

But the scripts I wrote seem easier than that.

lh3:
Not sure about what is the goal of this discussion. Anyway, if you want to pull a random set of reads, the recommended approach is (20% in this example):
```sh
samtools view -bs.2 in.bam > sampled.bam
```
This keeps pairing regardless of the sorting order of `in.bam`. It is nontrivial to do this with awk/etc if `in.bam` is coordinate sorted.

dcolligan:
@lh3 The goal of this discussion is to determine whether the scripts I wrote in this PR are necessary, or if samtools has this functionality built in.  What we want is to pull the first N alignments from a SAM or BAM file (where N usually is a small number like 10).  Based on my reading of the docs, it seems like there's currently no way to do this with samtools.  Grabbing a percentage of the alignments, per your example, doesn't guarantee a small number of reads if the file is large (we want small files for testing).

diekhans:
Danny Colligan <notifications@github.com> writes:
> But the scripts I wrote seem easier than that.

sounds like a great approach.

hannes-ucsc:
I think @dcolligan's approach is fine. Under the hood pysam is using the same library that the samtools command line utility is using. The difference is that with pysam, you don't get the overhead of  fork()/exec() another program, and, more importantly, that pysam allows you iterate over exactly the number of desired lines and then stop. The command line interface has no such functionality and to emulate it would require some mildly complicated heuristics for guessing a percentage given the desired  number of reads. So +1 from me.

lh3:
I still don't know what this test file is used for, but if you don't care about pairing or randomness, it is fine.

dcolligan:
@jeromekelleher sounds like your objections have been cleared up... can I get a +1 from you?

jeromekelleher:
OK, great, thanks for verifying this. I'm +1 here too.

@lh3, we basically just want a some small SAM/BAM files that we can include in the git repo for data-driven testing. They don't need to have any special properties, they just need to exercise our code and allow us to test it on real examples.

Issue #247, opened 2015-03-13
getReadAlignments method required


jeromekelleher:
The `datamodel.reads.ReadGroup` class needs a method called `getReadAlignments`. This should be directly analogous to the  `datamodel.variants.VariantSet.getVariants` method, and map directly to the `SearchReadsRequest` protocol method. Specifically, we should have
```python
getReadAlignments(self, referenceName, referenceId, start, end):
     # Get the corresponding reads from PySAM, convert to GA4GH objects and yield each one in turn 
     # to the calling function.
```

This should be tested in `datadriven/test_reads.py`. We should test for searching over all reads, and subsets of the reads within the file. Ideally, we would like to use a different library to provide an independent code path, but we might have to use pysam for the tests also. (For the VCF testing, we use pysam as our production code and pyvcf as an independent test, making sure that (a) we're not reusing the same code in our production code and test code; and (b) having an independent sanity check on our usage of pysam.)

macieksmuga:
@jeromekelleher sorry for asking this late, but why both the `referenceName` and `referenceId` arguments? From the implementation, it looks like at most one can be passed in, otherwise an exception is raised. This seems like a strange convention to me (not followed anywhere else in the codebase, as far as I can tell). Since `referenceName` can be computed internally from `referenceId`, is there some intended use of this method that would necessitate passing in the former as opposed to latter argument?

jeromekelleher:
Don't ask me @macieksmuga, I have no idea! This is coming from the schemas, which were written before I started working on the project.

pgrosu:
Maciek, Jerome - it already got removed in the following PR this past December, and only `referenceId` now remains:

https://github.com/ga4gh/schemas/pull/193

~p

macieksmuga:
thanks @pgrosu! The discussion in that PR provides the answer I was looking for. Will comment with this on the PR to be merged here...

pgrosu:
Gladly @macieksmuga - we're all helping each other :)

dcolligan:
Fixed in #249

PR #249, opened 2015-03-13
implement ReadGroup.getReadAlignments

dcolligan:
Issue #247


ga4gh/datamodel/reads.py
```diff
@@ -75,3 +76,80 @@ def toProtocolElement(self):
         readGroup.referenceSetId = None
         readGroup.sampleId = None
         return readGroup
+
+    def getReadAlignments(self, referenceName, referenceId, start, end):
+        """
+        Returns an iterator over the specified reads
+        """
+        # TODO use an index if it is found (accompanying .bai file)
```
jeromekelleher: We should mandate the existence of an index file.

ga4gh/datamodel/reads.py
```diff
@@ -75,3 +76,80 @@ def toProtocolElement(self):
...
+        # this will allow use of fetch / pileup / count methods
+        if referenceName is not None and referenceId is not None:
+            raise exceptions.BadReadsSearchRequestBothRefs()
+        elif referenceName is not None:
+            for readAlignment in self._samFile:
```
jeromekelleher: This is too slow, we have to use an index to the reads we need.

ga4gh/datamodel/reads.py
pgrosu: How big will this index file get and will it slow down with more files?

tests/datadriven/test_reads.py
```diff
@@ -5,10 +5,15 @@
...
+
 import ga4gh.protocol as protocol
+import ga4gh.exceptions as exceptions
 import ga4gh.datamodel.reads as reads
 import tests.datadriven as datadriven
 
+import pysam
+
 
 def testVariantSets():
```
jeromekelleher: testReadGroupSets

tests/datadriven/test_reads.py
```diff
@@ -24,6 +29,76 @@ def getDataModelClass(self):
...
+
+    def testGetReadAlignmentsByRef(self):
+        # test that getReadAlignments works when querying by either
+        # referenceName or referenceId
+        readGroupSet = self._gaObject
+        for readGroup in readGroupSet._readGroups:
+            self.resetSamFile(readGroup)
+            refNames = collections.defaultdict(int)
+            refIds = collections.defaultdict(int)
+            for read in readGroup._samFile:
```
jeromekelleher: The gaObjects should be treated as black boxes here. We'll need to open an independent SamFile for each of the readgroups here in this class. See the distinction between pyVCF and GA variants in the `test_variants` module. Ideally we'd like to use another library to get at the contents of the alignment files, but this probably isn't feasible.

jeromekelleher:
This is a good start, thanks @dcolligan. The main things I'd like to see changed are using indexes for the production code, and making the test code treat the GAObjects as black boxes (no reaching in and using the underlying SamFiles). One thing we might do is use the code that you currently have in the test module as an independent implementation of the searching code we need. This would be a good overall test I think, as we have a nice sanity check on how we're using the pysam indexes.

ga4gh/datamodel/reads.py
dcolligan: Someone correct me if I am wrong, but I don't think SAM files can use indexes.  Do we not want to support SAM files or BAM files without indexes?

jeromekelleher: Ah, I didn't realise this. We'll just have to support only BAMs in that case (which is fine, less to worry about!). We can't support stuff without indexes, as these will be huge files in general. There's no way we can support retrieval by iteration and filtering, it's just not an option.

dcolligan:

ga4gh/datamodel/reads.py
```diff
@@ -75,3 +89,59 @@ def toProtocolElement(self):
...
+        """
+        Returns an iterator over the specified reads
+        """
+        if referenceName is not None and referenceId is not None:
+            raise exceptions.BadReadsSearchRequestBothRefs()
+        if referenceId is not None:
+            referenceName = self._samFile.getrname(referenceId)
+        try:
+            readAlignments = self._samFile.fetch(referenceName, start, end)
+        except ValueError as exception:
```
jeromekelleher: I'm not sure that this `try` block will work as expected. I think `readAlignments` is an iterator, so we may well throw the low-level errors from the for-loop below, when we actually do the iteration. Dealing with errors coming from pysam/htslib is going to be a challenge, so it might be better to just leave this try-except out for now.

ga4gh/datamodel/reads.py
```diff
@@ -75,3 +89,59 @@ def toProtocolElement(self):
...
+        ret.alignment.position.referenceName = self._samFile.getrname(
+            read.reference_id)
+        ret.alignment.position.position = read.reference_start
+        ret.alignment.cigar = []
+        for operation, length in read.cigar:
+            gaCigarUnit = protocol.GACigarUnit()
+            gaCigarUnit.operation = self._reverseCigarMap[operation]
+            gaCigarUnit.operationLength = length
+            ret.alignment.cigar.append(gaCigarUnit)
+        ret.duplicateFragment = False
```
jeromekelleher: Can we get some TODOs on the values that have been left out, so we know what's correct and what's not been implemented yet?

ga4gh/datamodel/reads.py
```diff
@@ -75,3 +89,59 @@ def toProtocolElement(self):
         readGroup.referenceSetId = None
         readGroup.sampleId = None
         return readGroup
+
+    def getReadAlignments(self, referenceName, referenceId, start, end):
```
jeromekelleher: Maybe we should give these parameters defaults of `None` to make the testing code a little easier?

tests/datadriven/test_reads.py
```diff
@@ -24,7 +59,58 @@ def getDataModelClass(self):
...
+            for readGroup in readGroupSet._readGroups:
+                list(readGroup.getReadAlignments("a", 5, 0, None))
+
+    def testGetReadAlignments(self):
+        # test that searching with no arguments succeeds
+        readGroupSet = self._gaObject
+        for i, readGroup in enumerate(readGroupSet._readGroups):
+            alignments = list(readGroup.getReadAlignments(
+                None, None, None, None))
+            for j, alignment in enumerate(alignments):
```
jeromekelleher: If we set the default args for getReadAlignments, we could write
```python
for j, alignment in enumerate(readGroup.getReadAlignments()):
    # etc
```

tests/datadriven/test_reads.py
```diff
@@ -24,7 +59,58 @@ def getDataModelClass(self):
     def getProtocolClass(self):
         return protocol.GAReadGroupSet
 
-    def testFixMe(self):
+    def testGetReadAlignmentsBothRefs(self):
+        # test that querying by both referenceName and referenceId fails
+        with self.assertRaises(exceptions.BadReadsSearchRequestBothRefs):
+            readGroupSet = self._gaObject
+            for readGroup in readGroupSet._readGroups:
```
jeromekelleher: We should define a method `ReadGrouoSet.getReadGroups` rather than accessing a protected variable.

tests/datadriven/test_reads.py
```diff
@@ -24,7 +59,58 @@ def getDataModelClass(self):
...
+                list(readGroup.getReadAlignments("a", 5, 0, None))
+
+    def testGetReadAlignments(self):
+        # test that searching with no arguments succeeds
+        readGroupSet = self._gaObject
+        for i, readGroup in enumerate(readGroupSet._readGroups):
```
jeromekelleher: I'm slightly uneasy about relying on the ordering of the readGroups here. Perhaps we could just keep a dictionary mapping readGroupId to the list of reads, rather than two linked arrays (i.e. _readGroups and _reads)? That is, we have a dictionary self._reads which maps readGroupId to the reads in the SamFile.

tests/datadriven/test_reads.py
```diff
@@ -24,7 +59,58 @@ def getDataModelClass(self):
...
+                otherAlignment = self._reads[i][j]
+                self.assertAlignmentsEqual(alignment, otherAlignment)
+
+    def testGetReadAlignmentsRefName(self):
+        # test that searching with a reference name succeeds
         readGroupSet = self._gaObject
-        # TODO
-        self.assertIsNotNone(readGroupSet)
+        for i, readGroup in enumerate(readGroupSet._readGroups):
+            refNames = self._refNames[i]
```
jeromekelleher: A similar point for `self._refNames` depending on the ordering of the readGroups. Maybe a simple local class to gather together all the information we need about a SamFile together into one place would help here?

jeromekelleher:
This is great, thank @dcolligan! I think we need to work a bit more on the tests, but after that we should be good to merge.

Some general points on the tests:

- You are using a lot of low-level operations to get the reads/readgroups that we need. I think it would help a lot if we abstracted this up a bit and made a simple API to allow us get the reads that we need when we're testing.
- GAObjects should be black boxes, and their interface boundaries respected. No direct access to _protected variables --- add a getter function when you need to access these.

We need to do more checks on the alignment objects themselves. For a start, they need to be valid:
```python
def testAlignmentsValid(self):
    for gaAlignment in self.getAllGaAlignments():
        self.assertTrue(gaAlignment.validate(gaAlignment.toJsonDict()))
```

We need to verify that the alignment objects are equal, as well as checking that the length of the lists are the same.

We should verify the range searching mechanisms --- do we return exactly the same alignments, when we search over a variety of ranges? (0 to min +1, min to min + 1, min to median, median to max, max-1 to max, etc, etc.) Is the query empty when we search outside the range? (0 to min - 1, max + 1 to 2 * max, etc, etc). We should implement the search manually in the test cases, using the iteration based code that you previously used for SAM files.

tests/datadriven/test_reads.py
```diff
@@ -24,7 +59,58 @@ def getDataModelClass(self):
...
+            refName = refNames.keys()[0]
+            refCount = refNames[refName]
+            bigNumThatPysamWontChokeOn = 2**30
+            alignments = list(readGroup.getReadAlignments(
+                refName, None, 0, bigNumThatPysamWontChokeOn))
+            self.assertEqual(len(alignments), refCount)
+
+    def assertAlignmentsEqual(self, readOne, readTwo):
```
jeromekelleher: It would be good to be a bit clearer (what's readOne and readTwo?). I think gaRead and pysamRead is the most unambiguous way of stating it. Also, we'll want a lot more asserts here --- we should check every attribute on the GA read to make sure it is correct.

ga4gh/datamodel/reads.py
dcolligan: It's basically any field that is None or False (or otherwise has a default value from the protocol object).  Should I just not list those fields here?

jeromekelleher: I don't mind - anything that makes it easy to see what is done and what needs work. Maybe a '# TODO' at the end of each line?

dcolligan:
@jeromekelleher pushed new commit

tests/datadriven/test_reads.py
```diff
@@ -24,7 +64,133 @@ def getDataModelClass(self):
...
+            for readGroup in readGroupSet.getReadGroups():
+                list(readGroup.getReadAlignments("a", 5, 0))
+
+    def testGetReadAlignments(self):
+        # test that searching with no arguments succeeds
+        readGroupSet = self._gaObject
+        for readGroup in readGroupSet.getReadGroups():
+            alignments = list(readGroup.getReadAlignments())
+            readGroupInfo = self._readGroupInfos[readGroup.getSamFilePath()]
+            for i, gaAlignment in enumerate(alignments):
```
jeromekelleher: We could also use
```python
gaAlignment, pysamAlignment in zip(alignments, readGroupInfo.reads):
```
this is usually regarded as a bit more Pythonic that using an index variable. I don't feel strongly about it though, this code is still very clear.

jeromekelleher:
This is great @dcolligan, it looks like really solid code. The updates you've made are excellent and it's all very clear and readable. There's still a couple of things I'd like to see:

1. testValidateReads() - push all reads through validate. I think this is failing at the moment.
2. testSearchRanges() - systematically check the search on start and end coords, using the sorts of ranges I went through in the last set of comments. 
3. Everywhere you're asserting the number of alignments being equal to the expected number of alignments, I think we should also verify that the actual alignments are equal.

Once this is merged, we should file an issue to make some unit tests for the Sam utilities (SamFlags and Cigar) also.

dcolligan:
@jeromekelleher 
"testSearchRanges() - systematically check the search on start and end coords, using the sorts of ranges I went through in the last set of comments"
Doesn't this have us assume something about the data?  And isn't that bad?

jeromekelleher:
Well, (I think) we can find the minimum and maximum positions when we're reading in the reads, so we're just discovering this information. I guess in practice we'll be assuming a reasonable distance between max and min, but we can degrade gracefully to the case of 1 read if we think it's important.

ga4gh/converters.py
```diff
@@ -157,27 +143,35 @@ def toAlignedSegment(cls, read, targetIds):
     def toSamFlag(cls, read):
         flag = 0
         if read.numberReads:
-            flag += 0x1
+            reads.SamFlags.setFlag(
+                flag, reads.SamFlags.NUMBER_READS)
```
macieksmuga: Thank you! Code now human-readable :)

dcolligan:
@jeromekelleher pushed a new commit

The only issue I see is that the pysam `fetch` call doesn't seem to return a predictable number of results based on just the position attribute alone, so I don't know if it's going to be possible/easy to write a non-depending-significantly-on-data datadriven test to test the search ranges.

http://pysam.readthedocs.org/en/latest/usage.html#fetching-reads-mapped-to-a-region

"pysam.AlignmentFile.fetch() returns all reads overlapping a region sorted by the first aligned base in the reference sequence. Note that it will also return reads that are only partially overlapping with the region. Thus the reads returned might span a region that is larger than the one queried."

tests/datadriven/test_reads.py
```diff
@@ -5,18 +5,58 @@
...
+        def __init__(self, readGroupSetId, samFileName):
+            filename = os.path.split(samFileName)[1]
+            localId = os.path.splitext(filename)[0]
+            readGroupId = "{}:{}".format(readGroupSetId, localId)
+            self.id = readGroupId
+            self.samFile = pysam.AlignmentFile(samFileName)
+            self.reads = []
+            self.refNames = collections.defaultdict(list)
+            self.refIds = collections.defaultdict(list)
+            for read in self.samFile:
```
jeromekelleher: Can't we read the positions here, and work out what max and min are?

jeromekelleher:
I'm missing something here --- why do we care about `fetch` for this?

dcolligan:
Because `getReadAlignments` basically just passes arguments to pysam's `fetch`

jeromekelleher:
So is this for our test code or our production code? For our test code surely we can just filter the reads by hand?

dcolligan:
Huh?

My point is that I can't see how we are able to, given an arbitrary BAM file, do any kind of meaningful range testing (using `getReadAlignments` which calls into pysam's `fetch`).

`fetch(refName, start, end)` seems like it should return all alignments in refName with POS between start (inclusive) and end (exclusive).  But the tests indicate that is not how it works, and I'm unsure how it does at this point.  I suspect that it has something to do with "partially overlapping" reads that the pysam docs mention, but I'm not sure what those are.

jeromekelleher:
Well that sounds pretty serious - if we can't use pysam.fetch to get the reads we need then we can't implement the protocol. Lemme have a look.

dcolligan:
Er, I think now "partially overlapping" means if any point in the range of `start` and `end` is in the the range of read's `position` and `position + len(alignedSequence)`.  The tests now pass in the update I just pushed.

But the point remains that these tests are very brittle and data-dependent.  These tests will fail if the first and last reads position are sufficiently close together.  My question to you is: how could we come up with a testing strategy that reliably tested the `fetch` method's ranges but didn't depend on specific aspects of data in the BAM files that we are using?

jeromekelleher:
Phew, that's a relief! OK, I think we should merge this now-ish and file issues on anything else we see. This is great work, thanks @dcolligan. @macieksmuga, can you review this also please --- this is important stuff and we need to get more eyes on it.

WRT to the brittleness, I don't think it's an issue in practice to be honest. Real BAM files contain zillions of reads over huge intervals. We can make the tests more robust if/when we hit issues as we add more test data. We have to assume that `fetch` works correctly I think, we're really just sanity testing our own usage of it here.

+1

macieksmuga:
@jeromekelleher @dcolligan will review today.

ga4gh/datamodel/reads.py
```diff
@@ -13,6 +13,68 @@
...
+        8: protocol.GACigarOperation.SEQUENCE_MISMATCH,
+    }
+
+    @classmethod
+    def ga2int(cls, value):
+        return cls.cigarMap[value]
+
+    @classmethod
+    def int2ga(cls, value):
+        return cls.reverseCigarMap[value]
```
macieksmuga: I know it's nitpicking, but why not bake the inverse relationship of cigarMap and reverseCigarMap into the code
via lazy initialization, as follows:
```
reverseCigarMap = None
[...]
def int2ga(cls, value):
    if cls.reverseCigarMap is None:
         cls.reverseCigarMap = {v: k for k, v in cls.cigarMap.items()}
    return cls.reverseCigarMap[value]
```

ga4gh/datamodel/reads.py
```diff
@@ -23,14 +85,21 @@ def __init__(self, id_, dataDir):
...
-                filename = os.path.split(path)[1]
-                localId = filename.split(".")[0]
-                readGroupId = "{}:{}".format(self._id, localId)
-                readGroup = ReadGroup(readGroupId, path)
-                self._readGroups.append(readGroup)
+        # we only support BAM files right now;
+        # SAM files (which can't have index files) would be too slow
+        pattern = "*.bam"
+        for path in glob.glob(os.path.join(self._dataDir, pattern)):
+            filename = os.path.split(path)[1]
```
macieksmuga: I personally prefer `os.path.basename(path)` unless the other part of the split will be used as well.

tests/datadriven/test_reads.py
```diff
@@ -5,18 +5,58 @@
...
+                refId = read.reference_id
+                self.refIds[refId].append(read)
+                self.reads.append(read)
+
+    def __init__(self, readGroupSetId, baseDir):
+        super(ReadGroupSetTest, self).__init__(readGroupSetId, baseDir)
+        self._readGroupInfos = {}
+        for samFileName in glob.glob(
+                os.path.join(self._dataDir, "*.bam")):
+            self._readSam(readGroupSetId, samFileName)
```
macieksmuga: why the private method call here? Wouldn't this single statement be equivalent and simpler?
```
self._readGroupInfos[samFileName] = self.ReadGroupInfo(
    readGroupSetId, samFileName)
```

macieksmuga:
Very nice! Thank you, Danny :)

Reading over the `pysam` `fetch` documentation, I don't think it's creating a problem for testing: We can (should!) curate the data we're testing with to ensure we have a pretty representative set (and to cover any cases that we may have experienced problems with during development for regression testing), and the function will always return any reads that (partly or fully) overlap the specified region.

There's also the `pileup()` function in `pysam` if we need a column slice of all the reads that align on a particular base within a specified range - this may be closer to the kind of overlap guarantee that @dcolligan was referring to earlier, but I don't know if any of the ga4gh server functionality requires this.

+1

ga4gh/datamodel/reads.py
dcolligan: Because there's no guarantee those patterns will be future-compatible with changes to the spec.  I'd rather encode things explicitly.

tests/datadriven/test_reads.py
dcolligan: arguably, but I was trying to follow the pattern established in the variant equivalent test

dcolligan:
Pushed a new commit with @macieksmuga 's suggestions

ga4gh/datamodel/reads.py
macieksmuga: Huh? We're making a lot of assumptions that may not be future-compatible with changes to the spec (think all those GA prefixes :)) Postulating that a 1-1 invertible map stays that way seems about as safe an assumption to make as can be, and not hand-coding both directions seems like a less error-prone approach in general.

jeromekelleher: I'm probably with @macieksmuga here --- we can assume the map will remain a bijection I think!

ga4gh/datamodel/reads.py
```diff
@@ -76,3 +158,66 @@ def toProtocolElement(self):
         readGroup.referenceSetId = None
         readGroup.sampleId = None
         return readGroup
+
+    def getReadAlignments(self, referenceName=None, referenceId=None,
```
macieksmuga: See https://github.com/ga4gh/schemas/pull/193 for a revision to this API that strips the referenceName argument. While both `referenceName` and `referenceId` continue to be supported in the v0.5.1 version of the API (thus I'm not suggesting we change this here), we should also not spend too much time supporting the case where both arguments are supplied, as it's going to go away soon.

dcolligan:
Pushed new commit to address @macieksmuga and @jeromekelleher 's concerns

Issue #250, opened 2015-03-16
tests/unit/test_end_to_end.py should use SimulatedBackend.


jeromekelleher:
As part of the process of removing the dependency on wormtable, we need to remove the usage of the WormtableTestFixture.

dcolligan:
Fixed in #254

Issue #188, opened 2015-02-25
Rename test_end_to_end.py


dcolligan:
Now that we have an end_to_end test bucket, test_end_to_end.py seems like it should be renamed.  Suggestions?

dcolligan:
Fixed in #254

PR #254, opened 2015-03-16
Improved testing for simulated backends.

jeromekelleher:
- Adds an extraConfig dictionary to `configure` to allow us to easily
  configure the tests.
- Refactored test_views to get rid of malformed requests, and set up
  full request validation.
- Renamed test_end_to_end to test_simulated_stack and removed the
  wormtable dependency there.

Fixes issues #250 and #188.

### DIFF
--- a/tests/unit/test_simulated_stack.py
+++ b/tests/unit/test_simulated_stack.py
@@ -1,5 +1,5 @@
 """
-End-to-end tests for each backend configuration. Sets up a server with
+End-to-end tests for the simulator configuration. Sets up a server with
 the backend, sends some basic queries to that server and verifies results
 are as expected.
 """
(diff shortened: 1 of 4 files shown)


tests/unit/test_simulated_stack.py
```diff
@@ -8,41 +8,55 @@
 from __future__ import unicode_literals
 
 import unittest
+from nose.tools import nottest  # TMP until we fix Calls
```
dcolligan: TMP = temporary?

tests/unit/test_simulated_stack.py
```diff
@@ -1,5 +1,5 @@
 """
```
dcolligan: Yay on file rename.  This will close #188

dcolligan:
Great job.  +1

tests/unit/test_simulated_stack.py
jeromekelleher: I've changed this to a TODO.

Issue #253, opened 2015-03-16
Silence spam output from noisy tests


dcolligan:
A recently added test is spewing out extraneous debug output.  This shouldn't happen -- silence this test!

```
$ nosetests -v tests/datadriven/
...
tests.datadriven.test_variants.VariantSetTest.testVariantSetMetadata:1kgPhase1 ... ok
[W::_vcf_parse_format] FORMAT 'GT' is not defined in the header, assuming Type=String
[W::_vcf_parse_format] FORMAT 'GT' is not defined in the header, assuming Type=String
[W::_vcf_parse_format] FORMAT 'GT' is not defined in the header, assuming Type=String
tests.datadriven.VariantSetTest.testProtocolElementValid:1kgPhase3 ... ok
...
```

One way to do this might be creating a `@SuppressOutput` decorator in `tests.utils` that redirects stdout and stderr to `/dev/null` for the running duration of the test and then restores the file handles at the end.

dcolligan:
Alternatively, fix the data file (looks like one of the 1kg VCF files has a malformed header line)

jeromekelleher:
These warnings are coming from htslib, which would be nice to suppress in general; is there a way to silence it through pysam?

dcolligan:
The error looks like it comes from this line in htslib:

https://github.com/samtools/htslib/blob/f3e1602196bbf03f426dfb363a4841932a042194/vcf.c#L1522

It looks like it's writing it directly to stderr without going through any kind of logging subsystem.  So there are probably no hooks in pysam or htslib to redirect / silence.

dcolligan:
@jeromekelleher do you know what line in our code is accessing the htslib code that is emitting the error?

dcolligan:
Nevermind, seems to be `DataDrivenTest.__init__` where `self._gaObject` is initialized

dcolligan:
Redirecting stdout and stderr doesn't seem to do it... this diff still emits the output...
```
     def __init__(self, setId, baseDir):
         self._setId = setId
         self._dataDir = os.path.join(baseDir, setId)
+        origStdout = sys.stdout
+        origStderr = sys.stderr
+        devnull = open(os.devnull, 'w')
+        sys.stdout = devnull
+        sys.stderr = devnull
         self._gaObject = self.getDataModelClass()(
             self._setId, self._dataDir)
+        sys.stdout = origStdout
+        sys.stderr = origStderr
```
[...]

PR #255, opened 2015-03-16
get rid of spam error lines in htslib-using test

dcolligan:
- Add SuppressOutput decorator

Issue #253

### DIFF
--- a/ga4gh/datamodel/__init__.py
+++ b/ga4gh/datamodel/__init__.py
@@ -5,3 +5,52 @@ data in existing formats into GA4GH protocol types.
 from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
+
+import contextlib
+import os
+import sys
+
+
+class DataModelObject(object):
+    """
+    Base class for all data model object types
+    """
+    def __init__(self, lowLevelOutputSuppression=True):
+        self.lowLevelOutputSuppression = lowLevelOutputSuppression
+
+    def setLowLevelOutputSuppression(self, lowLevelOutputSuppression):
+        self.lowLevelOutputSuppression = lowLevelOutputSuppression
+
+    @contextlib.contextmanager
+    def suppressLowLevelOutput(self):
+        # I would like to use sys.stdout.fileno() and sys.stderr.fileno()
+        # here instead of literal fd numbers, but nose does something like
+        # sys.stdout = StringIO.StringIO() when the -s flag is not enabled
+        # (to capture test output so it doesn't get entangled with nose's
+        # display) so the sys.stdout and sys.stderr objects are not able to
+        # be accessed here.
+        if self.lowLevelOutputSuppression:
+            try:
+                # redirect stdout and stderr to /dev/null
+                devnull = open(os.devnull, 'w')
+                origStdoutFd = 1
+                origStderrFd = 2
+                origStdout = os.dup(origStdoutFd)
+                origStderr = os.dup(origStderrFd)
+                sys.stdout.flush()
+                sys.stderr.flush()
+                os.dup2(devnull.fileno(), origStdoutFd)
+                os.dup2(devnull.fileno(), origStderrFd)
+                # enter the wrapped code
+                yield
+            finally:
+                # restore original file streams
+                devnull.flush()
+                os.dup2(origStdout, origStdoutFd)
+                os.dup2(origStderr, origStderrFd)
+                # clean up
+                os.close(origStdout)
+                os.close(origStderr)
+                devnull.close()
+        else:
+            yield
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -12,14 +12,16 @@ import glob
 
 import pysam
 
+import ga4gh.datamodel as datamodel
 import ga4gh.protocol as protocol
 
 
-class ReadGroupSet(object):
+class ReadGroupSet(datamodel.DataModelObject):
     """
     Class representing a logical collection ReadGroups.
     """
     def __init__(self, id_, dataDir):
+        super(ReadGroupSet, self).__init__()
         self._id = id_
         self._dataDir = dataDir
         self._readGroups = []
--- a/ga4gh/datamodel/references.py
+++ b/ga4gh/datamodel/references.py
@@ -11,16 +11,18 @@ import glob
 
 import pysam
 
+import ga4gh.datamodel as datamodel
 import ga4gh.protocol as protocol
 
 
-class ReferenceSet(object):
+class ReferenceSet(datamodel.DataModelObject):
     """
     Class representing ReferenceSets. A ReferenceSet is a set of
     References which typically comprise a reference assembly, such as
     GRCh38.
     """
     def __init__(self, id_, dataDir):
+        super(ReferenceSet, self).__init__()
         self._id = id_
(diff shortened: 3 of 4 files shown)


jeromekelleher:
This is good @dcolligan; if it works, and it's portable, I'd like to include something like this. (I've tried this sort of fiddling with fds in Python before and given up, so I'm a bit skeptical!)

Two main comments:

1. I think we'll want to do this more generally, so the method should be included in the main trunk code. We don't want htslib writing cryptic stuff to /var/log/apache/error.log, so we need a method of suppressing it. My favoured approach would be to file a feature request with htslib to make warnings configurable, but this would take a long to time filter through to us.
2. I think a context manager would be more suitable than a decorator here. What happens if there's an error during this operation? I.e., something like
```python
with suppressLowLevelOutput(stderr=True):
    varFile = pysam.VariantFile(...)
```
That is, we make the suppression  as localised as possible, and for as short as possible. We need to be _very_ paranoid about error propagation, closing fds, etc also.

So, this is very useful, but tricky. Perhaps we could come back to this later?

dcolligan:
(1) is solved by making a ga4gh wrapper module for pysam (or whatever other library we want to silence) where every call to the library goes through the suppression logic before being passed to/from pysam.

(2) is solved with a try/finally block, which I absolutely should put in.

jeromekelleher:
> (1) is solved by making a ga4gh wrapper module for pysam (or whatever other library we want to silence) where every call to the library goes through the suppression logic before being passed to/from pysam.

Overkill! There's a limited number of places where this will happen - we don't need to wrap every single call going into pysam. This is unimportant corner-case behaviour, we don't need loads to code to deal with it.

dcolligan:
So suppose we do want this decorator in the main trunk and not just in tests... where should it go?  Perhaps in a (gasp) decorators module?

jmarshall:
Have you considered talking to the htslib people about messages that you think are particularly spammy?  They may agree and just remove them...

dcolligan:
It's not really a question of good message / bad spammy message.  It's that they emit things to stdout or stderr at all.  Every call to fprintf would need to be changed into some kind of logging wrapper:

https://github.com/samtools/htslib/search?utf8=%E2%9C%93&q=fprintf

jmarshall:
Me, I don't disagree in general, but practicality trumps.

bgzip.c, htsfile.c, tabix.c aren't part of the library, so printfs there don't count.

In theory, adding `hts_verbose = 0;` to the start of your client program should shut up many of the printfs in the library, especially those that aren't reporting errors.  Where it doesn't, that should be reported as a bug.

diekhans:
If you cat shut up the library file ticket.   I know the htslib people we can get it fixed



... Sent from my computer phone

-----Original Message-----

dcolligan:
We really only use htslib via pysam, so this is really a question of whether pysam provides hooks to set `hts_verbose = 0`.  I'm not sure if it does, but so far I haven't found anything like that...

dcolligan:
Reading the pysam docs, this seems like a possible road that we could go down:

http://pysam.readthedocs.org/en/latest/usage.html#extending-pysam

But that seems like a hack.  It would be much better for there to just be some pysam call to set the verbosity level of htslib (if there isn't one already...).

jeromekelleher:
I disagree with the decorator pattern here @dcolligan, I think it would be better as a context manager. That way we can keep the suppression right next to the emitting code, and it will be absolutely clear what's happening. Also, we can make it configurable, so we can turn on-and-off suppression when we want to see what the warnings actually are (sometimes they _will_ be useful). This isn't that easy using a decorator.

Anyway, this isn't that important, and I don't think it's a priority.

Thanks for the input @jmarshall. There's actually  a wider issue here for how htslib reports errors: it's not hugely useful to us for an error message to be written to stderr. Is there any possibility of htslib returning error codes from key functions,  and having a `hts_strerror` like function?

diekhans:
Hstlib deriver from samtools has many error handling issues.  Last I looked it doesn't check malloc returns.  The need pull requests



... Sent from my computer phone

-----Original Message-----

dcolligan:
@jeromekelleher Configurability isn't an argument for either a contextmanager or decorator, I don't think.  In either event, there should probably be some flag in the `BaseConfig` like `SUPPRESS_LIB_OUTPUT = True` or something.  Neither is being "right next to" the emitting code an argument that I find convincing...  as the code in the PR shows, the decorator is right next to the emitting code.  Granted, this needed to be wrapped in its own method... is that what you are getting at?

In any event, if you want to have this functionality available outside of tests, which module should we put it in?

I disagree that this isn't a priority.  We have spam polluting our test output right now.  And this code addresses a concern that is more general than the particulars of pysam / htslib.

jeromekelleher:
```python
def _addFile(self, filename):
    with self.suppressLowLevelOutput():
        varFile = pysam.VariantFile(...)
```
`suppressLowLevelOutput()` is a method of `DataModelObject`, which is defined in `datamodel.__init__.py`. This should be the superclass of AbstractVariantSet, and have an instance variable and setters for `lowLevelOutputSuppression`, which allow us to turn on and off suppression.

If we use a decorator, we'd have to make a one-line method for each time we use it, e.g.

```python
def _addFile(self, filename):
   varFile = self._openVariantFile()

@whateverthedecoratoriscalled
def _openVariantFile(self)
    return  pysam.VariantFile(...)
```
This seems obscure and easily misunderstood to me.  It's not entirely clear how we tell the decorator when we should and shouldn't suppress also.

This situation seems to be ideally suited to the context manager paradigm.

pgrosu:
Jerome, by embedding checks in the code, I feel you will spend more time supporting it in the future than extending it.  It would be better to make it more modular (i.e. plugin-like) so that any post-processing happens on-the-fly on a per-needed basis.  And I agree with Mark that utilizing as much from other implementations (i.e. htslib) would minimize re-inventing the wheel.

dcolligan:
@jeromekelleher should we have a variable in the `serverconfig` for this?  If so, it seems like we're going to be passing configuration down through several layers... cli -> frontend -> backend -> datamodel.  We don't do this now, so this is going to be changing a lot of method signatures.  

Which gets me thinking if this is really the right model.  Do we really want the datamodel objects to be generally configurable?  To me, they are just container classes for state (plus functionality for translating file system objects to that state).  It seems like any configuration should be on the frontend/backend side of things.

I've pushed a commit on which I'm sure you'll have some comments.

jeromekelleher:
Do we have to look at this now @dcolligan? There's lots of stuff we do actually need to do _now_, and this debate isn't getting any of that stuff done. I've pushed a PR that makes the warnings go away, so maybe we can just forget about this?

jeromekelleher:
Update - I realise this is exactly what I asked for, but I'm still not convinced this fd dup will actually work portably.

dcolligan:
As far as I am concerned, not spamming the test run output is something we need to do now as well.

If you have fixed the issue in another pull request, we can put this PR off until later.

lh3:
Htslib only has a few global variables, one of which is `hts_verbose`, equal to 3 by default. It is supposed to suppress information, warnings and even error messages at value 0. If it does not work for some stderr info, you should let htslib developers know.

tk2:
Noted. Please send us pull requests (if you know the offending lines and fix) or issues if you need us to investigate the behaviour in htslib.

PR #257, opened 2015-03-17
Initial outline of protocol error testing.

jeromekelleher:
- Adds a simple script to decode error codes.
- Adds tests for bad page sizes, and adds a configuration key for
  default page size.
- Improves request validation error messages

### DIFF
--- a/convert_error_code.py
+++ b/convert_error_code.py
@@ -1,0 +1,33 @@
+"""
+Simple script to decode exception error codes. This translates
+the error code received by clients into an exception class.
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+
+import ga4gh.exceptions as exceptions
+
+
+def parseArgs():
+    parser = argparse.ArgumentParser(
+        description=(
+            "Converts an error code received by a clients to the  "
+            "corresponding exception class."))
+    parser.add_argument(
+        "errorCode", type=int,
+        help="The errorCode value in a GAException object.")
+    args = parser.parse_args()
+    return args
+
+
+def main():
+    args = parseArgs()
+    exceptionClass = exceptions.getExceptionClass(args.errorCode)
+    print(args.errorCode, exceptionClass, sep="\t")
+
+
+if __name__ == '__main__':
+    main()
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -33,6 +33,7 @@ class AbstractBackend(object):
         self._callSetIds = []
         self._requestValidation = False
         self._responseValidation = False
+        self._defaultPageSize = 100
 
     def getVariantSets(self):
         """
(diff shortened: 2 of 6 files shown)


convert_error_code.py
```diff
@@ -0,0 +1,33 @@
+"""
```
dcolligan: Should this be in the base directory and not the `scripts` directory?

ga4gh/exceptions.py
```diff
@@ -6,11 +6,25 @@
...
 from __future__ import unicode_literals
 
+import sys
 import zlib
+import inspect
 
 import ga4gh.protocol as protocol
 
 
+def getExceptionClass(errorCode):
```
dcolligan: If this is only used for debugging purposes, why not put it in `convert_error_code.py`?

ga4gh/backend.py
```diff
@@ -33,6 +33,7 @@ def __init__(self):
         self._callSetIds = []
         self._requestValidation = False
         self._responseValidation = False
+        self._defaultPageSize = 100
```
dcolligan: shouldn't this be set by the config?

tests/unit/test_protocol_errors.py
```diff
@@ -0,0 +1,105 @@
...
+    _app = None
+
+
+class TestFrontendErrors(unittest.TestCase):
+    """
+    Tests the frontend for various errors that can occur and verify
+    that the correct exception was raised by the error code sent
+    back.
+    """
+
```
dcolligan: style: delete newline

tests/unit/test_protocol_errors.py
```diff
@@ -0,0 +1,105 @@
+"""
```
dcolligan: Good tests, but it strikes me that these should be in a compliance suite.  Of course, we don't have one yet, so this is fine for the time being, but once we get one, we should consider whether moving these there (or duplicating them in the client suite) would be the better option...

ga4gh/exceptions.py
jeromekelleher: Because it's poking around in the internal structures of the `exceptions` module, and this sort of thing shouldn't be done across file boundaries. It'll end up being used in other places as well anyway.

tests/unit/test_protocol_errors.py
jeromekelleher: We can't mandate the error codes for other implementations. These are specific to us.

convert_error_code.py
jeromekelleher: It has to be in the base directory if it imports ga4gh modules.

ga4gh/backend.py
jeromekelleher: It is - see the DEFAULT_PAGE_SIZE key. This is just to give it an initial value. We could set all these to None I suppose, but what if we want to test this class in isolation?

tests/unit/test_protocol_errors.py
jeromekelleher: Done.

tests/unit/test_protocol_errors.py
dcolligan: Ah, the error codes are not going to be part of the ga4gh protocol, I guess?  That gives us a lot more flexibility then... although from a specification standpoint I'm not sure if that's the best idea...

dcolligan:
+1

tests/unit/test_protocol_errors.py
jeromekelleher: Yeah, the error codes are completely up to the implementation to use as they see fit. In general, the specification of error handling in the protocol is pretty much non-existent (which is definitely not a good idea!).

PR #258, opened 2015-03-17
Updated benchmark to use new code layout.

jeromekelleher:
Also added some CLI parameters to allow us to easily compare different
variant set implementations.


dcolligan:
+1

PR #259, opened 2015-03-17
Moved genotype unit tests into new module.

jeromekelleher:
As we remove wormtable, we need to make sure that we retain all the tests that we have in tests/unit/test_backends.py in some form. This PR starts the process by moving the trivially seperable unit tests for genotype conversion into an independent module for variant unit tests.


dcolligan:
+1

PR #260, opened 2015-03-17
Replace samtools index runCommand with pysam call

dcolligan:
pysam provides the capability to call samtools methods directly...
we should use that instead of using shell commands!


jeromekelleher:
+1.

PR #261, opened 2015-03-17
Updated 1000 genomes VCFs.

jeromekelleher:
Subsetting the 1000 genomes VCFs so we only use a fraction of the samples. This makes test cases run much more quickly.

This also 'solves' #253, as we no longer get htslib complaining during tests.


dcolligan:
I'm +1 on this provided it passes tests.

dcolligan:
This is great... order of magnitude improvement in test time and no spam test output!

Issue #241, opened 2015-03-12
Paging test required


jeromekelleher:
As part of the drive to remove Wormtable (as detailed in #210) we need to replace the tests that are currently embedded in the `TestWormtableBackend` with more modular tests. One of the tests we require is a test of the correctness of paging over the different sets of objects. We should create a new class in `tests/unit/test_backend.py` called `TestPaging(unittest.TestCase)`, which exercises the paging code across all the supported endpoints. This should use a `SimulatedBackend` as its test fixture.

The idea then would be to run some standard queries against the endpoints on this simulated backend (searchVariantSets and searchVariants for now) for a systematic selection of page sizes, and to verify that we always get back precisely the same results.

dcolligan:
To replace TestWormtableBackend, it looks like we're going to have to substitute `.tbi` files for the wormtable files.  How does one create those files?  (Having a `README` inside `ga4gh-example-data` with how you generated those files would be useful.)

macieksmuga:
@dcolligan I'm looking into this now.

macieksmuga:
`.tbi` files are tabix indices on the VCFs as read by pysam.VariantFile, and generating them is described in the now removed tabix part of `README.txt` (perhaps we should bring this part back?):
```
The tabix backend allows us to serve variants from an arbitrary VCF file.  The
VCF file must first be indexed with `tabix
<http://samtools.sourceforge.net/tabix.shtml>`_.  Many projects, including the
`1000 genomes project
<http://ftp.1000genomes.ebi.ac.uk/vol1/ftp/release/20130502/>`_, release files
with tabix indices already precomputed. 
```
`ga4gh-example-data` currently comes with these index files precomputed.

jeromekelleher:
Generating Tabix indexes is pretty standard - you're right though, we'll need a proper discussion of how we deal with indexes in the README. We should add this as an issue.

macieksmuga:
Ok to close this now?

PR #263, opened 2015-03-19
Paging test moved from wormtable to simulated backend.

macieksmuga:
New `TestAbstractBackend` class introduced in `unit/test_backends.py`,
gathers together backend implementation independent logic,
`TestSimulatedBackend` and `TestWormtableBackend` classes inherit from it.
`TestPaging` is a new subclass of `TestSimulatedBackend`. This setup
should provide a clean migration path out of Wormtable for testing backend.

 For now, I could only pull `TestVariantSets.testPagination()` out of 
`TestWormtableBackend`, as `SimulatedBackend` only provides mock objects
for `VariantSet`s. As `SimulatedBackend` grows, the other tests in `test_backends.py`
can be pulled out as well.

I could technically pull out `TestVariantSets.testIds()` now as well, but this doesn't fit
the new `TestPaging` class's focus on pagination. 

Addresses issue #241

### DIFF
--- a/tests/unit/test_backends.py
+++ b/tests/unit/test_backends.py
@@ -1,7 +1,7 @@
 """
-Tests for the backend objects. We instantiate local copies of the backends
-and invoke the entry points for the protocol methods. We do not set up
-any server processes or communicate over sockets.
+Tests for the backend objects. We instantiate local copies of
+the backends and invoke the entry points for the protocol methods.
+We do not set up any server processes or communicate over sockets.
 """
 from __future__ import division
 from __future__ import print_function
@@ -15,6 +15,7 @@ import unittest
 import itertools
 import subprocess
 
+import pysam
 import wormtable as wt
 
 import tests.utils as utils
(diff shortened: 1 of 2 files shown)


tests/unit/test_backends.py
```diff
@@ -93,26 +93,10 @@ def convertInfoValue(column, value):
...
-                os.path.join(self._variantsDir, relativePath))
-            self._tables[relativePath] = table
-            self._chromIndexes[relativePath] = table.open_index("CHROM")
-            self._chromPosIndexes[relativePath] = table.open_index("CHROM+POS")
-        self._backend = backend.FileSystemBackend(self._dataDir)
-
-    def tearDown(self):
-        for table in self._tables.values():
-            table.close()
+        self._backend = backend.AbstractBackend()
```
jeromekelleher: I don't think there's much point in instantiating an AbstractBackend here, as this will get over-written in the subclass anyway. We can just get rid of `setUp` here.

tests/unit/test_backends.py
```diff
@@ -177,6 +161,53 @@ def getCallSets(self, variantSetId, pageSize=100):
...
+    def testVariantSetPagination(self):
+        results = []
+        for pageSize in range(1, 10):
+            variantSetIds = [
+                variantSet.id for variantSet in self.getVariantSets(
+                    pageSize=pageSize)]
+            results.append(variantSetIds)
+        for result in results[1:]:
+            self.assertEqual(result, results[0])
+        # TODO repeat this for full object equality, not just ids
```
jeromekelleher: We might as well just do this for object equality now and get rid of this comment. We can replace this with something like
```python
results = []
for pageSize in range(1, 10):
    results.append(list(self.getVariantSets(pageSize=pageSize)))
for result in results[1:]:
    self.assertEqual(result, results[0])
```

tests/unit/test_backends.py
```diff
@@ -177,6 +161,53 @@ def getCallSets(self, variantSetId, pageSize=100):
...
 
+
+class TestSimulatedBackend(TestAbstractBackend):
+
+    def setUp(self):
+        self._backend = backend.SimulatedBackend(
+            numCalls=100, numVariantSets=10)
+        # TODO arbitrary values
+
+    def tearDown(self):
```
jeromekelleher: Might as well delete this if it doesn't do anything.

tests/unit/test_backends.py
```diff
@@ -177,6 +161,53 @@ def getCallSets(self, variantSetId, pageSize=100):
...
+        pass
+
+
+class TestPaging(TestSimulatedBackend):
```
jeromekelleher: Some docstring comments would be nice. Just a quick sentence explaining what we want to achieve in this test.

jeromekelleher:
This looks good, thanks @macieksmuga. I wonder if there's much point in trying to make a general superclass of our new tests and the Wormtable backend test --- we're aiming to delete TestWormtableBackend, so there's not much point in trying to share code nicely with it. 

The approach I would suggest here is to make a new class, `TestBackend(unittest.TestCase)` that is completely separate from the current inheritence tree. This should create a simulated backend, and test the code in backend.py. This means tapping in to the top-level calls like `searchVariantSets` and verifying that they return the same values as we would get if we accessed the VariantSets directly. That is, we run `self._backend.searchVariantSets()`, and compare this with `self._backend.getVariantSets()`.

Paging is actually the main thing we want to test within `backend.py`, so I think we should be a bit more comprehensive: the number of elements returned should always be <= the requested page size, we should never return 0 length pages, as well as the consistency checks we're doing here.

I guess this has taken us a long way from your actual PR - sorry! What you have done here is good and we can merge this, and move on with adding more tests if you prefer. I just thought it would be good to explain what I think we're trying to test in this module now (in the past, it was basically everything).

macieksmuga:
@jeromekelleher There were a couple of possible directions I could go in to resolve the issue, and this PR is so far a quick sketch to get your take on the matter. Thus I really appreciate your comments above.f

As you mentioned, up to this point this module has tested... kind of everything (including data-driven-like tests). Thus, I am sympathetic to your suggestion to convert this to a collection of paging-oriented tests against `SimulatedBackend`, and unless you comment further on the below issue/suggestion, I will suspend this PR for now and implement what you suggested above.

But I have one conceptual problem with this approach, and it's the reason why I structured the test class hierarchy the way I did: such a testing regimen would then exclude unit-test coverage of the `FileSystemBackend` class. Admittedly this is a pretty lightweight class that mostly hooks up lower-level access to variants, reads and references (with the bulk of the code now in `AbstractBackend`), but I still felt uncomfortable cutting it out of the unit tests.

With the current structure in `test_backends.py`, I could write unit tests specifically against methods in every class in the `backend` hierarchy separately. The next step of this approach would be to transition `TestWormtableBackend` to `TestFileSystemBackend`, and switch its setUp method to generate/point at the `.tbi` indices, an approach I believe @dcolligan started to investigate.

jeromekelleher:
Ah, now I get it! Yes, everything you're saying here makes sense @macieksmuga, and I agree it would be nice to get some sort of coverage on the `FileSystemBackend` class. We can make a dataset out of the test data in `tests/data` which has exactly the right structure and all the files are already indexed. I take back the points about the class hierarchy above, you're spot on.

macieksmuga:
Thanks @jeromekelleher! I'll go ahead with the current structure, beef up `SimulatedBackend` to enable more tests against it (that effectively end up testing `AbstractBackend`, I belatedly realized), and fill out the gaps you've pointed out elsewhere.

macieksmuga:
Thinking about this a little bit more, I'll just make `TestAbstractBackend` use the `SimulatedBackend` and remove the now redundant `TestSimulatedBackend` class altogether (after all, the purpose of `SimulatedBacked` was precisely to enable running unit tests against the methods in `AbstractBackend`)

jeromekelleher:
Sounds great @macieksmuga, I'm looking forward to seeing the results!

macieksmuga:
@jeromekelleher I spent some time trying to implement what I outlined above, and fairly quickly discovered that many of the existing tests in `test_backends.py` are indeed more data-driven than unit: For example, switching over to a `pysam.VariantFile` backed `TestFileSystemBackend` ran right into the current issue with `HtslibVariantSet.getVariants()` not implementing restriction by `callSetIds`.

With this experience behind me, I'm now inclined to follow your initial proposal, namely unit-test all the methods implemented in `AbstractBackend` with an emphasis on paging, and leave coverage in the unit-tests of the `FileSystemBackend` to a simple check that it loads OK, and that the various `_*SetIds` are correct.

macieksmuga:
@jeromekelleher I've implemented all the unit tests I can for `backends.py`, but am running across this:
```
AttributeError: 'HtslibVariantSet' object has no attribute 'getCallSets'
```
This is a method that's instantiated both in the `Simulated-` and `WormtableVariantSet`s, so should have at least a trivial implementation in `HtslibVariantSet` (that possibly throws a `NotImplementedException` ? - I added it in for now).

Anyway, I'd strongly prefer to leave this test in and have somebody (perhaps myself) write the actual missing method rather than try to excise the test out, even though it will break the CI build. Let me know what you think about this.

tests/unit/test_backends.py
```diff
@@ -22,6 +23,211 @@
...
+        variant set.
+        """
+        request = protocol.GASearchCallSetsRequest()
+        request.variantSetIds = [variantSetId]
+        return self.resultIterator(
+            request, pageSize, self._backend.searchCallSets,
+            protocol.GASearchCallSetsResponse, "callSets")
+
+    def testGetVariantSets(self):
+        """
```
jeromekelleher: Minor nit: comments for test cases shouldn't be docstrings. See STYLE.md

tests/unit/test_backends.py
```diff
@@ -22,6 +23,211 @@
...
+        but without using that endpoint.
+        """
+        request = protocol.GASearchVariantSetsRequest()
+        responseStr = self._backend.runSearchRequest(
+            request.toJsonString(), protocol.GASearchVariantSetsRequest,
+            protocol.GASearchVariantSetsResponse, "variantSets",
+            self._backend.variantSetsGenerator)
+        response = protocol.GASearchVariantSetsResponse.fromJsonString(
+            responseStr)
+        self.assertTrue(type(response) is protocol.GASearchVariantSetsResponse)
```
jeromekelleher: Use `isinstance` here.

tests/unit/test_backends.py
```diff
@@ -22,6 +23,211 @@
...
+
+    def testSearchVariantSets(self):
+        """
+        Test with empty request object.
+        """
+        request = protocol.GASearchVariantSetsRequest()
+        responseStr = self._backend.searchVariantSets(request.toJsonString())
+        response = protocol.GASearchVariantSetsResponse.fromJsonString(
+            responseStr)
+        self.assertTrue(type(response) is protocol.GASearchVariantSetsResponse)
```
jeromekelleher: ditto here.

tests/unit/test_backends.py
```diff
@@ -22,6 +23,211 @@
...
+
+    def testVariantSetPagination(self):
+        results = []
+        for pageSize in range(1, 100):
+            variantSetIds = [
+                variantSet.id for variantSet in self.getVariantSets(
+                    pageSize=pageSize)]
+            results.append(variantSetIds)
+        for result in results[1:]:
+            print(result)
```
jeromekelleher: remove this print please

jeromekelleher:
This looks great, thanks @macieksmuga. For `getCallSets` I would add a trivial implementation in the superclass so your tests pass. This is being tackled in #286.

macieksmuga:
thanks @jeromekelleher! pushed new version with your corrections.

jeromekelleher:
Great! +1, assuming CI passes.

macieksmuga:
yeah, well, 'CORS it won't ;)

Issue #221, opened 2015-03-09
Searching by CallSetIds not supported by HtslibVariantSet


jeromekelleher:
We need to support searching by CallSetIds in the HtslibVariantSet. The implementation should roughly mirror the WormtableVariantSet version: maintain a map of callSetIds to sample indexes, and only retrieve the genotypes for these indexes.

See the pysam source for details and documentation: https://github.com/pysam-developers/pysam/blob/master/pysam/cbcf.pyx

We also require testing for this, and the case in which we do not specify the callSetIds. We should tetst the results using PyVCF. We must test
1. That we get back the correct information for each genotype;
2. We correctly get all genotypes when we specify no callSetIds;
3. Check that we get back the correct calls in the correct order when we specify some callSetIds. See `testSeachByCallSetIds` in `tests/unit/test_backends.py` for similar tests.

shajoezhu:
i am working on this

shajoezhu:
hey @jeromekelleher are callSetIds for variants? or samples?

pgrosu:
Joe, a `CallSet` has its own reference to a `sampleId`:

```
/**
A `CallSet` is a collection of variant calls for a particular sample.
It belongs to a `VariantSet`. This is equivalent to one column in VCF.
*/
record CallSet {

  /** The call set ID. */
  string id;

  /** The call set name. */
  union { null, string } name = null;

  /** The sample this call set's data was generated from. */
  union { null, string } sampleId;
```

Paul

shajoezhu:
cool! thanks. just wanna make sure

pgrosu:
Sure, you're welcome :)  It took me a while as well to clarify the definitions for myself.  Here's a diagram accompanied by a few definitions that clarified things for me (which can be found on http://ga4gh.org/#/api):

![variants](https://cloud.githubusercontent.com/assets/6555937/6646028/7bb83cac-c997-11e4-82eb-e454f790ab3f.png)

A GAVariant represents a change in DNA sequence relative to some reference. For example, a variant could represent a SNP or an insertion. Variants belong to a GAVariantSet. This is equivalent to a row in VCF.

* __VariantSet__ is all the variant information on a specified population
* __Variant__ (row) is a single variant (e.g. "a change A to T at offset 5,227,002 on chr 11", which causes sickle cell anemia)
* __CallSet__ (column) is the genotype information for a single tissue sample
[...]

PR #264, opened 2015-03-19
resolving issue #221

shajoezhu:
reworked on #221, this should be a cleaner version now. ready for revising

### DIFF
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -8,7 +8,6 @@ from __future__ import unicode_literals
 
 import datetime
 import glob
-import itertools
 import os
 import random
 
@@ -451,6 +450,7 @@ class HtslibVariantSet(AbstractVariantSet):
         self._created = protocol.convertDatetime(datetime.datetime.now())
         self._chromFileMap = {}
         self._metadata = None
+        self._callSetIds = None
         for pattern in ['*.bcf', '*.vcf.gz']:
             for filename in glob.glob(os.path.join(vcfPath, pattern)):
                 self._addFile(filename)
@@ -468,6 +468,7 @@ class HtslibVariantSet(AbstractVariantSet):
             self._metadata = metadata
         else:
             if self._metadata != metadata:
+                # TODO CHANGE EXCEPTION
                 raise Exception(expMsg)
 
     def getNumVariants(self):
@@ -477,9 +478,30 @@ class HtslibVariantSet(AbstractVariantSet):
         # TODO How do we get the number of records in a VariantFile?
         return 0
 
+    def _updateCallSetIds(self, variantFile):
+        """
+        Updates the call set IDs based on the specified variant file.
+        """
+        checkConsistency = True
+        if self._callSetIds is None:
+            self._callSetIds = []
+            checkConsistency = False
+
+        for sample in variantFile.header.samples:
+            if checkConsistency:
+                if sample not in self._callSetIds:
+                    # TODO CHANGE EXCEPTION
+                    raise Exception("Inconsistent sample names in VCF")
+            else:
+                if sample in self._callSetIds:
+                    # TODO CHANGE EXCEPTION
+                    raise Exception("CrappyVcfException")
+                self._callSetIds.append(sample)
+
     def _addFile(self, filename):
         varFile = pysam.VariantFile(filename)
         if varFile.index is None:
+            # TODO CHANGE EXCEPTION
             raise Exception("VCF/BCF files must be indexed")
         for chrom in varFile.index:
             # Unlike Tabix indices, CSI indices include all contigs defined
@@ -488,11 +510,33 @@ class HtslibVariantSet(AbstractVariantSet):
             # overlapping errors.
             if not isEmptyIter(varFile.fetch(chrom)):
                 if chrom in self._chromFileMap:
+                    # TODO CHANGE EXCEPTION
                     raise Exception("cannot have overlapping VCF/BCF files.")
                 self._updateMetadata(varFile)
+                self._updateCallSetIds(varFile)
                 self._chromFileMap[chrom] = varFile
 
-    def convertVariant(self, record):
+    def _convertGaCall(self, recordId, name, pysamCall):
+        call = protocol.GACall()
+        call.callSetId = recordId
+        call.callSetName = name
+        call.genotype = list(pysamCall.allele_indices)
+        # TODO: phaseset is not set by pysam
+        call.phaseset = None
+        call.genotypeLikelihood = []
+        for key, value in pysamCall.iteritems():
+            if key == 'GL' and value is not None:
+                call.genotypeLikelihood = list(value)
+            elif key != 'GT':
+                call.info[key] = _encodeValue(value)
+        return call
+
+    def convertVariant(self, record, callSetIds):
+        """
+        Converts the specified pysam variant record into a GA4GH Variant
+        object. Only calls for the specified list of callSetIds will
+        be included.
+        """
         variant = protocol.GAVariant()
         # N.B. record.pos is 1-based
         #      also consider using record.start-record.stop
@@ -513,21 +557,16 @@ class HtslibVariantSet(AbstractVariantSet):
             variant.alternateBases = list(record.alts)
         # record.filter and record.qual are also available, when supported
         # by GAVariant.
+
         for key, value in record.info.iteritems():
             if value is not None:
                 variant.info[key] = _encodeValue(value)
+
         variant.calls = []
         for name, call in record.samples.iteritems():
-            c = protocol.GACall()
-            c.genotype = _encodeValue(call.allele_indices)
-            for key, value in record.info.iteritems():
-                if key == 'GT':
-                    c.genotype = list(call.allele_indices)
-                elif key == 'GL' and value is not None:
-                    c.genotypeLikelihood = value
-                elif value is not None:
-                    c.info[key] = _encodeValue(value)
-            variant.calls.append(c)
+            if name in callSetIds:
+                variant.calls.append(self._convertGaCall(
+                    record.id, name, call))
         return variant
 
     def getVariants(self, referenceName, startPosition, endPosition,
@@ -539,15 +578,15 @@ class HtslibVariantSet(AbstractVariantSet):
         if variantName is not None:
             raise exceptions.NotImplementedException(
                 "Searching by variantName is not supported")
-        if callSetIds is not None:
-            raise exceptions.NotImplementedException(
-                "Specifying call set ids is not supported")
+        if callSetIds is None:
+            callSetIds = self._callSetIds
+
         if referenceName in self._chromFileMap:
             varFile = self._chromFileMap[referenceName]
-            cursor = varFile.fetch(referenceName, startPosition, endPosition)
-            return itertools.imap(self.convertVariant, cursor)
-        else:
-            return []
+            cursor = varFile.fetch(
+                referenceName, startPosition, endPosition)
+            for record in cursor:
+                yield self.convertVariant(record, callSetIds)
 
     def getMetadata(self):
         return self._metadata
--- a/tests/datadriven/test_variants.py
+++ b/tests/datadriven/test_variants.py
@@ -13,6 +13,7 @@ import vcf
 import ga4gh.protocol as protocol
 import ga4gh.datamodel.variants as variants
 import tests.datadriven as datadriven
+import tests.utils as utils
 
 
 def testVariantSets():
@@ -30,7 +31,7 @@ class VariantSetTest(datadriven.DataDrivenTest):
 
     def __init__(self, variantSetId, baseDir):
         super(VariantSetTest, self).__init__(variantSetId, baseDir)
-        self._variants = []
+        self._variantRecords = []
         self._referenceNames = set()
         # Read in all the VCF files in datadir and store each variant.
         for vcfFile in glob.glob(os.path.join(self._dataDir, "*.vcf.gz")):
@@ -46,9 +47,10 @@ class VariantSetTest(datadriven.DataDrivenTest):
         self._vcfVersion = metadata["fileformat"]
         self._infos = vcfReader.infos
         self._formats = vcfReader.formats
+        self.vcfSamples = vcfReader.samples
         for record in vcfReader:
-            self._variants.append(record)
             self._referenceNames.add(record.CHROM)
+            self._variantRecords.append(record)
 
     def getDataModelClass(self):
         return variants.HtslibVariantSet
@@ -56,11 +58,77 @@ class VariantSetTest(datadriven.DataDrivenTest):
     def getProtocolClass(self):
         return protocol.GAVariantSet
 
-    def verifyVariantsEqual(self, gaVariants, pyvcfVariants):
+    def _floatsAgreeWithinTolerance(self, a, b, decimalPoints=7):
+        afloat = float(a)
+        bfloat = float(b)
+        return (abs(afloat - bfloat) <= 10**(-decimalPoints) * abs(bfloat))
+
+    def _compareTwoListFloats(self, a, b):
+        for ai, bi in zip(a, b):
+            if not self._floatsAgreeWithinTolerance(ai, bi):
+                return False
+        return True
+
+    def _verifyInfoEqual(self, gaObjectInfo, pyvcfInfo):
+        def _assertEquivalentGaVCFValues(gaValue, pyvcfValue):
+            if isinstance(pyvcfValue, str):
+                self.assertEqual(gaValue, pyvcfValue)
+            elif isinstance(pyvcfValue, (int, bool)):
+                self.assertEqual(gaValue, str(pyvcfValue))
+            elif isinstance(pyvcfValue, float):
+                self._floatsAgreeWithinTolerance(gaValue, pyvcfValue)
+            elif pyvcfValue is None:
+                self.assertEqual(gaValue, ".")
+            else:
+                raise Exception(key, (
+                    " values are inconsistent",
+                    "between ga4ghObject and pyvcf!"))
+
+        for key, value in pyvcfInfo.iteritems():
+            if isinstance(value, list):
+                self.assertEqual(len(gaObjectInfo[key]), len(value))
+                for gaValue, pyvcfValue in zip(
+                  gaObjectInfo[key], pyvcfInfo[key]):
+                    _assertEquivalentGaVCFValues(gaValue, pyvcfValue)
+            else:
+                self.assertEqual(len(gaObjectInfo[key]), 1)
+
+    def _verifyVariantCallEqual(self, gaCall, pyvcfCall):
+        genotype, phaseset = variants.convertVCFGenotype(
+                pyvcfCall.data.GT, pyvcfCall.phased)
+        self.assertEqual(gaCall.callSetId, pyvcfCall.site.ID)
+        self.assertEqual(gaCall.callSetName, pyvcfCall.sample)
+        self.assertEqual(gaCall.genotype, genotype)
+        # TODO: Need to check the phaseset!
+        # gaCall.phaseset is currently not implemented?
+        # self.assertEqual(gaCall.phaseset,phaseset)
+        if len(gaCall.genotypeLikelihood) > 0:
+            self.assertTrue(self._compareTwoListFloats(
+                gaCall.genotypeLikelihood, pyvcfCall.data.GL))
+        else:
+            self.assertNotIn("GL", pyvcfCall.data)
+        for key, value in gaCall.info.items():
+            if key != "GT" and key != "GL":
+                if isinstance(value[0], (list, tuple)):
+                    self._compareTwoListFloats(value[0], getattr(
+                        pyvcfCall.data, key))
+                elif isinstance(value[0], float):
+                    self._compareTwoFloats(value[0], getattr(
+                        pyvcfCall.data, key))
+
+    def _verifyVariantsEqual(self, gaVariants, pyvcfVariants):
         """
         Verifies that the lists of GA4GH variants and pyvcf variants
         are equivalent.
         """
+        def _verifyVariantCalls():
+            for gaCall in gaVariant.calls:
+                self.assertTrue(protocol.GACall.validate(
+                    gaCall.toJsonDict()))
+                self.assertIn(gaCall.callSetName, pyvcfCallMap)
+                pyvcfCall = pyvcfCallMap[gaCall.callSetName]
+                self._verifyVariantCallEqual(gaCall, pyvcfCall)
+
         self.assertEqual(len(gaVariants), len(pyvcfVariants))
         for gaVariant, pyvcfVariant in zip(gaVariants, pyvcfVariants):
             pyvcfInfo = pyvcfVariant.INFO
@@ -73,6 +141,7 @@ class VariantSetTest(datadriven.DataDrivenTest):
             if "END" in pyvcfInfo:
                 end = pyvcfInfo["END"]
             self.assertEqual(gaVariant.end, end)
+            self._verifyInfoEqual(gaVariant.info, pyvcfInfo)
             alt = pyvcfVariant.ALT
             # PyVCF does something funny when no ALT allele is provided.
             # TODO we should clarify exactly what this means.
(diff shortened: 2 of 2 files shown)


ga4gh/datamodel/variants.py
```diff
@@ -490,9 +506,42 @@ def _addFile(self, filename):
...
                 self._chromFileMap[chrom] = varFile
 
-    def convertVariant(self, record):
+    def convertVariant(self, record, callSetIds):
+        """
+        Converts the specified pysam variant record into a GA4GH Variant
+        object. Only calls for the specified list of callSetIds will
+        be included.
+        """
+        def _convertGaCall(recordId, recordInfo, call):
```
jeromekelleher: I think convertCall is probably important enough to get a top-level method. These local functions are useful at times, but are a little over-used here. As a rule of thumb, if you're not actually using the context of the surrounding function then it should probably be a normal method.

ga4gh/datamodel/variants.py
```diff
@@ -490,9 +506,42 @@ def _addFile(self, filename):
...
+            c = protocol.GACall()
+            """
```
jeromekelleher: Delete this multi-line string.

ga4gh/datamodel/variants.py
```diff
@@ -490,9 +506,42 @@ def _addFile(self, filename):
...
+            callSetId
+            callSetName
+            genotype
+            phaseset
+            genotypeLikelihood
+            info
+            """
+            c.callSetId = recordId
+            c.callSetName = name
+            # c.genotype = _encodeValue(call.allele_indices)
```
jeromekelleher: genotype is a list of ints, so your code below is correct. Delete this comment.

ga4gh/datamodel/variants.py
```diff
@@ -490,9 +506,42 @@ def _addFile(self, filename):
...
+            # ga4gh_api.html#/schema/org.ga4gh.GACall
+            # says the genotype type is array of int?
+            c.genotype = list(call.allele_indices)
+            # TODO: phaseset is not set by pysam
+            c.phaseset = None
+            c.genotypeLikelihood = []
+            for key, value in call.iteritems():
+                if key == 'GL' and value is not None:
+                    c.genotypeLikelihood = [v for v in value]
+            c.info = recordInfo
```
jeromekelleher: This is a bug I think: are you adding all of the info fields for the variant into the Call? The call.info field is for info tags on the call itself. All you'll have here is something like:
```python
for key, value in call.iteritems():
    if key == 'GL':
        if value is not None:
            c.genotypeLikelihood = list(value)
    else:
          c.info[key] = _encodeValue(value)
```

ga4gh/datamodel/variants.py
```diff
@@ -513,21 +562,18 @@ def convertVariant(self, record):
...
-            c.genotype = _encodeValue(call.allele_indices)
-            for key, value in record.info.iteritems():
-                if key == 'GT':
-                    c.genotype = list(call.allele_indices)
-                elif key == 'GL' and value is not None:
-                    c.genotypeLikelihood = value
-                elif value is not None:
-                    c.info[key] = _encodeValue(value)
-            variant.calls.append(c)
+            # Filtering out invariant calls
```
jeromekelleher: Delete these comments.

ga4gh/datamodel/variants.py
```diff
@@ -539,15 +585,15 @@ def getVariants(self, referenceName, startPosition, endPosition,
         if variantName is not None:
             raise exceptions.NotImplementedException(
                 "Searching by variantName is not supported")
-        if callSetIds is not None:
-            raise exceptions.NotImplementedException(
-                "Specifying call set ids is not supported")
+        if callSetIds is None:  # This will get all Variants
```
jeromekelleher: We can lose this comment (and it's inaccurate)

tests/datadriven/test_variants.py
```diff
@@ -56,11 +57,65 @@ def getDataModelClass(self):
...
 
-    def verifyVariantsEqual(self, gaVariants, pyvcfVariants):
+    def _verifyInfoEqual(self, gaObjectInfo, pyvcfInfo):
+        def _compareTwoFloats(a, b):
+            afloat = float(a)
+            bfloat = float(b)
+            return (abs(afloat - bfloat) <= 0.001 * bfloat)
+
+        def _compareGaValuePyvcfValue(gaValue, pyvcfValue):
+            if type(pyvcfValue) is str:
```
jeromekelleher: The canonical way to do this is `isinstance(pyvcfValue, str)`

tests/datadriven/test_variants.py
```diff
@@ -56,11 +57,65 @@ def getDataModelClass(self):
...
+                self.assertEqual(gaValue, pyvcfValue)
+            elif type(pyvcfValue) is int or type(value) is bool:
+                self.assertEqual(gaValue, str(pyvcfValue))
+            elif type(pyvcfValue) is float:
+                _compareTwoFloats(gaValue, pyvcfValue)
+            elif pyvcfValue is None:
+                self.assertEqual(gaValue, ".")
+            else:
+                print (gaValue, pyvcfValue)
```
jeromekelleher: We should raise an exception here.

tests/datadriven/test_variants.py
```diff
@@ -56,11 +57,65 @@ def getDataModelClass(self):
...
+            genotype, phaseset = variants.convertVCFGenotype(
+                    pyvcfCall.data.GT, pyvcfCall.phased)
+            self.assertEqual(gaCall.callSetId, pyvcfCall.site.ID)
+            self.assertEqual(gaCall.callSetName, pyvcfCall.sample)
+            self.assertEqual(gaCall.genotype, genotype)
+            # TODO: Need to check the phaseset!
+            # gaCall.phaseset is currently not implemented?
+            # self.assertEqual(gaCall.phaseset,phaseset)
+            if len(gaCall.genotypeLikelihood) > 0:
+                gaCallgenotypeLK = [(
```
jeromekelleher: We should use the approximate comparison for floats here.

tests/datadriven/test_variants.py
```diff
@@ -84,19 +139,63 @@ def verifyVariantsEqual(self, gaVariants, pyvcfVariants):
...
-    def testSearchAllVariants(self):
-        allVariants = []
+            pyvcfCallMap = {}
+            for call in pyvcfVariant:
+                pyvcfCallMap[call.sample] = call
+            _verifyVariantCalls(gaVariant.calls, pyvcfCallMap, pyvcfInfo)
+            self._verifyInfoEqual(gaVariant.info, pyvcfInfo)
+
+    def _verifyGaVariantsSample(self, gaVariants, sampleIds):
+        for v in gaVariants:
```
jeromekelleher: We should probably use `variant` here instead of `v`: a lot of people _really_ don't like single letter variables!

ga4gh/datamodel/variants.py
shajoezhu: check

ga4gh/datamodel/variants.py
shajoezhu: check

ga4gh/datamodel/variants.py
shajoezhu: check

ga4gh/datamodel/variants.py
shajoezhu: check

ga4gh/datamodel/variants.py
shajoezhu: check

ga4gh/datamodel/variants.py
shajoezhu: check

tests/datadriven/test_variants.py
shajoezhu: check

tests/datadriven/test_variants.py
shajoezhu: check

tests/datadriven/test_variants.py
shajoezhu: check

tests/datadriven/test_variants.py
shajoezhu: check

ga4gh/datamodel/variants.py
dcolligan: @shajoezhu just so you know, an email gets sent out to everyone that is subscribed to a thread (like this PR) every time anyone makes a comment on it.  I would advise leaving out these "check" comments.  It doesn't really add anything to the discussion for anyone else, and we presume you see the comments already!

ga4gh/datamodel/variants.py
```diff
@@ -490,9 +506,30 @@ def _addFile(self, filename):
...
+        c.callSetId = recordId
+        c.callSetName = name
+        c.genotype = list(call.allele_indices)
+        # TODO: phaseset is not set by pysam
+        c.phaseset = None
+        c.genotypeLikelihood = []
+        for key, value in call.iteritems():
+            if key == 'GL' and value is not None:
+                c.genotypeLikelihood = list(value)
+            else:
```
jeromekelleher: We need to filter GT out of the info dict here: ```elif key != GT```

tests/datadriven/test_variants.py
```diff
@@ -56,11 +57,77 @@ def getDataModelClass(self):
...
 
-    def verifyVariantsEqual(self, gaVariants, pyvcfVariants):
+    def _compareTwoFloats(self, a, b):
+        afloat = float(a)
+        bfloat = float(b)
+        return (abs(afloat - bfloat) <= 0.001 * abs(bfloat))
+
+    def _compareTwoListFloats(self, a, b):
+        for ai, bi in zip(a, b):
+            if self._compareTwoFloats(ai, bi) is False:
```
jeromekelleher: ```if not self._compareTwoFloats(ai, bi)```

tests/datadriven/test_variants.py
```diff
@@ -56,11 +57,77 @@ def getDataModelClass(self):
...
+                self.assertEqual(gaValue, ".")
+            else:
+                raise Exception(key, (
+                    " values are inconsistent",
+                    "between ga4ghObject and pyvcf!"))
+
+        for key, value in pyvcfInfo.iteritems():
+            if isinstance(value, list):
+                self.assertEqual(len(gaObjectInfo[key]), len(value))
+                for gaValue, pyvcfValue in (
```
jeromekelleher: This should probably be something like
```python
for gaValue, pyvcfValue in zip(
     gaObjectInfo[key], pyvcfInfo[key]):
    _compareGaValuePyvcfValue()
```

tests/datadriven/test_variants.py
```diff
@@ -56,11 +57,77 @@ def getDataModelClass(self):
...
+                elif isinstance(value[0], float):
+                    self._compareTwoFloats(value[0], getattr(
+                        pyvcfCall.data, key))
+
+    def _verifyVariantsEqual(self, gaVariants, pyvcfVariants):
         """
         Verifies that the lists of GA4GH variants and pyvcf variants
         are equivalent.
         """
+        def _verifyVariantCalls(gaVariantCalls, pyvcfCallMap, pyvcfInfo):
```
jeromekelleher: Is `pyvcfInfo` used here?

jeromekelleher:
Thanks @shajoezhu, this is great! There's one small issue above, but I'm +1 after that. There's a few things outstanding that I think we should follow up with issues and subsequent PRs (for example, we should add a unitest like `assertAlmostEqual` for float comparisons), but we should try to merge this much first.

@dcolligan and @macieksmuga, can you review this also please?

ga4gh/datamodel/variants.py
```diff
@@ -477,6 +477,22 @@ def getNumVariants(self):
...
+        """
+        checkConsistency = True
+        if self._callSetIds is None:
+            self._callSetIds = []
+            checkConsistency = False
+
+        for sample in variantFile.header.samples:
+            if checkConsistency:
+                if sample not in self._callSetIds:
+                    raise Exception("Inconsistent sample names in VCF")
```
dcolligan: Create a custom exception here  instead of throwing something of class Exception

ga4gh/datamodel/variants.py
```diff
@@ -490,9 +506,30 @@ def _addFile(self, filename):
                 if chrom in self._chromFileMap:
                     raise Exception("cannot have overlapping VCF/BCF files.")
```
dcolligan: Same for the rest of these Exceptions throughout the file

tests/datadriven/test_variants.py
```diff
@@ -56,11 +57,77 @@ def getDataModelClass(self):
     def getProtocolClass(self):
         return protocol.GAVariantSet
 
-    def verifyVariantsEqual(self, gaVariants, pyvcfVariants):
+    def _compareTwoFloats(self, a, b):
```
dcolligan: Yeah this should be assertAlmostEqual or something, placed in the test superclass... check the python unittest documentation for the signature

tests/datadriven/test_variants.py
```diff
@@ -84,19 +152,52 @@ def verifyVariantsEqual(self, gaVariants, pyvcfVariants):
...
+
+    def testSearchAllVariants(self):
+        self._verifyVariantsCallSetIds(None, [])
+
+    def testSearchCallSetIdsSystematic(self):
+        for sampleIds in utils.powerset(self.vcfSamples, maxSets=10):
+            self._verifyVariantsCallSetIds(None, list(sampleIds))
+
+    def testVariantsValid(self):
+        end = 2**30  # TODO This is arbitrary, and pysam can choke. FIX!
```
dcolligan: it would be good to externalize this number as a property of the base class, since we seem to use it a lot in various tests

ga4gh/datamodel/variants.py
jeromekelleher: Good call. However, I think we should deal with this as a follow up issue. We need proper exceptions, but we also need to test that they are being thrown properly. I'll create an issue if you agree.

tests/datadriven/test_variants.py
jeromekelleher: We should really just fix this at the source instead. Before we send a number to pysam, we should check that it's not going to choke on it. There's a wider issue of sanitisation for pysam inputs too, I guess. Another issue?

dcolligan: I'm fine if we merge this and create follow-up issues for it if you're intent on merging.

jeromekelleher:
I've added a couple of issues following up on these points @dcolligan --- it's not that I'm intent on merging, it's just that these are big issues that need to be dealt with properly, and @shajoezhu's PR is going to get very big if we try to tackle them all here.

tests/datadriven/test_variants.py
macieksmuga: agree with @jeromekelleher, global fix needed, defer to another PR.

ga4gh/datamodel/variants.py
```diff
@@ -477,6 +477,22 @@ def getNumVariants(self):
...
+        if self._callSetIds is None:
+            self._callSetIds = []
+            checkConsistency = False
+
+        for sample in variantFile.header.samples:
+            if checkConsistency:
+                if sample not in self._callSetIds:
+                    raise Exception("Inconsistent sample names in VCF")
+            else:
+                self._callSetIds.append(sample)
```
macieksmuga: Is it possible for the same `sample` to be encountered more than once when `checkConsistency` is `False` (i.e., when `self._callSetIds` is being built up within this method)? If so, is adding it again the desired behavior?

ga4gh/datamodel/variants.py
```diff
@@ -490,9 +506,30 @@ def _addFile(self, filename):
                 if chrom in self._chromFileMap:
                     raise Exception("cannot have overlapping VCF/BCF files.")
                 self._updateMetadata(varFile)
+                self._updateCallSetIds(varFile)
                 self._chromFileMap[chrom] = varFile
 
-    def convertVariant(self, record):
+    def _convertGaCall(self, recordId, name, call):
+        c = protocol.GACall()
```
macieksmuga: in other similar places in the code we've used a lowercased entity name (minus the "GA") for the variable,
for example 
```
variant = protocol.GAVariant()
```
Thus I'd recommend  naming this `call` instead of `c`.

tests/datadriven/test_variants.py
```diff
@@ -56,11 +57,77 @@ def getDataModelClass(self):
     def getProtocolClass(self):
         return protocol.GAVariantSet
 
-    def verifyVariantsEqual(self, gaVariants, pyvcfVariants):
+    def _compareTwoFloats(self, a, b):
+        afloat = float(a)
+        bfloat = float(b)
+        return (abs(afloat - bfloat) <= 0.001 * abs(bfloat))
```
macieksmuga: Not loving the `0.001` magic number. Also, "compare" in the method signature suggests a comparison style return `(<,==,>)` as opposed to a bool. Suggest something like:
```
def floatsAgreeWithinTolerance(self, a, b, decimalPoints = 9):
    [...]
    return (abs(afloat - bfloat) <= 10**(-decimalPoints) * abs(bfloat))
```
If we're just worried about floating point math-induced differences, the tolerance can be much tighter than 3 sig digs.
Here's something I found that uses `FLT_EPSILON` to address this issue:
https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/

Apologies for the long and unwieldy suggested method name: I'm coming off an Objective-C programming project(I had to stymie my impulse to call it `[Utils compareFloat: a withFloat: b toSignificantDigits: decimalPoints];`). I welcome other suggestions for naming such a method.

tests/datadriven/test_variants.py
```diff
@@ -56,11 +57,77 @@ def getDataModelClass(self):
...
+
+    def _compareTwoListFloats(self, a, b):
+        for ai, bi in zip(a, b):
+            if not self._compareTwoFloats(ai, bi):
+                return False
+        return True
+
+    def _verifyInfoEqual(self, gaObjectInfo, pyvcfInfo):
+        def _compareGaValuePyvcfValue():
```
macieksmuga: it's an assertion, not a comparison. And I think I'd like to see values passed in explicitly, makes for less error prone code if somebody changes the variable names later in the code:
```
    def _assertEquivalentGaVCFValues(ga, pyvcf):
[...]
    #usage:
    _assertEquivalentGaVCFValues(gaValue, pyvcfValue)
```

tests/datadriven/test_variants.py
```diff
@@ -56,11 +57,77 @@ def getDataModelClass(self):
     def getProtocolClass(self):
         return protocol.GAVariantSet
 
-    def verifyVariantsEqual(self, gaVariants, pyvcfVariants):
+    def _compareTwoFloats(self, a, b):
+        afloat = float(a)
+        bfloat = float(b)
+        return (abs(afloat - bfloat) <= 0.001 * abs(bfloat))
+
+    def _compareTwoListFloats(self, a, b):
```
macieksmuga: ditto @dcolligan's above note for this.

tests/datadriven/test_variants.py
```diff
@@ -84,19 +152,52 @@ def verifyVariantsEqual(self, gaVariants, pyvcfVariants):
...
+        for variant in gaVariants:
+            self.assertEqual(len(variant.calls), len(sampleIds))
+            for call in variant.calls:
+                self.assertIn(call.callSetName, sampleIds)
+
+    def _verifyVariantsCallSetIds(self, searchVariants, searchsampleIds):
+        """
+        SearchVariants = None, this will get all variants
+        If searchsampleIds = None, this will get all samples
+        else if = [of something], this will iterate through the list
```
macieksmuga: oh, I see. `searchVariants` and `searchsampleIds` *(fix the camelCasing on that last one)* expect arrays or `None`.
Would it be simpler just to make the signature:
```
def _verifyVariantCallSetIds(self, searchVariants = [], searchSampleIds = [])
    """
    Leaving searchVariants empty will get all variants,
    leaving searchSampleIds will get all samples.
    """
```

tests/datadriven/test_variants.py
```diff
@@ -84,19 +152,52 @@ def verifyVariantsEqual(self, gaVariants, pyvcfVariants):
...
-            allVariants += variants
+            gaVariants = list(self._gaObject.getVariants(
+                referenceName, 0, end, searchVariants, searchsampleIds))
+            self._verifyGaVariantsSample(gaVariants, searchsampleIds)
+            gaCallSetVariants += gaVariants
             localVariants = [
-                v for v in self._variants if v.CHROM == referenceName]
-            self.verifyVariantsEqual(variants, localVariants)
-        self.assertEqual(len(allVariants), len(self._variants))
+                v for v in self._variantRecords if (v.CHROM == referenceName)]
```
macieksmuga: stylistic question for @jeromekelleher et al. - Is list comprehension preferable to `filter` in this a case? 
```
localVariants = filter(lambda v: v.CHROM == referenceName, self._variantRecords)
```

macieksmuga:
+1 modulo fixing a couple of minor line comments above. Thank you @shajoezhu!

ga4gh/datamodel/variants.py
shajoezhu: Hi @macieksmuga, good point. I don't think the same `sample` can appear twice. but we should always check. I think we can add an assertion, such that `sample` is not in `self._callSetIds` already, before append it?

tests/datadriven/test_variants.py
shajoezhu: Hi @dcolligan @jeromekelleher and @macieksmuga I am actually not sure about using the assertAlmostEqual from the unittest package, which check for round(a-b, 7) == 0, see https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertAlmostEqual, 

for example, this test fails when a and b small, 
```python
>>> a=0.00000005
>>> b=0.00000009
>>> round(a-b, 7) == 0
True
>>> abs(a - b) <= 0.001 * abs(b)
False
```
I am not sure, 0.001 was the best choice here. we could certainly increase the number of digits. but I am not sure using assertAlmostEqual was the best thing? maybe we should just have our own implementation of assertAlmostEqual ( as what I did )? but increase the precision? here is what i propose,

```python
def _assertAlmostEqual(self, a, b, precision=3):
    afloat = float(a)
    bfloat = float(b)
    return (abs(afloat - bfloat) <= 10**-precision * abs(bfloat))
```
what do you think?

tests/datadriven/test_variants.py
shajoezhu: :+1:

pgrosu:
Let's also clean up all the `GA` parts throughout all of the [server](https://github.com/ga4gh/server/) code, since that does not exist anymore in the [schemas](https://github.com/ga4gh/schemas/).

Thanks,
~p

macieksmuga:
@pgrosu @shajoezhu We need to keep the GA prefix for now, as it's part of the v0.5.1 spec, that we're developing this server version against. I'm removing it in the experimental 0.6a branch I'm working on, however.

pgrosu:
That makes sense and sounds reasonable Maciek.

Thanks,
Paul

ga4gh/datamodel/variants.py
jeromekelleher: Good point @macieksmuga. We should add a check for this and raise a CrappyVcfException (or something) if it happens.

tests/datadriven/test_variants.py
jeromekelleher: This is a subtle problem - I think we should move this to an issue so we can discuss it in isolation and not hold up this PR. I'll open an issue if we agree.

shajoezhu: agree

shajoezhu:
hi @dcolligan and @macieksmuga , I was wondering could you guys help to take another revise of this PR. I have done some work on issue 269. was hoping to submit another PR after merging this one. Thank you very much! 

Joe

tests/datadriven/test_variants.py
```diff
@@ -7,12 +7,12 @@
 
 import os
 import glob
-
```
jeromekelleher: We should keep this space, as we want to separate standard library imports from 3rd party packages --- see the note on imports in STYLE.md

ga4gh/datamodel/variants.py
jeromekelleher: Can we add this check please @shajoezhu? I think we should be ready to merge then, and add the extra issues for us to follow up on.

jeromekelleher:
Thanks for the updates @shajoezhu. Can you squash the commits down so this is ready to merge please? The builds are failing because of another issue (see #278 and #277), and hopefully this will be worked around soon.

jeromekelleher:
Great, thanks @shajoezhu! @dcolligan, can you check this out and merge if you're happy? There's a bunch of issues I want to make based on this, so it would be good to get it merged.

Issue #246, opened 2015-03-12
Implement reads/search


dcolligan:
The reads/search endpoint is not implemented on the server.  Change that.

dcolligan:
@jeromekelleher do we also want a ReadsSimulator?  What priority would that be?

jeromekelleher:
A `ReadsSimulator` would be great, and is certainly something we'll need. I've no real idea how this would work though---my read-data ignorance is starting to hurt!

jeromekelleher:
We should implement #247 before tackling this.

dcolligan:
@jeromekelleher can you add some reads to ga4gh-example-data ?

jeromekelleher:
Sure - if you point me in the direction of the files and give me a script to generate the subsets!

dcolligan:
Scripts are in #243

Files are here: http://hgdownload.cse.ucsc.edu/goldenPath/hg19/encodeDCC/wgEncodeUwRepliSeq/

(I picked the first two bam files for test data, so I would avoid those two)

dcolligan:
In `FileSystemBackend`, readGroupSetIds are indexed but there is no way in `GASearchReadsRequest` to specify a readGroupSetId.  Presumably, we should be indexing readGroupIds, which can be specified in a `GASearchReadsRequest`, (and which include the readGroupSetId in their id) correct?

jeromekelleher:
Yeah, we'll need to index all the readGroupIds so we can find them for queries. We'll still need all the readGroupSets though, as these need to be retrievable also.

dcolligan:
With regards to the nextPageToken with reads/search:

Is it possible for two alignments in a BAM file to be at the same starting position?  If so, setting the nextPageToken to include the position of the subsequent alignment will end up returning duplicate alignments.  In that case we would need to include the ids of alignments with the same starting position that the client has already consumed in the nextPageToken in order to filter them out on the server side.

If not, then the look-ahead method will work without the readAlignmentId array in the nextPageToken.

Fortunately (?), pysam's `fetch` doesn't include unmapped reads, so we don't need to worry about that complexity.

jeromekelleher:
[...]

PR #268, opened 2015-03-19
246 reads search

dcolligan:
This PR is branched off of the getReadAlignments implementation.

Tests are failing because of some validation stuff... is there any way for avro to tell us what field is invalid and why?

Issue #246

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -29,6 +29,8 @@ class AbstractBackend(object):
         self._referenceSetIds = []
         self._readGroupSetIdMap = {}
         self._readGroupSetIds = []
+        self._readGroupIds = []
+        self._readGroupIdMap = {}
         self._requestValidation = False
         self._responseValidation = False
         self._defaultPageSize = 100
@@ -205,6 +207,68 @@ class AbstractBackend(object):
         return self._topLevelObjectGenerator(
             request, self._variantSetIdMap, self._variantSetIds)
 
+    def readsGenerator(self, request):
+        # Local utility functions to save some typing
+        def getPosition(readAlignment):
+            return readAlignment.alignment.position.position
+
+        def getEndPosition(readAlignment):
+            return getPosition(readAlignment) + \
+                len(readAlignment.alignedSequence)
+
+        if len(request.readGroupIds) != 1:
+            raise exceptions.NotImplementedException(
+                "Read search over multiple readGroups not supported")
+        readGroupId = request.readGroupIds[0]
+        try:
+            readGroup = self._readGroupIdMap[request.readGroupIds[0]]
+        except KeyError:
+            raise exceptions.ReadGroupNotFoundException(readGroupId)
+        startPosition = request.start
+        equalPositionsToSkip = 0
+        if request.pageToken is not None:
+            startPosition, equalPositionsToSkip = self.parsePageToken(
+                request.pageToken, 2)
+        iterator = readGroup.getReadAlignments(
+            request.referenceName, request.referenceId, startPosition,
+            request.end)
+        readAlignment = next(iterator, None)
+        if request.pageToken is not None:
+            # First, skip any reads with position < startPosition
+            # or endPosition < request.start
+            while (getPosition(readAlignment) < startPosition or
+                   getEndPosition(readAlignment) < request.start):
+                readAlignment = next(iterator, None)
+                if readAlignment is None:
+                    raise exceptions.BadPageTokenException(
+                        "Inconsistent page token provided")
+            # Now, skip equalPositionsToSkip records which have position
+            # == startPosition
+            equalPositionsSkipped = 0
+            while equalPositionsSkipped < equalPositionsToSkip:
+                if getPosition(readAlignment) != startPosition:
+                    raise exceptions.BadPageTokenException(
+                        "Inconsistent page token provided")
+                equalPositionsSkipped += 1
+                readAlignment = next(iterator, None)
+                if readAlignment is None:
+                    raise exceptions.BadPageTokenException(
+                        "Inconsistent page token provided")
+        # The iterator is now positioned at the correct place.
+        while readAlignment is not None:
+            nextReadAlignment = next(iterator, None)
+            nextPageToken = None
+            if nextReadAlignment is not None:
+                if getPosition(readAlignment) == \
+                        getPosition(nextReadAlignment):
+                    equalPositionsToSkip += 1
+                else:
+                    equalPositionsToSkip = 0
+                nextPageToken = "{}:{}".format(
+                    getPosition(nextReadAlignment), equalPositionsToSkip)
+            yield readAlignment, nextPageToken
+            readAlignment = nextReadAlignment
+
     def variantsGenerator(self, request):
         """
         Returns a generator over the (variant, nextPageToken) pairs defined by
@@ -332,6 +396,15 @@ class SimulatedBackend(AbstractBackend):
             self._variantSetIdMap[variantSetId] = variantSet
         self._variantSetIds = sorted(self._variantSetIdMap.keys())
 
+        # Reads
+        readGroupSetId = "aReadGroupSet"
+        readGroupSet = reads.SimulatedReadGroupSet(readGroupSetId)
+        self._readGroupSetIdMap[readGroupSetId] = readGroupSet
+        for readGroup in readGroupSet.getReadGroups():
+            self._readGroupIdMap[readGroup.getId()] = readGroup
+        self._readGroupSetIds = sorted(self._readGroupSetIdMap.keys())
+        self._readGroupIds = sorted(self._readGroupIdMap.keys())
+
 
 class FileSystemBackend(AbstractBackend):
     """
@@ -366,7 +439,10 @@ class FileSystemBackend(AbstractBackend):
         for readGroupSetId in os.listdir(readGroupSetDir):
             relativePath = os.path.join(readGroupSetDir, readGroupSetId)
             if os.path.isdir(relativePath):
-                readGroupSet = reads.ReadGroupSet(
+                readGroupSet = reads.HtslibReadGroupSet(
                     readGroupSetId, relativePath)
                 self._readGroupSetIdMap[readGroupSetId] = readGroupSet
+                for readGroup in readGroupSet.getReadGroups():
+                    self._readGroupIdMap[readGroup.getId()] = readGroup
         self._readGroupSetIds = sorted(self._readGroupSetIdMap.keys())
+        self._readGroupIds = sorted(self._readGroupIdMap.keys())
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -67,23 +67,13 @@ class SamFlags(object):
         flagAttr |= flag
 
 
-class ReadGroupSet(object):
+class AbstractReadGroupSet(object):
     """
-    Class representing a logical collection ReadGroups.
+    The base class of a read group set
     """
-    def __init__(self, id_, dataDir):
+    def __init__(self, id_):
         self._id = id_
-        self._dataDir = dataDir
         self._readGroups = []
-        # we only support BAM files right now;
-        # SAM files (which can't have index files) would be too slow
-        pattern = "*.bam"
-        for path in glob.glob(os.path.join(self._dataDir, pattern)):
-            filename = os.path.basename(path)
-            localId = os.path.splitext(filename)[0]
-            readGroupId = "{}:{}".format(self._id, localId)
-            readGroup = ReadGroup(readGroupId, path)
-            self._readGroups.append(readGroup)
 
     def getReadGroups(self):
         """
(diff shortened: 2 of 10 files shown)


jeromekelleher:
Unfortunately, there's not way to find out which field is causing validation to fail. This sucks.

pgrosu:
Danny, Jerome -  can you replace `stderr=self.clientErrFile` with `stderr=subprocess.STDOUT`, and add the argument `shell=True` to [line 106 of test_gestalt.py](https://github.com/dcolligan/server/blob/3caa7388ec427a711ff428cfef1beac9c993a5f7/tests/end_to_end/test_gestalt.py#L106) so it looks like this:

```Python
  subprocess.check_call(
     splits, stdout=self.clientOutFile, stderr=subprocess.STDOUT, shell=True)
```

Then we can see why it fails in the Travis - after this line in the log - the next time it is run:

```Python
CalledProcessError: Command '['python', 'client_dev.py', '-v', '-O', 'reads-search',  
   'http://localhost:8001/v0.5.1']' returned non-zero exit status 1
```

Thanks,
Paul

dcolligan:
@jeromekelleher ok, pushed an update, but this has some serious defects.

Right now I have the nextPageToken constructed to send back every id in the readGroup that we've seen so far.  So, this isn't going to scale since we don't want to be sending the id of every readAlignment in the BAM file back and forth on every request.  We could send the startPos + length = endPos of every read along with the id so we can trim this list down: once request.start passes the endPos of any read, remove it since there's no chance it will be returned in the response.

Even if we pursued this path, this raises the issue of how we should construct/parse the nextPageTokens.  What characters should we use as dividers, what can we assume will be in the read ids, filenames, etc.

But, while correct, this approach is not the greatest by a long shot.  AFAICT pysam doesn't support resuming a certain partially fulfilled query in a BAM file.  If we can't think of anything better to do, it might be worth thinking about adding that support.

ga4gh/backend.py
```diff
@@ -374,10 +444,13 @@ def __init__(self, dataDir):
         for readGroupSetId in os.listdir(readGroupSetDir):
             relativePath = os.path.join(readGroupSetDir, readGroupSetId)
             if os.path.isdir(relativePath):
-                readGroupSet = reads.ReadGroupSet(
+                readGroupSet = reads.readGroupSetFactory(
```
jeromekelleher: Why the factory? We might get rid of the factory for variantSets soon, so I wouldn't be too literal about following the conventions there.

jeromekelleher:
This is quite tricky all right. I think there's a couple of things we can do to simplify though:

1. Lets support only searching over a single readGroup for now --- raise a NotImplementedException if the length of readGroupIds is not 1. I'm hoping this is going to be the only case soon (see https://github.com/ga4gh/schemas/pull/253). We're not implementing search over multiple readgroups correctly now in any case, as we should be returning _all_ reads in position sorted order.

2. There must be some simple solution that does not require us to store the IDs in the `nextPageToken`. For example, we can count the number of reads we have seen so far with the same `start` position, and embed this into the `nextPageToken`. That is, we have a pageToken of `(start, skipRecords)`. We position the iterator at `start`, and then discard `skipRecords` before we start yielding records back to the caller. This isn't optimal, but will probably work well enough in practise. What is the average number of reads that have the same start position in a BAM? Once this is less than the page size,  it'll perform well-enough in practice.

dcolligan:
wrt 1: How are we not implementing searching over multiple readGroups correctly?  This implementation has a lot of problems, but that seems to be something it actually does right!

wrt 2: your proposal was something like the original implementation I had, but "skipping" N reads requires starting at 0 and iterating through and discarding reads 0...N-1.  I like this even less than the implementation I have right now because it is very inefficient on the server side.

I think the problem with your exact proposal is that a read with startPosition A is probably going to be returned if the request.start attribute is A + 1 because pysam returns "overlapping reads."  That is, if A + 1 is between A and A + len(read) -- which it probably is -- it will be returned.

jeromekelleher:
From the [docs](http://ga4gh.org/documentation/api/v0.5.1/ga4gh_api.html#/schema/org.ga4gh.GASearchReadsResponse): "The list of matching alignment records, sorted by position. Unmapped reads, which have no position, are returned last." So, when we're searching over multiple sets we need to make sure we return the reads in the correct order _across sets_. This is complicated, so I'd rather not tackle it at all, unless we have to.

I'm not suggesting we start from the very, very start of the input range for each page, just from the start of the set of records with equal start values. OK, this will involve skipping some records, but I don't know on average how many. How many reads typically have exactly the same start coordinate in a BAM? If this is something like 100, then the approach I've sketched will be efficient enough.

dcolligan:
Well, we are additionally non-conforming to the docs because at present we are not returning any unmapped reads.

I'm still unsure how "just from the start of the set of records with equal start values" works.  Keep in mind the "overlapping reads" constraint.  Stated another way, if you have read with pos A, and `start` is B, the read will be returned if A + len(read) > B.  That is, reads with positions less than `start` can be returned even if you are "starting" at a position greater than those reads' start positions.

Perhaps some pseudocode would clarify what you're proposing?

jeromekelleher:
OK, fair enough. What about something like this:
```python
startPosition = request.start
readsToSkip = 0
    startPosition, readsToSkip = parsePageToken(pageToken)
iterator = readGroup.getReadAlignments(startPosition, request.end)
read = next(iterator, None)
skipped = 0
while read is not None and skipped < readsToSkip:
     if read.start == startPosition:
         skipped += 1
     read = next(iterator, None)
# Now we should be positioned at the correct place
currentPosition = startPosition
while read is not None:
     if read.start == currentPosition:
          readsToSkip += 1
     else:
          readsToSkip = 0
          currentPosition = read.start
      nextPageToken = (currentPosition, readsToSkip)   
      yield read, nextPageToken
      read = next(iterator, None)
```
I'm sure I've missed some important details here, but do you see what I'm getting at? In worst-case, where we've got zillions of reads at with start = x, this will perform terribly. I don't think this worst-case will happen too often though, and hopefully this approach will be good enough to get us started.

dcolligan:
Ok, so this suffers from exactly the issue I was describing.

Scenario:
read1: pos 1 len 100
read2: pos 2 len 100
read3: pos 3 len 100
request: start 0 end MAX pagesize 2

Since the start positions of the reads are always different, readsToSkip will always be 0.
The first fetch call of start=0, end=MAX will return read1, read2, read3
The nextPageToken will be (2, 0) (this is a bug, it should be (3, 0), but whatever)
The second fetch call of start=2, end=MAX will return read1, read2, read3

Oops!  We're returning the same reads each time, and we should never even return one read on two separate requests.

jeromekelleher:
OK, what about:
```
# snip
while read is not None and skipped < readsToSkip or read.start < startPosition:
     if read.start == startPosition:
         skipped += 1
     read = next(iterator, None)
# continue as before
```

dcolligan:
It seems like that would fail in the same way in the same scenario.

macieksmuga:
@jeromekelleher I think @dcolligan has a point about the dangers of using `startPosition` as part of the page token. Even if a terribly clever implementation is provided that assumes a certain ordering of reads, my experience with bio data files is that there's always a set of them somewhere that manages to fail to conform to your desired standard, but that everybody wants to use anyway.

As an alternative, why not just have the page token encode `readsToSkip` (effectively, a cursor into the set of result indices) - this, together with some standard hash of the request (sans page token) could be used to access an in-memory cached state of the request. Which/whether/how long to cache such partly filled requests (with, effectively, stateful cursor positions) would be the server implementation's business, and transparent to the client. As a side benefit, this same approach could be used across all multi-page queries, whether of reads, variants, or whatever, independent of the details of the underlying data being served.

jeromekelleher:
@macieksmuga Making the requests stateful would make our lives a _lot_ easier in some respects. It would be so great if we could just store a map of pageTokens to generator objects, so that we could just resume iteration over the low-level objects from where we left off. There would be some  low-level issues to resolve (like, will PySAM allow us to have multiple concurrent iterators over a single variant and alignment file?), and we'd also need to get a bit more defensive in terms of the number of concurrent clients we'll allow. These probably wouldn't be very big issues, and could be resolved easily enough.

The main problem with this is, we can only ever use _one process_ for this to work. We'd have to ensure that all requests coming from a single client get routed to the same worker process, so that the correct generator would be available. A huge advantage of the current stateless approach is that we can have any number of server processes running across different machines and the client does not need to know anything about it. Using standard load balancing approaches, requests for the first and second pages might end up being routed to completely different machines. For horizontal scaling, this property is invaluable.

So, I'd _really_ like to keep this process stateless.

jeromekelleher:
OK, I've had a play with this and I'm sure the approach I'm thinking about (but probably explaining really badly) works. I've sketched out a demo of this here: https://github.com/jeromekelleher/server/tree/reads-demo. Here is the diff from master: https://github.com/ga4gh/server/compare/master...jeromekelleher:reads-demo



I've tried this out using some real data, and I get the same output regardless of pageSize (the CLI is a very useful place to develop this sort of thing). There aren't that many reads that have the same position, but I guess this will be the case with most input data.

dcolligan:
Ok, yes, I agree that implementation works.  I'll incorporate it into my branch.

dcolligan:
Pushed a new change with @jeromekelleher 's paging algorithm

macieksmuga:
@jeromekelleher as long as the server is a front-end to indexed files, what you say makes perfect sense. Thanks for explaining your thinking behind this!
Some of the assumptions may change, however, as we implement a graph/database backend (for v0.6 schemas).

ga4gh/backend.py
```diff
@@ -217,6 +219,62 @@ def variantSetsGenerator(self, request):
...
+        readGroupId = request.readGroupIds[0]
+        try:
+            readGroup = self._readGroupIdMap[request.readGroupIds[0]]
+        except KeyError:
+            raise exceptions.ReadGroupNotFoundException(readGroupId)
+        startPosition = request.start
+        equalPositionsToSkip = 0
+        if request.pageToken is not None:
+            startPosition, equalPositionsToSkip = self.parsePageToken(
+                request.pageToken, 2)
```
macieksmuga: I believe I found a case where this algorithm fails, in that it could return reads outside the requested region.
Example:

Reads present on reference (with their ranges [start,end]):
`R1: [1, 100], R2: [2, 40], R3: [3, 100]`

Original client request:
Region: `[50,100]`, pagesize = `1`, no pageToken
 
Objects from `readGroup.getReadAlignment(…)`:
`R1, R3`
(R2 does not intersect requested range) so, first response is 
`[R1, pageToken=2:0]`

Next client request:
Region: `[50,100]`, pagesize = `1`, pageToken =` 2:0`

Objects from `readGroups.getReadAlignment(…)` (with startPosition reset from 50 to 2):
`R2, R3`
(R2 now intersects adjusted range), so second response is incorrect, contains spurious R2:
`[R2, pageToken=3:0]`

A solution would be to add an extra check in the first `while` loop below, to see if the read even intersects with the original requested region (substitute `readAlignment.end` here with whatever syntax is appropriate):
```
while getPosition(readAlignment) < startPosition or readAlignment.end < request.start:
    # etc...
```

jeromekelleher: Yes, very well spotted @macieksmuga! We should incorporate this check. We're going to need to think quite hard about how to test this --- I suspect a good start will be to make a read simulator which will generate all these nasty cases for us. This can be done after the beta though --- we're only promising initial support for reads in the beta release.

jeromekelleher:
This looks great, thanks @dcolligan. I'm +1 after we add the additional check that @macieksmuga pointed out.

ga4gh/backend.py
dcolligan: Testing should definitely not be done through a simulator (either the end to end one we have implemented or otherwise).  We should just create a discrete set of cases that exercise every branch.  But that's for another PR.

dcolligan:
Pushed a change with @macieksmuga 's improvement.

ga4gh/backend.py
macieksmuga: @dcolligan I agree! This kind of thing (specifically crafted datasets to test for a particular pitfall) should be compiled into a suite of regression tests.

ga4gh/backend.py
```diff
@@ -215,6 +217,68 @@ def variantSetsGenerator(self, request):
...
+        iterator = readGroup.getReadAlignments(
+            request.referenceName, request.referenceId, startPosition,
+            request.end)
+        readAlignment = next(iterator, None)
+        if request.pageToken is not None:
+            # First, skip any reads with position < startPosition
+            # or endPosition < startPosition
+            while (getPosition(readAlignment) < startPosition or
+                   getEndPosition(readAlignment) < startPosition):
```
macieksmuga: Since `startPosition` here may already be adjusted to something else than `request.start`, this line will need to read:
```
getEndPosition(readAlignment) < request.start):
```

macieksmuga: Otherwise +1

dcolligan:
My bad, fixed the `request.start` line @macieksmuga

macieksmuga:
+1
FWIW, this issue pulled me into the pagination weeds, and I'm glad for it. Thanks @jeromekelleher @dcolligan for being patient with me!

jeromekelleher:
I think this is ready to merge after a rebase.

dcolligan:
@jeromekelleher  merge me!

PR #274, opened 2015-03-20
Added avrotools and tests

dcolligan:
Intended uses of this tool: I was thinking that if the server is in DEBUG mode we could add information from ValidationTool to the request / response validation error info.

In any event, this is worth including in some form, since figuring out what is failing validation is very useful.

### DIFF
--- a/ga4gh/avrotools.py
+++ b/ga4gh/avrotools.py
@@ -1,0 +1,104 @@
+"""
+Provides additional functionality based on avro
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import json
+import avro.io
+
+import ga4gh.protocol as protocol
+
+
+class ValidationToolException(Exception):
+    """
+    Something that went wrong in the ValidationTool
+    """
+
+
+class ValidationTool(object):
+    """
+    Provides additional validation methods based on avro
+    """
+    @classmethod
+    def getInvalidFields(cls, class_, jsonDict):
+        """
+        Takes a ProtocolElement subclass class_ and a jsonDict instantiation
+        of that class.  Returns a dictionary whose keys are the names of
+        fields in the jsonDict that failed validation.  The values of the
+        returned dictionary are the jsonDict's values for those fields.
+
+        This is an expensive function to run and shouldn't be on the main
+        data path.
+
+        TODO is there an easier/faster way to do this than generating
+        an on-demand schema for every field we want to test?
+        """
+        if not issubclass(class_, protocol.ProtocolElement):
+            message = "class '{}' is not subclass of ProtocolElement".format(
+                class_.__name__)
+            raise ValidationToolException(message)
+        invalidFields = {}
+        schema = class_.schema
+        schemaJson = schema.to_json()
+        schemaJsonFields = schemaJson['fields']
+        cls._ensureNoExtraFields(schemaJsonFields, jsonDict)
+        for field in schemaJsonFields:
+            fieldName = field['name']
+            try:
+                fieldValue = jsonDict[fieldName]
+            except KeyError:
+                message = "jsonDict does not contain field '{}' for " \
+                    "class '{}'".format(fieldName, class_.__name__)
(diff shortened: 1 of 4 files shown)


ga4gh/avrotools.py
```diff
@@ -0,0 +1,83 @@
...
+                invalidFields[fieldName] = fieldValue
+        return invalidFields
+
+    @classmethod
+    def _createFieldSchema(cls, schema, fieldName):
+        schemaJson = schema.to_json()
+        # TODO this is an ugly hack
+        # I think GAReadAlignment.nextMatePosition's schema is not getting
+        # written correctly.
+        # In the schema its type attribute is:
```
jeromekelleher: This sounds like a bug to me. Can you verify that this is the case, file an issue if so?

jeromekelleher:
Very nice @dcolligan, I like it a lot.  It would be good to get an idea of what's going wrong with `nextMatePosition` and to try and fix that if possible.

Also, it would be nice to get a bit more test coverage. Some ideas:

- Test valid instances of all the types, and verify the dict is empty.
- Verify the exception state when we pass in a non-ProtocolElement.
- What happens when we send in a JSON dict for something completely different? I.e., what happens when we pass in the empty dict, or something with a whole bunch of extra fields?

dcolligan:
@jeromekelleher pushed a new commit with your suggestions

ga4gh/avrotools.py
dcolligan: Sure.  This is a schemas issue so I don't think it should hold up getting this change in.  We can work around it for the time being.

jeromekelleher: OK, agreed.

tests/unit/test_avrotools.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        for class_ in self.getProtocolElementSubclasses():
+            if class_ not in self.protocolElementSubclassesNonDefault:
+                self._assertGetInvalidFieldsResultEmpty(class_)
+
+    def testNonDefaultProtocolObjects(self):
+        """
+        Test that any protocol object that can not pass validation with
+        a default construction does so with additional help
+        """
```
jeromekelleher: Can we avoid the hard-coded fields here by instantiating a valid element using `tests.utils.InstanceGenerator`? We should then be able to loop over all classes automatically. Something like:
```python
for class_ in protocol.getProtocolClasses():
    instance = class()
    if not class_.validate(instance.toJsonDict()):
        validInstance = utils.InstanceGenerator()
        # Get the diff between the two dicts and use these to check your method
```
I'd like to define `protocol.getProtocolClasses()` and start using this whenever we're looping over all the subclasses of ProtocolElement. We could do this for all the tests above too, instead of just using `GACall`.

ga4gh/avrotools.py
dcolligan: Filed: https://github.com/ga4gh/schemas/issues/262

dcolligan:
@jeromekelleher added your suggestions; tests are now cleaner, but a bit slow (.5 secs to execute everything in `test_avrotools.py`).  I think that's fine.

tests/unit/test_avrotools.py
```diff
@@ -0,0 +1,55 @@
...
+import ga4gh.protocol as protocol
+import tests.utils as utils
+
+
+class TestValidationTool(unittest.TestCase):
+    """
+    Tests the ValidationTool
+    """
+    def testNonProtocolElement(self):
+        """
```
jeromekelleher: The comments are great, but can we change these to non-docstrings please (i.e., just ordinary comments)? Unittest uses the docstrings when you run nosetest -v, and it can make it hard to figure out what test is doing what.

jeromekelleher:
Excellent, thanks @dcolligan. I'm +1 modulo the tiny nit re docstrings.

dcolligan:
@jeromekelleher pushed update to address nit

jeromekelleher:
Thanks @dcolligan, merging.

PR #276, opened 2015-03-23
Adaptive pagesize and max request size.

jeromekelleher:
There are some serious issues with the current paging implementation:

1. It is trivial to perform a DOS/memory starvation attack. Simply provide a very large page size, and we will blindly try to fill it up until memory runs out.
2. There is no easy way to control latency. The only knob we have is the pageSize, but this is very crude as not all obejcts are equal. The pageSize controls the number of objects a page contains, _not_ the absolute size of the page in terms of bytes. Since objects like `Variant` can contain an arbitrarily large number of embedded objects (`Call` in the case of `Variant`) there's no way to know just how much memory we require.

Problem (1) could conceivably be tackled just by using the pageSize, but it would always be slightly tricky (what if there was 100k samples in the VCF?).  We could specify a MAX_PAGE_SIZE configuration variable, for example, but an administrator would have to work out roughly how much memory each object would require in worst-case and then figure out how many of these would correspond to a reasonable amount of memory. This puts a burden on the server admin, and is error prone. Problem (2) could also be tackled using this approach, but it's also quite tricky and error prone.

This is all a bit of a waste of time though, since all we really want to do is put a limit on the amount of *memory* each request is allowed to consume. This has the effect of solving (1) immediately, since an administrator can effectively cap the memory consumption per request. We also solve (2) which is good  for clients also, since they no longer need to worry about setting the appropriate page size to handle latency. The client knows they will be sent back appropriate sized packets of data, and does not need to worry about how to calculate the best page size for a given type of data. 

This PR adds a `MAX_RESPONSE_LENGTH` configuration variable, and sequentially encodes the JSON. Below we compare the time and memory requirements for requesting all calls in a range of 1000G data using a standard page size of 100 in the current master, and with a `MAX_RESPONSE_SIZE` of 1M in the current PR. 

Before:
```
$ /usr/bin/time -v python server_benchmark.py -c'*' --repeatLimit=1 --pageLimit=1000 1000g_2011.wt
31.87
        Command being timed: "python server_benchmark.py -c* --repeatLimit=1 --pageLimit=1000 1000g_2011.wt"
        User time (seconds): 32.13
        System time (seconds): 0.61
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:32.81
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 422648
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 258505
        Voluntary context switches: 1
        Involuntary context switches: 64
        Swaps: 0
        File system inputs: 0
        File system outputs: 48
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```
After:
```
$ /usr/bin/time -v python server_benchmark.py -c'*' --repeatLimit=1 --pageLimit=1000 1000g_2011.wt
20.69
        Command being timed: "python server_benchmark.py -c* --repeatLimit=1 --pageLimit=1000 1000g_2011.wt"
        User time (seconds): 21.00
        System time (seconds): 0.56
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:21.61
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 75784
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 251632
        Voluntary context switches: 1
        Involuntary context switches: 62
        Swaps: 0
        File system inputs: 0
        File system outputs: 0
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

We see we have reduced the CPU time by about 1/3 and reduced the memory footprint from ~420M to ~75M. This would seem like a clear win, and you might be wondering why I'm going to so much trouble to justify this change. There was a long discussion on this in #10, and there is the question of whether we should be making optimisations like this in the reference implementation.

So, pros:
1. We are (more) secure against memory starvation attacks.
2. We can control end-to-end latency much better.
3. Admins can get much better control over the memory usage of servers.
4. We are substantially faster and use less memory (over default page sizes on 1000G data).

Cons:
1. We break layering a little, and have some hand-written JSON encoding.

This PR is a proposal, and might be controversial. If people generally like it I'll update with some test cases for the new functionality.

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -34,6 +34,7 @@ class AbstractBackend(object):
         self._requestValidation = False
         self._responseValidation = False
         self._defaultPageSize = 100
+        self._maxResponseLength = 2**20  # 1 MiB
 
     def getVariantSets(self):
         """
@@ -67,12 +68,10 @@ class AbstractBackend(object):
         return values
 
     def runSearchRequest(
-            self, requestStr, requestClass, responseClass, pageListName,
-            objectGenerator):
+            self, requestStr, requestClass, responseClass, objectGenerator):
         """
         Runs the specified request. The request is a string containing
-        a JSON representation of an instance of the specified requestClass
-        in which the page list variable has the specified pageListName.
+        a JSON representation of an instance of the specified requestClass.
         We return a string representation of an instance of the specified
         responseClass in JSON format. Objects are filled into the page list
         using the specified object generator, which must return
@@ -90,19 +89,18 @@ class AbstractBackend(object):
             request.pageSize = self._defaultPageSize
         if request.pageSize <= 0:
             raise exceptions.BadPageSizeException(request.pageSize)
-        pageList = []
+        responseBuilder = protocol.SearchResponseBuilder(
+            responseClass, request.pageSize, self._maxResponseLength)
         nextPageToken = None
         for obj, nextPageToken in objectGenerator(request):
-            pageList.append(obj)
-            if len(pageList) >= request.pageSize:
+            responseBuilder.addValue(obj)
+            if responseBuilder.isFull():
                 break
-        response = responseClass()
-        response.nextPageToken = nextPageToken
-        setattr(response, pageListName, pageList)
-        responseDict = response.toJsonDict()
-        self.validateResponse(responseDict, responseClass)
+        responseBuilder.setNextPageToken(nextPageToken)
+        responseString = responseBuilder.getJsonString()
+        self.validateResponse(responseString, responseClass)
         self.endProfile()
-        return response.toJsonString()
+        return responseString
 
     def searchReadGroupSets(self, request):
         """
@@ -111,7 +109,7 @@ class AbstractBackend(object):
         """
         return self.runSearchRequest(
             request, protocol.GASearchReadGroupSetsRequest,
-            protocol.GASearchReadGroupSetsResponse, "readGroupSets",
+            protocol.GASearchReadGroupSetsResponse,
             self.readGroupSetsGenerator)
 
     def searchReads(self, request):
@@ -121,7 +119,7 @@ class AbstractBackend(object):
         """
         return self.runSearchRequest(
             request, protocol.GASearchReadsRequest,
-            protocol.GASearchReadsResponse, "reads",
+            protocol.GASearchReadsResponse,
             self.readsGenerator)
 
     def searchReferenceSets(self, request):
@@ -131,7 +129,7 @@ class AbstractBackend(object):
         """
         return self.runSearchRequest(
             request, protocol.GASearchReferenceSetsRequest,
-            protocol.GASearchReferenceSetsResponse, "referenceSets",
+            protocol.GASearchReferenceSetsResponse,
             self.referenceSetsGenerator)
 
     def searchReferences(self, request):
@@ -141,7 +139,7 @@ class AbstractBackend(object):
         """
         return self.runSearchRequest(
             request, protocol.GASearchReferencesRequest,
-            protocol.GASearchReferencesResponse, "references",
+            protocol.GASearchReferencesResponse,
             self.referencesGenerator)
 
     def searchVariantSets(self, request):
@@ -151,7 +149,7 @@ class AbstractBackend(object):
         """
         return self.runSearchRequest(
             request, protocol.GASearchVariantSetsRequest,
-            protocol.GASearchVariantSetsResponse, "variantSets",
+            protocol.GASearchVariantSetsResponse,
             self.variantSetsGenerator)
 
     def searchVariants(self, request):
@@ -161,7 +159,7 @@ class AbstractBackend(object):
         """
         return self.runSearchRequest(
             request, protocol.GASearchVariantsRequest,
-            protocol.GASearchVariantsResponse, "variants",
+            protocol.GASearchVariantsResponse,
             self.variantsGenerator)
 
     def searchCallSets(self, request):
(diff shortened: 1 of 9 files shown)


ga4gh/protocol.py
```diff
@@ -23,6 +24,63 @@ def convertDatetime(t):
...
+    """
+    A buffer to hold the string representation of protocol Response
+    objects. This allows us to sequentially build responses, and
+    reduces the number of live objects required to build response
+    objects.
+    """
+    def __init__(self, pageListName, pageSize, maxResponseLength=None):
+        self._pageListName = pageListName
+        self._pageSize = pageSize
+        self._maxResponseLength = 2**32
```
dcolligan: this constant is different from the one in backend.py ... is that a mistake?

macieksmuga: @dcolligan  I don't see a `maxResponseLength` constant (or anything similar) in `backend.py`... which one do you mean?

dcolligan:
Looks good.  The hand-encoding JSON looks straightforward enough and serves an important purpose to an extent that I don't really mind it.  I would like to see some unit tests for SearchResponseBuffer, though.

The only drawback I see is that if we have enormous objects, since the implementation guarantees we return at least one.  That would effectively get around this limit that the server admin sets.  I suppose we need that guarantee for the server to function; the alternative would be to just bail with an error (object too big to return for the max response length!) if a single object surpasses the max response length.

ga4gh/protocol.py
dcolligan: https://github.com/ga4gh/server/pull/276/files#diff-2dcd0fe257433a280068e998baf8eab4R37

jeromekelleher: Yeah, this is a bit of a hack. I'm just trying to cover the case when someone wants to make a `SearchResponseBuffer` but doesn't care about the `maxResponseLength`. I'll  just remove the default I think, it doesn't really help.

jeromekelleher:
Good point @dcolligan. I think we'd need to deal with that slightly differently, as we've already made the huge string by the time we've realised it's too big in this case. However, I think we'll probably want to break the layering a bit more at some point and start writing the embedded objects to the buffer directly (we're still using way too much memory). We can then catch this case directly and raise an exception when it happens.

I'll make some test-cases tomorrow if no-one objects to what's here by then.

ga4gh/backend.py
```diff
@@ -90,19 +91,18 @@ def runSearchRequest(
             request.pageSize = self._defaultPageSize
         if request.pageSize <= 0:
             raise exceptions.BadPageSizeException(request.pageSize)
-        pageList = []
+        response = protocol.SearchResponseBuffer(
```
macieksmuga: Interesting. So `response` is no longer explicitly an instance of `responseClass` as best as I can tell, even though it passes validation against the `responseClass` later on (with the caveat of converting the string to a dict via `json.loads`).

macieksmuga:
I'm finishing up with the test cases for the current `backend.py` today, would appreciate getting that PR in before you add more test cases to it, @jeromekelleher. 

I very much agree with the need for what you're proposing in this PR, but I'm having a bit harder time tracing the delegation of responsibility through the code with this setup (what can I say, layering helps). 

Thus, +1 on the concept behind this PR, I'll need to think about the implementation details just a tad more.

ga4gh/backend.py
jeromekelleher: Good point @macieksmuga --- would `SearchResponseBuilder` be better? i.e.,
```
responseBuilder = protocol.SearchResponseBuilder(...)
```

macieksmuga: @jeromekelleher Actually your suggestion gets to the core of my confusion. I'd have nothing against that, except that the new `protocol.SearchResponseB*er` doesn't take a `responseClass` argument (explicitly or implicitly, as far as I can tell).
Which makes me wonder... why do we even have that class in the first place?

jeromekelleher: Well, we're  assuming a certain structure of the response object in the `SearchResponseBuilder`, which the `SearchXResponse` class makes more explicit. Not much more explicit, mind, because we have to pass the `pageListName` variable around. Our lives would have been easier and we would have much cleaner code if all the `SearchXResponse` classes had the same structure and used `values` (or something) instead of `variants`, `callSets`, etc. etc. But they don't, so we need these ugly hacks.

We still need the class in any case though, because we want to validate against it: we should be able to verify that our results conform to the schema.

jeromekelleher:
I've pushed an update @dcolligan  and @macieksmuga. I've tackled some long-standing ugliness by making a new superclass for the `SearchRequest` and `SearchResponse` objects, and fiddled with the protocol processing code to identify these at schema generation time. Does this clarify things? 

I've also added some test cases. I think we should merge this after #274 as there will be some merge conflicts, and this uses some code @dcolligan introduces there.

dcolligan:
Code looks good, but tests are failing.

jeromekelleher:
Tests are failing because our CORS tests have started failing. I suggest a quick merge of #277 followed by analysis at #278.

macieksmuga:
Thanks @jeromekelleher! I think the intention behind some of the code is now clearer, and I applaud the removal of obvious redundancies (and potential sources for error), such as:
```
         return self.runSearchRequest(
             request, protocol.GASearchReferenceSetsRequest,
-            protocol.GASearchReferenceSetsResponse, "referenceSets",
+            protocol.GASearchReferenceSetsResponse,
             self.referenceSetsGenerator)
```

jeromekelleher:
Excellent! Are we ready to vote on this then?

dcolligan:
+1

Issue #271, opened 2015-03-20
Input data sanitisation required for pysam


jeromekelleher:
Pysam throws an exception if we provide numbers that are too large. We should fix this by finding out what the largest acceptable value is, and then setting our input values to this if any larger values are provided. More generally, we should sanitise the values that come in before we pass them to pysam. We should check the values of `referenceName`, `startPosition` and `endPosition` before we call `fetch` in both `getVariants` and `getReadAlignments`. Since this is shared functionality, we should push this into a superclass of the relevant classes, defined in `datamodel.__init__`.

For sanitisation, we can
1. Check that start and end are positive. Raise an exception if negative (or silently change to zero? It's arguable what the correct behaviour is here)
2. Check that start and end are < pysamMaxInteger. Silently change to pyamMaxInteger if greater.
3. Check that referenceName is a string of 'reasonable' length, and throw an exception if not. (Not sure what a reasonable length is: 1K?)

We should have unit tests for all of these too.

jeromekelleher:
Fixed in #279

PR #279, opened 2015-03-24
validation POC

dcolligan:
Issue #271

### DIFF
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -14,6 +14,7 @@ import pysam
 
 import ga4gh.protocol as protocol
 import ga4gh.exceptions as exceptions
+import ga4gh.datamodel as datamodel
 
 
 class SamCigar(object):
@@ -67,7 +68,7 @@ class SamFlags(object):
         flagAttr |= flag
 
 
-class AbstractReadGroupSet(object):
+class AbstractReadGroupSet(datamodel.DatamodelObject):
     """
     The base class of a read group set
     """
@@ -109,7 +110,7 @@ class SimulatedReadGroupSet(AbstractReadGroupSet):
         self._readGroups.append(readGroup)
 
 
-class HtslibReadGroupSet(AbstractReadGroupSet):
+class HtslibReadGroupSet(datamodel.PysamSanitizer, AbstractReadGroupSet):
     """
     Class representing a logical collection ReadGroups.
     """
@@ -207,7 +208,7 @@ class SimulatedReadGroup(AbstractReadGroup):
         return alignment
 
 
-class HtslibReadGroup(AbstractReadGroup):
+class HtslibReadGroup(datamodel.PysamSanitizer, AbstractReadGroup):
     """
     A readgroup based on htslib's reading of a given file
     """
@@ -231,6 +232,8 @@ class HtslibReadGroup(AbstractReadGroup):
             raise exceptions.BadReadsSearchRequestBothRefs()
         if referenceId is not None:
             referenceName = self._samFile.getrname(referenceId)
+        referenceName, start, end = self.sanitizeAlignmentFileFetch(
+                referenceName, start, end)
         # TODO deal with errors from htslib
         readAlignments = self._samFile.fetch(referenceName, start, end)
         for readAlignment in readAlignments:
--- a/ga4gh/datamodel/__init__.py
+++ b/ga4gh/datamodel/__init__.py
@@ -5,3 +5,84 @@ data in existing formats into GA4GH protocol types.
 from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
+
+import ga4gh.exceptions as exceptions
+
+
+class DatamodelObject(object):
+    """
+    Superclass of all datamodel types
+    """
+    def __init__(self):
+        # TODO move common functionality into this class from subclasses
+        pass
+
+
+class PysamSanitizer(object):
+    """
+    Provides various sanitization methods
+    """
+    samMin = 0
+    samMaxStart = 2**30 - 1
+    samMaxEnd = 2**30
+
+    vcfMin = -2**31
+    vcfMax = 2**31 - 1
+
+    maxStringLength = 2**10  # arbitrary
+
+    @classmethod
+    def sanitizeVariantFileFetch(cls, contig=None, start=None, stop=None):
+        if contig is not None:
+            contig = cls.sanitizeString(contig, 'contig')
+        if start is not None:
+            start = cls.sanitizeInt(start, cls.vcfMin, cls.vcfMax, 'start')
+        if stop is not None:
+            stop = cls.sanitizeInt(stop, cls.vcfMin, cls.vcfMax, 'stop')
+        if start is not None and stop is not None:
+            cls.assertValidRange(start, stop, 'start', 'stop')
+        return contig, start, stop
+
+    @classmethod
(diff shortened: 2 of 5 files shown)


dcolligan:
I've implemented all the functionality here, but I'm not sure what kind of interface we want.

It seems like validation shouldn't be a property of the datamodel objects themselves.  What we really want is to protect bad requests from going into pysam, right?

We should probably have a class hierarchy with a base class of `PysamRequest` or something which has a method `validate` that is overridden for each concrete subclass implementation.

Most validation frameworks are even more general.  Do we want to go even more general with this?

http://stackoverflow.com/questions/12235120/which-python-module-you-use-for-simple-data-validation

jeromekelleher:
I'm not sure that Validation is what we're trying to achieve here. We've already validated the data (it conforms to the protocol), so I don't think raising exceptions is what we want to do. We're just trying to

1) Protect the client from exceptions that will be raised if we send too-large values to pysam;
2) Protect lower-level code from being handed arbitrary strings fresh-in from the internet.

For 1, we just want to silently change numbers > n to n, where n is the largest value pysam will accept. For 2, we can either

1) Check for a given length and raise an exception if a passed in string is bigger than k;
2) Silently truncate the input string so it is of length k before passing it to pysam.

I don't have a strong opinion either way here. If we do decide to raise an exception on these types of issues, they should be placed in the `exceptions` module and somewhere in the hierarchy so they will be handled correctly when being passed back to the client.

dcolligan:
We're trying to identify and do something (TBD what) when a request passes schema validation but would fail a validation check that is a superset of schema validation.

Why don't we raise an exception in any of these cases instead of trying to guess what the intent of the user was?  Let's fail fast with informative messages so we don't bake in an implicit API that may change with later versions of pysam, etc.

Can you lay out your vision of the interface for this kind of "validation" (or whatever you want to call it)?

jeromekelleher:
I'm a user, and I send a perfectly valid request for variants, (start=0, end=2**63). I currently get back an error, which is completely cryptic and nothing at all to do with the protocol. We shouldn't extend the semantics of the protocol to throw errors when values we are provided that are perfectly legitimate happen to not work with our underlying implementation. For example, should we raise an error when we are provided with a referenceName that doesn't exist, or should we just return no values? This isn't really specified in the protocol, but at the moment the implicit assumption is we return an empty list.

All we really want to do here is:

1. Fix the problem we started off with;
2. Do something sensible when someone sends us a very long string maliciously.

So, something simple would be
```
def sanitisePysamCoordinates(start, end):
     if start > blahblah
          newstart = blahblah
     if end > blahblah
         newEnd = blahblah
    return newStart, newEnd
```

diekhans:
IMAO, the behavior should be a documented as part of the protocol.
So we decide what we think should be right and put in a schema
pull request.  We should have conformance tests as well.

For this particular ones:

#1 - out of range - I am unsure on this one. I think there should be
a special indicator for end of chromosome (NULL?).  If it's an error to
specify a number of the end, I somewhat think so.

#2 invalid referenceName should generate a clear error.  This
can be so confusing to debug otherwise.

jeromekelleher:
OK, my bad. I have been conflating two different topics here, which is very confusing. There's validation on the one hand, which we should tackle, and @dcolligan has made a good start on implementing here. However, we do need to have protocol level support for this also, and the behaviour for out-of-bounds inputs should be precisely specified.

On the other hand, there's a short term desire for the server to behave sensibly in the condition I've outlined above (end=2**63), and also not to allow potentially dangerous strings to be passed to lower level libraries. This is what I meant by 'sanitisation' originally.  I think we should try to tackle these two specific problems first, and then get some definitions of what should happen in out-of-bounds conditions in the protocol at a later date.

dcolligan:
Ok, so let's just solve the issue of sanitization in this PR.

This still leaves unanswered what the interface for that should be.  Putting this functionality in the `datamodel` hierarchy doesn't feel right to me.  What we're essentially asking for is a layer for the input data to pass through before it hits pysam, and pysam can be used in any number of places.

One heavyweight approach would be to wrap pysam in some module like `pysamwrapper` and then do the sanitization in that module.

Another, which I think I prefer, is to have some class like `PysamSanitizer` that provides class methods to sanitize data.  Of course, the downside of this would be that one might not remember to use it in the future before invoking pysam...

Thoughts?

macieksmuga:
Our `Htslib*` classes already effectively act as a wrapper for `pysam`, don't they? 
I like gathering any `pysam` cleanup related utilities into a single place, but prefer using a python module (`pysamUtils`?) to a class with only classmethods (the latter seems so... Java).

jeromekelleher:
A lightweight approach might be to just have methods `sanitizePysamInteger()` and `sanitizePysamString` in `DataModelObject` which is the superclass of VariantSet, ReadGroupSet, etc.  These take values are return sanitised values. We must also think about encoding for the strings, so this is actually not a bad interface. OK, we might forget to call them, but there aren't that many places we're actually interfacing with pysam.

macieksmuga:
I'm with @jeromekelleher on adding `sanitize*` methods to `DataModelObject`, although I'm wondering if this kind of functionality wouldn't be better served by a mix-in to be added to Htslib-specific classes as opposed to a global data model root object (that may see other backends with different limitations in the future).

jeromekelleher:
Sure, we can add in a `HtslibDataObject` class (or something like that) which does these things. I've no problem with multiple inheritance once the 'branch' classes are simple and don't involve much state.

dcolligan:
Pushed an update.  I decided to make them classmethods of a separate class since they are used in lots of places, not just classes derived from DatamodelObject.  Writing tests now...

ga4gh/datamodel/reads.py
```diff
@@ -158,6 +160,9 @@ def getReadAlignments(self, referenceName=None, referenceId=None,
             raise exceptions.BadReadsSearchRequestBothRefs()
         if referenceId is not None:
             referenceName = self._samFile.getrname(referenceId)
+        referenceName, start, end = \
```
jeromekelleher: I kinda liked @macieksmuga's idea of a mixin here. If we use the mixin idea we could have
```python
referenceName, start, end = self.sanitizeAlignmentFileFetch(
     referenceName, start, end)
```
which is a lot neater I think. By defining it as a mixin we can use the class where ever we want.

dcolligan:
Ok, made it a mixin.  Pushed a new update.

ga4gh/datamodel/reads.py
```diff
@@ -123,7 +125,7 @@ def __init__(self, id_, dataDir):
             self._readGroups.append(readGroup)
 
 
-class AbstractReadGroup(object):
+class AbstractReadGroup(datamodel.PysamSanitizer, object):
```
macieksmuga: Could you move `PysamSanitizer` mixin to `HtslibReadGoup`?

macieksmuga:
General request - could you move the `PysamSanitizer` mixin from the `AbstractWhatever` classes to their `HtslibWhatever` children? This way, we'd keep the pysam dependency constrained to the Htslib part of the server.

dcolligan:
Pushed change with @macieksmuga 's suggestions

jeromekelleher:
I like it, +1.

(I think string encoding is something we still need to deal with here, but let's sort that out in a follow up).

jeromekelleher:
Can you rebase this please @dcolligan --- we've got some merge conflicts.

dcolligan:
done

dcolligan:
@jeromekelleher merge me!

Issue #283, opened 2015-03-24
unittest like assertAlmostEqual required for data driven tests.


jeromekelleher:
The variant data driven tests currently use a method `_floatsAgreeWithinTolerance` to compare floats that should be approximately equal. We should change this to use an API that is compatible with unittest, and move it into the superclass. Unless there is a good reason otherwise, we should a method identical to the approach used by unittest.

dcolligan:
Fixed in #284

PR #284, opened 2015-03-24
adds assertAlmostEqual and assertNotAlmostEqual

dcolligan:
Issue #283

### DIFF
--- a/tests/datadriven/__init__.py
+++ b/tests/datadriven/__init__.py
@@ -170,6 +170,9 @@ class TestCase(object):
                 "{} is not less than or equal to {}".format(a, b))
 
     def assertRaises(self, exceptionType, func=None, *args, **kwargs):
+        """
+        Tests that calling func raises an exception of type exceptionType
+        """
         if func is None:
             return self.assertRaisesWith(exceptionType)
         exceptionRaised = False
@@ -184,6 +187,9 @@ class TestCase(object):
 
     @contextlib.contextmanager
     def assertRaisesWith(self, exceptionType):
+        """
+        Tests that a block of code raises an exception of type exceptionType
+        """
         exceptionRaised = False
         try:
             yield
@@ -193,6 +199,24 @@ class TestCase(object):
             raise AssertionError(
                 "exception of type {} not raised".format(
                     exceptionType.__name__))
+
+    def assertAlmostEqual(self, a, b, ndigits=7):
+        """
+        Assert that a and b are equal within ndigits digits
+        """
+        if round(a-b, ndigits) != 0:
+            message = "{} and {} not equal within {} digits".format(
+                a, b, ndigits)
+            raise AssertionError(message)
+
+    def assertNotAlmostEqual(self, a, b, ndigits=7):
+        """
+        Assert that a and b are not equal within ndigits digits
+        """
+        if round(a-b, ndigits) == 0:
+            message = "{} and {} equal within {} digits".format(
+                a, b, ndigits)
+            raise AssertionError(message)
 
 
 class DataDrivenTest(TestCase):
--- a/tests/datadriven/test_asserts.py
+++ b/tests/datadriven/test_asserts.py
@@ -108,3 +108,16 @@ class TestAsserts(unittest.TestCase):
         def func(*args, **kwargs):
             raise AssertionError()
         self.testCase.assertRaises(AssertionError, func, 1, a=2)
+
+    def testAssertAlmostEqual(self):
+        self.testCase.assertAlmostEqual(0.12345678, 0.12345679)
+        with self.testCase.assertRaises(AssertionError):
+            self.testCase.assertAlmostEqual(0.123456, 0.1234568)
+        with self.testCase.assertRaises(AssertionError):
+            self.testCase.assertAlmostEqual(0.12345678, 0.12345679, 10)
+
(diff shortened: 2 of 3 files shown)


tests/datadriven/__init__.py
```diff
@@ -194,6 +200,24 @@ def assertRaisesWith(self, exceptionType):
                 "exception of type {} not raised".format(
                     exceptionType.__name__))
 
+    def assertAlmostEqual(self, a, b, ndigits=7):
+        """
+        Assert that a and b are equal within ndigits digits
+        """
+        if round(a-b, ndigits) != 0:
```
macieksmuga: I'm worried that this implementation doesn't actually test the equality to within n significant digits. Consider something like `assertAlmostEqual(1.111111111E20, 1.111111112E20, 5)`, which should, to my mind, return True, but will return False in this implementation. I suggest something like:
```
if b!= 0 and round(abs((a-b)/b), ndigits) != 0
```
(handle case where `b == 0` separately)
I looked around to see if there was an easy way to just get the mantissa of a floating point number in python to avoid doing the division and couldn't find anything simple and consistent.

dcolligan: `round` is a built-in python function:
```
>>> help(round)
Help on built-in function round in module __builtin__:

round(...)
    round(number[, ndigits]) -> floating point number

    Round a number to a given precision in decimal digits (default 0 digits).
    This always returns a floating point number.  Precision may be negative.
```

The implementation is as is defined in the documentation for unittest: https://docs.python.org/2/library/unittest.html

macieksmuga: @dcolligan perhaps I mis-phrased my question. Do we want a `assertAlmostEqual` to test a comparison of two floats to within n significant digits, or n decimal points? 
I would argue for the former, but welcome feedback on the matter.

Consider:
```
>>> a = 1.11111111E20
>>> b = 1.11111112E20
>>> round(a-b, 5)
-999999995904.0
```

dcolligan: `round` only really makes sense when you are dealing with floating point numbers, and that's why round deals with decimal points.

But in any case, I got my inspiration for the implementation from the unittest spec... see the link above.

macieksmuga: @dcolligan point taken. But the mathematician in me bristles at the unittest spec implementation.

pgrosu: Hi Maciek,

I think you might be looking to have it like this:

```Python
def round_to_sigdigs( a_number, sig=7):
	return round(a_number, sig-int(floor(log10( a_number )))-1)
	
def assertAlmostEqual(self, a, b, ndigits=7):
	"""
	Assert that a and b are equal within ndigits digits
	"""
	if ( round_to_sigdigs(a, ndigits) - round_to_sigdigs(b, ndigits) ) != 0:
		 message = "{} and {} not equal within {} digits".format(
			 a, b, ndigits)
	raise AssertionError(message)
```

This is referenced from this:

http://stackoverflow.com/questions/3410976/how-to-round-a-number-to-significant-figures-in-python
	
Example:

```Python
>>> round_to_sigdigs(a,7) - round_to_sigdigs(b,7)
0.0
```
	
`p

macieksmuga: Clever, and perfect! 
Well... almost (caveat stackoverflow). Need to take `abs( a_number)` before applying `log10` to it, otherwise you error out on negative inputs. Thus:

```
import math

def round_to_sigdigs( a_number, sig=7):
    return round(a_number, sig-int(math.floor(math.log10(abs(a_number))))-1)
```
etc...

pgrosu: Cool!  Yeah,  I didn't think that we had negative numbers, but you're right that we should have it complete :)

thx,
~p

macieksmuga: @pgrosu I can assure you that negative numbers do indeed exist :)

All of the above being said, I'm going to defer to those of you who are already using `assertAlmostEqual` in the test code, as to whether this significant digit or the previously proposed decimal places implementation is more useful in practice...

jeromekelleher:
My take on this would be to either:

1. Copy unittest's implementation of this literally, and not worry too much about how correct it is (this has the advantage of being compatible, even if it's not entirely correct). 
2. Copy the implementation from GSL's (fcmp)[https://www.gnu.org/software/gsl/manual/html_node/Approximate-Comparison-of-Floating-Point-Numbers.html]. This is an implementation of Knuth's algorithm for approximate comparison from TAOCP.

I like @dcolligan's approach, as it fits (1) perfectly.

tests/datadriven/__init__.py
pgrosu: @macieksmuga, Hehehe, you're right, I realized I was thinking too locally on this one :)

shajoezhu:
I vote for (1) Copy unittest's implementation of this literally. In the test I ran in tests/datadriven/test_variants. the difference between a and b is either zero or a very small decimal number. so @dcolligan 's approach should work.

dcolligan:
Ok, so can I get some +1's / merge here?

jeromekelleher:
+1

macieksmuga:
FP arithmetic theory aside, I'm fine with whatever the principal consumers of this test code deem most useful. 
+1

Issue #265, opened 2015-03-19
GACall field phaseset unset 


shajoezhu:
currently using pysam tool to extract elements for GACall. (HtslibVariantSet)

jeromekelleher:
Can we get some more details on this please @shajoezhu?

bioinformed:
The pysam VCF/BCF code does not yet expose phase information.  As I
described in another thread, I am still considering how best to proceed:
https://github.com/ga4gh/server/issues/231#issuecomment-82643656

jeromekelleher:
We should fill this in ourselves in the mean time, so we have something reasonable. It's a bug if we're exporting a phaseset that's definitely wrong.

bioinformed:
Can you point me to the GA4GH API spec for how phase is communicated?

jeromekelleher:
I've never really understood what this means, but this is what we're going by:

http://ga4gh.org/documentation/api/v0.5.1/ga4gh_api.html#/schema/org.ga4gh.GACall

dcolligan:
Fixed in #285

PR #285, opened 2015-03-25
resolving issue #265

shajoezhu:
GACall field phaseset unset #265

### DIFF
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -516,13 +516,28 @@ class HtslibVariantSet(AbstractVariantSet):
                 self._updateCallSetIds(varFile)
                 self._chromFileMap[chrom] = varFile
 
-    def _convertGaCall(self, recordId, name, pysamCall):
+    def _convertGaCall(self, recordId, name, pysamCall, genotypeData):
         call = protocol.GACall()
         call.callSetId = recordId
         call.callSetName = name
-        call.genotype = list(pysamCall.allele_indices)
-        # TODO: phaseset is not set by pysam
+
+        # TODO:
+        # NOTE: THE FOLLOWING TWO LINES IS NOT THE INTENED IMPLEMENTATION,
+        ###########################################
         call.phaseset = None
+        call.genotype, call.phaseset = convertVCFGenotype(
+            genotypeData, call.phaseset)
+        ###########################################
+
+        # THEY SHOULD BE REPLACED BY THE FOLLOWING, ONCE NEW PYSAM
+        # RELEASE SUPPORTS phaseset. AS WELL AS REMOVING genotypeData
+        # FROM THE FUNCTION CALL
+
+        ###########################################
+        # call.genotype = list(pysamCall.allele_indices)
+        # call.phaseset = pysamCall.phaseset
+        ###########################################
+
         call.genotypeLikelihood = []
         for key, value in pysamCall.iteritems():
             if key == 'GL' and value is not None:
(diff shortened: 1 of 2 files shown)


jeromekelleher:
I agree it's not very nice bypassing the PySAM api here @bioinformed, but we do need something that is approximately correct now. We're aiming for a beta release next week, and I'd rather have something which is right most of the time rather than definitely wrong all of the time. Do you agree with the unit tests we've made for `convertVCFGenotype` in https://github.com/ga4gh/server/blob/master/tests/unit/test_variants.py ?

I agree the right way to do things is in pysam. Would you be happy if we lodged an issue stating that we should change over to pysam's phaseset once it's released?

bioinformed:
@jeromekelleher: If you adopt this hack, please do not use benchmark information from this implementation to make any serious decisions.  The next pysam release -- ready in about week -- will have proper support for phase.  

The reason for my caution and delay is that there are several "legacy" interpretations of how to encode phase information into VCF files.  In particular, GATK's UnifiedGenotyper uses the concept of implicit phase blocks that have boundaries at each unphased genotype.  This means that the first unphased genotype prior to a phased genotype is part of the same implicit phase set.  Because of this, I've been researching current implementations to ensure that pysam's support is as sensible as possible.

Regarding the unit tests, they look reasonable.  But realize that in BCF and VCF it is valid and possible to have genotypes like "0/1|2".  With BCF it is even murkier, were it is technically valid to have genotypes like "0|" and "0/1/2/3|" (where phase is encoded as a bit per allele).

jeromekelleher:
Is your concern with this (short-term) hack in terms of the accuracy of the output or that it will hurt performance @bioinformed? (Or indeed, that it's potentially dangerous?) If it's performance, then I'm happy enough to take a fairly hefty performance hit if we're outputting reasonably accurate information. I appreciate the fact that it's all very murky and I would _much_ rather that this was handled by pysam, but I would like the phaseset information we have for next week's release to be correct (at least, most of the time). I don't want to put pressure on you to get the pysam code ready based on our time scales.

bioinformed:
My concern is that short-term hacks have a knack of becoming long-term.  So long as the current approach is temporary and the performance impact is understood, then I don't have any real objections.

jeromekelleher:
OK, understood. Thanks for clearing this up @bioinformed, it's great having your input here.

@shajoezhu, can you put in a big, loud disclaimer in the code stating that this is a hack please? We'll open an issue to get rid of it after the forthcoming pysam release after this is merged.

pgrosu:
@jeromekelleher, @shajoezhu - seems like a lot of work (disclaimer, PR, etc.) for something that will be released in about a week.  Just a thought...

jeromekelleher:
Nice disclaimers @shajoezhu!

+1

ga4gh/datamodel/variants.py
```diff
@@ -516,13 +516,28 @@ def _addFile(self, filename):
...
+            genotypeData, call.phaseset)
+        ###########################################
+
+        # THEY SHOULD BE REPLACED BY THE FOLLOWING, ONCE NEW PYSAM
+        # RELEASE SUPPORTS phaseset. AS WELL AS REMOVING genotypeData
+        # FROM THE FUNCTION CALL
+
+        ###########################################
+        # call.genotype = list(pysamCall.allele_indices)
+        # call.phaseset = pysamCall.phaseset
```
bioinformed: call.phaseset = pysamCall.get('PS', '*')

jeromekelleher: Thanks @bioinformed, we should fix this. Has this been followed up by anyone @dcolligan? If not we should fix this using @bioinformed's code and close the original issue.

dcolligan: Those lines are still in the code, so I'm going to go with no, no one has fixed this yet.

Original issue is here: https://github.com/ga4gh/server/issues/291

david4096:
Closed with https://github.com/ga4gh/server/commit/d9117b8bc0a80f6f3fbca49f73d8141ba12a7feb

Issue #267, opened 2015-03-19
Client always sends extraneous last request


dcolligan:
Since we always return a pageToken regardless if we are at the end of the object stream, the client sees a non-null nextPageToken and decides to send one last request, which always returns an empty payload from the server.

Not sure if there is a way to fix this generally (or if it is even worth it in the specific case, since we would need to do some kind of look-ahead logic with iterators), but it's worth discussing.

```
$ python client_dev.py -vv -O variants-search http://localhost:8000/v0.5.1 -V 1000g_2013.wt -r 1 -c HG03279
... previous requests ...
ga4gh.client: POST http://localhost:8000/v0.5.1/variants/search
ga4gh.client: json request:
ga4gh.client: {
    "callSetIds": [
        "HG03279"
    ],
    "end": 4294967294,
    "pageSize": 100,
    "pageToken": "0:138397",
    "referenceName": "1",
    "start": 0,
    "variantName": null,
    "variantSetIds": [
        "1000g_2013.wt"
    ]
}
ga4gh.client: json response:
ga4gh.client: {
    "nextPageToken": null,
    "variants": []
}
```

jeromekelleher:
This is a bug in the server --- we shouldn't send back a non-null pageToken unless there are records left.

dcolligan:
It looks like `variantsGenerator` and `callSetsGenerator` are the only methods affected by this issue.

jeromekelleher:
We should fix `variantsGenerator` to use the same algorithm as reads (hopefully reusing some code too), as the same complications due to interval search can occur. `callSetsGenerator` will need to be fixed in isolation.

jeromekelleher:
Fixed in #295 and #304.

PR #290, opened 2015-03-25
Fix callSetsGenerator and variantsGenerator...

dcolligan:
... so that they send a null nextPageToken on
what should be their last response to the client.

Issue #267

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -191,6 +191,28 @@ class AbstractBackend(object):
                 nextPageToken = str(currentIndex)
             yield object_.toProtocolElement(), nextPageToken
 
+    def _lookAheadObjectGenerator(self, request, generator):
+        """
+        Ensures that when the generator has exhausted its objects a
+        null pageToken is returned to the client.
+        """
+        iterator = generator(request)
+        obj = None
+        nextObj = None
+        pageToken = None
+        nextPageToken = None
+        try:
+            while True:
+                if nextObj is not None:
+                    obj, pageToken = nextObj, nextPageToken
+                else:
+                    obj, pageToken = iterator.next()
+                nextObj, nextPageToken = iterator.next()
+                yield obj, pageToken
+        except StopIteration:
+            if obj is not None:
+                yield obj, None
+
     def readGroupSetsGenerator(self, request):
         """
         Returns a generator over the (readGroupSet, nextPageToken) pairs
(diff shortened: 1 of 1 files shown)


ga4gh/backend.py
```diff
@@ -222,13 +244,29 @@ def variantsGenerator(self, request):
...
+        """
+        Internal variants generator; returns (variant, nextPageToken) pairs
+        """
         variantSetIds = request.variantSetIds
         startVariantSetIndex = 0
         startPosition = request.start
         if request.pageToken is not None:
             startVariantSetIndex, startPosition = self.parsePageToken(
                 request.pageToken, 2)
-        for variantSetIndex in range(startVariantSetIndex, len(variantSetIds)):
```
jeromekelleher: Why the line break here after range?

jeromekelleher:
I think I'd rather deal with this at the source. It'll be confusing if we have some of our generators having one behaviour (correct null pageToken for the last value) and others having different behaviour which need a wrapper. We need to update `variantsGenerator` so that it deals with  variants overlapping the page boundary correctly, and we might as well use the code we're developing for reads to do this. We should just fix callSetsGenerator.

dcolligan:
How is that confusing?  I think the code is pretty easy to follow.

Plus, this gives us a general solution to a problem that more than one of the generators have now, and more are likely to have in the future.  This is way better than having every generator implementing look-ahead logic.

jeromekelleher:
I think #295 fixes the callSets stuff (using the general code for lists of objects), and we will need to review variants in any case (hopefully reusing the logic we have for reads).

jeromekelleher:
I think this has been superseded now --- shall we close this PR @dcolligan, or are there bits you'd like to bring forward?

dcolligan:
I do like the idea of having a general construct for implementing look-ahead to ensure that we send a null nextPageToken with the last data object yielded.  I think that this would still be worth implementing.  Do you agree?  It looks like we would only use this in `variantsGenerator` and `readsGenerator` right now (the only generators not using `_topLevelObjectGenerator`).

jeromekelleher:
Well, variantsGenerator is going to need to change to use the same algorithm as readsGenerator because the same issues with overlapping intervals can occur (albeit with a much lower probability). So,  we'll be using the same code for both of these in any case. 

I agree that implementing read-ahead differently for the different iterators is a bad idea. I would suggest that the next time we look at the paging code (which we're definitely doing to need to do when we tackle datasets) we incorporate something like you've defined here.

dcolligan:
Ok I'll close the PR but keep the branch around

Issue #270, opened 2015-03-20
Robust exception handling required for variants and 'bad-data driven tests"


jeromekelleher:
We need custom exceptions for all of the error conditions in the `datamodel/variants.py` file. We also need to test these exceptions, however, since there's no point in making custom exceptions if we don't test them.

One way we might to this is to introduce a new directory in `tests` called `bad-data`, which contains data that is designed to exercise the error cases in our code. We'll need some extra test infrastructure for this --- should we call these `bad-datadriven`, and reuse the datadriven infrastructure?

macieksmuga:
Since different exceptions would presumably be generated by different bad data, such tests would need to be structured differently than the `datadriven` ones so far (i.e., we can't just run all of the exception tests over the totality of the test data - they'd need to be matched: This bad data triggered that exception).

Maybe make a directory of `regression-data` in general, that contains data that will only be run on *certain* specified tests, either ones that are expected to throw an error, or some other outlier behavior that's been problematic in the past?

jeromekelleher:
Yes, that's a good point @macieksmuga --- we will need a different structure. We'd need some sort of map from bad-data file to expected exception. What if we had a directory of files for each exception (that is, any file in that directory is expected to throw a given exception)? Our directory tree of bad-data could mirror the exception hierarchy.

jeromekelleher:
After discussion, it seems that this is a bit ambitious and over complicated. I think the simplest approach is to have a new directory `tests/bad_data/variants` and put some test variant sets in there. These should have descriptive directory names, and contain one (or more) dodgy VCFs.

Then, we have a new test module, `tests/unit/test_bad_data.py`, and we have a bunch of standard unittest.TestCase subclasses. We then have some simple mechanism for associating these tests with some data files (perhaps a class variable containing the list of dodgy data set ids?). For example, we might have something like
```python
class TestVariantSetMetadataMismatch(BadDataTest):
    setIds = ["metadataMismatch1", "metadataMismatch2"]

    def testInstantiation(self):
        for variantSetId in self.setIds:
             path =  self.getFullPath(variantSetId)
             self.assertRaises(
                 exceptions.MetadataMisatchException, variants.HtslibVariantSet, path)
```

macieksmuga:
@jeromekelleher I like it, but I could also see the need for some additional mechanism to pass specific `setIds` to individual tests within a class.

dcolligan:
Fixed in #294

PR #294, opened 2015-03-26
resolving issue #270

shajoezhu:
creating test for exception handling when parsing vcf files


ga4gh/datamodel/variants.py
```diff
@@ -451,9 +451,14 @@ def __init__(self, variantSetId, vcfPath, use_bcf=True):
         self._chromFileMap = {}
         self._metadata = None
         self._callSetIds = None
+        fileList = []
```
jeromekelleher: We could just use counter here: numVariantFiles = 0 initially, and test this at the end.

ga4gh/datamodel/variants.py
```diff
@@ -487,22 +490,25 @@ def _updateCallSetIds(self, variantFile):
...
+            if len(self._callSetIds) != len(variantFile.header.samples):
+                raise exceptions.InconsistentCallSetIdException(
+                    variantFile.filename)
+
         for sample in variantFile.header.samples:
             if checkConsistency:
-                if sample not in self._callSetIds:
-                    # TODO CHANGE EXCEPTION
-                    raise Exception("Inconsistent sample names in VCF")
+                if (sample not in self._callSetIds):
```
jeromekelleher: We don't need the brackets here.

tests/unit/test_faultydata.py
```diff
@@ -0,0 +1,100 @@
...
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import unittest
+
+import ga4gh.datamodel.variants as variants
+import ga4gh.exceptions as exceptions
+
+
+class FaltyDataTest(unittest.TestCase):
```
jeromekelleher: FaultyData

tests/unit/test_faultydata.py
```diff
@@ -0,0 +1,100 @@
...
+    def setUp(self):
+        self.testDataDir = "tests/faultydata/variants"
+
+    def getFullPath(self, variantSetId):
+        return "{0}/{1}".format(self.testDataDir, variantSetId)
```
jeromekelleher: We should use `os.path.join()` for this.

jeromekelleher:
This looks good, thanks @shajoezhu. One thing we need to change I think is the exception hierarchy. I think we need a split between the exceptions that can occur during the processing of a client request and those that occur during the server startup. What about:
```
BaseServerException +
     RuntimeException
     DataException
```
where RuntimeException is the superclass of all the current exceptions (and contains all the code currently in BaseServerException), and DataException is the superclass of all the exceptions you want to define here that cover dodgy input data.

Any thoughts @dcolligan and @macieksmuga?

shajoezhu:
Something like this?
image

tests/unit/test_faultydata.py
```diff
@@ -0,0 +1,101 @@
...
+                variants.HtslibVariantSet, variantSetId, path)
+
+
+class TestDuplicateCallSetId(FaultyDataTest):
+    """
+    THIS SECTION IS CURRENTLY NOT WORKING
+    It returns the following error:
+
+    [E::bcf_hdr_add_sample] Duplicated sample name 'S1'
+    Aborted (core dumped)
```
dcolligan: Maybe a bug in pysam/htslib?

dcolligan:
I support the RuntimeException / DataException split.  It would also be nice to add a README in each directory with data files to say how they were generated (especially for us non-bio types that may need to reproduce this data someday!).

tests/unit/test_faultydata.py
jeromekelleher: Yep.

macieksmuga:
I also support the base split (RuntimeException / DataException), but wonder if the data exception hierarchy below that isn't a bit too much. For example, if I get an `OverlappingException`, I'd need to check in the hierarchy to know that it's a `VcfBcfFileException`... instead, would it be clearer to have something like `OverlappingVcfFilesException` (with the understanding that BCF is binary VCF)? 

Also, I'm not sure what the `MalFormedException` is buying us - could we cut it out of the hierarchy?

Oh, and agree with @dcolligan about for README's on how the test data was generated.

jeromekelleher:
Yeah, I'm generally with @macieksmuga here - we should probably simplify the hierarchy a little bit. READMEs would be very useful too.

It'll be a bit of a mess rebasing this on upstream/master - sorry about that!

shajoezhu:
Thanks for the comment guys! Here is the updated exception hierarchy. @jeromekelleher @dcolligan @macieksmuga 
image

tests/unit/test_faultydata.py
bioinformed: It is an htslib bug where adding a duplicate sample id triggers an `abort()` if verbosity >= 2.  There are quite a few other places where `abort()` is called instead of returning an error code.

See htslib/vcf.c lines 79-83:

        if (hts_verbose >= 2)
        {
            fprintf(stderr, "[E::%s] Duplicated sample name '%s'\n", __func__, s);
            abort();
        }

jeromekelleher: Thanks for tracking this down @bioinformed --- that's pretty nasty behaviour from htslib! Out of interest, do you know if are there any plans for better error handling from htslib?

ga4gh/exceptions.py
```diff
@@ -90,7 +90,14 @@ def toProtocolElement(self):
 #####################################################################
 
 
-class BadRequestException(BaseServerException):
+class RuntimeException(BaseServerException):
```
jeromekelleher: We should move the httpStatus stuff into `RuntimeException` from `BaseServerException` I think. The DataExceptions occur when the server is starting up, and result in something being communicated back to the user via stderr or logs. The idea of HttpStatus doesn't mean very much here. RuntimeExceptions occur when we are serving HTTP requests, and they need all of the machinery for http status and error codes. I guess we should keep `message` in `BaseServerException` so that we have uniform interface for creating error messages.

shajoezhu:
Good point! :+1:

tests/unit/test_faultydata.py
bioinformed: @jeromekelleher, I'm preparing a PR for htslib to address this and a few other really awful places where `abort()` is used.  I'll let you know whether it is accepted.

jeromekelleher: Thanks @bioinformed --- let me know if I can be of any help on this.

ga4gh/exceptions.py
```diff
@@ -180,6 +187,60 @@ def getMessage(self):
...
+
+class DataException(BaseServerException):
+    message = "Faulty data found."
+
+
+class EmptyDirException(DataException):
+    message = "Diretory empty, no VCF/BCF file was found"
+
+
+class MalFormedException(DataException):
```
jeromekelleher: I think `MalformedInputException` would be a bit more descriptive here.

ga4gh/exceptions.py
```diff
@@ -180,6 +187,60 @@ def getMessage(self):
...
+class NotIndexedException(MalFormedException):
+    def __init__(self, fileName):
+        self.fileName = fileName
+
+    def getMessage(self):
+        self.message = "File {} must be indexed.".format(self.fileName)
+        return self.message
+
+
+class OverlappingException(MalFormedException):
```
jeromekelleher: This name doesn't really tell me much when I don't know the context. What about `OverlappingVcfException`? For a user, this might be very mysterious also. What about:
```python
class OverlappingVcfException(MalformedInputException):
    """
    Exception thrown when two VCF files within a VariantSet directory contain records
    for the same contig. 
    """
    def __init__(self, fileName, contig):
         self.message = (
               "VCF file '{}' contains records for contig '{}'. Other files in this "
               "VariantSet have records for this contig, and overlapping VCFs "
              "are not permitted.")
```

jeromekelleher:
Looks good @shajoezhu, thanks. It would be nice to see some docstring comments on the exception classes to explain to developers when they are expected to be thrown (see the example above for OverlappingVcfException).

Other than that, I think we just need to rebase and we're ready to merge.

jeromekelleher:
@dcolligan, can you merge this please if you are happy with it?

ga4gh/exceptions.py
```diff
@@ -72,6 +71,21 @@ def getErrorCode(cls):
...
+#
+#####################################################################
+
+
+class RuntimeException(BaseServerException):
+    """
+    exceptions that can occur during the processing of a client request
+    """
+    httpStatus = -1
+    message = "Bad request"
```
dcolligan: change this message to be something like BaseServerException's, since we should never be throwing it

ga4gh/exceptions.py
dcolligan: I agree with @jeromekelleher 

In any event, I would object to the name because 'Malformed' is one word, so the camel-casing is wrong

ga4gh/exceptions.py
```diff
@@ -190,6 +198,84 @@ def getMessage(self):
...
+    """
+    message = "Faulty entry found in data file."
+
+
+class NotIndexedException(MalFormedException):
+    """
+    VCF/BCF files must be indexed, with the same prefix.
+    """
+    def __init__(self, fileName):
+        self.message = (
```
dcolligan: It looks like the message is being set to a tuple here rather than a string... that doesn't seem right

(same with many exceptions below)

ga4gh/exceptions.py
```diff
@@ -190,6 +198,84 @@ def getMessage(self):
...
+
+
+class InconsistentMetaDataException(MalFormedException):
+    """
+    Exception thrown when two VCF files within a VariantSet direcotry
+    contain different metadata entries.
+    """
+    def __init__(self, fileName):
+        self.fileName = fileName
+        self.message = "Metadata of {} is not consistent with other files"
```
dcolligan: I doubt this message is going to be formatted the way you intend.  It will probably just be the first line because of the line breaks.  Adding a backslash at the end of the first two lines would probably fix this.

(correct me if I am wrong...)

dcolligan: same comment for various other messages in this file

tests/faultydata/variants/duplicated_sampleid/README.md
```diff
@@ -0,0 +1,6 @@
+Data file example_1.vcf.gz was copied from tests/data/variants/example_1, then change sample id S2 to S1
+
+Index file was generated by
+```bash
```
dcolligan: bash is a shell, and you're not using any bash-specific utilities here, so the mention of bash is superfluous.

I would format it as just:

```
$ tabix example_1.vcf.gz
```

The dollar sign indicating a command prompt

tests/faultydata/variants/inconsist_sampleid/README.md
```diff
@@ -0,0 +1,11 @@
+1kgPhase1chr1.vcf.gz is a soft link to tests/data/variants/1kgPhase1/chr1.vcf.gz
+Index file 1kgPhase1chr1.vcf.gz.tbi is a soft link to tests/data/variants/1kgPhase1/chr1.vcf.gz.tbi
+
+To generate 1kgPhase1chr2_removed_HG00531.vcf.gz and its index file, do the following:
+```bash
```
dcolligan: same comment about bash here

tests/faultydata/variants/inconsist_sampleid2/README.md
```diff
@@ -0,0 +1,12 @@
+Create data set in tests/faultydata/variants/inconsist_sampleid before 
+
+1kgPhase1chr2_removed_HG00531.vcf.gz and its index file are soft links to tests/faultydata/variants/1kgPhase1chr2_removed_HG00531.vcf.gz
+
+To generate 1kgPhase1chr3_removed_HG00533.vcf.gz and its index file, do the following:
+```bash
```
dcolligan: same comment about bash here

dcolligan:
Not merging because I think there are a lot of problems, mostly with the exception messages.  Unless I am misinterpreting something, it seems like a lot of the exception messages will be other than what is intended.

@shajoezhu , please confirm that every exception message you added and/or modified is formatting the way that you intend.  I can take another look after you do that and fix my nits.

ga4gh/exceptions.py
jeromekelleher: Nope, this works @dcolligan. The strings are automatically concatenated and the brackets are just there to make line continuation work (which is the PEP recommended way).

ga4gh/exceptions.py
jeromekelleher: This one is wrong all right. You need a to wrap some brackets around it.

Every time you use a backslash a puppy dies @dcolligan...

dcolligan: Nobody appreciates my art!  Backslashes, decorators... what next?

jeromekelleher: I think we should start again and write this in [whitespace](http://en.wikipedia.org/wiki/Whitespace_%28programming_language%29). It would make all these style choices so much easier!

shajoezhu:
hi @dcolligan , I have updated this PR, could you take another look please? Many thanks!

ga4gh/exceptions.py
dcolligan: I learn something new everyday...

tests/faultydata/variants/duplicated_sampleid/README.md
```diff
@@ -0,0 +1,6 @@
+Data file example_1.vcf.gz was copied from tests/data/variants/example_1, then change sample id S2 to S1
+
+Index file was generated by
+```
```
dcolligan: Actually, one other thing... does it make sense to have these files as .md files and format them in markdown?

I'm not sure if we should be formatting the README files (aside from the ones in the root directory) as anything other than .txt ... @jeromekelleher thoughts?

shajoezhu: I wasn't sure about this either, just did it anyway. I think many text editors do support syntax highlighting for markdown. so if one views the README with an editor, I think it is quite clear?

dcolligan: I don't think we should assume one's editor has markdown support.  Mine doesn't, and every text editor supports plaintext!

But I want to hear @jeromekelleher 's thoughts on this, although my inclination is towards .txt files.

jeromekelleher: I don't have a strong opinion here. You don't need to understand markdown to read it as plain text. We're already using restructured text, markdown and plain text for the READMEs in various places, so we're not doing anything new here either way. It's up to you guys!

dcolligan: ¯\_(ツ)_/¯

PR #295, opened 2015-03-26
Widespread changes to CallSet handling.

jeromekelleher:
This PR fixes up CallSet handling throughout the code, as it was a very random mess up to this point. I've ended up making a bunch of other changes, while attempting to share code between various parts, and the PR has gotten a lot bigger than I intended.

Some important points:

- I've gotten rid of the SetIdMap and SetIds lists and dicts in AbstractBackend and just used an OrderedDict instead for all the important object types. There is a performance drawback here, but I don't think it'll be significant for the moment, and we can hopefully think about the right way to do this later.
- CallSet handling is taken care of entirely in AbstractVariantSet now, so the same code is shared by SimulatedVariantSet and HtslibVariantSet. There's a lot of refactoring here as well to push code into the superclass.
- I've disconnected the WormtableVariantSet completely from this hierarchy, and it's now ready to be excised.
- I've deleted some old test cases on CallSets that weren't worth keeping around as they were too tightly integrated with the wormtable backend.

What do we think? Sorry to make such a big disruptive change now, but the CallSet stuff was really in a mess and the only way to fix it was to make big changes. @macieksmuga, I don't think there's much point in de-wormtabling before this is merged.


ga4gh/backend.py
```diff
@@ -23,14 +24,9 @@ class AbstractBackend(object):
...
-        self._variantSetIds = []
-        self._referenceSetIdMap = {}
-        self._referenceSetIds = []
-        self._readGroupSetIdMap = {}
-        self._readGroupSetIds = []
-        self._callSetIdMap = {}
-        self._callSetIds = []
+        self._variantSetIdMap = collections.OrderedDict()
+        self._referenceSetIdMap = collections.OrderedDict()
+        self._readGroupSetIdMap = collections.OrderedDict()
```
pgrosu: `OrderedDict` will be very slow - below is a link to a quick explaination:

http://stackoverflow.com/questions/18422995/why-is-ordereddict-10x-slower-than-dict-and-list

~p

dcolligan: Using the old method might be the right thing to do if there is indeed a 10x slowdown.

jeromekelleher: I can change back to the old method if you think it's important --- might bring the size of the PR down a bit too. Your call.

dcolligan:
Generally looks good, but I'd like some more discussion about the performance implications.

ga4gh/backend.py
pgrosu: Before reverting any changes, it might be good to test it first and confirm that there is indeed a large slowdown.

Paul

jeromekelleher:
OK, I'll revert to the old way. We should do this properly later and not in a hurry. Gimme a minute and I'll update.

jeromekelleher:
OK, done. Moving wormtable has just made matters worse, and now it's impossible to see what's changed. I'm going to have a go at diswormtabling the code and see what happens.

dcolligan:
Yes!  Slay the beast!

jeromekelleher:
It's done! That was super, super easy, and we kept test coverage at 91%: yay us!

ga4gh/datamodel/variants.py
```diff
@@ -50,36 +49,106 @@ def convertVCFGenotype(vcfGenotype, vcfPhaseset):
 
 
 def variantSetFactory(variantSetId, relativePath):
-    if variantSetId.endswith(".wt"):
-        return WormtableVariantSet(variantSetId, relativePath)
-    else:
-        return HtslibVariantSet(variantSetId, relativePath)
+    # TODO remove this redundant factory.
```
dcolligan: Might as well do this now

ga4gh/backend.py
```diff
@@ -46,10 +44,6 @@ def getVariantSetIdMap(self):
         # TODO remove this --- why do we need direct access to the map?
```
dcolligan: Is there a reason we can't remove this one as well?

ga4gh/datamodel/variants.py
jeromekelleher: Done.

ga4gh/backend.py
jeromekelleher: No reason - done.

dcolligan:
Generally looks good.  We'll want to remove the `.wt` parts from `ga4gh-example-data` as well after this change.

dcolligan:
+1

macieksmuga:
@jeromekelleher I know it's merged and done already, but I'll want to spend a bit of time looking over this one.

pgrosu:
@macieksmuga, below are the two codes locations:

This is the "Widespread changes to CallSet handling":

https://github.com/jeromekelleher/server/tree/77988bfd274bc8376b86284d7bf9676c2106c116

The dewormtabled version is here:

https://github.com/jeromekelleher/server/tree/893d693d5db3776da355ffbe1baea3db14ad1a74

~p

PR #300, opened 2015-03-30
Removed wormtable references from README

jeromekelleher:
Also minor changes to change wormtable related values elsewhere.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ python:
   - "2.7"
 
 before_install:
-  - sudo apt-get update -qq
-  - sudo apt-get install -qq libdb-dev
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -3,8 +3,6 @@ python:
   - "2.7"
 
 before_install:
-  - sudo apt-get update -qq
```
dcolligan: why are we removing these lines?

dcolligan: er, I guess we don't need them for anything other than BDB?

dcolligan:
+1

.travis.yml
jeromekelleher: Yep.

PR #303, opened 2015-03-30
Improved client logging.

jeromekelleher:
Minor improvements to client logging, and brings the client module up to date to use recent improvements in the SearchRequest API.

### DIFF
--- a/ga4gh/client.py
+++ b/ga4gh/client.py
@@ -12,9 +12,6 @@ import logging
 
 import ga4gh.protocol as protocol
 
-# suppress warning about using https without cert verification
-requests.packages.urllib3.disable_warnings()
-
 
 class HttpClient(object):
     """
@@ -30,6 +27,9 @@ class HttpClient(object):
         self._key = key
 
         # logging config
+        # TODO we need to revisit this logging setup so that we can
+        # disentangle our logs from urllib3's.
+        logging.basicConfig()
         self._logger = logging.getLogger(__name__)
         if self._debugLevel == 0:
             logLevel = logging.WARN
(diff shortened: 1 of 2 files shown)


ga4gh/client.py
```diff
@@ -12,9 +12,6 @@
 
 import ga4gh.protocol as protocol
 
-# suppress warning about using https without cert verification
```
dcolligan: why take this out?

dcolligan: er, nm, I see it's called in the __init__

ga4gh/client.py
```diff
@@ -38,10 +36,13 @@ def __init__(self, urlPrefix, debugLevel, workarounds, key):
         else:
             logLevel = logging.DEBUG
         self._logger.setLevel(logLevel)
-        handler = logging.StreamHandler()
-        formatter = logging.Formatter('%(name)s: %(message)s')
-        handler.setFormatter(formatter)
-        self._logger.addHandler(handler)
+
+        requestsLog = logging.getLogger("requests.packages.urllib3")
```
dcolligan: Is this really a good idea?  We lose control over formatting our messages differently from the messages that urllib3 wants to emit

jeromekelleher: I was just hacking around here, as it's tricky getting the different loggers working together. This was the simplest way I could find that meant our own logs weren't getting written twice.

Totally open to ideas here if you can see a better way.

dcolligan: why were our logs getting written twice?

jeromekelleher: Dunno - seems like they were getting sent to the requests logger as well as the one we'd instantiated.

dcolligan: Ok, put a TODO here to disentangle our logger from the urllib3 logger, without printing duplicate messages.

jeromekelleher: I've put in a TODO here. I'm not particularly happy with what we've ended up with --- I wish I understood what was actually happening here. It'll do for now though I think.

dcolligan: @jeromekelleher reading this (and other python logging docs) will probably help: https://docs.python.org/2/howto/logging.html#logging-flow

Issue #301, opened 2015-03-30
Paging broken for variants-search


jeromekelleher:
Paging for variants/search is not working correctly. If we run

```
$ python client_dev.py -v variants-search http://localhost:8000/v0.5.1 -m 1 -V 1000g_2013 -r 2
```

We get an infinite set of results.

This should be solved by reusing the logic for reads/search.

jeromekelleher:
Fixed in #304.

Issue #180, opened 2015-02-24
Searching across all variantSets in client not working


jeromekelleher:
When we specify no variantSet arguments to `variants-search` we should get back all variants. At present, we seem to only be returning a subset. Here's an example using the example data:
```
$ python client_dev.py variants-search http://localhost:8000/v0.5.1 -V 1000g_2013 -r 1 | cut -f 2 | uniq
1000g_2013
$ python client_dev.py variants-search http://localhost:8000/v0.5.1 -V 1000g_2011 -r 1 | cut -f 2  | uniq
1000g_2011
$ python client_dev.py variants-search http://localhost:8000/v0.5.1 -r 1 | cut -f 2 | uniq       
1000g_2011
```
We should get 1000g_2011 and 1000g_2013 in the last example.

dcolligan:
Fixed in #304

PR #304, opened 2015-03-30
Simple fix for variants-search.

jeromekelleher:
This is a straightforward fix for a logic bug in the variants/search method. Issue #301 states that this should be fixed by reusing the code from reads/search. However, this proved to be a little complex, and it seemed better to fix this bug quickly in a simple way.

We have also reduced to handling exactly one variant set only. The existing implementation was incorrect (see issue #180), and we had almost no test coverage on the multi-container case, so it seemed better to explicitly only support the case that works.

This  solves #180.


dcolligan:
+1

Issue #299, opened 2015-03-30
Call data not being returned for variants/search


jeromekelleher:
When we run the callsets-search client command, no call data is returned (regardless of what is provided to the `callSetIds` parameter).

dcolligan:
Fixed in #309

PR #309, opened 2015-03-30
resolving issue #299, Call data not being returned for variants/search

shajoezhu:
also raise exception when callsetid can not be found in the vcf files

### DIFF
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -65,6 +65,9 @@ class CallSet(object):
         # TODO this should be in the superclass, DatamodelObject.
         return self._id
 
+    def getSampleName(self):
+        return self._sampleName
+
     def toProtocolElement(self):
         """
         Returns the representation of this CallSet as the corresponding
@@ -329,6 +332,13 @@ class HtslibVariantSet(AbstractVariantSet):
         # TODO How do we get the number of records in a VariantFile?
         return 0
 
+    def getCallSet(self, sampleName):
+        """
+        Returns the CallSet object for the specified sample name.
+        """
+        callSetId = self.getCallSetId(sampleName)
+        return self._callSetIdMap[callSetId]
+
     def _updateCallSetIds(self, variantFile):
         """
         Updates the call set IDs based on the specified variant file.
@@ -365,10 +375,11 @@ class HtslibVariantSet(AbstractVariantSet):
                 self._chromFileMap[chrom] = varFile
 
     def _convertGaCall(self, recordId, name, pysamCall, genotypeData):
+        callSet = self.getCallSet(name)
         call = protocol.GACall()
-        call.callSetId = recordId
-        call.callSetName = name
-
+        call.callSetId = callSet.getId()
+        call.callSetName = callSet.getSampleName()
+        call.sampleId = callSet.getSampleName()
         # TODO:
         # NOTE: THE FOLLOWING TWO LINES IS NOT THE INTENED IMPLEMENTATION,
         ###########################################
@@ -427,7 +438,7 @@ class HtslibVariantSet(AbstractVariantSet):
         variant.calls = []
         sampleIterator = 0  # REMOVAL
         for name, call in record.samples.iteritems():
-            if name in callSetIds:
+            if self.getCallSetId(name) in callSetIds:
                 genotypeData = sampleData[sampleIterator].split(
                     ":")[0]  # REMOVAL
                 variant.calls.append(self._convertGaCall(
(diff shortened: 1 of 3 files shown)


ga4gh/datamodel/variants.py
```diff
@@ -75,6 +75,17 @@ def toProtocolElement(self):
...
+        """
+        Returns a Global Alliance Call object that is suitable for this
+        CallSet.
+        """
+        gaCall = protocol.GACall()
+        gaCall.callSetId = self._id
+        gaCall.callSetName = self._sampleName
+        gaCall.sampleId = self._sampleName
+        return gaCall
+
```
macieksmuga: I'm not sure if this method is pulling its weight here, since its naming and placement is somewhat confusing. Its only added value is to fill out a couple of fields of a generic GACall object with the callSet info, and this could easily be accomplished below outside of a dedicated method.

ga4gh/datamodel/variants.py
```diff
@@ -335,9 +353,8 @@ def _addFile(self, filename):
                 self._chromFileMap[chrom] = varFile
 
     def _convertGaCall(self, recordId, name, pysamCall, genotypeData):
-        call = protocol.GACall()
-        call.callSetId = recordId
-        call.callSetName = name
+        callSet = self.getCallSet(name)
+        call = callSet.getGaCall()
```
macieksmuga: so, replace this line with:
```
call = protocol.GACall()
call.callSetId = callSet._id
call.callSetName = callSet._sampleName
gaCall.sampleId = callSet._sampleName
```
And just to check - should both the above always be set to the same thing? If so, why do we have two separate fields in the protocol definition?

jeromekelleher: I'm +0 on getting rid of the `getGaCall` method. However, we should have
```
call = protocol.GACall()
call.callSetId = callSet.getId()
call.callSetName = callSet.getSampleName()
call.sampleId = callSet.getSampleName()
```
These are protected variables, so we should never just reach in and use them (except from subclasses).

ga4gh/datamodel/variants.py
```diff
@@ -423,6 +440,11 @@ def getVariants(self, referenceName, startPosition, endPosition,
         # protocol version 0.6
         if callSetIds is None:
             callSetIds = []
+        else:
+            for call in callSetIds:
```
jeromekelleher: This variable should be `callSetId` rather than `call`.

ga4gh/datamodel/variants.py
```diff
@@ -423,6 +440,11 @@ def getVariants(self, referenceName, startPosition, endPosition,
...
+                if call not in self._callSetIds:
+                    raise exceptions.CallSetNotInVariantSetException(
```
jeromekelleher: Have we defined this exception? Also, I don't think we need to pass the dict of callSetIds to the exception. I would use 
```python
raise exceptions.CallSetNotInVariantSetException(self, callSetId)
```
The exception message could then be "CallSet '{}' not in VariantSet '{}'".format(callSetId, variantSet.getId())

jeromekelleher:
Looks good, thanks @shajoezhu. I don't have a strong opinion on the `getGaCall` method, I'm happy either way. There's a few other minor comments, but I'm +1 otherwise.

ga4gh/datamodel/variants.py
shajoezhu: This exception is defined. but i don't think it was actually ever used. it is currently in ga4gh/exception.py
```python
class CallSetNotInVariantSetException(NotFoundException):
    """
    Indicates a request was made for a callSet not in the actual variantSet
    """
    def __init__(self, callSetId, variantSetId):
        self.callSetId = callSetId
        self.variantSetId = variantSetId

    def getMessage(self):
        message = "callSet '{0}' not in variantSet '{1}'".format(
            self.callSetId, self.variantSetId)
        return message
```
i m inclined not to make changes in this PR, instead raise a minor separate issue, fix it after PR #294 is merged

jeromekelleher: You're right, sorry I should have looked this up. We still need to change the parameters though: it's expecting callSetId and variantSetId.

ga4gh/datamodel/variants.py
```diff
@@ -453,6 +464,11 @@ def getVariants(self, referenceName, startPosition, endPosition,
         # protocol version 0.6
         if callSetIds is None:
             callSetIds = []
+        else:
+            for callSetId in callSetIds:
+                if callSetId not in self._callSetIds:
+                    raise exceptions.CallSetNotInVariantSetException(
+                        callSetId, self.getId())
```
shajoezhu: fixed a bug here. was calling self._varientset. which was not actually an attribute of this class.

jeromekelleher:
This looks good now, thanks @shajoezhu.

+1.

PR #311, opened 2015-03-31
Changed to uncompressed tarball.

jeromekelleher:
The example data has been updated with an improved data set, and is available to download at http://www.well.ox.ac.uk/~jk/ga4gh-example-data.tar. It is an uncompressed tarball now, as all the internal data is already compressed. This PR updates the downloader script to take this into account.


dcolligan:
+1

PR #316, opened 2015-04-01
Fix validation errors for reads.

jeromekelleher:
Fixes multiple validation errors in the reads code (previous datadriven tests weren't actually checking the result of validate and so were passing). Also put in some comments for some parts of the read translation code that are especially murky.

Also adds some 1000G read data to our test set.


dcolligan:
Looks good.  As you can see, I made a lot of judgement calls in implementing reads, and they may not have been the correct ones!  I suspect we still have some ways to go in terms of implementing these things "the right way."

+1

jeromekelleher:
I didn't really appreciate how unhelpful the documentation was until I started looking at this! I think you did a great job @dcolligan --- we have something that mostly works where we had nothing before. We need expert help here to make sure we're interpreting the specs correctly.

Issue #305, opened 2015-03-30
Dockerize ga4gh server + example data


macieksmuga:
Once the first stable release is reached.

dcolligan:
What are the use cases of this?  Is this just to save users the trouble of having to download example data from a (jerome's) random server?

macieksmuga:
For the v0.5.1server, this will at first be a test of capability. 

However the intention is to deploy the server on realistically sized chunks of genomic data soon, and It will make sense to investigate various ways (docker containers being a prime candidate) of deploying the server+data across clusters/on the cloud. 

For the future v0.6 server, the datasets will likewise start growing fast, and they'll be spread across Sqlite databases as well as large FASTA files - so dockerization may be the right approach for deployment there too.

@jeromekelleher does this sound about right to you?

pgrosu:
@macieksmuga, is the use of Sqlite dbs been set in stone?  I only mention this because bottlenecks can appear, one specifically is the following:

http://www.sqlite.org/faq.html#q19

~p

macieksmuga:
@pgrosu the graph genome project will be using Sqlite for rapid development and minimizing of compatibility issues in a reference implementation, not necessarily for a final full-scale server: So yeah, it's not ideal for real-world deployment against a full-size population genome-sized dataset.

That being said, it shouldn't be too bad for small-scale deployments: The specific bottleneck you mention shouldn't be a problem, as the server will be read-only for the time being. And my experience with SQL database servers is that they can be nicely optimized to fit particular patterns of use as they become apparent.

pgrosu:
@macieksmuga, sounds good and agreed, though performance of concurrency for read-only can also be affected which depends on the configuration.  Below is a link to different configurations and how to maximize concurrency for read-only:

http://manski.net/2012/10/sqlite-performance/

Hope it helps,
~p

diekhans:
All database can encounter bottlenecks.  The fact that the
database will be read-only is a huge advantage in any approach.
It's a good choice for a reference implementation.

pgrosu:
@diekhans, I agree for now, but definitely not for the near future.  If we look just at the kilobases of sequencing per day per machine from a few years ago - which I'm sure will not plateau anytime soon, as things get more "affordable" - I imagine this data will feed eventually into GA4GH repositories on a daily basis:

![sequencing](https://cloud.githubusercontent.com/assets/6555937/6910020/34935c4e-d71b-11e4-9a6d-0f44594e83c5.png)

[...]

PR #326, opened 2015-04-02
Added Dockerfile

hershman:
This PR addresses Issue #305 by adding scripts/Dockerfile, which can be used to quickly launch a server instance.

### DIFF
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -1,0 +1,46 @@
+# Base image
+FROM ubuntu
+
+# Update the sources list
+RUN apt-get update
+
+# Install packages
+RUN apt-get install -y tar git curl nano wget dialog net-tools build-essential python python-dev python-distribute python-pip zlib1g-dev python-virtualenv apache2 libapache2-mod-wsgi
+
+# Enable wsgi module
+RUN a2enmod wsgi
+
+# Create cache directories
+RUN mkdir /var/cache/apache2/python-egg-cache && \
+    chown www-data:www-data /var/cache/apache2/python-egg-cache/
+
+# Set up GA4GH server
+RUN mkdir /srv/ga4gh
+WORKDIR /srv/ga4gh
+RUN virtualenv ga4gh-server-env
+RUN /bin/bash -c "source ga4gh-server-env/bin/activate"
+RUN pip install git+https://github.com/ga4gh/server
+
+# Install relevant sample data
+RUN wget http://www.well.ox.ac.uk/~jk/ga4gh-example-data.tar.gz && \
+    tar -zxf ga4gh-example-data.tar.gz && rm -r ga4gh-example-data/variants/*.wt
+
+# Write application.wsgi
+RUN echo "from ga4gh.frontend import app as application\nimport ga4gh.frontend as frontend\nfrontend.configure(\"/srv/ga4gh/config.py\")" > application.wsgi
+
+# Write config.py
+RUN echo "DATA_SOURCE = \"/srv/ga4gh/ga4gh-example-data\"" > config.py
+
+# Write new apache config
+WORKDIR /etc/apache2/sites-available
+RUN echo "<VirtualHost *:80>\n  ServerAdmin webmaster@localhost\n  DocumentRoot /var/www/html\n  ErrorLog ${APACHE_LOG_DIR}/error.log\n  CustomLog ${APACHE_LOG_DIR}/access.log combined\n  WSGIDaemonProcess ga4gh python-path=/srv/ga4gh/ga4gh-server-env/lib/python2.7/site-packages python-eggs=/var/cache/apache2/python-egg-cache\n  WSGIScriptAlias /ga4gh /srv/ga4gh/application.wsgi\n  <Directory /srv/ga4gh>\n    WSGIProcessGroup ga4gh\n    WSGIApplicationGroup %{GLOBAL}\n    Require all granted\n  </Directory>\n</VirtualHost>" > 001-ga4gh.conf
+
+# Configure apache to serve GA4GH site
+WORKDIR /etc/apache2/sites-enabled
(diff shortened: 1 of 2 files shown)


scripts/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+# Create cache directories
+RUN mkdir /var/cache/apache2/python-egg-cache && \
+    chown www-data:www-data /var/cache/apache2/python-egg-cache/
+
+# Set up GA4GH server
+RUN mkdir /srv/ga4gh
+WORKDIR /srv/ga4gh
+RUN virtualenv ga4gh-server-env
+RUN /bin/bash -c "source ga4gh-server-env/bin/activate"
+RUN pip install git+https://github.com/ga4gh/server
```
jeromekelleher: Should we install from git or from PyPI for the default Docker install? I would say PyPI probably.

scripts/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+# Install relevant sample data
+RUN wget http://www.well.ox.ac.uk/~jk/ga4gh-example-data.tar.gz && \
```
jeromekelleher: We're using an uncompressed tarball now - you can remove the .gzs and the wormtable deletes.

scripts/Dockerfile
hershman: The current version PyPI server has a wormtable dependency that won't
compile because of a missing header,

scripts/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+# Write application.wsgi
+RUN echo "from ga4gh.frontend import app as application\nimport ga4gh.frontend as frontend\nfrontend.configure(\"/srv/ga4gh/config.py\")" > application.wsgi
+
+# Write config.py
+RUN echo "DATA_SOURCE = \"/srv/ga4gh/ga4gh-example-data\"" > config.py
+
+# Write new apache config
+WORKDIR /etc/apache2/sites-available
+RUN echo "<VirtualHost *:80>\n  ServerAdmin webmaster@localhost\n  DocumentRoot /var/www/html\n  ErrorLog ${APACHE_LOG_DIR}/error.log\n  CustomLog ${APACHE_LOG_DIR}/access.log combined\n  WSGIDaemonProcess ga4gh python-path=/srv/ga4gh/ga4gh-server-env/lib/python2.7/site-packages python-eggs=/var/cache/apache2/python-egg-cache\n  WSGIScriptAlias /ga4gh /srv/ga4gh/application.wsgi\n  <Directory /srv/ga4gh>\n    WSGIProcessGroup ga4gh\n    WSGIApplicationGroup %{GLOBAL}\n    Require all granted\n  </Directory>\n</VirtualHost>" > 001-ga4gh.conf
```
jeromekelleher: Is there a way to split these long lines? It's a bit hard to read what's going on here.

scripts/Dockerfile
jeromekelleher: Ah OK. This will be fixed today, when we tag and upload 0.1.0a3.

jeromekelleher:
This is fantastic @hershman, thanks! This will make deployment a breeze, so I think we should definitely include it. I won't have a chance to test it out until next week, unfortunately, as we're concentrating on shipping the 0.1.0 beta release tomorrow.

Is `scripts` the right place for this to live everyone? We're currently using `scripts` as our location for internal-use-only development tools. The docker file is something different though. Should we make a new directory (say, `deploy`) for deployment stuff like this?

scripts/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN a2enmod wsgi
+
+# Create cache directories
+RUN mkdir /var/cache/apache2/python-egg-cache && \
+    chown www-data:www-data /var/cache/apache2/python-egg-cache/
+
+# Set up GA4GH server
+RUN mkdir /srv/ga4gh
+WORKDIR /srv/ga4gh
+RUN virtualenv ga4gh-server-env
```
jeromekelleher: Actually, is there any point in using a virtualenv inside a Docker image?

hershman:
[This blog post](http://www.markbetz.net/2014/01/17/python-if-you-have-docker-do-you-need-virtualenv/#comments) has a pretty good discussion about that issue. They seem undecided.

scripts/Dockerfile
hershman: [This blog post](http://www.markbetz.net/2014/01/17/python-if-you-have-docker-do-you-need-virtualenv/#comments) has a pretty good discussion about that issue. They seem undecided.

jeromekelleher: Interesting... I guess we should play with both options and see what works better for us.

scripts/Dockerfile
dcolligan: If this is just running bash commands, perhaps a here document? http://tldp.org/LDP/abs/html/here-docs.html

knoxcarey: There's been a lot of discussion of heredocs for Dockerfiles -- cf https://github.com/docker/docker/issues/1799 -- but they are still not supported as far as I can tell. It seems like the consensus is to use the ADD command to copy an external file into the image. That has the advantage of readability, but it loses the simplicity of having everything in the one file. It would also be possible to make the Dockerfile more readable by writing lines one at a time, but at the expense of generating an intermediate docker image for every single line added. That adds quite a lot of overhead in building the image.

dcolligan:
Closing this PR.  All PRs should now be made against the `develop` branch, not the `master` branch.

dcolligan:
(the way to do this is to push the same code with a new branch name and in the PR UI select to merge against develop instead of master -- develop will be made the default shortly)

jeromekelleher:
Sorry about this @hershman --- we're changing over to a new development branching model as discussed in #330, and your PR got caught in the middle. The structure of the documentation has changed too, and we no longer put detailed information in the README. There's a placeholder for your information in `docs/source/installation.rst`. Basically, if you copy your docs code in there and reopen the PR I think we should be ready to merge quickly.

Issue #296, opened 2015-03-26
Test generators more thoroughly


dcolligan:
We need some tests for the generator logic in `backend.py`.  I'm thinking of `readsGenerator` in particular.

dcolligan:
Fixed in #336

PR #313, opened 2015-03-31
Refactor and test readGenerator

dcolligan:
Issue #296

PR #336, opened 2015-04-06
Refactor and test readGenerator

dcolligan:
Issue #296

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -17,6 +17,167 @@ import ga4gh.exceptions as exceptions
 import ga4gh.datamodel.variants as variants
 
 
+def _parsePageToken(pageToken, numValues):
+    """
+    Parses the specified pageToken and returns a list of the specified
+    number of values. Page tokens are assumed to consist of a fixed
+    number of integers seperated by colons. If the page token does
+    not conform to this specification, raise a InvalidPageToken
+    exception.
+    """
+    tokens = pageToken.split(":")
+    # TODO define exceptions.InvalidPageToken and raise here.
+    if len(tokens) != numValues:
+        raise Exception("Invalid number of values in page token")
+    # TODO catch a ValueError here when bad integers are passed and
+    # convert this into the appropriate InvalidPageToken exception.
+    values = map(int, tokens)
+    return values
+
+
+def _getVariantSet(request, variantSetIdMap):
+    if len(request.variantSetIds) != 1:
+        if len(request.variantSetIds) == 0:
+            msg = "Variant search requires specifying a variantSet"
+        else:
+            msg = ("Variant search over multiple variantSets "
+                   "not supported")
+        raise exceptions.NotImplementedException(msg)
+    variantSetId = request.variantSetIds[0]
+    try:
+        variantSet = variantSetIdMap[variantSetId]
+    except KeyError:
+        raise exceptions.VariantSetNotFoundException(variantSetId)
+    return variantSet
+
+
+class IntervalIterator(object):
+    """
+    Implements generator logic for types which accept a start/end
+    range to search for the object
+    """
+    def __init__(self, request, containerIdMap):
+        self._request = request
+        self._containerIdMap = containerIdMap
+        self._badPageTokenExceptionMessage = (
+            "Inconsistent page token provided")
+        self._container = self._getContainer()
+        self._startPosition, self._equalPositionsToSkip = \
+            self._getIntervalCounters()
+        self._iterator = self._getIterator()
+        self._generator = self._internalIterator()
+
+    def __iter__(self):
+        return self._generator
+
+    def next(self):
+        obj = next(self._generator)
+        return obj
+
+    def _raiseBadPageTokenException(self):
+        raise exceptions.BadPageTokenException(
+            self._badPageTokenExceptionMessage)
+
+    def _getIntervalCounters(self):
+        startPosition = self._request.start
+        equalPositionsToSkip = 0
+        if self._request.pageToken is not None:
+            startPosition, equalPositionsToSkip = _parsePageToken(
+                self._request.pageToken, 2)
+        return startPosition, equalPositionsToSkip
+
+    def _internalIterator(self):
+        obj = next(self._iterator, None)
+        if self._request.pageToken is not None:
+            # First, skip any records with getStart < startPosition
+            # or getEnd < request.start
+            while (self._getStart(obj) < self._startPosition or
+                   self._getEnd(obj) < self._request.start):
+                obj = next(self._iterator, None)
+                if obj is None:
+                    self._raiseBadPageTokenException()
+            # Now, skip equalPositionsToSkip records which have getStart
+            # == startPosition
+            equalPositionsSkipped = 0
+            while equalPositionsSkipped < self._equalPositionsToSkip:
+                if self._getStart(obj) != self._startPosition:
+                    self._raiseBadPageTokenException()
+                equalPositionsSkipped += 1
+                obj = next(self._iterator, None)
+                if obj is None:
+                    self._raiseBadPageTokenException()
+        # iterator is now positioned to start yielding valid records
+        while obj is not None:
+            nextObj = next(self._iterator, None)
+            nextPageToken = None
+            if nextObj is not None:
+                if self._getStart(obj) == self._getStart(nextObj):
+                    self._equalPositionsToSkip += 1
+                else:
+                    self._equalPositionsToSkip = 0
+                nextPageToken = "{}:{}".format(
+                    self._getStart(nextObj), self._equalPositionsToSkip)
+            yield obj, nextPageToken
+            obj = nextObj
+
+
+class ReadsIntervalIterator(IntervalIterator):
+    """
+    An interval iterator for reads
+    """
+    def _getContainer(self):
+        if len(self._request.readGroupIds) != 1:
+            if len(self._request.readGroupIds) == 0:
+                msg = "Read search requires a readGroup to be specified"
+            else:
+                msg = "Read search over multiple readGroups not supported"
+            raise exceptions.NotImplementedException(msg)
+        readGroupId = self._request.readGroupIds[0]
+        try:
+            readGroup = self._containerIdMap[self._request.readGroupIds[0]]
+        except KeyError:
+            raise exceptions.ReadGroupNotFoundException(readGroupId)
+        return readGroup
+
+    def _getIterator(self):
+        iterator = self._container.getReadAlignments(
+            self._request.referenceId,
+            self._startPosition, self._request.end)
+        return iterator
+
+    @classmethod
+    def _getStart(cls, readAlignment):
+        return readAlignment.alignment.position.position
+
+    @classmethod
+    def _getEnd(cls, readAlignment):
+        return cls._getStart(readAlignment) + \
+            len(readAlignment.alignedSequence)
+
+
+class VariantsIntervalIterator(IntervalIterator):
+    """
+    An interval iterator for variants
+    """
+    def _getContainer(self):
+        return _getVariantSet(self._request, self._containerIdMap)
+
+    def _getIterator(self):
+        iterator = self._container.getVariants(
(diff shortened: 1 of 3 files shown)


dcolligan:
This is the same PR as #313 except this PR is made against the develop branch instead of master

tests/unit/test_response_generators.py
```diff
@@ -0,0 +1,282 @@
...
+    def setUp(self):
+        self.request = protocol.GASearchVariantsRequest()
+        self.backend = backend.AbstractBackend()
+        self.variantSetId = "variantSetId"
+
+    def testNoVariantsNotSupported(self):
+        # a request for no variant sets should throw an exception
+        self.request.variantSetIds = []
+        with self.assertRaises(exceptions.NotImplementedException):
+            iterator = self.backend.variantsGenerator(self.request)
```
jeromekelleher: Shouldn't the `iterator = ...` bit be outside of the `with` block?

ga4gh/backend.py
```diff
@@ -384,6 +318,95 @@ def setMaxResponseLength(self, maxResponseLength):
         """
         self._maxResponseLength = maxResponseLength
 
+    # private methods used to implement readsGenerator
+
+    def _getPosition(self, readAlignment):
```
jeromekelleher: These aren't really methods of the backend I think (``self isn't used, which is a good indication). I think it might be a good idea to make these as plain functions, and then pass these as parameters to your general methods for iterating over the intervals. I think it's probably worthwhile making a new class `IntervalIterator` to do this to avoid cluttering up the backend with the details of this. So, concretely, I'd image:
```python
def _getReadAlignmentStart(readAlignment):
    return blah blah

def _getReadAlignmentEnd(readAlignment):
    return blah blah

class ItervalPager(object):
    """
    Takes care of the details of paging over interval search results for reads and variant search.
    """
    def __init__(self, request, iterator, getStart, getEnd):
        self._request = request
        self._getStart = getStart
        # etc.
    # The rest of the interface is similar to what you have

"""

The advantage of this is, (1) we isolate the tricky interval paging logic into a single class, making it more easily testable; and (2) this should make it very easy to share code between the reads and variants search.

Does this make sense?

**Comment by jeromekelleher** [comment]
Looks good @dcolligan, thanks. I have a suggestion for a slightly different interface that might make things a little easier to generalise and test. What do you think?

**Comment by dcolligan** [comment]
what would be the definition of the `getStart` and `getEnd` methods for variants?  is it just returning `variant.start` and `variant.end`, respectively?

**Comment by jeromekelleher** [comment]
Exactly, yeah. The advantage is that we're abstracting the details of how we find the start and end of the intervals, so we can use the same code to work with variants and reads.

**Comment by dcolligan** [comment]
@jeromekelleher pushed update

File: ga4gh/backend.py [hunk 3]
```diff
@@ -384,6 +386,49 @@ def setMaxResponseLength(self, maxResponseLength):
         """
         self._maxResponseLength = maxResponseLength
 
+    # private methods used to implement readsGenerator
```
jeromekelleher: There's only one method here, so a quick docstring would be more appropriate I think. Ditto the other methods below.

ga4gh/backend.py
```diff
@@ -384,6 +386,49 @@ def setMaxResponseLength(self, maxResponseLength):
...
+        variantSetId = request.variantSetIds[0]
+        try:
+            variantSet = self._variantSetIdMap[variantSetId]
+        except KeyError:
+            raise exceptions.VariantSetNotFoundException(variantSetId)
+        return variantSet
+
+    # private methods common to various interval generators
+
+    def _getCounters(self, request):
```
jeromekelleher: I'm not totally convinced by this name... `getIntervalCounters` or `getIntervalSearchToken` or something? `_getCounters` could mean almost anything.

ga4gh/backend.py
```diff
@@ -214,113 +289,40 @@ def variantSetsGenerator(self, request):
             request, self._variantSetIdMap, self._variantSetIds)
 
     def readsGenerator(self, request):
```
jeromekelleher: Docstring please.

ga4gh/backend.py
jeromekelleher: Better idea: the only reason this needs to be a method of the Backend class at all is because it uses the `self.parsePageToken` method. However, there's no reason for `parsePageToken` to be a method at all. I suggest we make `parsePageToken` a top level function, and we move the functionality of this method into the `IntervalIterator` class. This will make things _very_ tidy and self contained.

ga4gh/backend.py
```diff
@@ -17,6 +17,81 @@
...
+    """
+    def __init__(self, request, iterator, getStart, getEnd):
+        self._request = request
+        self._iterator = iterator
+        self._getStart = getStart
+        self._getEnd = getEnd
+        self._badPageTokenExceptionMessage = (
+            "Inconsistent page token provided")
+
+    def iterate(self, startPosition, equalPositionsToSkip):
```
jeromekelleher: I don't really see much value in splitting up the `_positionIterator` and `_yieldObjects` part here; these are intrinsically linked, and I think it's clearer to see them explicitly laid out together in one method.

Also, I think it might be nice if we implemented the iterator protocol explicitly in this object, so that client code will be able to just call:
```python
return IntervalIterator(request, iterator, _getVariantStart, _getVariantEnd)
```

tests/unit/test_response_generators.py
```diff
@@ -0,0 +1,297 @@
...
+
+import ga4gh.backend as backend
+import ga4gh.exceptions as exceptions
+import ga4gh.datamodel.reads as reads
+import ga4gh.datamodel.variants as variants
+import ga4gh.protocol as protocol
+
+
+def generateVariant(position=0):
+    variant = protocol.GAVariant()
```
jeromekelleher: Why the argument if you're not using it?

tests/unit/test_response_generators.py
```diff
@@ -0,0 +1,297 @@
...
+    def testPositionIterator(self):
+        def getIterator():
+            # skipped because position < startPosition
+            yield generateReadAlignment(0)
+            # skipped because endPosition < request.start
+            yield generateReadAlignment(1, 'a')
+            # skipped because equalPositionsSkipped
+            yield generateReadAlignment(1, 'abcdefg')
+            # this one should be returned
+            yield generateReadAlignment(1, 'abcdefghijklmnop')
```
jeromekelleher: Can you put this string into a variable for testing against? It's too hard to read it to make sure it's the same string below.

jeromekelleher:
This is great, thanks @dcolligan! I have a few suggestions which I think will make it even nicer...

tests/unit/test_response_generators.py
```diff
@@ -0,0 +1,297 @@
...
+        self.backend = backend.AbstractBackend()
+        self.variantSetId = "variantSetId"
+
+    def testNoVariantsNotSupported(self):
+        # a request for no variant sets should throw an exception
+        self.request.variantSetIds = []
+        with self.assertRaises(exceptions.NotImplementedException):
+            self.backend.variantsGenerator(self.request)
+
+    def testMultipleVariantsNotSupported(self):
```
macieksmuga: shouldn't this be `testMultipleVariantSetsNotSupported`?

tests/unit/test_response_generators.py
```diff
@@ -0,0 +1,297 @@
...
+class TestVariantsGenerator(unittest.TestCase):
+    """
+    Tests the logic of variantsGenerator
+    """
+    def setUp(self):
+        self.request = protocol.GASearchVariantsRequest()
+        self.backend = backend.AbstractBackend()
+        self.variantSetId = "variantSetId"
+
+    def testNoVariantsNotSupported(self):
```
macieksmuga: `testNoVariantSetsNotSupported`?

ga4gh/backend.py
dcolligan: splitting them up is good because otherwise the method becomes pretty large.  smaller ones are easier to understand and test.  also they perform different functions: one positions the iterator to where it needs to be, and the other starts the process of yielding records.

we can't just `return IntervalInterator(...)` because `__init__` can't be a generator

dcolligan:
pushed a new commit

ga4gh/backend.py
jeromekelleher: You can if the class supports the iterator protocol (basically it needs an `__next__` method or something like that). I won't push the point here though, keep it this way if you prefer.

ga4gh/backend.py
```diff
@@ -214,113 +299,44 @@ def variantSetsGenerator(self, request):
...
-        equalPositionsToSkip = 0
-        if request.pageToken is not None:
-            startPosition, equalPositionsToSkip = self.parsePageToken(
-                request.pageToken, 2)
+        """
+        Returns a generator over the (read, nextPageToken) pairs defined
+        by the specified request
+        """
+        readGroup = self._getReadGroup(request)
+        startPosition, equalPositionsToSkip = _getIntervalCounters(request)
```
jeromekelleher: This should be done internally in the IntervalIterator class. This method should't do anything with the page token. This way ``_getIntervalCounters`` can be a method of IntervalIterator also, which is where it belongs logically I think.

ga4gh/backend.py
dcolligan: `__next__` is only for python3.  python 2 requires `next`.  We should probably avoid implementing this method if we're worried about compatibility (even in the distant future), in any event.

dcolligan: Er, I kind of take that back.  Python3 compatibility would just be adding a `__next__` method that calls the same underlying method as the existing `next` method.

ga4gh/backend.py
dcolligan: Maybe, but we need access to `startPosition` outside `IntervalIterator`, namely to create the iterator.

I think a better design would be to have subclasses of `IntervalIterator`, one for variants and one for reads at present, that could each implement their own methods instead of passing module-level functions into the `IntervalIterator` constructor.  Then `readsGenerator`, etc. would be a one-line function that just called the appropriate `IntervalIterator` subclass. Agree?

dcolligan:
@jeromekelleher pushed a new commit

ga4gh/backend.py
jeromekelleher: Ah, OK, I've just seen the reason why we need `startPosition`; sorry about that.

Agreed on the subclassing idea, that is a much nicer design.

dcolligan:
@jeromekelleher pushed a new commit

ga4gh/backend.py
```diff
@@ -17,6 +17,167 @@
...
+    # TODO define exceptions.InvalidPageToken and raise here.
+    if len(tokens) != numValues:
+        raise Exception("Invalid number of values in page token")
+    # TODO catch a ValueError here when bad integers are passed and
+    # convert this into the appropriate InvalidPageToken exception.
+    values = map(int, tokens)
+    return values
+
+
+def _getVariantSet(request, variantSetIdMap):
```
jeromekelleher: Might as well inline this into `VariantsIntervalIterator._getContainer` for symmetry with `ReadsIntervalIterator`.

jeromekelleher:
Superb; +1, modulo the minor suggestion above.

ga4gh/backend.py
dcolligan: But `AbstractBackend.callSetsGenerator` uses it as well

jeromekelleher: OK, disregard this then.

jeromekelleher:
@macieksmuga, are you happy with this PR? I suggest we merge it, and then update your big PR based on this.

dcolligan:
@macieksmuga pushed a change rebased on top of latest develop... could you merge if you're happy with it?

dcolligan:
Can we merge this soon?  It's been open for two weeks...

macieksmuga:
Yep, happy.
Sorry, was under mistaken impression I had already given it my +1

jeromekelleher:
Merged; thanks for this @dcolligan.

Issue #333, opened 2015-04-03
New branching process scripts and docs


dcolligan:
Create documentation for new branching process (master branch, develop branch, release branches)

Create scripts for making branching and merging according to that process easy and error-free

dcolligan:
Fixed in #337

PR #337, opened 2015-04-06
Added docs describing branch strategy

dcolligan:
Issue #333

### DIFF
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -106,14 +106,151 @@ are making:
 
     $ git remote add upstream https://github.com/ga4gh/server.git
 
+All development is done against the ``develop`` branch.  The ``master``
+branch is meant to be kept stable since it is the branch releases are
+based on -- don't touch it!  These are the two mainline branches.
+Our branching model is loosely based on the one described
+`here <http://nvie.com/posts/a-successful-git-branching-model/>`_.
 
-**TODO**
+All development should be done in a topic branch.  That is, a branch
+that the developer creates him or herself.  These steps will create
+a topic branch (replace ``TOPIC_BRANCH_NAME`` appropriately):
 
-- Walk a new developer through the process of fetching ``upstream/master``
-  and creating a new topic branch.
-- Walk them through pushing to their fork and making a pull request.
-- Explain why we do a lot of rebasing, and walk them through the
-  process of squashing commits.
+.. code-block:: bash
+
+    $ git fetch --all
+    $ git checkout develop
+    $ git merge --ff-only upstream/develop
+    $ git checkout -b TOPIC_BRANCH_NAME
+
+Topic branch names should include the issue number (if there is a tracked
+issue this change is addressing) and provide some hint as to what the
+changes include.  For instance, a branch that addresses the (imaginary)
+tracked issue with issue number #123 to add more widgets to the code
+might be named ``123_more_widgets``.
+
+At this point, you are ready to start adding, editing and deleting files.
+Stage changes with ``git add``.  Afterwards, checkpoint your progress by
+making commits:
+
+.. code-block:: bash
+
+    $ git commit -m 'Awesome changes'
+
+(You can also pass the ``--amend`` flag to ``git commit`` if you want to 
+incorporate staged changes into the most recent commit.)
+
+Once you have changes that you want to share with others, push your
+topic branch to GitHub:
+
+.. code-block:: bash
+
+    $ git push origin TOPIC_BRANCH_NAME
+
+Then create a pull request using the GitHub interface.  This pull request
+should be against the ``develop`` branch (this should happen automatically).
+
+At this point, other developers will weigh in on your changes and will
+likely suggest modifications before the change can be merged into
+``develop``.  When you get around to incorporating these suggestions,
+it is likely that more commits will have been added to the ``develop``
+branch.  Since you (almost) always want to be developing off of the 
+latest version of the code, you need to perform a rebase to incorporate
+the most recent changes from ``develop`` into your branch.
+
+.. code-block:: bash
+
+    $ git fetch --all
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Two ways @dcolligan:

1. From the project root, run `python setup.py build_sphinx`. The HTML docs are then built somewhere in the `build` directory.
2. Go to the `docs/source` directory and run `make`. The HTML is then in `_build`.

We should definitely add this to our development documentation.

docs/source/development.rst
```diff
@@ -106,14 +106,164 @@ are making:
...
+All development is done against the ``develop`` branch.  The ``master``
+branch is meant to be kept stable since it is the branch releases are
+based on -- don't touch it!  These are the two mainline branches.
+Our branching model is loosely based on the one described
+`<here http://nvie.com/posts/a-successful-git-branching-model/>`_.
+
+All development should be done in a topic branch.  That is, a branch
+that the developer creates him or herself.  These steps will create
+a topic branch (assuming that ``upstream`` points to 
+``https://github.com/ga4gh/server.git``) and ``origin`` points to
```
jeromekelleher: Is this parenthetical comment a bit confusing? We've just told the user to create a remote called upstream in the previous paragraph, and I think we can assume origin follows the usual convention.

docs/source/development.rst
```diff
@@ -106,14 +106,164 @@ are making:
...
+a topic branch (assuming that ``upstream`` points to 
+``https://github.com/ga4gh/server.git``) and ``origin`` points to 
+``https://github.com/YOUR_GITHUB_USERNAME/server.git``, which can be
+verified by running ``git remote -v``):
 
-**TODO**
+.. code-block:: bash
+
+    $ git fetch --all
+    $ git checkout develop
```
jeromekelleher: I usually do this as
```bash
$ git fetch upstream
$ git checkout upstream/develop
$ git checkout -b TOPIC_BRANCH_NAME
```
What are the advantages/disadvantages of the various approaches? I know there's a million different ways to do things in git, but I think it's worthwhile trying to figure out the most foolproof recipe we can to try and make the learning curve a little less steep.

docs/source/development.rst
```diff
@@ -106,14 +106,164 @@ are making:
...
+If you had tried to push the topic branch without using the force flag,
+it would have failed.  This is because non-force pushes only succeed when
+you are only adding new commits to the tip of the existing remote branch.
+When you want to do something other than that, such as insert commits
+in the middle of the branch history (what ``git rebase`` does), or modify a
+commit (what ``git commit --amend`` does) you need to blow away the remote
+version of your branch and replace it with the local version.  This is
+exactly what a force push does.
+
+**IMPORTANT**: Never use the force flag to push to ``upstream``.  Never use
```
jeromekelleher: Maybe use a ``.. warning::`` block here? 

http://sphinx-doc.org/markup/para.html

docs/source/development.rst
```diff
@@ -106,14 +106,164 @@ are making:
...
+
+Once your pull request has been merged into ``develop``, you can close
+the pull request and delete the remote branch in the GitHub interface.
+Locally, run this command to delete the topic branch:
+
+.. code-block:: bash
+
+    $ git branch -D TOPIC_BRANCH_NAME
+
+Customizing git's `configuration <http://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration>`_
```
jeromekelleher: I think we might be veering off topic a bit here with the last three paragraphs. I would definitely delete the bit about aliases --- you never know how these might interact with someone's environment.

jeromekelleher:
This is superb, thanks @dcolligan! Minor comments/questions inline, but definitely what we need.

docs/source/development.rst
dcolligan: There are two differences between your approach and my approach:
1) You fetch only updates from the upstream repo, whereas I fetch updates from all remote repos.
2) You don't create a local `develop` branch, whereas I do.

Advantages of my approach:
- no chance of a new user screwing up the fetch command by mistyping, or fetching the wrong remote repo... just fetch 'em all!  (fetch is a pretty easy command for newbies to screw up http://stackoverflow.com/questions/21537244/differences-between-git-fetch-and-git-fetch-origin-master )
- having an updated local `develop` branch saves a little typing when one wants to reference it

Disadvantages of my approach:
- little slower to fetch all repos (not really a problem in practice)
- need to maintain a local copy of develop (not a problem with my alias below)

dcolligan:
Pushed an update.

@jeromekelleher you wouldn't object to renaming `master` to `release`, would you? (sent you an email about this)

diekhans:
I suggest have a very standard, old school approach to branches.

The default trunk is where work gets submitted.  There is a
per-release branch and then tags on that branch for alpha, beta,
release, fixes, etc.

A release branch is made when it's time to stabilize.  Of
course, commits and be moved back to between trunk and release
branch as needed, but trunk is free to charge ahead with
development.

Mark

dcolligan:
Okay, maybe `stable` would be a better name then to not confuse it with the release branches.  (The default branch is already `develop`, where all changes are by default submitted.)

jeromekelleher:
I'm not hugely keen on changing the name of 'master', but I do see your point @dcolligan. So, the idea is that we have two main branches, 'stable' and 'develop'. All PRs go against develop, and stable is always in a production ready state. When we are moving towards a release, we create a release branch and stabilise on that branch. Once the release is ready, we merge it to stable and tag it. This release branch then lives on as where we make bug fixes.

How does this sound? Is this more-or-less what you have in mind @diekhans?

jeromekelleher:
Looks great, +1. We should clarify the branching and release model in a follow up PR.

macieksmuga:
Very nice.
+1 and thank you @dcolligan!

Issue #338, opened 2015-04-07
Continuous integration checks required for docs


jeromekelleher:
Documentation is built automatically now using read-the-docs, but there is no CI tests to ensure that this process works. We should include some checks in our travis configuration to make sure that the documentation at least builds without error. (This might be as simple as adding `python setup.py sphinx_build`)

dcolligan:
Fixed in #342

PR #342, opened 2015-04-07
Add docs to CI

dcolligan:
Issue #338


jeromekelleher:
This is a good start, but it doesn't actually check very much. We should turn warnings into errors, and this will help prevent people commiting in broken docs. We can do this by adding 
```
SPHINXOPTS     = -W
```
to the makefile. Can we do the same when running `sphinx_build` from setup.py? If not, we should use the makefile when running checks.

dcolligan:
@jeromekelleher pushed a new change

jeromekelleher:
Excellent; +1.

PR #345, opened 2015-04-08
Generalised the datamodel file system code.

jeromekelleher:
Addresses issue #232. 

The idea here is push code that is common to variants and the reads datamodel code into a new mixin,  `PysamDatamodelMixin`. 

Also fixes a bug in the exceptions for missing VariantSets.

### DIFF
--- a/ga4gh/datamodel/__init__.py
+++ b/ga4gh/datamodel/__init__.py
@@ -8,6 +8,22 @@ from __future__ import unicode_literals
 
 import ga4gh.exceptions as exceptions
 
+import atexit
+import glob
+import json
+import os
+import shutil
+import tempfile
+
+
+def _cleanupHtslibsMess(indexDir):
+    """
+    Cleanup the mess that htslib has left behind with the index files.
+    This is a temporary measure until we get a good interface for
+    dealing with indexes for remote files.
+    """
+    shutil.rmtree(indexDir)
+
 
 class DatamodelObject(object):
     """
@@ -18,9 +34,11 @@ class DatamodelObject(object):
         pass
 
 
-class PysamSanitizer(object):
+class PysamDatamodelMixin(object):
     """
-    Provides various sanitization methods
+    A mixin class to simplify working with DatamodelObjects based on
+    directories of files interpreted using pysam. This mixin is designed
+    to work within the DatamodelObject hierarchy.
     """
     samMin = 0
     samMaxStart = 2**30 - 1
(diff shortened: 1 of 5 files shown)


macieksmuga:
@jeromekelleher I like where you're going with this.
+1

ga4gh/datamodel/reads.py
```diff
@@ -110,22 +109,22 @@ def __init__(self, id_):
...
-class HtslibReadGroupSet(datamodel.PysamSanitizer, AbstractReadGroupSet):
+class HtslibReadGroupSet(AbstractReadGroupSet, datamodel.PysamDatamodelMixin):
     """
     Class representing a logical collection ReadGroups.
     """
     def __init__(self, id_, dataDir):
         super(HtslibReadGroupSet, self).__init__(id_)
-        self._dataDir = dataDir
-        # we only support BAM files right now;
-        # SAM files (which can't have index files) would be too slow
```
dcolligan: we can't keep this comment?

ga4gh/datamodel/reads.py
```diff
@@ -110,22 +109,22 @@ def __init__(self, id_):
         self._readGroups.append(readGroup)
 
 
-class HtslibReadGroupSet(datamodel.PysamSanitizer, AbstractReadGroupSet):
+class HtslibReadGroupSet(AbstractReadGroupSet, datamodel.PysamDatamodelMixin):
```
dcolligan: reversing this order is bad.  the classes are essentially checked in left to right order for method resolution order (MRO).  the true "base class" should always go last.

ga4gh/datamodel/__init__.py
```diff
@@ -86,3 +104,47 @@ def sanitizeString(cls, attr, attrName):
...
+        """
+        # ctime is in seconds, and we want milliseconds since the epoch
+        ctimeInMillis = int(os.path.getctime(directoryPath) * 1000)
+        self._creationTime = ctimeInMillis
+        self._updatedTime = ctimeInMillis
+
+    def _scanDataFiles(self, dataDir, patterns):
+        """
+        Scans the specified directory for files with the specified globbing
+        pattern and calls self._addDataFile for each. Raises an EmptyDirException
```
dcolligan: Line too long, broseph

PS if you use vim, there's a great utility to automatically format line wrapping http://vim.wikia.com/wiki/Automatic_word_wrapping

dcolligan:
Why aren't the methods you are adding going into DatamodelObject?

Putting a bunch of grab-bag methods into PysamSanitizer (or whatever it comes to be called) seems to be diluting the meaning of it as a class.  It is merely meant to serve as a shim between our code and pysam.

ga4gh/datamodel/reads.py
macieksmuga: While this shouldn't matter for genuine mixins (which have no business redefining symbols from the "true" base class), I support @dcolligan's reasoning. Good blog post about this [here](http://www.ianlewis.org/en/mixins-and-python).

dcolligan: Agreed about genuine mixins.  The problem I was running into (although I forget the details) was that python was throwing a TypeError complaining about the MRO when it hit the class definition.  The takeaway I basically came away from reading up on it was the left-to-right rule.

ga4gh/datamodel/reads.py
jeromekelleher: I don't think we'll ever support SAM, so it seemed misleading to me.

ga4gh/datamodel/reads.py
jeromekelleher: OK, good point. I'll change it back.

ga4gh/datamodel/__init__.py
jeromekelleher: Dammit, I thought I'd fixed this and pushed an update.

I've been looking for a text object to properly work with python docstring comments, so I can go `gqid` (or whatever it would be) and reformat the docs. Sadly, it doesn't seem to exist.

jeromekelleher:
OK, I've pushed an update on the smaller changes.

The reasoning for putting these methods into the `PysamDatamodelMixin` is because they don't belong as methods of `DatamodelObject`. We have DatamodelObjects that are simulated (and so have no connection at all with the file system), and we'll also have DatamodelObjects that are derived from a/some SQLlite DBs soon. So, I think it's a mistake to bury the logic of walking around a directory in the file system looking for some files in the DatamodelObject.

I don't think it's a grab-bag of methods really --- these are all the utilities that you need to use pysam as a DatamodelObject, in one handy place.

However, I'm not at all convinced that the design I have here is the correct one. Definitely open to suggestions on how to model this better...

jeromekelleher:
This is failing because of the docs build thing. Can you kick the CI off again once #348 has been merged please @dcolligan?

ga4gh/datamodel/reads.py
dcolligan: ok

dcolligan:
You're going to need to rebase it on top of `develop` to pick up the changes.  Running CI again on the same failing branch is pointless.

jeromekelleher:
Yes, I think you might have a point there @dcolligan!

jeromekelleher:
Can we vote on this one please?

macieksmuga:
+1

Issue #340, opened 2015-04-07
URL support for read data.


jeromekelleher:
We currently have some cludgy code to allow variant data be read from a remote URL source. The same method can applied to reads. We should try to abstract the details of how we do this so that the VariantSet and ReadGroupSet share the same code --- we could try to adapt (and rename) PysamSanitizer to do this.

~~We should resolve #339 first/also.~~

dcolligan:
What does `PysamSanitizer` have to do with this?

jeromekelleher:
Fixed in #345

PR #352, opened 2015-04-10
Test remote reads functionality

dcolligan:
Issue #340

### DIFF
--- a/ga4gh/datamodel/__init__.py
+++ b/ga4gh/datamodel/__init__.py
@@ -6,14 +6,14 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import ga4gh.exceptions as exceptions
-
-import atexit
 import glob
-import json
 import os
-import shutil
+import json
 import tempfile
+import shutil
+import atexit
+
+import ga4gh.exceptions as exceptions
 
 
 def _cleanupHtslibsMess(indexDir):
@@ -22,7 +22,8 @@ def _cleanupHtslibsMess(indexDir):
     This is a temporary measure until we get a good interface for
     dealing with indexes for remote files.
     """
-    shutil.rmtree(indexDir)
+    if os.path.exists(indexDir):
+        shutil.rmtree(indexDir)
 
 
 class DatamodelObject(object):
@@ -111,7 +112,6 @@ class PysamDatamodelMixin(object):
         DatamodelObject. This is derived from the ctime of the specified
         directoryPath.
         """
-        # ctime is in seconds, and we want milliseconds since the epoch
         ctimeInMillis = int(os.path.getctime(directoryPath) * 1000)
         self._creationTime = ctimeInMillis
         self._updatedTime = ctimeInMillis
@@ -147,4 +147,4 @@ class PysamDatamodelMixin(object):
             os.chdir(cwd)
             atexit.register(_cleanupHtslibsMess, indexDir)
         if numDataFiles == 0:
-            raise exceptions.EmptyDirException(dataDir)
+            raise exceptions.EmptyDirException(dataDir, patterns)
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -13,6 +13,7 @@ import pysam
 
 import ga4gh.protocol as protocol
 import ga4gh.datamodel as datamodel
+import ga4gh.exceptions as exceptions
 
 
 class SamCigar(object):
@@ -114,6 +115,7 @@ class HtslibReadGroupSet(datamodel.PysamDatamodelMixin, AbstractReadGroupSet):
     """
     def __init__(self, id_, dataDir):
         super(HtslibReadGroupSet, self).__init__(id_)
+        self._dataDir = dataDir
         self._readGroups = []
         self._setAccessTimes(dataDir)
         self._scanDataFiles(dataDir, ["*.bam"])
@@ -213,9 +215,11 @@ class HtslibReadGroup(datamodel.PysamDatamodelMixin, AbstractReadGroup):
     """
     def __init__(self, id_, dataFile):
         super(HtslibReadGroup, self).__init__(id_)
-        self._setAccessTimes(dataFile)
         self._samFilePath = dataFile
-        self._samFile = pysam.AlignmentFile(dataFile)
+        try:
+            self._samFile = pysam.AlignmentFile(dataFile)
+        except ValueError:
+            raise exceptions.FileOpenFailedException(dataFile)
 
     def getSamFilePath(self):
         """
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -271,6 +271,7 @@ class HtslibVariantSet(datamodel.PysamDatamodelMixin, AbstractVariantSet):
     """
     def __init__(self, id_, dataDir):
         super(HtslibVariantSet, self).__init__(id_)
+        self._dataDir = dataDir
         self._setAccessTimes(dataDir)
         self._chromFileMap = {}
         self._metadata = None
--- a/tests/end_to_end/remoteTestData/reads/remoteTest/urls.json
+++ b/tests/end_to_end/remoteTestData/reads/remoteTest/urls.json
@@ -1,0 +1,5 @@
+{
+    "urls": [
+        "http://0.0.0.0:8002/wgEncodeUwRepliSeqBg02esG1bAlnRep1_sample.bam"
+    ]
+}
--- a/tests/end_to_end/server.py
+++ b/tests/end_to_end/server.py
@@ -1,146 +1,216 @@
 """
-Assists in executing tests with a running server on localhost
+Servers to assist in testing
 """
 from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
 import logging
-import requests
-import shlex
-import subprocess
+import os
+import random
 import tempfile
-import unittest
+import shlex
+import string
+import subprocess
+
+import requests
 
 import tests.utils as utils
 
 
-class ServerTest(unittest.TestCase):
-    """
-    Manages server bootup and shutdown for a test
-    """
-    def run(self, *args, **kwargs):
-        self.setupServer()
-        try:
-            self.startServer()
-            super(ServerTest, self).run(*args, **kwargs)
-        finally:
-            self.shutDownServer()
+ga4ghPort = 8001
+remotePort = 8002
 
-    def setupServer(self):
+
+class ServerForTesting(object):
+    """
+    The base class of a test server
+    """
+    def __init__(self, port):
         # suppress requests package log messages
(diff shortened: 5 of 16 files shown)


dcolligan:
Pysam's remote read functionality seems to just not work.  On instantiating a remote BAM file, it retrieves the file, but then just hangs and doesn't return control to the client... like there's an indefinite sleep call in pysam or something.
```
>>> a = 'http://localhost:8008/HG00533.mapped.ILLUMINA.bwa.CHS.low_coverage.20120522.bam'
>>> import pysam
>>>
>>> b = pysam.AlignmentFile(a)
[knet_seek] SEEK_END is not supported for HTTP. Offset is unchanged.
^C^C^C^D
^Z
[1]+  Stopped                 python
```
Investigating...

jeromekelleher:
What HTTP server are you using here @dcolligan? It would be useful to see the logs for this transation. Keep us posted in any case.

I don't see why we need a PR for this though: why not the code in #345?

(ps CTRL+\ is your friend in these situations when stuff is hanging in low-level code)

dcolligan:
I'm just doing something like `python -m SimpleHTTPServer 8008` to serve the files.  @jeromekelleher could you do me a favor and post a small bam file on your (presumably, well-functioning) server?

jeromekelleher:
OK, I've untarred the example data, so you can have some 1000G BAMS at http://www.well.ox.ac.uk/~jk/ga4gh-example-data/reads/low-coverage/

I don't have any access to the logs on this machine though, so I think it would probably be better to set up a local Apache for your own sanity.

My guess here is that SimpleHTTPServer doesn't understand the Accept-Ranges header, and that the error handling on the client side isn't dealing with this.

dcolligan:
@jeromekelleher yeah, you're right.  Another local web server that comes with OSX that handles it correcly is:
```
twistd -no web --path=.
```

dcolligan:
Filed this issue with pysam: https://github.com/pysam-developers/pysam/issues/107

jeromekelleher:
Twisted is a good choice --- we can use that to make a HTTP server that'll inject some errors. We'll have to be able to deal with network errors robustly if we're going to use remotely accessed data, so we'll need a way of simulating these.

dcolligan:
Yeah, we're probably going to need to make another end-to-end style test that boots up both the server and a twistd server that serves data files.  In the server's config, we will need to pass `DATA_SOURCE` as being some directory under `test` which has a hierarchy that includes some test `url.json` files.  Not sure if we need any more tests than that... that's quite a bit of work already.

dcolligan:
@jeromekelleher the main test is still failing (not sure why... my best guess is the lack of a content-disposition header served by twistd, but that's probably not it), but at this point I bet you would probably have comments on how this is progressing in any event

ga4gh/datamodel/__init__.py
```diff
@@ -112,9 +113,13 @@ def _setAccessTimes(self, directoryPath):
         directoryPath.
         """
         # ctime is in seconds, and we want milliseconds since the epoch
-        ctimeInMillis = int(os.path.getctime(directoryPath) * 1000)
-        self._creationTime = ctimeInMillis
-        self._updatedTime = ctimeInMillis
+        if os.path.exists(directoryPath):
```
jeromekelleher: I don't think this works as intended --- the directory will always exist, even if the constituent files are remote. It's arguable what  the ctime of the directory actually means here, but this is a bigger issue. I would just leave this as it is.

ga4gh/datamodel/variants.py
```diff
@@ -269,6 +269,7 @@ class HtslibVariantSet(datamodel.PysamDatamodelMixin, AbstractVariantSet):
     """
     def __init__(self, id_, dataDir):
         super(HtslibVariantSet, self).__init__(id_)
+        self._dataDir = dataDir
```
jeromekelleher: Perhaps we should just move this up the inheritance chain, and add some getters?

ga4gh/datamodel/reads.py
```diff
@@ -217,7 +218,10 @@ def __init__(self, id_, dataFile):
         super(HtslibReadGroup, self).__init__(id_)
         self._setAccessTimes(dataFile)
         self._samFilePath = dataFile
-        self._samFile = pysam.AlignmentFile(dataFile)
+        try:
+            self._samFile = pysam.AlignmentFile(dataFile)
+        except ValueError:
```
jeromekelleher: I don't suppose we get any information back in the exception do we?

ga4gh/datamodel/reads.py
```diff
@@ -217,7 +218,10 @@ def __init__(self, id_, dataFile):
...
+            raise exceptions.RemoteFetchFailedException(dataFile)
```
jeromekelleher: Isn't this overly specific? This applies to all files, remote or local.

tests/end_to_end/remoteTestData/reads/remoteTest/urls.json
```diff
@@ -0,0 +1,5 @@
+{
+    "urls": [
+        "http://localhost:8002/bam.bam"
```
jeromekelleher: Why don't we reuse the existing files in `tests/data` rather than adding in new data just for this test?

tests/end_to_end/server.py
```diff
@@ -1,146 +1,202 @@
...
+class TwistedServerForTesting(ServerForTesting):
+    """
+    A twistd test server
+    """
+    def __init__(self, path):
+        super(TwistedServerForTesting, self).__init__(twistdPort)
+        self.path = path
+
+    def getCmdLine(self):
+        cmdLine = "twistd -no web -p {} --path={}".format(
```
jeromekelleher: I think it would be better if we created a Twisted HTTP server in code, and ran this in a subprocess. We're implicitly depending on Twisted here anyway (which we'll need to add to requirements.txt), and it would be good to have direct access to the HTTP server we're using. You should be able to make a HTTP server using twisted in a couple of lines of code.

So, to be clear, I think we should make a function that runs a twisted HTTP server and run this using multiprocessing, rather than calling the shell command `twistd`.

jeromekelleher:
Yeah, it's generally looking good @dcolligan. I think it would help a lot if we make our own HTTP server using Twisted rather than just calling twistd. This should help you track down where the problems are happening.

ga4gh/datamodel/variants.py
dcolligan: Sure, but to where?  `AbstractVariantSet` would not eliminate the redundancy, and `DatamodelObject` is too general.  We could do something in `PysamDatamodelMixin`, but that doesn't have any `__init__` method...

dcolligan: Er actually `DatamodelObject` would work.

dcolligan: I take it back.  `DatamodelObject`s aren't necessarily backed by directories because it can have simulated subclasses.  (I suppose we could initialize the dataDir with None in the case of simulated, etc. subclasses that didn't need a dataDir, but that seems ugly.)  The mixin does seem like the logical place to put it, but again, I don't know if we want to "initialize" a mixin.

ga4gh/datamodel/reads.py
dcolligan: Not much, just this string: "could not open file (mode='r') - is it SAM/BAM format?"

dcolligan: Which in and of itself is misleading in our case

dcolligan:
@jeromekelleher pushed a new update; tests work now.  I think we should push the custom twistd server to another issue.  Same with testing remote variants.  (Btw, my previous problem was I was starting up the servers in the wrong order.)  This push also incorporates the changes from #363 ... I can rebase it on top of that after it is merged.

dcolligan:
hm... the new test works locally but is failing on travis...

ga4gh/datamodel/variants.py
jeromekelleher: I agree that we shouldn't have an init in mixins (that way madness lies). Let's leave it as it is and see if we can tidy the concepts up a bit in our next refactor.

jeromekelleher:
Looks good @dcolligan --- I'm happy to merge once the CI passes.

dcolligan:
Hmm, it looks like pysam just doesn't fetch remote files correctly in Travis' environment.  See https://github.com/dcolligan/server/blob/340_remote_reads/tests/end_to_end/test_fail.py where I have two tests that do identical things -- that is, fetches a file from a twistd web server -- only one uses pysam and one doesn't.  Only the test that uses pysam is failing in that file in CI: https://travis-ci.org/dcolligan/server/builds/59627286

I think I'm going to take this course of action here:
1) file a bug with pysam (although it's going to be really annoying for them to debug since Travis' environment is a black box)
2) move the remote fetch test to a suite of tests that isn't run on travis

@jeromekelleher agree?

dcolligan:
looks like remote fetches on Travis also fail with a direct call to samtools without going through pysam

https://travis-ci.org/ga4gh/server/builds/59639424

diekhans:
I can fork and help put together a pysam/hstlib that produces better error message.

jeromekelleher:
I'm not sure that Travis is the problem @dcolligan --- I've just tried this on my Debian and linux mint machines and it's failed on both. Have you tried this on a linux box?

dcolligan:
I just logged into an AWS ubuntu instance and booted up twistd in one terminal and did a `pysam.AlignmentFile` fetch an other.  Both that and a `samtools view` fetch of a bam file off the twistd server worked fine.

Did you try doing that?  What are the specs of your environments @jeromekelleher ?

jmarshall:
[knet_seek] SEEK_END is not supported for HTTP. Offset is unchanged.

This message comes from samtools's or htslib's _knetfile.c_, which doesn't implement seeking relative to EOF.  (It's likely only doing this to check that the BAM file has an EOF trailer block; not being able to check this shouldn't be ruining reading the file, but _knetfile.c_'s error handling may not be ideal.)

Anyway, htslib will soon be able to do http/etc I/O via libcurl instead of _knetfile.c_, and with this backend SEEK_END will be supported (provided the server allows range requests) and error-checking is greatly improved.  See htslib's [libcurl branch](https://github.com/samtools/htslib/tree/libcurl).  This will be landing in **develop** and new htslib/samtools/etc releases soon, but in the meantime you might like to build pysam against this **libcurl** branch, configuring with `‑‑enable‑libcurl`, and see if the problem persists.

dcolligan:
Tests are passing now... the last issue had something to do with this, I think: https://github.com/samtools/samtools/issues/382

Ready for final review / merge.

ga4gh/datamodel/__init__.py
jeromekelleher: I think we should just remove the condition here --- it doesn't actually do anything since the directory will always exist.

jeromekelleher:
I'm happy to merge once the issue above is sorted out.

jeromekelleher:
Merging, thanks @dcolligan.

PR #353, opened 2015-04-14
protocols updated to reflect schemas commit e6d6074

macieksmuga:
This is the first of a short series of PA's intended to update the protocol definitions from
0.5.1 to the most recent version found on ga4gh/schemas.
Since graph reference support introduced a host of changes, this PR aims to bring
the server protocols up to the last commit before graph support was merged into the schemas repo.

This specific PR includes removing the GA prefix, introduction of new ontology entities, as well as minor semantic changes (especially with union types). 

The state of this PR is very rough, and several tests are currently failing (they've been decorated so as to be skipped). The intention, as discussed with @jeromekelleher and @dcolligan is to fix these failing tests through a followup series of issues/short PR's.

### DIFF
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -12,7 +12,6 @@ import os
 import pysam
 
 import ga4gh.protocol as protocol
-import ga4gh.exceptions as exceptions
 import ga4gh.datamodel as datamodel
 
 
@@ -23,15 +22,15 @@ class SamCigar(object):
     # see http://pysam.readthedocs.org/en/latest/api.html
     # #pysam.AlignedSegment.cigartuples
     cigarStrings = [
-        protocol.GACigarOperation.ALIGNMENT_MATCH,
-        protocol.GACigarOperation.INSERT,
-        protocol.GACigarOperation.DELETE,
-        protocol.GACigarOperation.SKIP,
-        protocol.GACigarOperation.CLIP_SOFT,
-        protocol.GACigarOperation.CLIP_HARD,
-        protocol.GACigarOperation.PAD,
-        protocol.GACigarOperation.SEQUENCE_MATCH,
-        protocol.GACigarOperation.SEQUENCE_MISMATCH,
+        protocol.CigarOperation.ALIGNMENT_MATCH,
+        protocol.CigarOperation.INSERT,
+        protocol.CigarOperation.DELETE,
+        protocol.CigarOperation.SKIP,
+        protocol.CigarOperation.CLIP_SOFT,
+        protocol.CigarOperation.CLIP_HARD,
+        protocol.CigarOperation.PAD,
+        protocol.CigarOperation.SEQUENCE_MATCH,
+        protocol.CigarOperation.SEQUENCE_MISMATCH,
     ]
 
     @classmethod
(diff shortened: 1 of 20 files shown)


dcolligan:
+1

ga4gh/datamodel/reads.py
```diff
@@ -174,7 +173,7 @@ class SimulatedReadGroup(AbstractReadGroup):
     def __init__(self, id_):
         super(SimulatedReadGroup, self).__init__(id_)
 
-    def getReadAlignments(self, referenceName=None, referenceId=None,
```
jeromekelleher: Will the full method declaration on one line now?

ga4gh/datamodel/reads.py
```diff
@@ -225,13 +224,14 @@ def getSamFilePath(self):
         """
         return self._samFilePath
 
-    def getReadAlignments(self, referenceName=None, referenceId=None,
+    def getReadAlignments(self, referenceId=None,
```
jeromekelleher: Ditto here --- no point in breaking lines unless we need to.

ga4gh/datamodel/reads.py
```diff
@@ -225,13 +224,14 @@ def getSamFilePath(self):
...
                           start=None, end=None):
         """
         Returns an iterator over the specified reads
         """
-        if referenceName is not None and referenceId is not None:
-            raise exceptions.BadReadsSearchRequestBothRefs()
+        # TODO If referenceId is None, return against all references,
+        # including unmapped reads.
```
jeromekelleher: Need a TODO here for referenceName also.

tests/unit/test_schemas.py
```diff
@@ -76,11 +76,14 @@ def getTypicalValue(self, cls, fieldName):
         typ = field.type
         if isinstance(typ, avro.schema.UnionSchema):
             t0 = typ.schemas[0]
-            if (isinstance(t0, avro.schema.PrimitiveSchema) and
-                    t0.type == "null"):
-                typ = typ.schemas[1]
-            else:
-                raise Exception(err)
+            t1 = typ.schemas[1]
+            if (isinstance(t0, avro.schema.PrimitiveSchema)):
```
jeromekelleher: extra brackets on all these conditions here.

jeromekelleher:
This is perfect, thanks @macieksmuga. A surgical GA-ectomy if I ever saw one!

+1 modulo the minor comments above.

ga4gh/datamodel/reads.py
macieksmuga: @jeromekelleher I'm not sure I understand what you're asking - unless we're to re-introduce referenceName as an argument to this method (which we could, if you think it's a good idea), I don't believe there's a TODO to add for it: If referenceId is not none, the reference name is automatically looked up within the body of the method.

macieksmuga:
@jeromekelleher other than the above question about referenceName in getReadAlignments, I think I've addressed all your issues.

ga4gh/datamodel/reads.py
jeromekelleher: You're right, sorry. I should have read ahead a bit here.

jeromekelleher:
Thanks @macieksmuga, merging.

Can you open issues for each of the disabled unittests please?

Issue #356, opened 2015-04-16
Re-enable &  fix unit tests in test_schemas.py for schemas_e6d6074


macieksmuga:
Resulting from schema update in PR #353

macieksmuga:
Fixing this issue is uncovering inconsistencies in the Avro schemas themselves since 0.5.1. I'll shortly create a PR in ga4gh/schemas to address all the ones I've found.

dcolligan:
Fixed in #515

PR #359, opened 2015-04-16
Fixed issues flagged by failing tests in test_schemas

macieksmuga:
All the problems in test_schemas came from inconsistencies in the schema definitions themselves. I'll be creating a separate pull request on those based on my experiences here. While I needed to edit `_protocol_definitions.py` by hand, fixing the underlying schema problems in the future will make such manual adjustments unnecessary.

Addresses Issue #356


jeromekelleher:
Not sure if I agree with this; can you give us a bit more detail on why they were breaking, and what you have changed?

macieksmuga:
Specifically: An array member of a record was being defaulted to null, and inconsistent duplicate definitions of  the `Experiment` record were being pulled from two places (`metadata` and `reads`). I made a [pull request](https://github.com/ga4gh/schemas/pull/286) against schemas to address these issues there, but for now edited the `_protocol_definitions` file by hand to make the tests pass.

@jeromekelleher I do agree, however, that this may not be worth pulling - this issue will resolve itself automatically once we update to the corrected schemas. Shall I close this PR then?

jeromekelleher:
I think so @macieksmuga, let's close this PR and fix the problem at the source.

macieksmuga:
Closing PR, since underlying problem is addressed in this [schemas PR](https://github.com/ga4gh/schemas/pull/286)

PR #370, opened 2015-04-24
Make process_schemas.py easier to read / use

dcolligan:
- fix some text wrapping bugs
- refactor code so we're not computing whitespace everywhere
- make the version argument be literally accepted when the -i flag is used

### DIFF
--- a/scripts/process_schemas.py
+++ b/scripts/process_schemas.py
@@ -127,42 +127,38 @@ class SchemaClass(object):
         output = "\n".join(textwrap.wrap(jsonData)) + "\n"
         return output
 
-    def formatRequiredFields(self):
+    def writeRequiredFields(self, outputFile):
         """
-        Returns a string encoding the set of required fields (i.e those
-        fields that do not have a default value.
+        Writes a string encoding the set of required fields (i.e those
+        fields that do not have a default value)
         """
         fields = []
         for field in self.getFields():
             if not field.has_default:
                 fields.append(field)
-
-        if len(fields) < 2:
-            string = "set(["
-            for field in fields:
-                string += '"{0}"'.format(field.name)
-            string += "])"
+        if len(fields) < 1:
+            self._writeWithIndent('requiredFields = set([])', outputFile)
         else:
-            string = "set([\n"
+            self._writeWithIndent('requiredFields = set([', outputFile)
             for field in fields:
-                string += (" " * 8) + '"{0}",\n'.format(field.name)
-            string += (" " * 4) + "])"
-        return string
+                string_ = '"{0}",'.format(field.name)
+                self._writeWithIndent(string_, outputFile, 2)
+            self._writeWithIndent('])', outputFile)
 
     def writeConstructor(self, outputFile):
         # Force using slots to avoid the overhead of a dict per object;
         # when a query returns hundreds of thousands of calls this can
         # save a hundred megabytes or more.
-        print("    __slots__ = ['",
-              textwrap.fill(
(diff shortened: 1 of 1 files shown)


scripts/process_schemas.py
```diff
@@ -266,7 +288,10 @@ def writeHeader(self, outputFile):
         print(file=outputFile)
         print("import avro.schema", file=outputFile)
         print(file=outputFile)
-        versionStr = self.version[1:]  # Strip off leading 'v'
+        if self.version[2] == '.':
+            versionStr = self.version[1:]  # Strip off leading 'v'
+        else:
+            versionStr = self.version
         print("version = '{0}'".format(versionStr), file=outputFile)
```
saupchurch: This may be a little pedantic, but if we are stripping the leading 'v' shouldn't we check on a leading 'v' instead of assuming the position of the point?  Like this, we'll have trouble here once we get to v10.#.#

pgrosu: +1 @saupchurch, not pedantic at all, on the contrary a very valid, fail-safe point.

scripts/process_schemas.py
```diff
@@ -201,44 +207,60 @@ def write(self, outputFile):
...
             self.writeConstructor(outputFile)
         elif isinstance(self.schema, avro.schema.EnumSchema):
             # TODO make a proper Python enum here using the Python 3.4 enum?
             for symbol in self.schema.symbols:
-                string = '    {0} = "{0}"'.format(symbol, symbol)
-                print(string, file=outputFile)
+                string = '{0} = "{0}"'.format(symbol, symbol)
+                self._writeWithIndent(string, outputFile)
+
+    def _writeWithIndent(self, string_, outputFile, indentLevel=1):
```
jeromekelleher: Very nice, excellent abstraction. We seem to just use `outputFile` as a parameter all the time: do we ever write to more than one file? Perhaps we should make a an instance variable `self._outputFile`, which might tidy things up even more?

scripts/process_schemas.py
jeromekelleher: I agree here also --- what's the intent of this section @dcolligan?

jeromekelleher:
Looks great, +1 modulo minor comments above.

scripts/process_schemas.py
dcolligan: The intent is to allow the user to set the `version = ...` line in `_protocol_definitions.py` without having to edit the file directly.  But yes, I agree the test is brittle.  I'll fix that.

scripts/process_schemas.py
dcolligan: I tried making that refactor, but here's the problem: the `SchemaClass` instance does not know if it should open that file for writing or appending.  That is, it doesn't know if it is the first `SchemaClass` instance created, and why should it have to know anything about the order of other instances' creation?  Aha, so you say, why not just pass in the file stream?  Well, `SchemaClass` instantiation and writing are done in two different places, and the file is only opened when the later is called.  So if we wanted to eliminate all of the `outputFile` argument passing (which I agree is annoying), we would probably need to change the interface.  I'm not super opposed to doing that, but my main intent with this change was to just make the file more readable so nobody has to count white space to debug the output, etc.

dcolligan:
@jeromekelleher  Pushed new commit.

scripts/process_schemas.py
jeromekelleher: OK, I'm good with that.

jeromekelleher:
Thanks @dcolligan, merging.

Issue #373, opened 2015-04-28
no instructions on running process_schemas


diekhans:
a grep of the tree shows no information on how to run process_schemas

I have updated the schemas in the annotation branch and need to rebuild.

dcolligan:
Running with the `--help` flag should give you a list of options to pass in.  What additional kind of docs do you think we should have on this tool?

diekhans:
Yea, --help makes sense.  I guess mentioning it in the developer
doc and specific how it's supposed to be used for this project
would help people come up to speed.

jeromekelleher:
Good point @diekhans: we should add a subsection on this in the Organisation section of development.rst. This should basically just explain the basic idea of how and why we generate code from the Avro IDL schemas, and an example of running process_schemas.py against a released and local version of the schema definitions.

dcolligan:
Fixed in #377

PR #377, opened 2015-04-28
Added directions about process_schemas.py to docs

dcolligan:
Issue #373

### DIFF
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -292,3 +292,19 @@ the ``ga4gh`` directory in the project root. Within this package, the
 functionality is split between the ``client``, ``server``, ``protocol`` and
 ``cli`` modules.  The ``cli`` module contains the definitions for the
 ``ga4gh_client`` and ``ga4gh_server`` programs.
+
+An important file in the project is ``ga4gh/_protocol_definitions.py``.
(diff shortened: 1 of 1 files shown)


dcolligan:
@diekhans can I get your +1 on this, since you filed the issue?

jeromekelleher:
Looks good @dcolligan, although I feel like it could be fleshed out a bit. I agree @diekhans should have the last word though.

jeromekelleher:
ps. apologies for a very vague comment! I'm fine with this as-is if @diekhans is happy with it.

diekhans:
I think we need to explain what `desired version' means in the contact of the server.

dcolligan:
Alright, added a bit about that.  If you don't like this phrasing @diekhans please specify how you would prefer it.

diekhans:
This doesn;t really say what desiredVersion is:
    Where ``desiredVersion`` is the version that will be written to the ``_protocol_definitions.py`` file.

is it a git tag, sha1 hash, etc.  what are the valid values?

dcolligan:
@diekhans updated the code based on your concerns

dcolligan:
@diekhans can I get a review here?

diekhans:
looks good

Issue #369, opened 2015-04-23
Run process_schemas, etc. successfully on latest schemas repo


dcolligan:
Get rid of all the errors in the schemas repo by running a modified process_schemas (and other tests) on it, and fixing all the errors.

Depends on #366

dcolligan:
Fixed in #515

PR #379, opened 2015-04-28
Update protocol to schemas/dab002b3

dcolligan:
Just creating this PR to get the CI.  Lots of tests to fix still...

Issue #369


dcolligan:
@jeromekelleher do you think the simplifying assumptions in test_schemas:SchemaTest.getTypicalValue  should be gotten rid of?  They're definitely inconsistent with the most recent schemas.

pgrosu:
Travis seems to have failed.  Is this critical?

jeromekelleher:
@dcolligan --- we'll need to update this method all right. I don't think it needs a super-general solution though, whatever fits the new patched-up version of the protocol is good.

@pgrosu --- did you read the (single sentence) description of this PR before commenting? It clearly states that CI was expected to fail. We're trying to run a software development project here, and this sort of noise isn't helping.

pgrosu:
@jeromekelleher Ah!  Sorry I missed that small detail, since usually I get these via email first to review, but lately I never get updates from @dcolligan, and got to this by looking through the open PRs for something else.  I was just posting it as a friendly comment in case it can be merged even if it failed, if the errors are not critical. I was working under the assumption that any PRs usually work before being pushed, but no problem :)  

~p

tests/unit/test_schemas.py
```diff
@@ -236,23 +187,26 @@ def validateClasses(self, factory):
...
-    @unittest.skipIf(protocol.version.startswith("0.6"), "")
     def testValidateDefaultValues(self):
         self.validateClasses(self.getDefaultInstance)
 
-    @unittest.skipIf(protocol.version.startswith("0.6"), "")
     def testValidateTypicalValues(self):
-        self.validateClasses(self.getTypicalInstance)
+        self.validateClasses(avrotools.CreationTool.getInstance)
 
+    # TODO fix ReadAlignment.alignments decode type ambiguity!
```
dcolligan: @jeromekelleher there is a nasty bug here which has to do with some ambiguity in the schemas.  What happens here is that we get nondeterministic failures when `testValidateRandomValues` is enabled when it hits `ReadAlignment` or any type that has `ReadAlignment` embedded in it somewhere.  The reason for this is that `ReadAlignment.alignment`'s signature is:
```
union { null, LinearAlignment, GraphAlignment } alignment = null;
```
When we try and decode a `ReadAlignment` dict into a protocol object, we need to decide what type to interpret the `dict['alignment']` as.  But, unfortunately, the decision is already made for us in `ReadAlignment.getEmbeddedType`, which is obviously wrong when we want to decode the value as a `GraphAlignment`:
```
    @classmethod
    def getEmbeddedType(cls, fieldName):
        embeddedTypes = {
            'alignment': LinearAlignment,  # ugh
            'nextMatePosition': Side,
        }

        return embeddedTypes[fieldName]
```
I guess the solution to this previously was that there was only one non-null type in a union.  But the schemas no longer reflect that convention.

How should we proceed here?

ga4gh/avrotools.py
```diff
@@ -14,6 +14,68 @@
...
+class CreationToolException(Exception):
+    """
+    Something went wrong in the CreationTool
+    """
+
+
+class CreationTool(object):
+    """
+    Provides methods for creating and populating objects with data
+    """
```
jeromekelleher: Maybe we could make this a real class with some state, rather than collection of class methods? Something like 'InstanceGenerator`, which takes a class/schema as an argument at instantiation time, and then has methods like `getTypicalInstance` or `getRandomInstance`. I'm not hugely keen on the classmethod approach --- as @macieksmuga pointed out somewhere else, it's very.... Java.

tests/unit/test_schemas.py
jeromekelleher: Crud; this is a nasty one. We'll have to think about this... The `embeddedTypes` short cut was always a nasty hack, and I guess it's after catching up with us now.

macieksmuga: I spent much of Friday thinking about this along with other issues having to do with mixing "classic" and graph modes into the same "do it all" entities. While I'm mulling how to make a schemas PR to unbundle such objects,  how about something like this for now:
```
union { null, LinearAlignment } alignment = null;
union {null, GraphAlignment } graphAlignment = null;
```
with the intention of only ever having one of these non-null - we're already using this kind of logic all over the place.

ga4gh/avrotools.py
macieksmuga: @jeromekelleher I like your idea!

tests/unit/test_schemas.py
dcolligan: That's certainly one way to do it.  I'll submit a PR to schemas to that effect.

macieksmuga: @dcolligan thanks! I wanted your weigh-in on this before going ahead with a full PR.

ga4gh/avrotools.py
dcolligan: Adding a `SchemaTools` class (it doesn't just generate based on schemas, it checks against them, etc.) is a huge refactor.  I'll submit that in another branch.

dcolligan:
I don't want to go forward with this until #409 is merged, but after that I will try the update again at a more recent schema commit that incorporates the changes I've made to the schemas repo.

dcolligan:
Just pushed an update here with protocol based off `schemas/dab002b3`.  This is a schemas version that incorporates my changes to the schemas repo.  The good news is that key tests (`test_schemas`, etc.) are now passing.  The bad news is that there are still a lot of tests to fix.

dcolligan:
@jeromekelleher I was working on this, but then I stepped back and wondered if it was worth the effort.  Is the schema going to stabilize on a graph/classic split soon?  If so, it doesn't make much sense to update our code to the mixed schema (which is this PR).

jeromekelleher:
It's hard to know @dcolligan, it's all a bit up in the air right now. Hopefully we'll get a little clarity on what the situation with the graph/non-graph schemas is soon. Until then I would park this.

bwalsh:
Hi all,
Just checking in to see if there is anything I can contribute to un-block this.

Briefly:
* ready to start work on integration of an evidence ontology to fulfill /genotypephenotype/search
* dependency on https://github.com/ga4gh/schemas/pull/298, 15 commits ahead of  schemas/dab002b3

How are others working with the newer schema  updates?

Much appreciated.

jeromekelleher:
We're hoping to get the schemas updated soon to a slimmed down version, which we'll then fully implement here. This is happening at ga4gh/schemas#352. Once that completes, hopefully it'll be reasonably easy to add the changes for the g2p work. I guess the simplest thing to do while we're adding the g2p features to the server is to keep a g2p branch here on the server repo, so that this can be updated easily and we can keep it in sync with the head of develop. I've created the g2p branch, for this. We can rename this, or use a different method if you prefer; this is just a suggestion.

In the mean time, I think the simplest thing to do so you can start making progress is the following:

1) Run process_schemas.py on the version of the schema that you want to import.
2) Manually copy the classes that you are interested in from `_protocoldefinitions.py` and copy these into the `_protocoldefinitions.py` in your g2p branch.

This is a horrible kludge, I know, but it's the best way I can think of to get you guys making progress. If you're having trouble with this, contact me or @dcolligan via email and we'll try to help you out.

bwalsh:
Thanks very much for creating the g2p branch.  To get things kicked off, I'll create a PR on that branch with a single test.  That way I can course correct if there are any snafus in the workflow.

Regarding manual editing of `_protocoldefinitions.py`, yes that's what I thought would be the most pragmatic thing.  Not ideal, but it will keep forward progress going.   I also used it as an opportunity to re-enforce within the g2p community the coupling of schema changes with accompanying reference implementation and tests.

bwalsh:
Guys:
Thanks for your support.  
I've got the scaffolding for the g2p/search setup with [this commit](https://github.com/ohsu-computational-biology/server/commit/1af901354d95874b7ba42a8c3a5e940c5683811e). 

There are additions to the following:
```
#	modified:   ga4gh/_protocol_definitions.py
#	modified:   ga4gh/backend.py
#	modified:   ga4gh/frontend.py
#	modified:   tests/unit/test_views.py
```
It simply returns an empty set for queries.  All of the tests work.

I don't necessarily want to create a  PR for this since there is no data returned.

However, I wonder if you could give the commit a quick review and let me know if you would do anything different.

dcolligan:
Closing this since there's a better schema version to target

dcolligan:
New PR is here https://github.com/ga4gh/server/pull/515

Issue #382, opened 2015-05-01
better error responses for bad requests


acthp:
Error responses for bad requests look like

    {blah} is not a valid instance of <class 'ga4gh._protocol_definitions.GASearchVariantsRequest'>

which is less than helpful. It would be good to return why the request is not valid, like 'referenceName is undefined', etc.

jeromekelleher:
I agree, this would be a useful feature. I think we can now do this using the Avro technology that @dcolligan developed. How feasible would it be to add some extra information, and what could we provide @dcolligan?

dcolligan:
Yeah, we could definitely improve error messages.  `avrotools` can't provide something like 'referenceName is undefined', but it can provide the subset of the JSON hierarchy that is erroneous.

dcolligan:
Fixed in #383

PR #383, opened 2015-05-01
Improve request/response failure error messages

dcolligan:
Error messages now look like:
```
"Request '{u'referenceId': None, u'end': None, u'readGroupIds': [], u'pageSize': None, u'pageToken': None, u'start': u'thisIsWrong'}' is not a valid instance of <class 'ga4gh._protocol_definitions.SearchReadsRequest'>; invalid fields: {u'start': u'thisIsWrong'}"
```

Issue #382

### DIFF
--- a/tests/unit/test_exceptions.py
+++ b/tests/unit/test_exceptions.py
@@ -79,11 +79,56 @@ class TestExceptionConsistency(unittest.TestCase):
 
     def testInstantiation(self):
         for class_ in self._getExceptionClasses():
-            numInitArgs = len(inspect.getargspec(class_.__init__).args) - 1
-            args = ['arg' for _ in range(numInitArgs)]
+            # some exceptions are becoming too complicated to instantiate
+            # like the rest of the exceptions; just do them manually
+            if class_ == exceptions.RequestValidationFailureException:
+                wrongString = "thisIsWrong"
+                objClass = protocol.SearchReadsRequest
+                obj = objClass()
+                obj.start = wrongString
+                jsonDict = obj.toJsonDict()
+                args = (jsonDict, objClass)
+            elif class_ == exceptions.ResponseValidationFailureException:
+                objClass = protocol.SearchReadsResponse
+                obj = objClass()
+                obj.alignments = [protocol.ReadAlignment()]
+                obj.alignments[0].alignment = protocol.LinearAlignment()
+                obj.alignments[0].alignment.mappingQuality = wrongString
+                jsonDict = obj.toJsonDict()
+                args = (jsonDict, objClass)
+            else:
+                numInitArgs = len(inspect.getargspec(
+                    class_.__init__).args) - 1
+                args = ['arg' for _ in range(numInitArgs)]
             instance = class_(*args)
             self.assertIsInstance(instance, exceptions.BaseServerException)
             message = instance.getMessage()
             self.assertIsInstance(message, basestring)
             self.assertGreater(len(message), 0)
             self.assertEqual(instance.getErrorCode(), class_.getErrorCode())
+
+    def testValidationFailureExceptionMessages(self):
+        # RequestValidationFailureException
+        wrongString = "thisIsWrong"
+        objClass = protocol.SearchReadsRequest
+        obj = objClass()
+        obj.start = wrongString
+        jsonDict = obj.toJsonDict()
+        instance = exceptions.RequestValidationFailureException(
+            jsonDict, objClass)
+        self.assertIn(
+            "invalid fields: {u'start': u'thisIsWrong'}",
+            instance.message)
+        self.assertEqual(instance.message.count(wrongString), 2)
+
+        # ResponseValidationFailureException
+        objClass = protocol.SearchReadsResponse
+        obj = objClass()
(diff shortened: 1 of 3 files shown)


tests/unit/test_exceptions.py
```diff
@@ -79,6 +79,12 @@ def testCodeInvariants(self):
 
     def testInstantiation(self):
         for class_ in self._getExceptionClasses():
+            # some exceptions are becoming too complicated to instantiate
+            # like the rest of the exceptions; just do them manually
+            if class_ in [
+                    exceptions.RequestValidationFailureException,
+                    exceptions.ResponseValidationFailureException]:
+                continue
```
jeromekelleher: `continue`s make me ill; can't we work around this, and keep the nice loop? What's the issue that we hit with these two?

tests/unit/test_exceptions.py
```diff
@@ -87,3 +93,28 @@ def testInstantiation(self):
             self.assertIsInstance(message, basestring)
             self.assertGreater(len(message), 0)
             self.assertEqual(instance.getErrorCode(), class_.getErrorCode())
+
+    def testInstantiationManual(self):
```
jeromekelleher: We should make this more specific: `testValidationFailureExceptions` seems good.

jeromekelleher:
+1 modulo comments above.

tests/unit/test_exceptions.py
diekhans: Agreed.  Continues are worse than gotos. At least gotos have an obvious label.

I simplify stuff like this by breaking the loop in to functions.

dcolligan: @jeromekelleher the point of this test is to test that we can create Exception objects without error.  Some exception objects, however, accept arguments, so we generate dummy arguments (always the string 'arg') to pass into the object.  In the case of *ValidationFailureException, those exceptions take a json string and a class as their two arguments, so attempting to pass in 'arg' causes an exception.  That's why we need to special-case these two.  I don't see any other way around this.

jeromekelleher: Well, what about just creating a different args list for these classes and carrying on with the test?

dcolligan: To be clear, you would not be fine with a continue, but you would be fine with this?
```
if class_ in [...]:
 args = oneThing
else:
 args = default
```
That seems like a very similar construct...

jeromekelleher: Each `continue` corresponds to at least ten dead puppies...

Anyway, I think
```python
args = defaultValue
if class_ in exceptionList:
    args = someThingElse
# Run the test with args...
```
is far better than not bothering running the test at all for the classes that are slightly different.

macieksmuga:
@jeromekelleher @diekhans huh? `continue`, when used (properly) at the beginning of a loop is a perfectly clear flow control technique (unlike the wild-west of `goto`). I wasn't aware it was considered "bad". Cf. discussion here:

http://programmers.stackexchange.com/questions/58237/are-break-and-continue-bad-programming-practices

But feel free to educate me.

jeromekelleher:
@macieksmuga I've never seen a case where
```python
for x in something:
     if not condition(x):
          break
      loop_body()
```
was better than
```python
for x in something:
    if condition(x):
         loop_body()
```
The second option makes it far clearer that we are conditionally evaluating `loop_body()` based on the value of x. Visual structure is important in Python, and we lose this link between visual structure and the semantics of the code by using `continue`. The only reason to use `continue` is so your code doesn't become indented too deeply, and this is a key tell-tale that your code is terrible in the first place.

*At least* ten dead puppies!

macieksmuga:
@jeromekelleher how can a litter-full of dead puppies not move my heart? Consider me convinced :)

dcolligan:
@jeromekelleher pushed update, puppies saved

tests/unit/test_exceptions.py
```diff
@@ -79,9 +79,36 @@ def testCodeInvariants(self):
 
     def testInstantiation(self):
         for class_ in self._getExceptionClasses():
-            numInitArgs = len(inspect.getargspec(class_.__init__).args) - 1
-            args = ['arg' for _ in range(numInitArgs)]
-            instance = class_(*args)
+            # some exceptions are becoming too complicated to instantiate
```
jeromekelleher: I actually quite liked the old idea of keeping these two in their own test, where we tested the messages of the errors that we got back in detail. To me, this test is just to make sure we instantiate correctly and that we get back a string of some sort when we as for it. So, I would take the detailed tests for the ValidationFailureExceptions out of this loop, and spin them back into their own test. Then, I would have here something like
```python
if class_ == exceptions.RequestValidationFailureException:
    args = ["{}", protocol.SearchReadsRequest()]
elif # etc
    args = #etc
else:
    numInitArgs = #etc
    args = ['arg' for _ in range(numInitArgs)]
instance = class_(*args)
# tests
```

dcolligan:
@jeromekelleher pushed update, no net puppy gain/loss this time

jeromekelleher:
Thanks @dcolligan, merging.

Issue #385, opened 2015-05-04
Comment out unimplemented paths in frontend.py


dcolligan:
The only difference in functionality will be that now only the paths that are really implemented will show up in the HTML display at /.

dcolligan:
Fixed in #386

PR #386, opened 2015-05-04
Take out unused paths

dcolligan:
Issue #385 

This is so the paths don't appear in the HTML display, which can be confusing to a new user when s/he tries to hit them and they don't work.  We're not losing any functionality here, and can look up this commit and incorporate its reversion when we actually do want to implement these paths.

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -257,21 +257,6 @@ def indexRedirect(version):
     return index()
 
 
-@app.route('/<version>/references/<id>', methods=['GET'])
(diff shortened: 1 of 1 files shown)


jeromekelleher:
+1

macieksmuga:
+1 I'm doing the equivalent path cleaning in the graph branch. Will make for a cleaner merge between the two down the line.

macieksmuga:
@dcolligan d'you mind rebasing your branch on top of the latest develop so I can do a clean merge?

Issue #384, opened 2015-05-01
Create /current alias for version number


dcolligan:
It's tedious to type in the version number into the URL when we are only supporting one version.  We should create an alias like '/latest' or '/current' so we can request resources like `http://localhost:8080/current/variants/search`

diekhans:
+1  this is good for demos, etc.

jeromekelleher:
+1

dcolligan:
Fixed in #387

PR #387, opened 2015-05-04
Add 'current' url version alias

dcolligan:
Issue #384

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -29,6 +29,15 @@ class Version(object):
     """
     A major/minor/revision version tag
     """
+    currentString = "current"
+
+    @classmethod
+    def isCurrentVersion(cls, versionString):
+        if versionString == cls.currentString:
+            return True
+        return (Version.parseString(versionString) ==
+                Version.parseString(protocol.version))
+
     @classmethod
     def parseString(cls, versionString):
         versions = versionString.strip('vV').split('.')
@@ -228,9 +237,8 @@ def handleFlaskPostRequest(version, flaskRequest, endpoint):
     at at the specified version. Invokes the specified endpoint to
     generate a response.
     """
-    if Version.parseString(version) != Version.parseString(protocol.version):
+    if not Version.isCurrentVersion(version):
         raise exceptions.VersionNotSupportedException()
-
     if flaskRequest.method == "POST":
         return handleHttpPost(flaskRequest, endpoint)
     elif flaskRequest.method == "OPTIONS":
--- a/tests/unit/test_views.py
+++ b/tests/unit/test_views.py
@@ -191,3 +191,8 @@ class TestFrontend(unittest.TestCase):
     def testWrongVersion(self):
         path = '/v0.1.2/variantsets/search'
         self.assertEqual(404, self.app.options(path).status_code)
+
+    def testCurrentVersion(self):
+        path = '/{}/variantsets/search'.format(
+            frontend.Version.currentString)
+        self.assertEqual(200, self.app.options(path).status_code)


diekhans:
+1

saupchurch:
Do we expect the reference server handle multiple simultaneous versions of the schema?  I'm +1 for implementing 'current' URL for now to make development and demo work easier (although the implementation is a bit more 'ignore version' than 'use the current version'), but perhaps at some point we should consider moving the version number from the URL to the result JSON and stabilize the URL.

dcolligan:
@saupchurch we only handle a single version right now.  Handling multiple versions would be a lot of work (but may be something that happens in the distant future).

diekhans:
Give how hard it is to get the resource to do one version, I
would say that multiple version should not be an immediate goal.

saupchurch:
If there is only a single supported version then it seems unnecessary to have it in the URL at all as it can either be gleaned from the code or reported statically on the server home page.  Making the URL dependent on version just breaks things when it gets changed which is what drives us to replace it with 'current' in the first place.

diekhans:
The version does change, so it's an indication that it has.

Not all server instances running will be using the same version.

It's good planning for the future.

It is also an example of how version support can be done
for other servers.

saupchurch:
OK +1

dcolligan:
We want an easy, programatic way to indicate to the client that they're using the wrong version before they get some low-level error that is unintelligible to them.  Using `current` is easier to use, but also foregoes this check, so really should only be used be people that know what they're doing (i.e. what version they are on).

ga4gh/frontend.py
```diff
@@ -228,7 +230,9 @@ def handleFlaskPostRequest(version, flaskRequest, endpoint):
     at at the specified version. Invokes the specified endpoint to
     generate a response.
     """
-    if Version.parseString(version) != Version.parseString(protocol.version):
+    if version == Version.currentString:
```
jeromekelleher: I think this logic should be internalised into the Version class. What if we had a method on version called `isCurrentVersion` which takes a string, which can be either 'current' or a version string and returns a Boolean?

jeromekelleher:
I'm +1 for this change (modulo the nitpicks above). 

However, I think we should be using 'Content-Type' as the definitive API versioning mechanism, as the Matchmaker team have done. They use a content-type which is something like 'json+mme-v1.0' (sorry, I'm doing this from memory, and can't find the slides!), which is far better than a URL prefix I think. Among other advantages, this allows the client and server negotiate a protocol version that they both understand and seems a lot more flexible than URLs.

Anyway, this would need to be specified on the schemas repo before it should be implemented here, so it's not directly relevent to this PR.

diekhans:
Suggest investigating if this Content-Type approach is used
elsewhere.  It's appealing, although I am a bit worried about
unexpected downsides.  Generally content type is fairly generic
format.

diekhans:
whoever designed &format=json doesn't understand HTTP

pgrosu:
You prefer to define a new protocol - I believe `JSTP` and `JTP` are taken :)  I understand but it's pretty much the standard these days with many APIs over the web.

jmarshall:
> I think we should be using 'Content-Type' as the definitive API versioning mechanism

:+1:

The first page of Google search results for `content-type versioning` make for good reading.

dcolligan:
If anyone wants to file an issue on schemas for content-type versioning (although I find the idea a little strange myself... seems like a higher bar for custom clients to implement), be my guest.  This change is about an incremental improvement in the URL versioning scheme.

jeromekelleher:
Agreed @dcolligan --- sorry for veering off on a tangent there.

dcolligan:
@jeromekelleher fixed nit

diekhans:
ticket for collection HTTP/ReST specification issue is here:

       https://github.com/ga4gh/schemas/issues/300

jeromekelleher:
Thanks @dcolligan, merging.

PR #388, opened 2015-05-04
Make GET /{version} return the HTML page as well

dcolligan:
Rationale:
Some might think /{version} not / is the base url and fire off a browser request to it.  Easy mistake to make.


jeromekelleher:
Maybe add '/current' also, following #387?

+1 either case.

dcolligan:
`current` is interpreted as the version argument, so there's no need for an additional route

Issue #374, opened 2015-04-28
References support needed.


jeromekelleher:
We currently do not support the references section of the API, which we should redress. This can act as a meta-issue for references support, and can be closed once the all of the sub-issues have been resolved.

~~1. Choose an alternative (pip installable) FASTA file parsing library, for data driven tests.~~
2. Implement `toProtocolElement` for `ReferenceSet` and `Reference` in ga4gh/datamodel/references.py, and add data driven tests for this functionality (using the library chosen above). 
3. Implement `getBases` (or whatever seems appropriate) in `datamodel/references.py` as the low-level equivalent of the `ListReferenceBasesRequest` method. This should have `start` and `end` parameters, and return a string. Create datadriven tests for this functionality, and check corner cases. Add lots of examples of FASTA files.
4. Add a `ReferenceSimulator` that generates random sequence in a reproducible manner.
5. Add support for the ListReferenceBases queries in `backend.py`. Add tests for this functionality in all the appropriate places.
6. Update the ga4gh-example-data to include the relevant subsets of the GRC references for the 1000G data (but see #312).

dcolligan:
Wrt (1):
```
$ pip search fasta
...
fastavro            - Fast iteration of AVRO files
pyfaidx             - pyfaidx: efficient pythonic random access to fasta
                      subsequences
pyfasta             - fast, memory-efficient, pythonic (and command-line) access
                      to fasta sequence files
fastools            - FASTA/FASTQ analysis and manipulation toolkit.
multifastadb        - present a collection of indexed fasta files as a single
                      source
pyfastaq            - Script to manipulate FASTA and FASTQ files, plus API for
                      developers
tfasta              - Parses fasta files using templates and creates formatted
                      fasta.
fastinterval        - Interval class and fasta access
oldowan.fasta       - Read and write FASTA format.
nebfa               - Fasta file parser.
fastaq              - fastAQ is a very and super lightweight package for working
                      with FASTA/FASTQ sequences
fastac              - Compiler for FASTA files and a FASTA-based DNA scripting
                      language.
fasta               - The fasta python package enables you to deal with
                      biological sequence files easily
phytab_splitter     - Splits Phytab files into FASTA
```
Do we have any reason for picking one of these over the other?

dcolligan:
PageRank says this one https://pypi.python.org/pypi/pyfasta/

jeromekelleher:
I'd try out pyfasta and see how it goes. We're looking for:

[...]

PR #390, opened 2015-05-04
References support

dcolligan:
Issue #374


dcolligan:
submitting a PR since I'm sure @jeromekelleher will have thoughts on the code thus far

jeromekelleher:
Frankly, I'm a bit in the dark @dcolligan. I don't think it's possible to turn an arbitrary collection of FASTA files into a GA4GH ReferenceSet, so we're going to have to add some metadata. I don't have any experience with FASTA, so I don't have any idea of how this should be done.

@diekhans have you any thoughts here? What's the right way to do this? A single FASTA file corresponding to a ReferenceSet, which contains a bunch of sequences, or a directory full of files, each containing 1 or more sequences? How do we specify the required metadata? Do we attempt to decipher the metadata line in the FASTA file, or do we just require the existence of a 'metadata.json' file (for example) holding the information that we need?

Anyone else want to weigh in here?

dcolligan:
Trying to parse the first line of a FASTA file for metadata is madness.  There's absolutely no standardization there, and if we invent our own ad-hoc format it should be at least a bit more structured.

diekhans:
> Trying to parse the first line of a FASTA file for metadata is madness.

Welcome to bioinformatics ;-)

This is one of the huge problems GA4GH is trying to fix.

I think the right approach is to have libraries handle the common
work and then have commands tailored to specific sources of data.

Even though GFF3 is better standardized than FASTA, it's going
to be far easier to write a gencode_gff3 import command than
works with a specific data set than tring to have some that is
parameterized to handle special cases semantics.  I would rather
have lots of specific commands than complex `soft coding' of a
general one.

We have libraries and people provide commands to get their 
data to the right format.

jeromekelleher:
Some specific questions for you above @diekhans...

pgrosu:
It's not as bad as people think.  Standardization is your friend and yes @diekhans is right :)  We fixed this problem in industry by mapping to controlled vocabulary and sometimes restructuring it.  Anyone submitting references would pre-process them before submitting them to the server to integrate with other data.

dcolligan:
@jeromekelleher pushed an updated change.  Want your opinion on how we should abstract the stack of the listReferenceBases code path, or if it's worth it at all for just one server method.

jeromekelleher:
I had a quick look at the `listReferenceBases` code path @dcolligan, and not crazy about the current setup. My main concern is the flask dependency within `backend.py`: I think backend.py should be able to operate completely independently of Flask.  I think we should try to follow the pattern of the rest of the objects, and create a dictionary out of the Flask params within frontend.py, and then pass this on to the appropriate handler in the backend where it can be checked to see if it is an instance of  `ListReferenceBasesRequest`. This is slightly different to the search requests (where we pass the data as a string), but I think this is OK.

dcolligan:
@jeromekelleher alright, pushed an update; let me know what you think

ga4gh/frontend.py
```diff
@@ -207,6 +214,24 @@ def handleHttpPost(request, endpoint):
     return getFlaskResponse(responseStr)
 
 
+def handleList(id_, endpoint, request):
+    """
+    Handles the specified HTTP GET request, mapping to a list request
+    """
+    requestArgs = request.args
```
jeromekelleher: Might as well just include `request.args` in the list of parameters here.

ga4gh/backend.py
```diff
@@ -209,6 +214,83 @@ def getReadGroupSets(self):
...
+        return self.runGetRequest(self._referenceSetIdMap, id_)
+
+    def listReferenceBases(self, id_, requestArgs):
+        # parse arguments
+        try:
+            reference = self._referenceIdMap[id_]
+        except KeyError:
+            raise exceptions.ObjectWithIdNotFoundException(id_)
+        start = 0
+        end = datamodel.PysamDatamodelMixin.fastaMax
```
jeromekelleher: Is there anyway that we can use the validation machinery here to avoid some of this messiness? Can't we create a `ListReferenceBasesRequest` object out of the requestArgs dictionary? We then have full validation, we reuse the general machinery for parsing and life is good.

jeromekelleher:
Generally looks good @dcolligan, modulo the minor comments above.  I'll need to spend some time to play with it properly and go through the code more thoroughly though.

ga4gh/backend.py
dcolligan: `protocol.fromJsonDict` just assigns a value if there is an attribute name match; it doesn't check the type.  (Not sure if we want to change this.)

That is, we need to cast `requestArgs.start` to an int from a string ourselves.

jeromekelleher: Ah right; we can't call `validate` on the dict as it's just a string:string map, and there's no type conversion done. OK, if there are no other GET requests with multiple typed arguments (are there, or will there be?) then I guess what you have is as good as we need. If there are, then I think it's worthwhile thinking about how to generalise this code, and getting automated parsing and type checking of these values from the schema. We can open an issue on this if you would prefer not to slow this PR down any more.

dcolligan: Yeah, I'm generally an advocate of "incremental progress with less-than-perfect code, and refactor as it becomes a problem" over "implement clean, very general abstractions before they are necessarily widely needed."  (As has been said in other contexts before, the later is so... java.)  Let's go forward with this PR and file and issue to generalize list request abstractions.

There is still this puzzling CI failure to fix, though.

jeromekelleher:
I've had a quick play with this @dcolligan, and I think there are still some issues:

1. When I run `referencesets-search` on the client, the `referenceIds` attribute is empty in the returned

tests/datadriven/test_references.py
```diff
@@ -31,23 +32,89 @@ class ReferenceSetTest(datadriven.DataDrivenTest):
...
+        Container class for information about a reference
+        """
+        def __init__(self, referenceSetId, fastaFileName):
+            filename = os.path.split(fastaFileName)[1]
             referenceId = "{}:{}".format(
                 referenceSetId, filename.split(".")[0])
-            self._idFastaFileMap[referenceId] = pysam.FastaFile(path)
+            self.id = referenceId
+            self.fastaFile = pysam.FastaFile(fastaFileName)
+            self.bases = self.fastaFile.fetch(self.fastaFile.references[0])
```
jeromekelleher: Is this reading all the bases into memory? If so, we can't do this. GRCH37 is 3GB...

dcolligan: Well, let's not include GRCH37 in our test data files, then!

jeromekelleher: Oops, sorry! My bad, didn't realise this was the test path. This brings up another issue: shouldn't we be using some other library to access the FASTA data (i.e. not pysam) as a sanity check?

dcolligan: Maybe, but it's no worse than what we do in other test modules :)

dcolligan:
Alright, I finally fixed the CI (the issue was that the CI was passing locally but failing on travis because the `.gz` file was not checked in and therefore not present when the test was run on travis, but that wasn't obvious to me since `*.gz` is a pattern in `.gitignore`).

tests/datadriven/test_references.py
jeromekelleher: Only for reads, and there because we have to essentially. For variants we use PyVCF.

jeromekelleher:
What's the status with this @dcolligan? Have you tried it with large-scale data? If so, I'll have another play with it and hopefully we can get it merged in the next couple of days.

dcolligan:
@jeromekelleher `references-get` works for me.  Maybe you're forgetting to pass in an id (it's now a required field for that command, not an optional field -- I changed it because this seems to make more sense to error out early on the client)

dcolligan:
@jeromekelleher pushed a new update that fixes `referencesets-search` not having the `referenceIds` attribute set

dcolligan:
@jeromekelleher that ftp link you posted above doesn't seem to be responsive... do you have any other examples of "real world data"?  also, what other FASTA files do you think we should add to the datadriven tests?  Unless there are some corner cases we're covering, I don't really see the value...

jeromekelleher:
OK, the NCBI server is on your side of the water and should be accessible: ftp://ftp-trace.ncbi.nih.gov/1000genomes/ftp/technical/reference/phase2_reference_assembly_sequence/. There's some real-world references here, which we're going to need to support soon, so it'll be good to get some experience of working with them.

dcolligan:
Ok, I loaded the NCBI reference (~3G uncompressed) into the reference server and it seems to work.  The only thing I noticed is that startup of the server is pretty slow, which is due to generating the MD5 checksums.  Precomputing these would definitely speed things up, but I'm not sure if we want to go down that road.  Thoughts?

I needed to split the given NCBI FASTA file up into several in order to get it to map to our model.  See `scripts/split_fasta.py` for a tool I added to help do that.

jeromekelleher:
Excellent, thanks @dcolligan. For now, I would disable md5 sums --- we'll need to store these, and a bunch of other metadata about the references, so we might as well do that all in one go in a follow up. Recomputing md5s on 3G files each time the server starts up would be very frustrating for users, not to mention bad for scale-out performance, where we want to be able to quickly spin up new instances.

Do we really need to split the FASTA up? Can't we reuse the convention from VCFs, where we have one or more files, and we automatically map to the file in question based on the query?

dcolligan:
@jeromekelleher just pushed a change that disabled Md5 computation.  Server startup is much faster.

Re: FASTA file splitup: The `ListReferenceBasesRequest` request is to `GET /references/{id}/bases` and has three attributes: `end`, `start`, and `pageToken`.  There is no attribute for indexing into a specific sequence within a reference.

jeromekelleher:
I think we can work around that issue easily enough @dcolligan, but we should merge this much now. It's arguable whether we should support single fasta references or not: if someone really wants it, let them open an issue on it.

This looks good and works well; thanks @dcolligan. It's starting to look like a real server now!

+1.

PR #393, opened 2015-05-06
Graph reference server, starter batch

macieksmuga:
graph reference server initial code, consisting of endpoints:
sequences/search, joins/search, sequences/{id}/bases, allele/{id}


macieksmuga:
Making this public at long last. Have fun @jeromekelleher @dcolligan

Issue #394, opened 2015-05-06
avrotools refactor


dcolligan:
Make a `SchemaTool` class in `avrotools.py` that exposes all of the current classmethod functionality in `avrotools` as instance methods.  Move functionality in `tests.utils` into `SchemaTool`.  Refactor `test_schemas` to use `SchemaTool`.

dcolligan:
Fixed in #397

PR #397, opened 2015-05-07
Refactor avrotools

dcolligan:
Issue #394


ga4gh/avrotools.py
```diff
@@ -152,9 +223,194 @@ def _getInvalidFields(cls, expectedSchema, datum):
...
+            "int": cls._integerGenerator(),
+            "long": cls._longGenerator(),
+            "string": cls._stringGenerator(),
+            "float": cls._floatGenerator(),
+            "double": cls._doubleGenerator()
+        }
+        return typeToGenerator
+
+    @classmethod
+    def _valueFromSetGenerator(cls, possibilities=[]):
```
dcolligan: Ugh.  mutable default type.  will fix.

ga4gh/avrotools.py
```diff
@@ -152,9 +223,194 @@ def _getInvalidFields(cls, expectedSchema, datum):
...
+    def _booleanGenerator(cls):
+        """
+        Creates a generator which returns only True and False.
+        """
+        gen = cls._valueFromSetGenerator([True, False])
+        while True:
+            yield gen.next()
+
+    @classmethod
+    def _integerGenerator(cls, min=-2 ** 31, max=2 ** 31):
```
dcolligan: these default arguments (and others) should be changed to use the avro constants

ga4gh/avrotools.py
```diff
@@ -152,9 +223,194 @@ def _getInvalidFields(cls, expectedSchema, datum):
...
+                result = self._getInvalidFields(field.type, value)
+                if result != self.sinkValue:
                     dic[key] = result
             if len(dic):
                 return dic
-        return cls.sinkValue
+        return self.sinkValue
+
+
+class RandomSchemaTool(object):
```
jeromekelleher: This sounds like a tool for generating random schemas to me. What about `RandomInstanceGenerator`? I'm a bit suspicious of anything ending in 'tool' --- it sounds a bit vague and generic... Likewise, I would prefer something like `InstanceValidator` rather than `SchemaTool` above.

ga4gh/avrotools.py
```diff
@@ -152,9 +223,194 @@ def _getInvalidFields(cls, expectedSchema, datum):
...
+        self.class_ = class_
+        self.schema = class_.schema
+
+    def getInstance(self):
+        generator = self._generateRandomInstance(self.schema)
+        jsonDict = generator.next()
+        instance = self.class_.fromJsonDict(jsonDict)
+        return instance
+
+    def _generateRandomInstance(self, schema):
```
jeromekelleher: Why the `schema` argument here? The same value is available as `self.schema`.

dcolligan: Yes, but this is a recursive function.

ga4gh/avrotools.py
dcolligan: I'm fine with changing this name, but unfortunately `SchemaTool`'s name needs to be pretty generic because its function is pretty generic.  It just takes a schema and then does a variety of things with it, depending on the methods called... validation, generating an instance, etc.  A less descriptive name is the consequence of moving towards a "one class that has all of the schema-related methods in it" approach, which is what this refactor is all about (as opposed to the "too java-y" classmethods approach)!

dcolligan:
@jeromekelleher pushed new commit

ga4gh/avrotools.py
jeromekelleher: Ahem, yes indeed it is, my bad.

jeromekelleher:
+1 --- this is good refactor and improves readability.

dcolligan:
I'm not proposing we change anything else about this PR (this should be merged asap), but @jeromekelleher do you think the generator semantics of RandomInstanceGenerator's methods give us anything?  It's a bit awkward to have those semantics different than SchemaTool's semantics.

jeromekelleher:
Merged --- thanks @dcolligan.

To answer your question, I was always slightly uneasy about the generator semantics of these. I'm perfectly happy to see these refactored if you want to do it.

Issue #406, opened 2015-05-11
pyvcf module is returning empty metadata 


MollyZhang:
In tests/data_driven/test_variants.py, it imports the module vcf to parse meta data and compares that result to pysam vcf meta data parser. However, the vcf module returns an empty metadata dictionary, and the tests in line 332-347 pass because they are not run at all.

jeromekelleher:
This seems to work fine for me @MollyZhang --- I've put in some prints (say at line 333) to print out the infoKeys, and all the metadata seems to be there. Can you be more specific about what you're seeing?

shajoezhu:
Hi @MollyZhang , when I add line ` print(infoKey)` to line 333, and ran,

    $ nosetests tests/datadriven/test_variants.py -vs

I get the printout something like
```
tests.datadriven.test_variants.VariantSetTest.testVariantSetMetadata:example_1 ... AC
AF
AN
BaseQRankSum
DP
DS
Dels
FS
HaplotypeScore
InbreedingCoeff
MLEAC
MLEAF
MQ
MQ0
MQRankSum
QD
RPA
RU
ReadPosRankSum
STR
ok
```

It seems working fine for me as well.

MollyZhang:
@jeromekelleher @shajoezhu @dcolligan 
I found out that I used the pyvcf module wrong. 
basically I did this `vcf.Reader(vcfFile)` instead of `vcf.Reader(filename=vcfFile)`
and the former returns an empty metadata. (Although I would think it's more intuitive to allow for the first way to read a file)
I didn't double check the code, sorry for the trouble...so this issue can be closed.

dcolligan:
That is unintuitive... consider filling an issue with pysam if you think it's warranted.

diekhans:
If the vcf documentation indicates vcf.Reader(vcfFile) should
work and it doesn't, please file a ticket with the pyvcf project.

MollyZhang:
@diekhans 
the pyvcf documentation is like this:
`class vcf.Reader(fsock=None, filename=None, compressed=None, prepend_chr=False, strict_whitespace=False)`
it's not wrong, just not very intuitive.

Issue #396, opened 2015-05-07
HtslibVariantSet._getMetadataFromVcf() is not parsing metadata correctly


MollyZhang:
The "_getMetadataFromVcf()" in HtslibVariantSet class doesn't seem to parse the metadata correctly. 

Here is a line of output in self._metadata after parsing:

VariantSetMetadata({"info": {}, "description": "", "number": "1", "value": "", "key": "FORMAT.DS", "type": "Float", "id": ""})

the "description" key,  for example, has an empty " " value while vcf file header has English words as description, shown here:

##FORMAT=\<ID=DS,Number=1,Type=Float,Description="Genotype dosage from MaCH/Thunder">

bioinformed:
Please let me know if this is a pysam/htsib bug.  If so, please send a
minimal test case and I will fix asap.

jeromekelleher:
When this code was being written, we weren't able to access the description in `VariantMetadata` from pysam/cbcf.pyx @bioinformed. It seemed like it hadn't been implemented (we should have filed a bug report, sorry). Has this been implemented now, or is there a different way to access the description field?

dcolligan:
@bioinformed I just sent you an email with a test case

bioinformed:
Thanks for the heads up.  `description` will be an attribute of `VariantMetadata` in the next pysam release.  I've also added a full mapping interface to `VariantHeaderRecord` so that access is more natural (i.e. doesn't require iterating through `attrs`).

dcolligan:
@bioinformed Great!

david4096:
Closed with https://github.com/ga4gh/server/pull/407

PR #407, opened 2015-05-11
updated vcf metadata parser to get description field 

MollyZhang:
solving issue #396 by adding description field to metadata
TODO: 
testVariantSetMetadata() currently uses python module pyvcf to test against pysam, however, pyvcf is not reading any metadata at all. filed an separate issue #406

### DIFF
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -474,17 +474,23 @@ class HtslibVariantSet(datamodel.PysamDatamodelMixin, AbstractVariantSet):
         ret = []
         header = varFile.header
         ret.append(buildMetadata(key="version", value=header.version))
-        for formatKey, value in header.formats.items():
-            if formatKey != "GT":
-                ret.append(buildMetadata(
-                    key="FORMAT.{}".format(value.name), type=value.type,
-                    number="{}".format(value.number)))
-                # NOTE: description is not currently implemented as a member
-                # of VariantMetadata in pysam/cbcf.pyx
-        for infoKey, value in header.info.items():
-            ret.append(buildMetadata(
-                key="INFO.{}".format(value.name), type=value.type,
-                number="{}".format(value.number)))
-            # NOTE: description is not currently implemented as a member
-            # of VariantMetadata in pysam/cbcf.pyx
+        formats = header.formats.items()
+        infos = header.info.items()
+        # TODO: currently ALT field is not implemented through pysam
+        # NOTE: contigs field is different between vcf files,
+        # so it's not included in metadata
+        # NOTE: filters in not included in metadata unless needed
+        for prefix, content in [("FORMAT", formats), ("INFO", infos)]:
+            for contentKey, value in content:
+                attrs = dict(value.header.attrs)
+                # TODO: refactor description at next pysam release
+                # since description will be implemented as a member of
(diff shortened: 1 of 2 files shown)


ga4gh/datamodel/variants.py
```diff
@@ -474,17 +474,25 @@ def buildMetadata(
...
+        # NOTE: filters in not included in metadata unless needed
+        keys = ["FORMAT", "INFO"]
+        for content in [formats, infos]:
+            for contentKey, value in content:
+                attrs = dict(value.header.attrs)
+                # TODO: refactor description at next pysam release
+                # since description will be implemented as a member of
+                # VariantMetadata
+                description = attrs.get('Description', '').strip('"')
+                key = "{0}.{1}".format(keys[0], value.name)
```
dcolligan: Having `keys[0]` and `keys.pop(0)` for indexing in a loop is unintuitive.  Try something like this:
```
for i, content in enumerate([formats, infos]):
...
           key = "{0}.{1}".format(keys[i], value.name)
...
```

dcolligan:
Good job refactoring redundant code into a common code path.  Are you going to write tests for this code?

MollyZhang:
@dcolligan I updated the code based on your comment
As to tests, Since the pyvcf isn't working, I am undecided on wether to write a metadata parser or to wait till the pyvcf bug (#406) is fixed. 
Also the ALT field in metadata is not included because of the pysam bug you were emailing @bioinformed about, I wonder when is the next pysam release? should I wait for that one or file a seperate issue for ALT filed in metadata after pysam is released?

bioinformed:
My plan is to submit my fixes for release in pysam over the weekend.  I'll
check with Andreas to see if he is okay with this timing.

ga4gh/datamodel/variants.py
```diff
@@ -474,17 +474,24 @@ def buildMetadata(
...
-            # NOTE: description is not currently implemented as a member
-            # of VariantMetadata in pysam/cbcf.pyx
+        formats = header.formats.items()
+        infos = header.info.items()
+        # TODO: currently ALT field is not implemented through pysam
+        # NOTE: contigs field is different between vcf files,
+        # so it's not included in metadata
+        # NOTE: filters in not included in metadata unless needed
+        keys = ["FORMAT", "INFO"]
+        for i, content in enumerate([formats, infos]):
```
jeromekelleher: Since we're only using `i` to access the `keys` variable, it would be better to have something like
```python
for prefix, content in [("FORMAT", formats), ("INFO", infos)]:
     # same as before except we replace `keys[i]` with `prefix`.
```

jeromekelleher:
This is a nice refactor @MollyZhang, thanks. I agree with @dcolligan though, it would be good to get some test coverage for this. I can't reproduce #406, so I think we should be able to add some quick tests on the description field fairly easily.

MollyZhang:
@dcolligan and @jeromekelleher 
I added and pushed tests for description field.

dcolligan:
+1

bioinformed:
Fixes are merged in the pysam master branch.  I'm checking on release dates.

jeromekelleher:
Great, thanks @MollyZhang. +1.

jeromekelleher:
Thanks @bioinformed, keep us posted!

Issue #404, opened 2015-05-11
avrotools: eliminate generator patterns


dcolligan:
From discussion in #397

dcolligan:
Fixed in #409

PR #409, opened 2015-05-12
Eliminate avrotools generator patterns

dcolligan:
Also do a lot of refactoring

Issue #404

### DIFF
--- a/ga4gh/avrotools.py
+++ b/ga4gh/avrotools.py
@@ -5,433 +5,528 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import itertools
 import random
 import string
-
-import avro
 
 import ga4gh.protocol as protocol
 
 
-class SchemaToolException(Exception):
+INT_MIN_VALUE = -(1 << 31)
+INT_MAX_VALUE = (1 << 31) - 1
+LONG_MIN_VALUE = -(1 << 63)
+LONG_MAX_VALUE = (1 << 63) - 1
+
+
+class AvrotoolsException(Exception):
     """
-    Something that went wrong in the SchemaTool
+    Something that went wrong in the avrotools classes
     """
 
 
-class SchemaTool(object):
+class AvroTool(object):
     """
-    Provides additional validation methods based on avro
-
-    Some code adapted from
-    https://github.com/apache/avro/blob/trunk/lang/py/src/avro/io.py#L101
+    Base class for avrotool classes
     """
-    sinkValue = 'aRandomStringBecauseWeCanNotUseNone'
-    INT_MIN_VALUE = -(1 << 31)
-    INT_MAX_VALUE = (1 << 31) - 1
-    LONG_MIN_VALUE = -(1 << 63)
-    LONG_MAX_VALUE = (1 << 63) - 1
-
     def __init__(self, class_):
         self.class_ = class_
-        self._ensureProtocolSubclass()
+        self.assertProtocolSubclass()
         self.schema = class_.schema
-        self.typeToGenerator = None
 
+    def assertProtocolSubclass(self):
+        if not issubclass(self.class_, protocol.ProtocolElement):
+            message = "class '{}' is not subclass of ProtocolElement".format(
+                self.class_.__name__)
+            raise AvrotoolsException(message)
+
+
+class Validator(AvroTool):
+    """
+    Provides methods for validating schemas
+    """
     def getInvalidFields(self, jsonDict):
         """
         Takes a jsonDict instantiation of class class_.  If the jsonDict is
-        valid for that class, return an empty dictionary.  Otherwise, return a
-        multilevel dictionary of the invalid data members and only the
+        valid for that class, return an empty dictionary.  Otherwise,
+        return a multilevel dictionary of the invalid data members and only the
         invalid data members.
 
         This is an expensive function to run and shouldn't be on the main
         data path.
         """
-        self._ensureNoExtraFields(jsonDict)
-        fields = self._getInvalidFields(self.schema, jsonDict)
-        if fields == self.sinkValue:
-            return {}
-        else:
-            return fields
+        validator = SchemaValidator(self.class_)
+        invalidFields = validator.getInvalidFields(jsonDict)
+        return invalidFields
+
+
+class Creator(AvroTool):
+    """
+    Provides methods for creating instances of protocol classes
+    according to different rules
+    """
+    def __init__(self, class_):
+        self.class_ = class_
+        self.assertProtocolSubclass()
+        self.schema = class_.schema
 
     def getTypicalInstance(self):
         """
         Return a typical instance of the class
         """
-        jsonDict = self._getTypicalInstance(self.schema)
-        instance = self.class_.fromJsonDict(jsonDict)
-        return instance
+        return self._getInstance(TypicalInstanceCreator)
+
+    def getRandomInstance(self):
+        """
+        Return a random instance of the class
+        """
+        return self._getInstance(RandomInstanceCreator)
+
+    def getInvalidInstance(self):
+        """
+        Return an invalid instance of the class
+        """
+        return self._getInstance(InvalidInstanceCreator)
 
     def getDefaultInstance(self):
         """
         Return an instance of the class with required values set
         """
-        # TODO this could be more efficient
-        defaultInstance = self.class_()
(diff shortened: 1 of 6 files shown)


ga4gh/avrotools.py
```diff
@@ -23,105 +26,65 @@ class SchemaToolException(Exception):
...
-        if fields == self.sinkValue:
-            return {}
-        else:
-            return fields
+        Return a random instance of the class
+        """
+        return self._getInstance(RandomInstanceCreator)
 
-    def getTypicalInstance(self):
+    def getBogusInstance(self):
```
jeromekelleher: Nitpick: I'd prefer 'Invalid' here to 'Bogus'. I think it would be clearer what the intention of the method is.

ga4gh/avrotools.py
```diff
@@ -129,309 +92,419 @@ def _ensureProtocolSubclass(self):
...
 
-    def __init__(self, class_):
-        self.class_ = class_
-        self.schema = class_.schema
+    def handleEnum(self, schema):
+        return random.choice(schema.symbols)
+
+    def handleArray(self, schema, length=10):
+        toReturn = []
+        for _ in xrange(length):
```
jeromekelleher: `xrange` goes away in Python 3, so we should just stick to `range`. The performance issues aren't worth worrying about in this context.

ga4gh/avrotools.py
```diff
@@ -129,309 +92,419 @@ def _ensureProtocolSubclass(self):
...
+        return self.handleString(schema.size)
 
-    def __init__(self, class_):
-        self.class_ = class_
-        self.schema = class_.schema
+    def handleEnum(self, schema):
+        return random.choice(schema.symbols)
+
+    def handleArray(self, schema, length=10):
+        toReturn = []
```
jeromekelleher: A list comprehension would probably be nice here:
```python
return [self.handleSchema(schema.items) for _ in range(length)]
```

ga4gh/avrotools.py
```diff
@@ -129,309 +92,419 @@ def _ensureProtocolSubclass(self):
...
+        for _ in xrange(length):
+            item = self.handleSchema(schema.items)
+            toReturn.append(item)
+        return toReturn
+
+    def handleMap(self, schema, size=10):
+        toReturn = {}
```
jeromekelleher: Dict comprension:
```python
return {self.handleString():self.handleSchema() for _ in range(size)}
```

ga4gh/avrotools.py
```diff
@@ -129,309 +92,419 @@ def _ensureProtocolSubclass(self):
...
+            value = self.handleSchema(schema.values)
+            toReturn[key] = value
+        return toReturn
+
+    def handleUnion(self, schema):
+        chosenSchema = random.choice(schema.schemas)
+        return self.handleSchema(chosenSchema)
 
+    def handleRecord(self, schema):
+        record = {}
```
jeromekelleher: Dict comprehension would be nice here too.

ga4gh/avrotools.py
```diff
@@ -129,309 +92,419 @@ def _ensureProtocolSubclass(self):
...
+        raise SchemaToolException()
+
+    def handleBoolean(self):
+        return True
+
+    def handleString(self):
+        return 'aString'
+
+    def handleBytes(self):
+        return 'someBytes'
```
jeromekelleher: Should have a 'b' prefix here for Python 3 compatibility:
```python
return b'someBytes'
```

I know we're not supporting Python 3 right now, but I think it's worth doing these little things to make the job easier when we do change.

ga4gh/avrotools.py
```diff
@@ -129,309 +92,419 @@ def _ensureProtocolSubclass(self):
...
+    def handleFixed(self, schema):
+        return 'x' * schema.size
+
+    def handleEnum(self, schema):
+        return schema.symbols[0]
+
+    def handleArray(self, schema):
+        val1 = self.handleSchema(schema.items)
+        val2 = self.handleSchema(schema.items)
+        arr = [val1, val2]
```
jeromekelleher: How about
```python
return [self.handleSchema(schema.items) for _ in range(2)]
```

ga4gh/avrotools.py
```diff
@@ -129,309 +92,419 @@ def _ensureProtocolSubclass(self):
...
+        return arr
+
+    def handleMap(self, schema):
+        value = self.handleSchema(schema.values)
```
jeromekelleher: I guess this is a subjective question of style, but I don't think this is any more readable than
```python
return {'key': self.handleSchema(schema.values)}
```

ga4gh/avrotools.py
```diff
@@ -129,309 +92,419 @@ def _ensureProtocolSubclass(self):
...
+        mapping = {'key': value}
+        return mapping
+
+    def handleUnion(self, schema):
+        for memberSchema in schema.schemas:
+            if memberSchema.type == 'null':
+                continue
+            return self.handleSchema(memberSchema)
+
```
jeromekelleher: This is definitely dead puppy time... Why not
```python
for memberSchema in schema.schemas:
    if memberSchema.type != 'null':
         return self.handleSchema(memberSchema)
```
This is much clearer to me, and one less line of code. I think the semantics should be documented here: we return an instance of the first non-null schema within the union.

ga4gh/avrotools.py
```diff
@@ -129,309 +92,419 @@ def _ensureProtocolSubclass(self):
...
+    """
+    schemaRequired = set(
+        ['record', 'enum', 'map', 'array', 'union',
+         'request', 'error', 'error_union', 'fixed'])
+
+    def __init__(self, class_):
+        self.class_ = class_
+        self.schema = class_.schema
+
+    def getInstance(self):
```
jeromekelleher: Docstring this please.

ga4gh/avrotools.py
```diff
@@ -129,309 +92,419 @@ def _ensureProtocolSubclass(self):
...
+        jsonDict = self.handleSchema(self.schema)
+        instance = self.class_.fromJsonDict(jsonDict)
+        return instance
+
+    def getFieldValue(self, fieldName):
```
jeromekelleher: Docstring.

ga4gh/avrotools.py
```diff
@@ -129,309 +92,419 @@ def _ensureProtocolSubclass(self):
...
+        fieldType = None
+        for field in self.schema.fields:
+            if field.name == fieldName:
+                fieldType = field.type
+                break
+        value = self.handleSchema(fieldType)
+        return value
+
+    def handleSchema(self, schema, extra=None):
```
jeromekelleher: Docstring. Since we don't have docstrings for the individual `handleX` methods we should provide a good overview of how this setup works in this function.

jeromekelleher:
This looks very nice, thanks @dcolligan. I like the general approach, and the reuseable Avro type switch mechanism is ingenious. There are two general points that come through for me:

1. We need more docstrings. In general, every public method of a class should have a docstring. This would be very tedious in this case to have docstrings for all the handleX methods, but we can still explain how this all works a lot better.
2. The code here is a bit verbose in places for my taste --- I've highlighted some places where we can cut out a lot of lines of code using list and dictionary comprehensions, increasing readability (IMO).

ga4gh/avrotools.py
dcolligan: I agree, but we also have something in `SchemaTool` called `getInvalidFields` which does something completely different, so anything else with `invalid` in the name is confusing.  (Unless we want to split `SchemaTool` into a verification class and a creation class, but you seemed against that previously.)

dcolligan:

ga4gh/avrotools.py
jeromekelleher: No, I think I'd be all for splitting SchemaTool into two different (more specific) classes. As I said before, I'm suspicious of any class called Tool, as it seems a bit generic. Apologies if I gave the opposite impression before.

jeromekelleher:
Looks great, thanks @dcolligan. I'm +1 modulo the Bogus vs Invalid discussion above.

dcolligan:
@jeromekelleher alright, pushed an update that split `SchemaTool` into two classes, `Validator` and `Creator` -- there's nothing named 'bogus' now

jeromekelleher:
Thanks @dcolligan, +1.

Issue #391, opened 2015-05-05
Dense variants not handled correctly by released server


jeromekelleher:
Users of the released version of the server have encountered problems in which dense variants (i.e., those in which we have more than one variant overlapping a given position) are present. This is because we don't handle interval search correctly, which has been resolved since release in  the develop branch. 

Suggested resolution:

1. Makes some tests for dense variants (we can update the variant simulator to generate many variants at a given position) to verify they are handled correctly in the 'develop' branch.
1. Make a release branch called 0.1-release from the current stable branch.
2. Backport the changes required to implement interval search on variants (plus new tests) by cherry-picking the relevant commits.
3. Release a new version. Should we call this 0.1.0b2 or 0.1.0?

diekhans:
>  4. Release a new version. Should we call this 0.1.0b2 or 0.1.0?

b2 == beta 2 to me

suggest the original is tagged on the branch as 0.1.0 and this one
tagged a 0.1.1

This is the major.minor.bugfix  release numbering approach.  It 
is clean and easy to understand (and parse).

this is an entertaining distraction
  http://en.wikipedia.org/wiki/Software_versioning

The fact that you have to do this kind of professional releasing
management at all is exciting.

jeromekelleher:
OK, let's go to 0.1.1 for this one so. We can make a copy of the tag for 0.1.0 so that we have a record of it (although I don't see much point in pushing a copy to PyPI).

The original motivation of using a1, a2, b1, b2 etc was to use the Python standard for managing alpha and beta version numbers. In retrospect, we should have tagged 0.1.0 instead of 0.1.0b1 as the official release. Oh well, we'll know better next time!

diekhans:
I like that convention for alpha/beta ; but I think 0.1.0 is a real release..

Hopefully this will not be like some OSS projects where they work for
years and avoid ever releasing a 1.0 ;-)

dcolligan:
I think we should also include this change in the new release: https://github.com/ga4gh/server/pull/386

jeromekelleher:
@dcolligan: sure, I'm happy to roll in #386 too.

dcolligan:
@jeromekelleher are you taking this on?

jeromekelleher:
@dcolligan I think the person with the requisite git-foo is you. You should have the permissions on PyPI etc, so just go for it I'd say. One thing though: do we have test coverage on dense variants?

Slightly revised plan:
1. make a new branch 0.1-release from stable;
[...]

PR #410, opened 2015-05-13
Bugfix release

dcolligan:
- Bump to version ~~0.1.0~~ 0.1.1
- Refactor and test backend generators
- Take out unused paths
- added sphinx exception to gitignore

#391

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -95,7 +95,7 @@ coverage.xml
 *.log
 
 # Sphinx documentation
-docs/source/ga4ghDocsBuild/
+docs/source/_build/
 
 # PyBuilder
 target/
--- a/ga4gh/__init__.py
+++ b/ga4gh/__init__.py
@@ -2,4 +2,4 @@
 Reference implementation of the GA4GH APIs.
 """
 
-__version__ = '0.1.0'
+__version__ = '0.1.1'
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -17,6 +17,167 @@ import ga4gh.exceptions as exceptions
 import ga4gh.datamodel.variants as variants
 
 
+def _parsePageToken(pageToken, numValues):
+    """
+    Parses the specified pageToken and returns a list of the specified
+    number of values. Page tokens are assumed to consist of a fixed
+    number of integers seperated by colons. If the page token does
+    not conform to this specification, raise a InvalidPageToken
+    exception.
+    """
+    tokens = pageToken.split(":")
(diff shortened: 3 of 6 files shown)


jeromekelleher:
Thanks for this @dcolligan. Some minor issues:

1. I think we should make a release of 0.1.0 first where all we do is bump the version number. This is our real 'initial' release, which we have for archival purposes. The release that you're making here will then be 0.1.1, the first bugfix release.
2. Is it possible to pull in the original commits into the PR where the work was done, and not just all of them squashed together? I think it'll be hard to keep track of what's gone into the release branch if we don't do this.

dcolligan:
@jeromekelleher Ok, I changed the version number here to 0.1.1.  I suppose it is possible to separate out the commits here, but I don't see much point in that.  They have different commit shas than the cherry-picked commits both because the commit history they are working off is different and the cherry-picks weren't clean and I needed to resolve some conflicts manually.  I listed the commits in `server/develop` that I cherry picked in the commit message, which hopefully will satisfy you.  I also submitted a new PR for the 0.1.0 version bump here: https://github.com/ga4gh/server/pull/413

jeromekelleher:
Looks good, thanks @dcolligan. Good git-fu on pulling out the commits and documenting their source.

One last point: shouldn't we be making these pull requests against the 0.1-release branch rather than stable? The idea is we can make more and smaller commits onto the release branch, and then tag from there. We only merge back to `stable` after the release is made and we're fully happy everything works.

dcolligan:
If there were multiple contributors that needed to commit code over time to a release branch, then yes, it would make sense to commit against 0.1-release.  But as it stands, we're only putting one commit into stable and I'm the only author, so I don't think it makes that much difference.

jeromekelleher:
OK, fair enough. I would prefer to see this done on a release branch in future, but you're right that it makes no difference now. I'm think we should merge this and release it, as soon as the 0.1.0 has been pushed to PyPI.

+1.

dcolligan:
Ok, 0.1.0 has been uploaded to pypi... this can be merged now

jeromekelleher:
Merged, thanks @dcolligan. Great work on the git-foo. Fancy rolling the 0.1.1 release from this (pypi + github)?

dcolligan:
Done and done
https://github.com/ga4gh/server/releases/tag/v0.1.1
https://pypi.python.org/pypi/ga4gh/0.1.1

jeromekelleher:
Superb!

PR #411, opened 2015-05-13
Update dockerhub repo name to point to org.

afirth:
Add example of linked containers. Add demo script
[Rendered docs](http://umms-ga4gh-server.readthedocs.org/en/2015-05-13_update_docker_docs/installation.html#deployment-on-docker)

### DIFF
--- a/scripts/demo_example.py
+++ b/scripts/demo_example.py
@@ -1,0 +1,29 @@
+#!/usr/bin/env python
+"""
+Example of client query
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+import ga4gh.client as client
+import ga4gh.protocol as protocol
+
+
+def runDemo():
+    httpClient = client.HttpClient("http://server:80/current")
+    request = protocol.SearchVariantsRequest()
+    request.variantSetIds = ["1kg-phase1"]
+    request.referenceName = "2"
+    request.start = 33100
+    request.end = 34000
+    for variant in httpClient.searchVariants(request):
+        print(
+            variant.referenceName, variant.start, variant.end,
+            variant.referenceBases, variant.alternateBases,
+            sep="\t")
+
+
+if __name__ == '__main__':
+    runDemo()
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -168,13 +168,13 @@ If you already have a dataset on your machine, you can download and deploy the a
 
 .. code-block:: bash
 
-  $ docker run -e GA4GH_DATA_SOURCE=/data -v /my/ga4gh_data/:/data:ro -itd -p 8000:80 --name ga4gh_server afirth/ga4gh_apache_server:prod
+  $ docker run -e GA4GH_DATA_SOURCE=/data -v /my/ga4gh_data/:/data:ro -d -p 8000:80 --name ga4gh_server afirth/ga4gh-server:latest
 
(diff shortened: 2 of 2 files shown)


scripts/demo_example.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+from __future__ import print_function
+
+import ga4gh.client as client
+import ga4gh.protocol as protocol
+
+httpClient = client.HttpClient("http://server:80/v0.6.e6d6074")
```
dcolligan: If this is working off the `develop` branch, you can use the `current` alias instead of having the protocol version in the URL

scripts/demo_example.py
```diff
@@ -0,0 +1,17 @@
...
+request = protocol.SearchVariantsRequest()
+request.variantSetIds = ["1kg-phase1"]
+request.referenceName = "2"
+request.start = 33100
+request.end = 34000
+for variant in httpClient.searchVariants(request):
+    print(
+        variant.referenceName, variant.start, variant.end,
+        variant.referenceBases, variant.alternateBases,
+        sep="\t")
```
dcolligan: Usually python files have the main code encapsulated in a `if __name__ == '__main__':` block so that importing the module doesn't run the code.

scripts/demo_example.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+from __future__ import print_function
```
dcolligan: Just as a matter of consistency / style, you should probably include the same header lines that we do on all of our python files:
```
```

scripts/demo_example.py
afirth: This example is straight from the demo docs, all I added is the shebang line [and the /current alias]. Is it worth fleshing out? If so, would you mind submitting a PR against this branch?
Thanks for the /current alias, that was upsetting me too and then I found #387

scripts/demo_example.py
afirth: as above

scripts/demo_example.py
dcolligan: True, but that's not python code thats in our repo, as this file is.  @jeromekelleher what do you think?

scripts/demo_example.py
jeromekelleher: I'm with @dcolligan here --- the code checked into our repo here should conform to our style guidelines and be consistent with the rest of our code base. I agree on both the imports and the `__main__` bit.

jeromekelleher:
This looks good @afirth, thanks very much. I think it's good to merge once you update the example script. It would be good to add the standard imports (these are nasty Python 2/3 gotchas, and others may copy this script before working with it further). Also, I'd like to see a `main()` function just for consistency, and a quick module docstring explaining what the file is.

afirth:
everyone happy with comment resolution? if so, i will squash commits

dcolligan:
Looks like you addressed my concerns, but usually we only give out +1's to the commits that are actually going to be committed (in this case, the post-squash ones) since we read +1 as "this is okay to commit to mainline."

jeromekelleher:
Yep, I'm +1 once the commits are squashed.

afirth:
OK, squashed. Recommend you use the mergable tag for that so that you can see the comment fixes separately (compare links don't work across rebase), but I suppose it doesn't matter much. Anyway, this should be good to go.

PR #412, opened 2015-05-13
Use current version alias for demo docs

afirth:


dcolligan:
-1

Unfortunately, the `current` alias is not in the pypi version (the code that is in the latest pypi release is in the `stable` branch).

afirth:
Fair enough. My concern is that these are the only examples in the documentation and they are broken for 0.6.*. Anyway, closing for now

Issue #403, opened 2015-05-11
Simple OAUTH authentication provider


jbu:
For testing any Oauth work, and for simple out-of-the-box deployment, we want a simple oauth provider to ship with the server. Really just a basic web server that can respond correctly to authentication against a list of user:passwd pairs.

Issue #31, opened 2014-10-30
OAuth support required


jeromekelleher:
The reference implementation needs support for [OAuth](http://oauth.net/) to allow us to perform authentication. There are many problems to be resolved in this, and this issue should provide a central location for the discussion around this topic. 

Some questions:

1. What library should we use?
2. Do we use OAuth 1.0 or 2.0?
3. Lots of other questions I would ask if I knew more about OAuth.

Hopefully we can flesh out the actual requirements for the reference server here, and make a concrete plan for implementing these.

jbu:
Hi everyone,

After the call yesterday, the consensus seems to be: 
* Use OIDC
* Go with https://github.com/rohe/pyoidc as the bulk of the implementation.
* Stick to static client registration.

dglazer:
Glad to see momentum here. Was there also discussion on yesterday's call about the plan for authentication and access control?

For authentication -- per http://openid.net/2015/04/17/openid-connect-certification-program/, it looks like our timing is good here -- apparently several big identity providers (including Microsoft and Google) have recently certified their compliance with OIDC. Assuming pyoidc works as advertised, that means people should be able to use existing credentials to authenticate with the reference server -- great.

For access control -- server operators still need a way to say "this user can see this object". See the earlier comments (Oct 31 2014) from @delagoya and me on possible simple answers.

jeromekelleher:
This sounds like a good plan @jbu, thanks.

@dglazer, for authentication, we're hoping to support a few OIDC providers, time allowing. Initially we'll be keeping it simple a using a local provider. For authorisation, we'll be keeping it very simple. Initially, yes/no access to the repository, and after than, yes/no access to a dataset. That will be enough to get us started.

dglazer:
Thanks @jeromekelleher - starting simple sounds great.

Making sure I understand:
- for authN, having a "local provider" implies there's some way for users to get their id/credentials added to that local provider. Is the idea to just have a server-side text file of names and non-secure passwords, or to build an account management UI, or am I misunderstanding?
- for access control, is the idea to have a manually-maintained server-side text file with a list of authorized ids (initially a repository-level list; later adding dataset-level granularity)? If so, SGTM.

jeromekelleher:
@dglazer, well, the local provider we have is just for testing and development. For deployments, we'll use a real authentication provider. Who the auth provider is and how we set up the login process is to be decided. There is no plan to have a password management UI.

[...]

PR #414, opened 2015-05-18
403 simple oauth provider

jbu:
#403. Ready for a bit of review. This is a simple docker file to run an example OIDC provider from pyoidc. You can start it (alongside the gh4gs server) using 'docker-compose up'. 

Now looking more at adding RP capability to the main server.


jeromekelleher:
This looks good @jbu, thanks. I wonder if we really need Docker for this though. If I'm interpreting things correctly, we're running a provider using the example code from PyOIDC, which we could just bake into a script that we run locally (say, `scripts/oauth_provider.py`). To me this would be more straightforward and transparent than pushing this logic into a Docker image, and it would make it easier to work with the files it depends on.

However, it's entirely possible I'm not understanding this very well, so I'm not closed to the idea to using Docker for this.

dcolligan:
I agree with @jeromekelleher.  It makes sense to add Docker-related stuff to our codebase for packaging functionality to deploy so other people can use our code without having to perform excessive configuration.  But it seems like it's breaking the pure-python simplicity of our codebase to introduce docker into other areas.  It makes sense to hook any auth things into the existing code more closely: we're going to want to run tests with this, etc.

Looking back on the old OAuth issue here https://github.com/ga4gh/server/issues/31 it seems like basically the only point that was decided upon was to use OAuth 2.  I'd like to have a better idea of what our use cases are before I'm ready to incorporate any oauth code into the repo.

jbu:
I'll have a look at doing this without docker. @dcolligan, the direction of the discussion yesterday is that using OIDC would allow us to make the first steps without precluding more difficult use cases in the future. The uses cases I'm aware of are a straight-forward RP-OP authorisation flow, and a federated system where a central 'client' server has static authorisation with a set of ga4gh servers (which might just as well be handled by tls client certs).

jbu:
I've removed docker, but have still used virtualenv to help keep the requirements for this server isloated. The ./run.sh script sets everything up and runs the server.

oidc-provider/run.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+virtualenv .
```
jeromekelleher: We need to add `./bin/activate` here to start the venv.

jeromekelleher:
Excellent, thanks @jbu. This looks a lot easier to work with to me, and it worked fine on my system once I added the bit above. I it would be good to add a quick README explaining what this directory is, where the code came from, and how we can run it.

After that, I'm +1. Once we're all happy with the PR and it's ready to be merged, we'll ask you to rebase the PR against the current state of the 'develop' branch and squash the commits.

@bartgrantham and @knoxcarey, would you like to weigh in here?

@dcolligan, @jbu and others are working on adding OAuth support to the server. This is a critical feature for people with non-public data (which is almost everyone), so that they can selectively share their data. We're coordinating with members of the security working group (@bartgrantham and @knoxcarey) to make sure that what we're doing here conforms to their recommendations.

oidc-provider/run.sh
jbu: Yes! I must not have pushed that. Will do now.

jeromekelleher:
Thanks for the update @jbu, +1.

dcolligan:
commits need to be squashed

oidc-provider/simple_op/src/provider/authn/__init__.py
```diff
@@ -0,0 +1,30 @@
+import importlib
```
dcolligan: It would be good to add our usual imports to the top of any significant (that is, non-init) python file:

```
```

jbu: All this code is a direct copy from the examples in pyoidc. I think that ideally we should leave it as is to prevent anything like a fork. In the best possible world I'd not check the code in here, but instead have a script that fetches pyoidc and extracts this example (the docker version did this). @jeromekelleher would this make more sense to you?

jeromekelleher: I agree with @jbu --- since this is a direct copy of pyoidc I don't think we should edit it in any way. I'm not hugely keen on downloading a new copy each time, as this will inevitably break some time in the future. At least if we have the code checked in here, we're all working from the same auth provider code for development. 

I think we should squash the commits on this and merge ASAP.

jbu:
Commits have been squashed. I'll leave it to others ( @jeromekelleher ?) to merge.

jeromekelleher:
Thanks @jbu. The convention here is that no-one merges their own code, and in practice either me, @dcolligan or @macieksmuga do the merging. Let's see if @dcolligan is happy with this being merged, and go from  there.

dcolligan:
+1

jeromekelleher:
Merged: thanks @jbu!

Issue #375, opened 2015-04-28
Need to generalise data input mechanisms.


jeromekelleher:
We currently use DATA_SOURCE which points to a directory containing subdirectories `reads`, `variants` and `references`. One possibility would be to have a `REFERENCES` config variable that is a list of directories (or SQLite DBs for graph mode?), each containing a single `ReferenceSet`. We could have another variable `DATASETS` which contains a list of directories, each of which has the required `variants` and `reads` subdirectories. 

We should also consider the issues of auth down the line, and how we associate permissions with datasets (or at lease, choose a mechanism which will allow for this).

dcolligan:
Why can't we have DATA_SOURCE point to a directory of structure:
```
/root
  /dataset1
    /reads
    /variants
    /references
  /dataset2
 ...
```
The three things that a dataset is supposed to enforce (access control, billing, provenance) seem to be applicable to all three data types, correct?

jeromekelleher:
This is what we started out with initially, but various people thought that it was important that each dataset is associated with it's own independent directory. I don't really have a strong opinion on this: I think we should play this by ear, and see what works out as the simplest approach when we start adding ACLs to the datasets.

One thing though is that `references` are *not* contained in a dataset. These are definitely a separate thing (we expect lots of datasets, and one or two referencesets, at most).

dcolligan:
Ok, why don't we keep the single DATA_SOURCE variable and have it point to a directory tree of this structure:
```
/root
  /references
  /dataset1
    /reads
    /variants
  /dataset2
    /reads
    /variants
  ...
```
That is, any directory under the root not named "references" is considered a dataset.  This seems simpler to me than adding variables to our server config, and we're imposing a directory structure on the end user either way, so not go with the simpler one?

jeromekelleher:
I'm happy with this, so let's go with your approach unless someone shouts out with a good reason not to.

dcolligan:
Fixed in #420

PR #420, opened 2015-05-20
375 datasets

dcolligan:
Mandate single dataset for server data

Issue #375


ga4gh/backend.py
```diff
@@ -184,30 +182,31 @@ class AbstractBackend(object):
...
         """
-        return list(self._variantSetIdMap.values())
+        return list(self.getDataset().getVariantSetIdMap().values())
 
     def getReadGroupSets(self):
         """
         Returns the list of ReadGroupSets in this backend.
         """
-        return list(self._readGroupSetIdMap.values())
+        return list(self.getDataset().getReadGroupSetIdMap().values())
```
jeromekelleher: These methods are only used by the frontend code for displaying the dataset summaries. We should just move the `getReadGroupSets` and `getVariantSets` methods into DataSet, and have the frontend code work directly with the dataset object.

ga4gh/backend.py
```diff
@@ -184,30 +182,31 @@ class AbstractBackend(object):
...
-        self._readGroupSetIds = []
-        self._readGroupIds = []
-        self._readGroupIdMap = {}
         self._requestValidation = False
         self._responseValidation = False
         self._defaultPageSize = 100
         self._maxResponseLength = 2**20  # 1 MiB
+        self._dataset = datasets.AbstractDataset()
+
+    def getDataset(self):
```
jeromekelleher: Very good; I assume the idea here is that this will take a dataset Id in the future (when we have more than one dataset), and we'll use this to get the right dataset?

ga4gh/datamodel/datasets.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    def __init__(self):
+        self._variantSetIds = []
+        self._variantSetIdMap = {}
+        self._readGroupSetIds = []
+        self._readGroupSetIdMap = {}
+        self._readGroupIds = []
+        self._readGroupIdMap = {}
+
+    def getDirectory(self):
```
jeromekelleher: It would be nice to docstring these methods in the abstract class. It's not blindingly obvious what the IDs and maps are for, so we should explain that through these comments.

jeromekelleher:
This a great PR, thanks @dcolligan! A few minor nits above, but very much in favour of this in general.

ga4gh/backend.py
dcolligan: Yes, there will have to be modifications made when the number of datasets > 1.

dcolligan:
@jeromekelleher just pushed an update addressing your concerns

jeromekelleher:
Superb; +1.

jeromekelleher:
Can you rebase this please @dcolligan?

dcolligan:
@jeromekelleher done

jeromekelleher:
Excellent; thanks @dcolligan. Merging.

Issue #417, opened 2015-05-19
Data validation tool


jbu:
@jeromekelleher has some ideas about this.

dcolligan:
Is this the same as the compliance suite?  Or is this something else?

jeromekelleher:
The idea here is to provide a way to verify that the input provided to the server makes sense. The intention is similar to `apachectl configtest`: provide the server admin a way to check that their configuration makes sense _before_ they try to deploy it. This is helpful because (a) it can be hard to pull error messages from web server logs (especially with the way htslib does things); and (b) we want to find out if things are going to fail _before_ a client starts making requests. 

We can do simple things like make sure that the config file is well formed and all the arguments make sense. However, I think it would be valuable to look at things a bit more deeply too. It would be good to go through all of the variant sets, readgroup sets and reference sets, and make sure that they are well formed. Do we have the right types of files in the directories? Can we open them successfully? A lot of these checks are performed automatically when we load up the server, so we should be able to reuse this code quite easily. Once we have checked over all the input data, we can then present a summary of this information on stdout, where the admin can take a look. I think this would be very helpful for people arranging their data into the required hierarchies.

I think it would also be useful to perform more indepth analysis of the data files. For example, are the reads and the variants in a dataset against the correct reference? Do we have this reference in the repository? Do (a sample of) the REF values in the VCFs correspond to the bases at this position in the reference sequence? One of the principle benefits of GA4GH for me could be that users won't need to worry about this sort of thing any more: a GA4GH repository should be self consistent and have integrity. We just need to provide the tools to make sure that this is true on the server side.

dcolligan:
Cool, sounds like a good idea, and also something we can iterate on (it's probably not essential to get all of the advanced checks in from the get-go).

diekhans:
GA4GH will badly need a data validation tool.  This should eventually be a stand alone set of  tools that can do some level validation against the data. This is different  than conformance checking against an API implementation.

pgrosu:
I'm very carefully working on this, which means it will take some time to implement it properly together my other activities that are also taking up my time.  Basically I'm trying to allow for flexibility in the types of tests we want to run.  So please keep implementing your version to satisfy any immediate needs.

jbu:
I suppose we need to define what the immediate needs are, and probably some sort of framework so @pgrosu and I can pull in the same direction. Paul, what have you done so far? @jeromekelleher, do you have a simple immediate 'story' we could start with?

[...]

PR #423, opened 2015-05-26
Framework for server config validator

jbu:
Framework for the server validator work. see #417.

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -10,11 +10,16 @@ from __future__ import unicode_literals
 import time
 import argparse
 import sys
+import logging
+import unittest
+import unittest.loader
+import unittest.suite
 
 import ga4gh.client as client
 import ga4gh.protocol as protocol
 import ga4gh.converters as converters
 import ga4gh.frontend as frontend
+import ga4gh.configtest as configtest
 
 
 # the maximum value of a long type in avro = 2**63 - 1
@@ -816,3 +821,62 @@ def client_main(parser=None):
     else:
         runner = args.runner(args)
         runner.run()
+
+##############################################################################
+# Configuration testing
+##############################################################################
+
+
+class SimplerResult(unittest.TestResult):
+    """
+    The TestResult class gives formatted tracebacks as error messages, which
+    is not what we want. Instead we just want the error message from the
+    err praram. Hence this subclass.
+    """
+    def addError(self, test, err):
+        self.errors.append((test,
+                            "{0}: {1}".format(err[0].__name__, err[1])))
+
+    def addFailure(self, test, err):
+        self.failures.append((test,
+                              "{0}: {1}".format(err[0].__name__, err[1])))
+
+
+def configtest_main(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser(
+            description="GA4GH server configuration validator")
+    parser.add_argument(
+        "--config", "-c", default='DevelopmentConfig', type=str,
+        help="The configuration to use")
+    parser.add_argument(
+        "--config-file", "-f", type=str, default=None,
+        help="The configuration file to use")
+    args = parser.parse_args()
+    configStr = 'ga4gh.serverconfig:{0}'.format(args.config)
+
+    configtest.TestConfig.configStr = configStr
+    configtest.TestConfig.configFile = args.config_file
+    configtest.TestConfig.configEnv = "GA4GH_CONFIGURATION"
+
+    loader = unittest.TestLoader()
+    tests = loader.loadTestsFromModule(configtest)
+    results = SimplerResult()
+    tests.run(results)
+
+    logging.basicConfig(level=logging.INFO)
+    log = logging.getLogger(__name__)
+    log.info('{0} Tests run. {1} errors, {2} failures, {3} skipped'.
+             format(results.testsRun,
+                    len(results.errors),
+                    len(results.failures),
+                    len(results.skipped)))
+    for result in results.errors:
+        if result is not None:
+            log.critical('Error: {0}: {1}'.format(result[0].id(), result[1]))
+    for result in results.failures:
+        if result is not None:
+            log.critical('Failure: {0}: {1}'.format(result[0].id(), result[1]))
+    for result in results.skipped:
+        if result is not None:
+            log.info('Skipped: {0}: {1}'.format(result[0].id(), result[1]))
--- a/ga4gh/configtest.py
+++ b/ga4gh/configtest.py
@@ -1,0 +1,55 @@
+"""
+Tests for the server configuration. Run via configtest_dev.py
+
+Tests are standard python unittest tests.
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import unittest
+import flask
+import os
+
+
+class TestConfig(unittest.TestCase):
+    """
+    This class implements configuration checks for the GA4GH server. It is
+    structured as a set of unit tests, and relies on the unittest module.
+    However, as some of the tests require state injection (through command
+    line arguments) we need to break the usual unittest pattern and expose
+    some configuration as class attributes that can be set before tests are
+    run.
+    """
+
+    configStr = None
+    configEnv = None
+    configFile = None
+
+    def setUp(self):
+        """
+        Create a flask app to be used in the tests. Use the supplied class
+        attributes for configuration
+        """
+        self.app = flask.Flask(__name__)
+        try:
+            self.app.config.from_object(self.configStr)
+        except Exception:
+            self.fail("Cannot create app from {0}".format(self.configStr))
+        if os.environ.get(self.configEnv) is not None:
+            self.app.config.from_envvar(self.configEnv)
+        if self.configFile is not None:
+            self.assertTrue(os.path.exists(self.configFile),
+                            "config file does not exist")
+            self.app.config.from_pyfile(self.configFile)
+
+    def test_config_parameters(self):
+        """
+        A simple test that configuration parameters are of the correct types
+        and ranges
+        """
+        self.assertIsInstance(self.app.config["REQUEST_VALIDATION"],
+                              bool,
+                              'REQUEST_VALIDATION is not a boolean')
+        self.assertIsNotNone(self.app.config["DATA_SOURCE"],
+                             'DATA_SOURCE must be set')
--- a/configtest_dev.py
+++ b/configtest_dev.py
@@ -1,0 +1,7 @@
+"""
+Simple shim for running the configuration testing program during development.
+"""
+import ga4gh.cli
+
+if __name__ == "__main__":
+    ga4gh.cli.configtest_main()
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,7 @@ setup(
     entry_points={
         'console_scripts': [
             'ga4gh_client=ga4gh.cli:client_main',
+            'ga4gh_configtest=ga4gh.cli:configtest_main',
             'ga4gh_server=ga4gh.cli:server_main',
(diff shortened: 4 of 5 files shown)


ga4gh/cli.py
```diff
@@ -816,3 +820,72 @@ def client_main(parser=None):
...
+    else:
+        return None
+
+
+def configtest_main(parser=None):
+    test_config = TestConfigExists(parser)
+    test_config.setUp()
+    tests = ['test_config_exists']
+    results = map(lambda t: run_test(test_config, t), tests)
+    print('{0}/{1} tests passed'.format(results.count(None), len(results)))
```
dcolligan: we shouldn't have any `print` calls in the ga4gh module code ... these should be calls to some kind of logging tool (even if it is something as simple as calling a method that just calls print)

ga4gh/cli.py
```diff
@@ -816,3 +820,72 @@ def client_main(parser=None):
...
+
+    def test_config_exists(self):
+        self.assertIsInstance(self.app.config["REQUEST_VALIDATION"],
+                              types.BooleanType,
+                              'REQUEST_VALIDATION is not a boolean')
+        self.assertIsNotNone(self.app.config["DATA_SOURCE"],
+                             'DATA_SOURCE must be set')
+
+
+def run_test(test_obj, test):
```
dcolligan: we use camelCase variable naming conventions in our code rather than this_style

ga4gh/cli.py
```diff
@@ -816,3 +820,72 @@ def client_main(parser=None):
...
+    line arguments) we need to break the usual unittest pattern and instead
+    implement our own simple test runner.
+    """
+
+    def __init__(self, parser=None):
+        self.parser = parser
+
+    def setUp(self):
+        self.app = flask.Flask(__name__)
+        parser = self.parser if self.parser is not None \
```
dcolligan: @jeromekelleher claims that line-continuation backslashes kill puppies, so I would avoid this

I don't know if we have an opinion on ternary operators.  Personally, I avoid them, but I'd be willing to hear what other people think...

ga4gh/cli.py
```diff
@@ -816,3 +820,72 @@ def client_main(parser=None):
...
+    However, as some of the tests require state injection (through command
+    line arguments) we need to break the usual unittest pattern and instead
+    implement our own simple test runner.
+    """
+
+    def __init__(self, parser=None):
+        self.parser = parser
+
+    def setUp(self):
+        self.app = flask.Flask(__name__)
```
dcolligan: I don't know if the configuration checker should ever start up a Flask instance; that seems outside the scope of what it's supposed to do...

ga4gh/cli.py
```diff
@@ -816,3 +820,72 @@ def client_main(parser=None):
...
+            else argparse.ArgumentParser(
+                description="GA4GH server configuration validator")
+        parser.add_argument(
+            "--config", "-c", default='DevelopmentConfig', type=str,
+            help="The configuration to use")
+        parser.add_argument(
+            "--config-file", "-f", type=str, default=None,
+            help="The configuration file to use")
+        args = parser.parse_args()
+        config_str = 'ga4gh.serverconfig:{0}'.format(args.config)
```
dcolligan: should be a camelCase var ... same throughout the rest of your changes

ga4gh/cli.py
jbu: It kills puppies, but it passes the pep8 linter.

ga4gh/cli.py
jbu: The idea is to reuse the config reader in the flask app. I could do that more manually, but might end up either using flask internals, or re-writing code. Another alternative is to pull the config reader out of the app into something more standalone and test it there.

ga4gh/cli.py
jbu: righto.

ga4gh/cli.py
jbu: adding the logging module. Will print to stdout to start with.

ga4gh/cli.py
jbu: So no 'N806 variable in function should be lowercase' from flask8?

ga4gh/cli.py
jbu: Happy to be guided on ternary operators. I found them more readable once I got used to them, particularly as it becomes obvious where and when a variable is being assigned.

ga4gh/cli.py
jeromekelleher: We adopted camelCase for this project because that's the convention that the GA4GH protocol uses. I would have preferred snake_case, but having mixed conventions is much worse than one you don't like. I'm not sure where flake8 stands at the moment with camelCase identifiers, if it decides to start checking this, we'll need to update our Travis script.

ga4gh/cli.py
jeromekelleher: The pep8 tool doesn't _currently_ have any issues with backslash line continuations, but pep8 is actually pretty clear on this:

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line  continuation.

So, I think we should avoid backslashes where we can. I think this is such an occasion. What about
```python
    parser = (
        self.parser 
        if self.parser is not None 
        else argparse.ArgumentParser(
            description="GA4GH server configuration validator"))
```
I quite like ternary operators, but I find it hard to write PEP8 compliant code using them, for just this reason. I would tend to favour the simpler
```python
    parser = self.parser
    if parser is None:
          parser = ArgumentParser(...)
```
in this situation.

jeromekelleher:
Looks good, thanks @jbu. I agree we should reuse the Flask configuration parsers to get our data in, otherwise we'll be duplicating a lot of their code. Reusing the unittest framework to do the tests is very sneaky - I like it! 

The only issue I have here is that I think we should take the real work out of the 'cli' module. Can you move all the code where we do the actual testing and so on into a new module called 'ga4gh/configtest.py' please?

jbu:
So I found there was a module configuration test that wasn't passing. I added configtest.py to the notional 'libraries' module as it's the best fit currently. Or, I could start a new module.

ga4gh/configtest.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import division
```
jeromekelleher: Can we get a module docstring here please? Doesn't have to be anything elaborate, just a quick description of what the module is for.

ga4gh/configtest.py
```diff
@@ -0,0 +1,48 @@
...
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
```
jeromekelleher: I'd be happier if we could keep all of the argparse stuff in the cli module. The reasoning for this is that it makes it a lot easier for us to keep our CLI applications consistent if we try to share as much code as possible between then. There's a quite a lot of sharing between the apps at the moment, as we use functions to create command line parsers that have the same arguments and so on.

ga4gh/configtest.py
```diff
@@ -0,0 +1,48 @@
...
+
+class TestConfigExists(unittest.TestCase):
+    """This class implements configuration checks for the GA4GH server. It is
+    structured as a set of unit tests, and relies on the unittest module.
+    However, as some of the tests require state injection (through command
+    line arguments) we need to break the usual unittest pattern and instead
+    implement our own simple test runner.
+    """
+
+    def __init__(self, parser=None):
```
jeromekelleher: Following on from the point above, how about we pass configFile and configClass as arguments to this constructor, and take the command line argument parsing bit out of the setUp function below?

ga4gh/cli.py
```diff
@@ -816,3 +818,37 @@ def client_main(parser=None):
...
+    try:
+        getattr(testObj, test)()
+    except AssertionError as e:
+        return test, e
+    else:
+        return None
+
+
+def configtest_main(parser=None):
+    testConfig = configtest.TestConfig(parser)
```
jeromekelleher: Maybe we could use `unittest.loadTestsFromModule` or similar to discover the tests, so we don't have to register each one?

jbu: I rejected that originally, but perhaps prematurely. Will have another look.

ga4gh/cli.py
```diff
@@ -816,3 +818,59 @@ def client_main(parser=None):
...
+    except Exception as e:
+        results = [('setUp failed', e)]
+    else:
+        results = run_tests(testConfig, tests)
+
+    logging.basicConfig(level=logging.INFO)
+    log = logging.getLogger(__name__)
+    log.info('{0}/{1} tests passed'.format(results.count(None),
+                                           len(results)))
+    for r in results:
```
dcolligan: we tend to frown on single-letter variables here, unless they are counting variables (which this one is not)

ga4gh/cli.py
```diff
@@ -816,3 +818,59 @@ def client_main(parser=None):
...
+    configStr = 'ga4gh.serverconfig:{0}'.format(args.config)
+    testConfig = configtest.TestConfig(configStr,
+                                       'GA4GH_CONFIGURATION',
+                                       args.config_file)
+    tests = ['test_config_parameters']
+    try:
+        # setUp can fail, so we wrap it in a try to help it look like
+        # the other tests.
+        testConfig.setUp()
+    except Exception as e:
```
dcolligan: another single-letter variable

ga4gh/cli.py
```diff
@@ -816,3 +818,59 @@ def client_main(parser=None):
...
+
+def run_tests(testObj, tests):
+    """
+    Run a list of tests on a testObj
+    :param testObj: The object under test
+    :param tests: the list of test methods to run
+    :return: a list of results, either None, or a tuple of the test name and
+    reason for failure (raised exception)
+    """
+    def run_test(o, t):
```
dcolligan: single-letters...

ga4gh/cli.py
```diff
@@ -816,3 +818,59 @@ def client_main(parser=None):
     else:
         runner = args.runner(args)
         runner.run()
+
+##############################################################################
+# Configuration testing
+##############################################################################
+
+
+def run_tests(testObj, tests):
```
jeromekelleher: All functions should be camelCase also (the '_main' functions were the only exception to this, which I keep forgetting to fix!). Sorry, should have spotted this earlier.

ga4gh/configtest.py
jbu: done. I don't know why it's not showing here.

jbu:
have added test discovery. I had a good look at how one might pass state to unit tests using the unittest module, and the best I could come up with was to just set class variables in the test class before running any tests. There were potentially other ways involving subclassing bits of the test framework, but they started to get into the framework internals. None of the options are fantastic, but at least this one is simple.

jeromekelleher:
I like it @jbu, very tidy. Shall we squash these commits and aim to merge this much, or would you prefer to add some more functionalilty first?

jbu:
I'll squash these (once travis goes green). We can then build on that.

jbu:
Ready for merge.

jeromekelleher:
+1.

@dcolligan, can you merge please, if you are happy with this?

ga4gh/cli.py
```diff
@@ -816,3 +821,62 @@ def client_main(parser=None):
...
+##############################################################################
+# Configuration testing
+##############################################################################
+
+
+class SimplerResult(unittest.TestResult):
+    """
+    The TestResult class gives formatted tracebacks as error messages, which
+    is not what we want. Instead we just want the error message from the
+    err praram. Hence this subclass.
```
pgrosu: Probably might be helpful to substitute `praram` with `param`.

PR #434, opened 2015-06-03
Adding ga2lastgraph, validating server responses

adamnovak:
I've written a ga2lastgraph tool that will dump a referenceset (or actually, currently, all referencesets on a server, due to the server's lack of filtering) to an attempt at LastGraph format, which the Bandage tool can visualize.

It is used thusly:

```
ga2lastgraph --outputFile test.lastgraph http://ec2-54-149-188-244.us-west-2.compute.amazonaws.com/cactus-brca2/v0.6.g
```

It pulls down the sequence data as well, which can result in a lot of round trips for a large graph. Additionally, the page sizes the server is willing to serve seem to slow down the processing of the sequences and joins.

There doesn't seem to be a good way to accomplish this export in a streaming fashion, because sequences need to be split where joins attach (since LastGraph depicts a sequence graph, not a side graph).

In addition to the export tool, I have added validation of server responses, and a warning that the client will log when the server returns invalid Avro data. This, surprisingly, seems to happen quite a lot.

### DIFF
--- a/ga4gh/datamodel/graphs.py
+++ b/ga4gh/datamodel/graphs.py
@@ -211,7 +211,6 @@ class GraphDatabase(object):
         with sidegraph.SideGraph(self._dbFile, self._dataDir) as sg:
             count = sg.searchSequencesCount()
             rawSequences = sg.searchSequences(limits)
-
             sequences = []
             for rawSeq in rawSequences:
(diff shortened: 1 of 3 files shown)


adamnovak:
No more tests fail than in the existing graph branch; someone needs to fix the read search response so it validates.

Issue #376, opened 2015-04-28
Dataset support required


jeromekelleher:
Currently, we implicitly only support one dataset. We need to support an arbitrary number of datasets. This issue acts as a meta-issue and can be closed once all of the dependent issues have been resolved.

1. ~~Choose a configuration mechanism suitable for multiple datasets (collections of reads and variants) and references. Issue #375 has been created for this. The server should insist on exactly one dataset being specified while the rest of the code is being updated.~~
3. Update the example data and the documentation to reflect this new structure. 
2. Generalise backend.py to allow for multiple datasets. *Important* We do not need to support queries spanning multiple datasets, and should require exactly one dataset be specified. (A change in the schemas to make this official will be included any month now --- see ga4gh/schemas#253)

dcolligan:
@jeromekelleher do we have a precise definition of what a dataset is (supposed to be)?  Why are datasets even necessary?  Couldn't one keep adding directories under `/variants`, etc. if one wanted to serve more variants?

diekhans:
It is not clear what a DataSet is to GA4GH.  The intention was not
documentation and it's use is inconsistent.  It's burred in reads.avdl,
but it appears to apply to other data types.

Might be worthwhile finding out who add this can get them to
document what it is.

pgrosu:
@diekhans, @jeromekelleher, @dcolligan So a Dataset was added as a logical grouping for a collection of one or more of the following - just think of a dataset as a study that was performed and which joins/groups all the relevant data associated with it:

- ReadGroup
- ReadGroupSet
- VariantSet 
- FeatureSet 

This first introduced in [schemas PR #33](https://github.com/ga4gh/schemas/pull/33), and is available at the following link:

https://github.com/ga4gh/schemas/pull/33

Hope it helps,
Paul

pgrosu:
And yes you will need querying multiple datasets, especially for metadata among other things.  For a review of the different reasons why, below is a link to [pull-request 253](https://github.com/ga4gh/schemas/pull/253) from schemas:

https://github.com/ga4gh/schemas/pull/253

~p

dcolligan:
Seems like @jeromekelleher filed a clarification issue a while back: https://github.com/ga4gh/schemas/issues/248

dcolligan:
Fixed in #442

PR #442, opened 2015-06-08
Dataset support

dcolligan:
TODO:
- the end-to-end tests are failing because our default for variants-search is to first to a variantsets-search to request all of the variantsets of the server.  With multiple datasets, however, we must also specify what dataset's variantSets we want (and we have a self-imposed restriction right now of only being able to request one dataset).  Knowing this datasetId is impossible without querying the server before the variantsets-search request, so we need to request it from the server.  There is a `/datasets/search` path specified in the most current schema, but it is not in our `_protocol_definitions.py`.  We can't get the response object into `_protocol_definitions.py` without running `process_schemas.py` on the most recent schemas, which is a questionable idea in and of itself (see #379).  An alternative is to just cut and paste the response object into `process_schemas.py`, resulting in an additionally-hacked `process_schemas.py` (which may be the best solution).  How to proceed here? 
- tests for multiple datasets

Issue #376

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -127,6 +127,10 @@ class RequestFactory(object):
         request.referenceName = self.args.referenceName
         return request
 
+    def createSearchDatasetsRequest(self):
+        request = protocol.SearchDatasetsRequest()
+        return request
+
     def createListReferenceBasesRequest(self):
         request = protocol.ListReferenceBasesRequest()
         request.start = self.args.start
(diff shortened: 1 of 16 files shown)


jeromekelleher:
A quick scan says this looks good @dcolligan, thanks. I think the simplest solution right now is to append the definition of the dataset stuff you need on to the end of ``_protocoldefinitions.py``. We'll have some pretty big schema upheaval soon (hopefully), so lets just work around the short term issues whatever way we can.

dcolligan:
@jeromekelleher ok, went ahead and did that; pushed new commit

ga4gh/cli.py
```diff
@@ -357,10 +361,18 @@ def __init__(self, args):
         super(SearchVariantsRunner, self).__init__(args)
         request = RequestFactory(args).createSearchVariantsRequest()
         # if no variantSets have been specified, send a request to
-        # the server to grab all variantSets and then continue
+        # the server to grab all variantSets in the first dataset
```
jeromekelleher: Why the first data set here? Shouldn't we do this for all datasets for consistency?

tests/unit/test_backends.py
```diff
@@ -208,6 +213,21 @@ def testRunListReferenceBases(self):
...
+        # no datasetIds attr
+        request = protocol.SearchReadsRequest()
+        self._backend._getDatasetFromRequest(request)
+
+        # datasetIds attr
+        request = protocol.SearchVariantSetsRequest()
+        with self.assertRaises(exceptions.NotExactlyOneDatasetException):
+            self._backend._getDatasetFromRequest(request)
+        request.datasetIds = ['dataset1']
+        self._backend._getDatasetFromRequest(request)
```
jeromekelleher: Maybe check that this is the expected value? Or perhaps we should have another separate test for this?

jeromekelleher:
Looks great, thanks @dcolligan! I've a few very minor comments above, but I'm +1 overall.

ga4gh/cli.py
dcolligan: Doesn't seem to make sense if we're only restricted to one-dataset-only queries on the server side

dcolligan: Furthermore, this series of queries is intended for people who are under-specifying what they want, not our, erm, "power users"

dcolligan:
@jeromekelleher pushed an update addressing your concern about the test

ga4gh/cli.py
jeromekelleher: I don't see the distinction here @dcolligan --- if we leave the `variantSets` option unset, we're saying search over all variant sets. To me, this means
```python
for datasetId in getDatasets():
    for variantSet in getVariantSets(datasetId):
         # do stuff for the variant set
```

jeromekelleher:
What's the latest here @dcolligan?

dcolligan:
@jeromekelleher pushed an update that will hopefully satisfy your concerns

jeromekelleher:
Looks great, thanks @dcolligan. 

+1. Since no one has weighed in up to this point I'm going to merge now.

Issue #449, opened 2015-06-15
Server returns landing page when any non-mapped URL requested


jeromekelleher:
The server does not return an error when we try to access a nonexistent URL, but instead returns the index landing page. We should return a 404.

Steps to reproduce:

1. Run the latest development server;
2. Put something like http://localhost:8000/thisShould404 into your web browser.

dcolligan:
Fixed in #452

PR #452, opened 2015-06-15
Certain non-mapped URLs now return 404s

dcolligan:
Issue #449

Kinda weird this wasn't caught by the `tests.unit.test_views:TestFrontend.test404sReturnJson` test.  Not sure what accounts for the behavior discrepancy between the test and requesting the nonexistant page in a browser...

### DIFF
--- a/tests/unit/test_views.py
+++ b/tests/unit/test_views.py
@@ -131,10 +131,16 @@ class TestFrontend(unittest.TestCase):
         return response
 
     def test404sReturnJson(self):
-        path = utils.applyVersion('/doesNotExist')
-        response = self.app.get(path)
-        protocol.GAException.fromJsonString(response.get_data())
-        self.assertEqual(404, response.status_code)
+        paths = [
+            '/doesNotExist',
+            utils.applyVersion('/doesNotExist'),
+            utils.applyVersion('/reads/sea'),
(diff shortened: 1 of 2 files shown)


jeromekelleher:
It would be nice to see some tests to go with this --- we should figure what happened here and why there's a difference between the browser and not. Now that we're adding authentication, we need to be absolutely rock solid sure about the URL mapping code.

dcolligan:
Figured it out... stupid mistake

tests/unit/test_views.py
```diff
@@ -131,7 +131,7 @@ def sendReferenceBasesList(self, id_=None):
         return response
 
     def test404sReturnJson(self):
-        path = utils.applyVersion('/doesNotExist')
+        path = '/doesNotExist'
```
jeromekelleher: Maybe we could get some more examples here? I.e. things which have versions included, things which don't, things which are prefixes of legal URLs, legal URLS that have a suffix (.e.g., GET /variantsets/id/should404) . We should be testing a whole pile of different things here to make sure they 404.

dcolligan:

jeromekelleher:
Excellent, thanks @dcolligan.

+1. Merging.

Issue #431, opened 2015-06-02
API endpoints defined in schema but not yet implemented should return 501


dcolligan:
They currently return a 404

Based on discussion in #430

dcolligan:
This would probably best be handled by also creating a `getMethods` in `_protocol_definitions` (which we can do without waiting for an error-free re-generation from a recent schema, since it will go at the bottom, separate from the rest of the data) which is used (along with `postMethods`) to check against the requested path in an error handler in `frontend.py`

dcolligan:
@jeromekelleher did you have any ideas for creating the `getMethods` attribute in `_protocol_definitions` ?  We can't generate it in the same way since those endpoints don't correspond to a request/response class.

jeromekelleher:
I was hoping there would be a way of getting at these using Avro. There is a method definition within the schema IDL, so I'd have hoped that the Avro library would provide a way of accessing it. I've gotten used to being disappointed by Avro though, so it might be as well to just define the functions in the frontend by hand.

dcolligan:
Fixed in #454

PR #454, opened 2015-06-16
Make unimplemented paths defined in schema 501

dcolligan:
As of schema version 84927c4cc9a74a3db310026de8b31cdca8bbcfd0

Issue #431 

It would probably be better to use the flask route syntax instead of the `re` module, but I'm not sure if Flask supports matching routes outside an app context...

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -13,6 +13,7 @@ import datetime
 import flask
 import flask.ext.cors as cors
 import humanize
+import werkzeug
 
 import ga4gh
 import ga4gh.backend as backend
@@ -24,6 +25,32 @@ SEARCH_ENDPOINT_METHODS = ['POST', 'OPTIONS']
 
 
 app = flask.Flask(__name__)
+
+
+class NoConverter(werkzeug.routing.BaseConverter):
+    """
+    A converter that allows the routing matching algorithm to not
+    match on certain literal terms
+
+    This is needed because if there are e.g. two routes:
+
+    /<version>/callsets/search
+    /<version>/callsets/<id>
+
+    A request for /someVersion/callsets/search will get routed to
+    the second, which is not what we want.
+    """
+    def __init__(self, map, *items):
+        werkzeug.routing.BaseConverter.__init__(self, map)
+        self.items = items
+
+    def to_python(self, value):
+        if value in self.items:
+            raise werkzeug.routing.ValidationError()
+        return value
+
+
+app.url_map.converters['no'] = NoConverter
 
 
 class Version(object):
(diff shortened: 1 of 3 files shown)


jeromekelleher:
This looks good, but I'm a bit worried about the regex stuff. I know it's a bit tedious, but would it be simpler to just put in the handler functions and 501 from all the ones that we haven't implemented?

dcolligan:
@jeromekelleher alright, did that (which required a little flask hacking)

tests/unit/test_views.py
```diff
@@ -269,3 +272,46 @@ def testCurrentVersion(self):
...
+            '/mode/<id>',
+            '/datasets/<id>',
+            '/readgroupsets/<id>',
+            '/readgroups/<id>',
+        ]
+
+        def runRequest(method, path):
+            requestPath = path.replace('<id>', 'someId')
+            versionedPath = utils.applyVersion(requestPath)
+            print(method, versionedPath)
```
jeromekelleher: Stray print here.

ga4gh/frontend.py
```diff
@@ -384,8 +403,139 @@ def searchDatasets(version):
         version, flask.request, app.backend.searchDatasets)
 
 
+# The below paths have not yet been implemented
+
+
+@app.route('/<version>/callsets/<no(search):id>')
```
jeromekelleher: I don't get it: why the <no(search):id> bit?

dcolligan: If there are two routes, 
```
/<version>/callsets/search
/<version>/callsets/<id>
```
a request to `/someVersion/callsets/search` will be routed to the later, which is not what we want.  Therefore, we have this converter to explicitly say that the `<id>` attribute of the later should not match on a literal `search`, thus routing the request correctly.

dcolligan:
@jeromekelleher pushed update removing print statement

ga4gh/frontend.py
jeromekelleher: Ugh, this is pretty horrible. So, we can never have an object with ID = 'search'? Is this a standard RESTish issue do you think, or something new we've made up ourselves?

There's not a lot we can do about it though. Can you add a quick comment on the Converter class to explain why we need this? I'm happy to merge then.

dcolligan: I don't think it's terribly unusual to have certain reserved paths in REST (in any event, it's defined by the schema, so our hands are tied somewhat).  Alternatives would be (if `foo` is the resource):

- have an attribute in the POST payload to `/foo` specifying the action (this seems like overloading paths = bad)
- have `/foo` for CRUD methods and `/foos` for more advanced ones over all the `foo` objects like search (but foo vs foos distinction could be seen as confusing)... that is, if we even ever implement CRUD methods
- organize the URL space differently and have a `/search/foos` endpoint (seems to break the object-then-action URL simplicity)
- have some other piece of data be sent along with the usual payload to `/foo`, like a header or something (confusing)
- just assume a POST to `/foo` means we want to search (possible, but would make any CRUD implementation backwards-incompatible)

None of these seem as attractive as what we're doing now... in any event, `search` seems like a pretty improbable ID (although, I don't doubt that someone, somewhere...)

dcolligan:
@jeromekelleher pushed update with comment

jeromekelleher:
Thanks @dcolligan, merging.

ga4gh/frontend.py
jeromekelleher: Fair enough; thanks for the analysis.

Issue #455, opened 2015-06-16
Unable to use referenceId, start fields in /reads/search API call


danvk:
Following the [demo instructions][1], I can query _all_ reads:

```
$ curl --data '{"readGroupIds":["low-coverage:HG00534.mapped.ILLUMINA.bwa.CHS.low_coverage.20120522"]}' --header 'Content-Type: application/json' http://localhost:8000/v0.5.1/reads/search
(many results!)
```

If I try to query a particular range, however, I get an error:

```
$ curl --data '{"readGroupIds":["low-coverage:HG00534.mapped.ILLUMINA.bwa.CHS.low_coverage.20120522"], "referenceId": "1", "start": 129650, "stop": 129670}' --header 'Content-Type: application/json' http://localhost:8000/v0.5.1/reads/search | jq .
{
  "errorCode": 2019025445,
  "message": "only one of referenceId and referenceName can be specified"
}
```

This is odd, since I didn't specify `referenceName` in the request (in fact, it's not even a valid field in [`SearchReadsRequest`][2]). Dropping the `referenceId` field, I get an even stranger error:

```
$ curl --data '{"readGroupIds":["low-coverage:HG00534.mapped.ILLUMINA.bwa.CHS.low_coverage.20120522"], "start": 129650, "stop": 129670}' --header 'Content-Type: application/json' http://localhost:8000/v0.5.1/reads/search | jq .
{
  "errorCode": 3424984409,
  "message": "Internal Server Error"
}
```

Looking at the server logs, I see this:

```
  File "pysam/calignmentfile.pyx", line 572, in pysam.calignmentfile.AlignmentFile.getrname (pysam/calignmentfile.c:8026)
ValueError: reference_id 129650 out of range 0<=tid<86
```

i.e. it looks like the `start` value is being interpreted as a `referenceId`!

I'd appreciate some help figuring out why this API endpoint is behaving so oddly. Thanks!

[1]: http://ga4gh-reference-implementation.readthedocs.org/en/latest/demo.html
[2]: https://github.com/ga4gh/schemas/blob/84927c4cc9a74a3db310026de8b31cdca8bbcfd0/src/main/resources/avro/readmethods.avdl#L16

dcolligan:
Thanks for finding these issues.  I'll have a PR up to fix this soon.

dcolligan:
Fixed in https://github.com/ga4gh/server/releases/tag/v0.1.2

Thanks for the excellent bug report, @danvk !

Issue #433, opened 2015-06-02
this query get wrong server response back.


jingchunzhu:
This following query against umd get 2520 variants back, many are not within the start and end interval
python client_dev.py -vv variants-search --variantSetIds umd --referenceName 17 --start 41227747  --end 41244748 --pageSize 1000 http://ec2-54-148-207-224.us-west-2.compute.amazonaws.com:8000/v0.6.e6d6074
2520 variants

The most strange thing is that the xena browser client (in javascript) also get the "same"(same in the sense that the same number of variants returned) wrong answer back.

however,  curl command seems to get the correct response back.

curl --data '{ "callSetIds": [],  "end": 41244748, "referenceName": "17",  "start":41227747,  "variantSetIds": ["umd”], “pageSize”:1000}’ --header 'Content-Type: application/json' http://ec2-54-148-207-224.us-west-2.compute.amazonaws.com:8000/v0.6.e6d6074/variants/search

jingchunzhu:
All three clients are now consistently getting the incorrect response back. 


pageSize parameter in the  curl command does not  seem to be used. once you manually add pageToken parameter, it also triggers the incorrect response back from the server. 

curl --data '{ "callSetIds": [], "end": 41244748, "referenceName": "17", "start":41227747, "variantSetIds": ["umd”], “pageSize”:1000}’ --header 'Content-Type: application/json' http://ec2-54-148-207-224.us-west-2.compute.amazonaws.com:8000/v0.6.e6d6074/variants/search

dcolligan:
Which version of the server is this?  Is it an old one that has the bug we fixed in this update?  https://github.com/ga4gh/server/releases/tag/v0.1.1

It looks like you are running 0.1.0b1, not 0.1.1.  Try updating and see what happens.

jingchunzhu:
For querying umd database on chr17 between position 41227747 to 41244748 the server found 2520 variants, many are not in this interval.  

here are just a random copy of paste a few that are outside (both start and end) the interval. 

variant_start     variant_end
41226453        41226454
41226456        41226457
41226473        41226474
41226487        41226488
41226487        41226488
41226487        41226488
41226487        41226488
41226487        41226488
41226487        41226488

You should see the same thing using the python client.


server implementation used:  http://ec2-54-148-207-224.us-west-2.compute.amazonaws.com:8000/v0.5.1  

GA4GH reference server 0.1.1

Protocol version 0.5.1

Operations available

Method	Path
GET	/v0.5.1
[...]

PR #456, opened 2015-06-16
Fixes for reads/search

dcolligan:
- fix referenceId and referenceName not getting passed correctly
- sanitize pysam getrname calls

Issue #455 

The first error was probably introduced when we backported the paging changes to `stable` without taking into account the API signature change.

We should bundle this change, along with the paging fix in #433, into a new bugfix release, IMHO.

We should also forward-port the sanitization changes.


jeromekelleher:
Looks good, thanks @dcolligan. I agree about the process --- we should get a bugfix release out ASAP.

On the process though, I think we should start using a release branch. We should keep the stable branch just for stuff we've actually released, not stuff we're about to release soon. I've created the release-0.1 branch for this; can you change this PR so that it's against that branch please?

PR #457, opened 2015-06-17
Updated interval paging algorithm + better tests.

jeromekelleher:
This addresses issue #433.

Several subtle issues weren't dealt with correctly in the existing interval iteration code. This has been dealt with by adding a more abstract testing layer which lets us work directly on raw intervals without needing to interact with VCF and BAM at all. This includes a variety of systematic and random tests. These are somewhat slow at the moment, but they can probably be improved by using a proper interval tree for searches.


dcolligan:
shouldn't this target the `release-0.1` branch and then be forward-ported?

dcolligan:
also, is it really a good idea to introduce randomness into our tests?

jeromekelleher:
I don't think it matters which way around we do this: we can cherry pick this commit on to ``release-0.1`` after it's merged into develop, or do it the other way around. I'll re-open the PR against release-0.1 if you  think it's important to do it this way.

I think random tests are fantastic. Random data is a million times better at thinking up awkward corner cases than I am, and I've found loads of subtle bugs by using random data tests. The worst that can happen by using random tests is that we get a failure that's hard to reproduce; but, at least we know that a bug exists then, and it's usually possible to reproduce the problem with a bit of patience. Of course random tests aren't a replacement for good systematic testing, but I think they're an essential part of any good test suite.

dcolligan:
No, I don't have a strong opinion one way or the other on porting direction.  This is fine.

ga4gh/backend.py
```diff
@@ -55,71 +55,99 @@ def _getVariantSet(request, variantSetIdMap):
...
+        self._searchAnchor = searchAnchor
+        self._distanceFromAnchor = objectsToSkip
+        self._searchIterator = self._search(searchAnchor, self._request.end)
+        obj = next(self._searchIterator)
+        if searchAnchor == self._request.start:
+            # This is the initial set of intervals, we just skip forward
+            # objectsToSkip positions
+            for _ in range(objectsToSkip):
+                obj = next(self._searchIterator)
+        else:
```
dcolligan: Couldn't `searchAnchor < self._request.start` or `searchAnchor > self._request.start` at this point with this logic?  Would either of those be a problem?

tests/unit/test_intervals.py
```diff
@@ -0,0 +1,251 @@
...
+from __future__ import unicode_literals
+
+import unittest
+import random
+
+import ga4gh.backend as backend
+
+
+def setUp():
+    # set the random seed to this is reproducable.
```
dcolligan: wording in this comment is weird

tests/unit/test_intervals.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+class IntervalSet(object):
+    """
+    Class representing a set of half-closed intervals.
+    """
+    def __init__(self, start, end, intervals):
+        self.start = start
+        self.end = end
+        self.intervals = sorted(intervals)
```
dcolligan: this should probably sort only by start coordinate, not both start and end, as AFAICT that's not a requirement of returned variants or reads

dcolligan:
couple minor qualms, but in general +1

ga4gh/backend.py
jeromekelleher: No, because we set searchAnchor = request.start for the first set of intervals (i.e., those with start <= request.start), and we don't update it until start becomes > request.start. This is the signalling condition to let us know we are in the initial phase.

tests/unit/test_intervals.py
jeromekelleher: Updated.

tests/unit/test_intervals.py
jeromekelleher: Good point; updated.

jeromekelleher:
Thanks for the comments @dcolligan, updated pushed.

dcolligan:
Great, just merged.  Can you backport this to the `release-0.1` branch?  After you do that we can release an update.

jeromekelleher:
Backporting done in #461 @dcolligan (git is freakin awesome!), so I think we're ready for a bugfix release. Would you like to do the honours?

PR #458, opened 2015-06-17
Fixes for reads/search

dcolligan:
Reopening #456 against the release-0.1 branch

- fix referenceId and referenceName not getting passed correctly
- sanitize pysam getrname calls

Issue #455

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -141,6 +141,7 @@ class ReadsIntervalIterator(IntervalIterator):
 
     def _getIterator(self):
         iterator = self._container.getReadAlignments(
+            self._request.referenceName,
             self._request.referenceId,
             self._startPosition, self._request.end)
         return iterator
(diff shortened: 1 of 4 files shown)


jeromekelleher:
+1, merging. As you said in #456, we should forward port the sanitising and tests to develop too.

bioinformed:
Are you sure you want to be changing the numeric values of a reference ids
to clamp them within a standard range?  It seems like a recipe for massive
confusion, since you're effectively changing the query and thus changing
data.  Maybe I'm missing something fundamental, but a simple type and range
check would seem the safer and more correct way to go.

dcolligan:
@bioinformed this change doesn't do that, but other functions in the sanitizer do.  This change merely errors with a more informative error message if the referenceId would cause pysam to error anyways.

bioinformed:
Ah-- I must have misread the partial diff.

Issue #450, opened 2015-06-15
Simulated references and reference sets need to be updated


jeromekelleher:
At the moment, we cannot specify the number of references and reference sets when  we create a simulated backend. We need to provide a much more flexible interface (simular to variant sets) so that we can

1. Specify the number of reference sets;
2. specify the number of references per reference set;
3. generate random bases, for an arbitrarily large region _without_ storing the bases directly. This can be done using a similar method to the approach used for variants, where we use the position (plus another seed) to seed the generator.

Md5checksums aren't really supported as yet, so we can ignore these.

We should update the `test_simulated_stack.py` tests to thoroughly exercise this interface.

dcolligan:
Fixed in #464

PR #464, opened 2015-06-25
Update simulated references and reference sets

dcolligan:
Issue #450


ga4gh/datamodel/references.py
```diff
@@ -68,11 +69,13 @@ class SimulatedReferenceSet(AbstractReferenceSet):
...
     """
-    def __init__(self, id_):
+    def __init__(self, id_, randomSeed=0, numReferences=1):
         super(SimulatedReferenceSet, self).__init__(id_)
-        referenceId = "{}:srsone".format(id_)
-        reference = SimulatedReference(referenceId)
-        self._referenceIdMap[referenceId] = reference
+        for i in range(numReferences):
+            referenceId = "{}:srs{}".format(id_, i)
+            reference = SimulatedReference(referenceId, randomSeed)
```
jeromekelleher: Does every Reference get the same random seed then? It would be good to give different references different seeds so they are not all identical. The easiest way to do that is to make a local random generator, seeded with randomSeed and use that to generate seeds for the Reference objects.

ga4gh/datamodel/references.py
```diff
@@ -135,9 +138,19 @@ class SimulatedReference(AbstractReference):
...
+        self._randomSeed = randomSeed
+        self._randomGenerator = random.Random()
+        self._randomGenerator.seed(self._randomSeed)
+        self._length = length
+        bases = []
+        for _ in range(length):
+            choice = random.choice(self.choices)
+            bases.append(choice)
+        self.bases = ''.join(bases)
         self._md5checksum = "lol"
```
jeromekelleher: Might as well just calculate the md5 here?

jeromekelleher:
Look good, thanks @dcolligan. +1 apart from some minor quibbles above.

dcolligan:

jeromekelleher:
Looks good, thanks @dcolligan.

Issue #441, opened 2015-06-08
LRU cache for file handles


jeromekelleher:
Currently, the reference server opens a htslib backed file handle for each FASTA, VCF and BAM file that it serves on startup. When a incoming request is processed, the queries are run on the relevant files using pysam. This has the advantage of simplicity, and means that we don't need to open the file each time a request comes in (which would be prohibitively expensive). However, this approach scales poorly with increasing numbers of VCF and BAM files: (1) server startup becomes very slow; (2) we use a lot of memory keeping files we don't need open; (3) we start hitting OS limits on the number of open file descriptors.

A good compromise between keeping files open all the time and opening a file each time we receive a query is to maintain a least recently used cache of data file handles. We have a configurable limit on the number of concurrently open data files, and each time we receive a request we ask a central cache to open a file for us. If it is already open, it returns the file handle; if not, it opens it for us, possibly closing another file first.

**Possible implementation**

- The cache should operate at the level of the `datamodel.PysamDatamodelMixin` class, which is a mixin superclass of all datamodel objects (VariantSet, ReadGroupSet and ReferenceSet) offering an interface to htslib backed data files. This class can have a class method that takes a path (possibly a URL) and a method and returns an opened file handle. The cache itself should probably be an instance of another class (.e.g. `PysamFileHandleCache`) to avoid making the mixin too complicated.
- `PysamFileHandleCache` has a method `getFileHandle(self, url, method)`, which returns the opened file handle. The simplest (non-caching) implementation of this would be 
```python
     def getFileHandle(self, url, openMethod):
         return openMethod(url)
```
- We then remove the instance variables holding the file handles in `HtslibReadGroup`, `HtslibVariantSet` and `HtslibReferenceSet`, and dynamically get the file handles using the cache instead when processing queries in `getVariants`, etc.

This is only a suggested implementation, and may well change considerably as the actual coding is done. 

**Complications**

- This is slightly complicated by the fact that all the data for (e.g.) VariantSets is not contained in a single file. We sometimes have data spread across multiple files in a directory (usually, one per chromosome) and then figure out which file we need to query at run time. The `_chromFileMap` is used for this in `HtslibVariantSet`. We can easily change this to map from chromosome to the file's URL, and use this to open the file at run time. However, we'll still need to open the files at startup to examine their contents and build up the map. Initially, it's probably easiest to skip the cache at this point, and just open the file, read the information we need, and close it again.
[...]

PR #468, opened 2015-06-30
Cache for open file handles

eelsther:
This is the implementation for the open file handles cache (issue #441)
Please review and let me know if you see any problems with the proposed implementation.

The size of the cache (maximum number of open files) can be set in serverconfig.py with the variable CACHE_MAX_SIZE currently set to 50.

Thanks!

Esther

### DIFF
--- a/ga4gh/datamodel/__init__.py
+++ b/ga4gh/datamodel/__init__.py
@@ -12,6 +12,7 @@ import json
 import tempfile
 import shutil
 import atexit
+import collections
 
 import ga4gh.exceptions as exceptions
 
@@ -24,6 +25,89 @@ def _cleanupHtslibsMess(indexDir):
     """
     if os.path.exists(indexDir):
         shutil.rmtree(indexDir)
+
+
+class PysamFileHandleCache(object):
+    """
+    Cache for opened file handles. We use a deque which has the
+    advantage to have push/pop operations in O(1) We always add
+    elements on the left of the deque and pop elements from the right.
+    When a file is accessed via getFileHandle, its priority gets
+    updated, it is put at the "top" of the deque.
+    """
+
+    def __init__(self):
+        self._cache = collections.deque()
+        self._memoTable = dict()
+        # Initialize the value even if it will be set up by the config
+        self._maxCacheSize = 50
+
+    def setMaxCacheSize(self, size):
+        """
+        Sets the maximum size of the cache
+        """
+        if size <= 0:
+            raise ValueError(
+                "The size of the cache must be a strictly positive value")
+        self._maxCacheSize = size
+
+    def _add(self, dataFile, handle):
+        """
+        Add a file handle to the left of the deque
+        """
+        self._cache.appendleft((dataFile, handle))
+
+    def _update(self, dataFile, handle):
+        """
+        Update the priority of the file handle. The element is first
+        removed and then added to the left of the deque.
+        """
+        self._cache.remove((dataFile, handle))
+        self._add(dataFile, handle)
+
+    def _removeLru(self):
+        """
+        Remove the least recently used file handle from the cache.
+        The pop method removes an element from the right of the deque.
+        Returns the name of the file that has been removed.
+        """
+        (dataFile, handle) = self._cache.pop()
+        handle.close()
+        return dataFile
+
+    def getCachedFiles(self):
+        """
+        Returns all file names stored in the cache.
+        """
+        return self._memoTable.keys()
+
+    def getFileHandle(self, dataFile, openMethod):
+        """
+        Returns handle associated to the filename. If the file is
+        already opened, update its priority in the cache and return
+        its handle. Otherwise, open the file using openMethod, store
+        it in the cache and return the corresponding handle.
+        """
+        if dataFile in self._memoTable:
+            handle = self._memoTable[dataFile]
+            self._update(dataFile, handle)
+            return handle
+        else:
+            try:
+                handle = openMethod(dataFile)
+            except ValueError:
+                raise exceptions.FileOpenFailedException(dataFile)
+
+            self._memoTable[dataFile] = handle
+            self._add(dataFile, handle)
+            if len(self._memoTable) > self._maxCacheSize:
+                dataFile = self._removeLru()
+                del self._memoTable[dataFile]
+            return handle
+
+
+# LRU cache of open file handles
+fileHandleCache = PysamFileHandleCache()
 
 
 class DatamodelObject(object):
@@ -187,3 +271,6 @@ class PysamDatamodelMixin(object):
             atexit.register(_cleanupHtslibsMess, indexDir)
         if numDataFiles == 0:
             raise exceptions.EmptyDirException(dataDir, patterns)
+
+    def getFileHandle(self, dataFile):
+        return fileHandleCache.getFileHandle(dataFile, self.openFile)
--- a/ga4gh/datamodel/references.py
+++ b/ga4gh/datamodel/references.py
@@ -175,14 +175,17 @@ class HtslibReference(datamodel.PysamDatamodelMixin, AbstractReference):
     def __init__(self, id_, dataFile):
         super(HtslibReference, self).__init__(id_)
         self._fastaFilePath = dataFile
-        self._fastaFile = pysam.FastaFile(dataFile)
-        numReferences = len(self._fastaFile.references)
+        fastaFile = self.openFile(dataFile)
+        numReferences = len(fastaFile.references)
         if numReferences != 1:
             raise exceptions.NotExactlyOneReferenceException(
                 self._id, numReferences)
-        self._refName = self._fastaFile.references[0]
-        # refData = self._fastaFile.fetch(self._refName)
+        self._refName = fastaFile.references[0]
+        # refData = fastaFile.fetch(self._refName)
         self._md5checksum = "TODO"  # hashlib.md5(refData).hexdigest()
+
+    def openFile(self, dataFile):
+        return pysam.FastaFile(dataFile)
 
     def getFastaFilePath(self):
         """
@@ -192,11 +195,12 @@ class HtslibReference(datamodel.PysamDatamodelMixin, AbstractReference):
 
     def getBases(self, start=None, end=None):
         start, end = self.sanitizeFastaFileFetch(start, end)
-        bases = self._fastaFile.fetch(self._refName, start, end)
+        bases = self.getFileHandle(self.getFastaFilePath()).fetch(
+                      self._refName, start, end)
         return bases
 
     def getLength(self):
-        return self._fastaFile.lengths[0]
+        return self.getFileHandle(self.getFastaFilePath()).lengths[0]
 
     def getName(self):
-        return self._fastaFile.references[0]
+        return self.getFileHandle(self.getFastaFilePath()).references[0]
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -24,6 +24,7 @@ import requests
 
 import ga4gh
 import ga4gh.backend as backend
+import ga4gh.datamodel as datamodel
 import ga4gh.protocol as protocol
 import ga4gh.exceptions as exceptions
 
@@ -200,6 +201,9 @@ def configure(configFile=None, baseConfig="ProductionConfig",
     if configFile is not None:
         app.config.from_pyfile(configFile)
     app.config.update(extraConfig.items())
+    # Setup file handle cache max size
+    datamodel.fileHandleCache.setMaxCacheSize(
+        app.config["FILE_HANDLE_CACHE_MAX_SIZE"])
     # Setup CORS
     cors.CORS(app, allow_headers='Content-Type')
     app.serverStatus = ServerStatus()
--- a/ga4gh/serverconfig.py
+++ b/ga4gh/serverconfig.py
@@ -29,6 +29,8 @@ class BaseConfig(object):
     SIMULATED_BACKEND_NUM_REFERENCES_PER_REFERENCE_SET = 1
     SIMULATED_BACKEND_NUM_ALIGNMENTS_PER_READ_GROUP = 2
 
+    FILE_HANDLE_CACHE_MAX_SIZE = 50
+
 
 class DevelopmentConfig(BaseConfig):
     """
--- a/tests/unit/test_file_handle_cache.py
+++ b/tests/unit/test_file_handle_cache.py
@@ -1,0 +1,66 @@
+"""
+Tests the file handle cache
+"""
+
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import os
+import shutil
+import tempfile
+import unittest
+import uuid
+
+import ga4gh.datamodel as datamodel
+
+
+class TestFileHandleCache(datamodel.PysamFileHandleCache, unittest.TestCase):
+    def __init__(self, *args, **kwargs):
+        super(TestFileHandleCache, self).__init__()
+        unittest.TestCase.__init__(self, *args, **kwargs)
+
+    def setUp(self):
+        self._tempdir = tempfile.mkdtemp(prefix="ga4gh_file_cache",
+                                         dir=tempfile.gettempdir())
+
+    def _getFileHandle(self, dataFile):
+        def openMethod(dataFile):
+            return open(dataFile, 'w')
+        return self.getFileHandle(dataFile, openMethod)
+
+    def testGetFileHandle(self):
+        def genFileName(x):
+            return os.path.join(self._tempdir, str(uuid.uuid4()))
+
+        # Set cache size to 9 files max
+        self.setMaxCacheSize(9)
+
+        # Build a list of 10 files and add their handles to the cache
+        fileList = map(genFileName, range(0, 10))
+
+        for f in fileList:
+            handle = self._getFileHandle(f)
+            self.assertEquals(self._cache.count((f, handle)), 1)
+
+        self.assertEquals(len(self._memoTable), len(self._cache))
+
+        # Ensure that the first added file has been removed from the cache
+        self.assertEquals(filter(lambda x: x[0] == fileList[0], self._cache),
+                          [])
+
+        topIndex = len(self._cache) - 1
+
+        # Update priority of this file and ensure it's no longer the
+        # least recently used
+        self.assertEquals(self._cache[topIndex][0], fileList[1])
+        self._getFileHandle(fileList[1])
+        self.assertNotEqual(self._cache[topIndex][0], fileList[1])
+        self.assertEquals(self._cache[0][0], fileList[1])
+
+    def testSetCacheMaxSize(self):
+        self.assertRaises(ValueError, self.setMaxCacheSize, 0)
+        self.assertRaises(ValueError, self.setMaxCacheSize, -1)
+
+    def tearDown(self):
+        shutil.rmtree(self._tempdir)
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -13,7 +13,6 @@ import pysam
 
 import ga4gh.protocol as protocol
 import ga4gh.datamodel as datamodel
-import ga4gh.exceptions as exceptions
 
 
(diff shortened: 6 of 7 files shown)


ga4gh/datamodel/__init__.py
```diff
@@ -25,6 +28,28 @@ def _cleanupHtslibsMess(indexDir):
...
         shutil.rmtree(indexDir)
 
+class PysamFileHandleCache(object):
+    """
+    Cache for opened file handles
+    We use a deque which has the advantage to have push/pop operations in O(1)
+    We always add elements on the left of the deque and pop elements from the right.
+    When a file is accessed, its priority gets updated, it is put at the "top" of the deque.
+    """
+    cache = collections.deque()
```
dcolligan: This should be declared in an `__init__` method, not as a class-level attribute

ga4gh/datamodel/__init__.py
```diff
@@ -25,6 +28,28 @@ def _cleanupHtslibsMess(indexDir):
...
+        self.cache.appendleft((dataFile, handle))
+
+    def update(self, dataFile, handle):
+        try:
+            self.cache.remove((dataFile, handle))
+        except ValueError:
+            self.add(dataFile, handle)
+
+    def remove_lru(self):
+        (dataFile, handle) = self.cache.pop()
```
dcolligan: parenthesis unnecessary here

ga4gh/datamodel/__init__.py
```diff
@@ -56,6 +81,12 @@ class PysamDatamodelMixin(object):
 
     maxStringLength = 2**10  # arbitrary
 
+    # LRU cache of open file handles
+    cache = PysamFileHandleCache()
+
+    # memoization table for the function opening file handles
+    memo_table = dict()
```
jeromekelleher: camelCase please.

ga4gh/datamodel/__init__.py
```diff
@@ -187,3 +218,21 @@ def _scanDataFiles(self, dataDir, patterns):
             atexit.register(_cleanupHtslibsMess, indexDir)
         if numDataFiles == 0:
             raise exceptions.EmptyDirException(dataDir, patterns)
+
+    def getFileHandle(self, dataFile):
+        try:
```
dcolligan: the try / except pattern should only be used for exceptional conditions.  I don't think a cache miss qualifies.

Instead use the `if dataFile in self.memo_table: ... else: ...` pattern

ga4gh/datamodel/__init__.py
```diff
@@ -56,6 +81,12 @@ class PysamDatamodelMixin(object):
 
     maxStringLength = 2**10  # arbitrary
 
+    # LRU cache of open file handles
+    cache = PysamFileHandleCache()
```
jeromekelleher: This should be a protected variable, so ``_cache``. Also `_memoTable` below.

ga4gh/datamodel/__init__.py
jeromekelleher: Is there a reason for not using somthing like
```python
if dataFile in self._memoTable:
     handle = self.memo_table[dataFile]
     self._cache.update(dataFile, handle)
else:
     handle = # rest of it.
return handle
```
I know there are subtle errors with concurrency that can be avoided by using the `try/except` approach instead of testing if the key is in the dictionary, but I tend to find the latter approach more readable. I don't think we're ever going to have multiple threads in this process, it's just not worth it in Python.

ga4gh/datamodel/__init__.py
dcolligan: The other class-level attributes of the mixin are immutable; I'm not sure if adding mutable class-level attributes is going to give us one file cache or N file caches (where N = the number of instantiations of objects using the mixin).  Pretty sure it's the former, but we should confirm that in a test to be sure we are getting the desired behavior.

I guess what bugs me about this pattern is that we are combining the singleton pattern (cache) with the mixin pattern.  Separation of concerns seems to say that they should be distinct... @jeromekelleher what do you think?

ga4gh/datamodel/__init__.py
jeromekelleher: This should also be a private variable so, ``_cache``.

ga4gh/serverconfig.py
```diff
@@ -26,6 +26,7 @@ class BaseConfig(object):
     SIMULATED_BACKEND_VARIANT_DENSITY = 0.5
     SIMULATED_BACKEND_NUM_VARIANT_SETS = 1
 
+    CACHE_MAX_SIZE = 50
```
dcolligan: this should be renamed to specify which cache it is, since we might have more than one and "cache" is ambiguous

what about `FILE_HANDLE_CACHE_MAX_SIZE`?

ga4gh/datamodel/__init__.py
```diff
@@ -25,6 +28,28 @@ def _cleanupHtslibsMess(indexDir):
...
+    def add(self, dataFile, handle):
+        self.cache.appendleft((dataFile, handle))
+
+    def update(self, dataFile, handle):
+        try:
+            self.cache.remove((dataFile, handle))
+        except ValueError:
+            self.add(dataFile, handle)
+
+    def remove_lru(self):
```
jeromekelleher: camelCase please.

dcolligan:
Build is failing... use `python script/run_tests.py` to run the same tests that travis does on your local machine

ga4gh/datamodel/__init__.py
jeromekelleher: I tend to agree --- we're not encapsulating the cache concept very well here, and spreading the implementation across the two classes. More detailed comments in a minute when I figure it out.

dcolligan:
Also the cache doesn't seem threadsafe at all... @jeromekelleher you don't think this is a concern?

ga4gh/datamodel/__init__.py
```diff
@@ -187,3 +218,21 @@ def _scanDataFiles(self, dataDir, patterns):
             atexit.register(_cleanupHtslibsMess, indexDir)
         if numDataFiles == 0:
             raise exceptions.EmptyDirException(dataDir, patterns)
+
+    def getFileHandle(self, dataFile):
+        if dataFile in self.memo_table:
```
jeromekelleher: Here, we're splitting the job of the cache between the cache class and the mixin class, with both the memoisation table and the cache being updated to implement the LRU cache. All the cache functionality should be encapsulated in the cache. The getFileHandle() is a good pattern here, and we should keep this API. However, this should be a direct call to the cache class, with no logic.

We need a singleton instance of PysamFileHandleCache, and we also need to be able to modify the `maxCacheSize` externally during configuration. So, I think the simplest option is to have a module level variable called ``fileHandleCache``, which is created at module load time. I.e, near the top of the module (just after the definition of PysamFileHandleCache I guess, we have
```python
# Global singleton to cache file handles.
fileHandleCache = PysamFileHandleCache()
```

Then, during `configure` we can call
```python
datamodel.fileHandleCache.setMaxCacheSize(app.config["CACHE_MAX_SIZE"])
```

This method here, should then just be 
```
return fileHandleCache.getFileHandle()
```

This encapsulate things nicely. What do we think? @dcolligan, how does this design seem to you?

ga4gh/datamodel/__init__.py
```diff
@@ -12,8 +12,11 @@
 import tempfile
 import shutil
 import atexit
+import collections
+
 
 import ga4gh.exceptions as exceptions
+import ga4gh.serverconfig as serverconfig
```
jeromekelleher: We don't want to import this here, as this only gives us static configuration --- see below for a suggestion on how to restructure to give us dynamic config.

jeromekelleher:
This looks good, thanks @eelsther. The majority of this is excellent, and I think we can improve the rest easily enough using the slight restructuring suggested above. I think that encapsulating all the cache functionality into one class will make it easier to see what's going on.

We should also get some unit tests on the cache class, to make sure that it does what it's supposed to do.

ga4gh/datamodel/__init__.py
jeromekelleher: @dcolligan --- lets not worry about thread safety. It's generally a lot easier to parallelise using processes in Python, and it tends to work better too. For a reference implementation I think we can say 1 thread, always.

dcolligan: We don't really have a good framework for using singleton objects, like this cache, throughout the code... now might be a good time to add some kind of IOC functionality?

Although, if usage of the cache is confined to just this module, that may not be necessary.  @jeromekelleher  your proposal seems fine to me, modulo thread-safety

jeromekelleher: I don't think we'll need to access this singleton outside the module, so this module-level variable should be fine. We can look at more sophisticated stuff later if the need arises.

eelsther:
Thanks to both of you for the feedback! I am currently doing some fixes taking into accounts your comments. It should be pushed on my branch later today.

ga4gh/datamodel/__init__.py
eelsther: Just  a quick question about modifying the cache size during configuration:
during the tests, this configuration part doesn't seem to happen so we still need a default value for maxCacheSize when the module is loaded... do you have any suggestion?

dcolligan: If you're testing the cache itself, some low value should suffice, like 3.  But this value should not be the default for production.  Change `serverconfig` to set the various values.

ga4gh/datamodel/__init__.py
```diff
@@ -26,6 +27,59 @@ def _cleanupHtslibsMess(indexDir):
         shutil.rmtree(indexDir)
 
 
+class PysamFileHandleCache(object):
+    """
+    Cache for opened file handles We use a deque which has the
```
jeromekelleher: Missing full-stops in this docstring?

ga4gh/datamodel/__init__.py
```diff
@@ -26,6 +27,59 @@ def _cleanupHtslibsMess(indexDir):
...
+    the "top" of the deque.
+    """
+
+    def __init__(self):
+        self._cache = collections.deque()
+        self._memoTable = dict()
+        # Initialize the value even if it will be set up by the config
+        self._maxCacheSize = 50
+
+    def setMaxCacheSize(self, size):
```
jeromekelleher: Docstring these public methods please.

ga4gh/datamodel/__init__.py
```diff
@@ -26,6 +27,59 @@ def _cleanupHtslibsMess(indexDir):
...
+        self._maxCacheSize = size
+
+    def add(self, dataFile, handle):
+        self._cache.appendleft((dataFile, handle))
+
+    def update(self, dataFile, handle):
+        self._cache.remove((dataFile, handle))
+        self.add(dataFile, handle)
+
+    def removeLRU(self):
```
jeromekelleher: Two things here:

1. Does adding this method make the `getFileHandle` method easier to understand? I would say, probably not.
2. If we do keep the method, we should change it to `removeLru` for consistency with existing conventions.

ga4gh/datamodel/__init__.py
```diff
@@ -26,6 +27,59 @@ def _cleanupHtslibsMess(indexDir):
...
+        (dataFile, handle) = self._cache.pop()
+        handle.close()
+        return dataFile
+
+    def getFileHandle(self, dataFile, openMethod):
```
jeromekelleher: It's not entirely clear what methods are part of the public interface and not here. The convention we've adopted is to have internal helper methods as private (i.e. `_removeLru`, and public methods should have docstring comments to document their usage).

ga4gh/datamodel/references.py
```diff
@@ -175,15 +175,18 @@ class HtslibReference(datamodel.PysamDatamodelMixin, AbstractReference):
     def __init__(self, id_, dataFile):
         super(HtslibReference, self).__init__(id_)
         self._fastaFilePath = dataFile
-        self._fastaFile = pysam.FastaFile(dataFile)
-        numReferences = len(self._fastaFile.references)
+        _fastaFile = self.openFile(dataFile)
```
jeromekelleher: s/_fastaFile/fastaFile

ga4gh/frontend.py
```diff
@@ -203,6 +203,9 @@ def configure(configFile=None, baseConfig="ProductionConfig", extraConfig={}):
     if configFile is not None:
         app.config.from_pyfile(configFile)
     app.config.update(extraConfig.items())
+    # Setup file handle cache max size
+    ga4gh.datamodel.fileHandleCache.setMaxCacheSize(
```
jeromekelleher: Change this to `datamodel.fileHandle...`, and add an import at the top for `import ga4gh.datamodel as datamodel`.

tests/unit/test_file_handle_cache.py
```diff
@@ -0,0 +1,52 @@
...
+from ga4gh import datamodel
+
+import unittest
+import tempfile
+import os
+
+import ga4gh.frontend as frontend
+
+
+class TestFileHandleCache(datamodel.PysamFileHandleCache, unittest.TestCase):
```
jeromekelleher: Why do we need to inherit from the `PysamFileHandleCache` here? I would have thought it would be simpler to just create some instances.

tests/unit/test_file_handle_cache.py
```diff
@@ -0,0 +1,52 @@
...
+Tests the file handle cache
+"""
+
+from ga4gh import datamodel
+
+import unittest
+import tempfile
+import os
+
+import ga4gh.frontend as frontend
```
jeromekelleher: Why do we need to import the frontend here? I think it would be better to test this class in isolation.

tests/unit/test_file_handle_cache.py
```diff
@@ -0,0 +1,52 @@
+"""
+Tests the file handle cache
+"""
+
+from ga4gh import datamodel
```
jeromekelleher: Please change this import to `import ga4gh.datamodel as datamodel`, and move to below the standard library imports. (See the STYLE.md file for details on this completely arbitrary convention)

tests/unit/test_file_handle_cache.py
```diff
@@ -0,0 +1,52 @@
...
+            return open(dataFile, 'w')
+        return self.getFileHandle(dataFile, openMethod)
+
+    def testGetFileHandle(self):
+        def removeFiles(l):
+            for f in l:
+                os.remove(f)
+
+        def genFileName(x):
+            return os.path.join(tempfile.gettempdir(),
```
jeromekelleher: I don't think we should be using a private method in tempfile. What about creating a `NamedTemporaryFile` instance and returning its name?

tests/unit/test_file_handle_cache.py
jeromekelleher: Ah, I see, we're doing a white-box test here... I think it would be better to add a public method like `getCachedFiles` or something to the `FileHandleCache` which returns the list of file names that are currently in the cache. This will let us test the external functionality without needing to access the internal state of the object, and coupling our test cases with the implementation.

tests/unit/test_file_handle_cache.py
```diff
@@ -0,0 +1,52 @@
...
+        fileHandles = map((lambda x: (x, self._getFileHandle(x))), fileList)
+
+        try:
+            # Ensure that the first added file has been removed from the cache
+            self.assertEquals(self._cache.count(fileHandles[0]), 0)
+            # Ensure that all other files are in the cache
+            for i in range(1, 10):
+                self.assertEquals(self._cache.count(fileHandles[i]), 1)
+
+            removeFiles(fileList)
```
jeromekelleher: Shouldn't this sort of cleanup be done in a `tearDown` method?

jeromekelleher:
This is looking good, thanks @eelsther. I think we need to make the test cases a bit more comprehensive though. We should test a few different scenarios of the cache, like what the cache looks like after adding each file. Do we remove the least recently used, every time? If we re-add the same file, is it treated like a new file? Do we raise an exception if we try to set a max cache size <= 0? What happens if we set the maxCacheSize to a value less than the current size of the cache?

tests/unit/test_file_handle_cache.py
eelsther: Yes,  that what I intended to do, but NamedTemporaryFile opens the file, and we want that to be done by getFileHandle. We are just interested here in the file names.
But I agree that it is not a very good workaround, I'm going to think of another solution.

jeromekelleher: Any reason you can't open a second handle to this via the file name? It might be a bit tricky if you're doing IO, but I think it'll probably work if you're just opening the file.

eelsther: Anyway, I just switched to using uuid for file names, as it is very unlikely that an already existing  temporary file would have the same name.

ga4gh/datamodel/__init__.py
```diff
@@ -26,6 +27,90 @@ def _cleanupHtslibsMess(indexDir):
...
+        self._maxCacheSize = 50
+
+    def setMaxCacheSize(self, size):
+        """
+        Sets the maximum size of the cache
+        """
+        if size <= 0:
+            raise ValueError(
+                "The size of the cache must be a strictly positive value")
+        else:
```
jeromekelleher: We don't need the `else` here.

tests/unit/test_file_handle_cache.py
```diff
@@ -0,0 +1,58 @@
...
+        self.assertEquals(self._cache[topIndex][0], fileList[1])
+        self._getFileHandle(fileList[1])
+        self.assertNotEqual(self._cache[topIndex][0], fileList[1])
+        self.assertEquals(self._cache[0][0], fileList[1])
+
+    def testSetCacheMaxSize(self):
+        self.assertRaises(ValueError, self.setMaxCacheSize, 0)
+        self.assertRaises(ValueError, self.setMaxCacheSize, -1)
+
+    def tearDown(self):
```
jeromekelleher: I'm slightly worried about this leaking files into /tmp --- don't we leave one file without deleting it at the moment? How about this pattern. In `setUp` we create a temporary directory using tempdir.mkdtemp, which we store as `self.tmpDir`. We can then create as many files as we like within this directory (which we can then just name sequentially I guess) within tests. Then, `tearDown` calls `shutil.rmtree(self.tmpDir)`, and we are guaranteed not to leak any files, regardless of whether tests succeed or fail.

jeromekelleher:
Very clear, thanks @eelsther. I think we're ready to merge, modulo my minor quibbles above. Can you squash the commits down once you've addressed these issues please?

tests/unit/test_file_handle_cache.py
eelsther: Yes good idea, this is indeed clearer to create a dedicated temp dir for that and just delete it afterwards.

eelsther:
Hi Jerome,
Here are the squashed commits. It also contains some commits from the develop branch because I merged the upstream branch regularly to stay up-to-date 
Oddly, I had to do a merge from my own branch before pushing this commit... 

Cheers,
Esther

tests/unit/test_file_handle_cache.py
```diff
@@ -0,0 +1,62 @@
+"""
+Tests the file handle cache
+"""
```
jeromekelleher: Sorry, should have spotted this earlier. We're missing some standard  `__future__` imports here. These are essential for future Python 3 compatibility.

jeromekelleher:
We need to squash this down to one commit  @eelsther. This is done by rebasing against upstream/develop. See http://ga4gh-reference-implementation.readthedocs.org/en/develop/development.html#github-workflow for a quick overview.

There's one small issue I spotted above, which should be resolved also. After that, we're good to go I think!

eelsther:
Sorry, I should have read this part of the documentation before doing anything!
I have now done the rebase against the develop branch, following the instructions in the guide.
Please let me know if there is anything left to be done in terms of rebase/merge (I'm not sure if what I've done is enough in order for the branch to be merged).

jeromekelleher:
Excellent! Thanks @eelsther, I think we're ready to merge. @dcolligan, can you push the green button if you're happy with this please?

+1.

Issue #453, opened 2015-06-16
Datasets not found correctly by _getDatasetFromRequest


jeromekelleher:
Currently, we try to get a dataset for all incoming search requests via the `_getDatasetFromRequest` method. This works fine for request where a `datasetIds` attribute is included, but is broken for others (searchVariants for example). Currently, if we have more than one dataset, the searchVariants method fails mysteriously. There are no plans to include a datasetIds attribute in the `searchVariants` method, so the premise of this function is flawed (which I should have caught in review; my bad).

When we have a dataset attribute, for example in readGroupSets generator, we can use `_getDatasetFromRequest`, but removing the check to see if it has the `datasetIds` attribute (as it shouldn't be called on any request object that doesn't have a datasetIds attribute).

When we do not have a dataset attribute (such as in searchVariants), we need some method of finding the dataset that a particular variant set belongs to; there are two basic options I think:

1. Maintain a map of variantSetIds to datasets in the backend; or 
2. Make a hierarchical ID scheme such that we can always derive the datasetId from a variantSetId. For example, if we make variantSetIds in the form `{dataSetId}:{localVariantSetId}`, we can decompose the datasetId quickly, allowing us to find the relevant dataset.

I think (2) is the best and simplest option. We'll need to revise the ID generation code throughout the datamodel code, and systematise it. This is overdue though.

dcolligan:
I like (1) since it requires less changes to the wire protocol.  Plus, this should only be a temporary measure until we bring the schemas up to date, right? (Which, I know, could be a long time...)  (1) also only works if we can assume duplicate variant set ids are not present, which I think is a fair assumption.

jeromekelleher:
I don't think there are any changes to the schema in the pipelines that would make any difference to this @dcolligan. We're not changing the wire protocol in any way: ids are server internal values, which are opaque to users. We can do whatever we want with them. 

I think it's a bit onerous in any case to require that all the variant set IDs (== directory name) within different data sets are different. Suppose I have a hundred datasets, and each of these contains 10 variant sets. I would like to put these in directories called vs1, vs2, ..., vs10 within their respective datasets. If we leave IDs as they are, then I can't do this and I have to invent unique names for variant sets across all datasets. I've already started hitting on this, and it's a pain. I want to be able to move datasets in and out by just mv'ing them. By requiring globally unique directory names for variantsets I end up having to do lots of renaming, which is very tedious.

dcolligan:
[...]

PR #474, opened 2015-07-07
Use compoundIds for variantSets and readGroupSets

dcolligan:
A compound id is the former id plus the dataset id it belongs to

Issue #453

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -222,18 +222,39 @@ class AbstractBackend(object):
         self._datasetIdMap = {}
         self._datasetIds = []
 
+    def _getDatasetFromReadsRequest(self, request):
+        if len(request.readGroupIds) != 1:
+            raise exceptions.NotImplementedException(
+                "Exactly one read group id must be specified")
+        compoundId = request.readGroupIds[0]
+        dataset = self._getDatasetFromCompoundId(compoundId)
+        return dataset
+
+    def _getDatasetFromVariantsRequest(self, request):
+        variantSetIds = request.variantSetIds
+        if len(variantSetIds) != 1:
+            raise exceptions.NotImplementedException(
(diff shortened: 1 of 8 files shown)


ga4gh/backend.py
```diff
@@ -222,18 +222,41 @@ def __init__(self):
...
+        return dataset
+
+    def _getDatasetFromCompoundId(self, compoundId):
+        splits = compoundId.split(':')
+        datasetId = splits[0]
+        dataset = self._datasetIdMap[datasetId]
+        return dataset
+
     def _getDatasetFromRequest(self, request):
         if hasattr(request, "datasetIds"):
```
jeromekelleher: It would be better to just have 
```python
assert hasattr(request, "datasetIds")
```
at the top of the function.

ga4gh/backend.py
```diff
@@ -222,18 +222,41 @@ def __init__(self):
...
+            raise exceptions.NotImplementedException(
+                "Exactly one variant set id must be specified")
+        compoundId = request.variantSetIds[0]
+        dataset = self._getDatasetFromCompoundId(compoundId)
+        return dataset
+
+    def _getDatasetFromCompoundId(self, compoundId):
+        splits = compoundId.split(':')
+        datasetId = splits[0]
+        dataset = self._datasetIdMap[datasetId]
```
jeromekelleher: Shouldn't we check if the datasetId exists and raise a NotFoundException here? Also, we should add some test cases to test for this situation.

jeromekelleher:
Looks good, thanks @dcolligan. I think we need to handle the error case a bit better, but looks great otherwise.

dcolligan:

jeromekelleher:
Excellent; thanks @dcolligan.

+1.

dcolligan:
@jeromekelleher merge me!

Issue #472, opened 2015-07-06
Fix 'operations available' part of html display


dcolligan:
The 'operations available' part of the html display doesn't make sense anymore because

1) The `@app.route` arguments are no longer literal URLs, so the html display ends up displaying weird flask routing syntax
2) Not all of the routes tagged with an `@app.route` decorator are operational; many are not yet implemented

Ideally, we should create another decorator that wraps `@app.route` that can take two additional, optional arguments:

1) a string representing the URL to display
2) a flag indicating if the route is not yet implemented

dcolligan:
Fixed in #475

PR #475, opened 2015-07-07
Only display live routes on html page

dcolligan:
Issue #472 

'Operations available' section now looks like:
```
Method	Path
GET	/0.6.e6d6074/references/<id>
GET	/0.6.e6d6074/references/<id>/bases
GET	/0.6.e6d6074/referencesets/<id>
POST	/0.6.e6d6074/callsets/search
POST	/0.6.e6d6074/datasets/search
POST	/0.6.e6d6074/readgroupsets/search
POST	/0.6.e6d6074/reads/search
POST	/0.6.e6d6074/references/search
POST	/0.6.e6d6074/referencesets/search
POST	/0.6.e6d6074/variants/search
POST	/0.6.e6d6074/variantsets/search
```

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -11,6 +11,7 @@ import os
 import datetime
 import socket
 import urlparse
+import functools
 
 import flask
 import flask.ext.cors as cors
@@ -32,6 +33,8 @@ SEARCH_ENDPOINT_METHODS = ['POST', 'OPTIONS']
 SECRET_KEY_LENGTH = 24
 
 app = flask.Flask(__name__)
+assert not hasattr(app, 'urls')
+app.urls = []
 
 
 class NoConverter(werkzeug.routing.BaseConverter):
@@ -156,22 +159,8 @@ class ServerStatus(object):
         Returns the list of (httpMethod, URL) tuples that this server
         supports.
         """
-        urls = []
-        rules = app.url_map.iter_rules()
-        excluded_methods = ('OPTIONS', 'HEAD')
-        excluded_rules = (
-            '/', '/flask-api/static/<path:filename>',
-            '/static/<path:filename>')
-        version = Version.getVersionForUrl(protocol.version)
-        for rule in rules:
-            for method in rule.methods:
-                if (method not in excluded_methods and
-                        rule.rule not in excluded_rules):
-                    versionedRule = rule.rule.replace(
-                        "<version>", version)
-                    urls.append((versionedRule, method))
-        urls.sort()
-        return urls
+        app.urls.sort()
+        return app.urls
 
     def getDatasetIds(self):
         """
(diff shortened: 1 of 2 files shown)


ga4gh/frontend.py
```diff
@@ -25,6 +26,7 @@
 
 
 app = flask.Flask(__name__)
+urls = []
```
jeromekelleher: No so sure about this. Do we _really_ need a global variable to do this? If we do, then I would all it `_urls` to ensure that it's module private.

ga4gh/frontend.py
```diff
@@ -359,67 +382,67 @@ def indexRedirect(version):
         raise exceptions.PathNotFoundException()
 
 
-@app.route('/<version>/references/<id>')
+@DisplayedRoute('/<version>/references/<id>')
```
jeromekelleher: Don't decorators normally start with a lowercase letter?

jeromekelleher:
I'm a bit ambivalent about this one. It seems to me we're adding a lot of complexity to the frontend to support some fairly minor functionality. Does the admin really need to know what URL endpoints are supported? We only need this because we're not fully implementing the spec, but that's changing quickly. Can't we just say we're implementing the spec at version x (when x is fixed, which I agree is a problem at the moment), and leave it at that?

ga4gh/frontend.py
dcolligan: Not if they're a class

ga4gh/frontend.py
dcolligan: Unless there's some other way of sharing information between different parts of `frontend` that I'm not thinking of... 

I'll try and attach the information to some part of the `app` variable, but that may or may not work.

dcolligan:
I think having the right routes reflected to the admin is a really helpful debugging tool for our users.  We might not be fully implementing the spec for a while (whatever the spec ends up being!).  Let's give our users the clearest idea of what works and what doesn't in the mean time; I think that's important.

I also like the decorator abstraction.  It cleans up a bit of the repetition in the decorators on the routes and gives us some more flexibility in what we want to do.

dcolligan:
@jeromekelleher pushed change that eliminated the global variable

ga4gh/frontend.py
```diff
@@ -32,6 +33,7 @@
 SECRET_KEY_LENGTH = 24
 
 app = flask.Flask(__name__)
```
jeromekelleher: Just in the interest of paranoia, can we put in an `assert not hasattr(app, 'urls')` here? This is worryingly close to something Flask might use itself, so I don't want to clobber this, now or in the future.

jeromekelleher:
OK, I buy these arguments @dcolligan. Thanks for clearing out the global variable. I'm +1 modulo my paranoia above.

dcolligan:
@jeromekelleher pushed an update with the assert

jeromekelleher:
+1. Merging.

PR #476, opened 2015-07-07
Converted identifiers to snake case

hjellinek:
At Jerome Kelleher's suggestion, I have worked my way out from the ```_protocol_definitions.py``` file and converted from camelCase to snake_case all of the identifiers that are (1) used in the bodies of requests or responses, and (2) touched by the unit tests.  This is a significant chunk of the camelCase identifiers in the code, though not 100% of them.

All of the unit tests that succeeded in the unmodified snake_case branch succeed after I made these changes.

(BTW, this is my first contribution to the code, so in my ignorance I may have broken something - for instance, the docs. Let me know if so.)


docs/source/demo.rst
```diff
@@ -95,7 +95,7 @@ equivalent command using:
 .. code-block:: bash
 
     $ source ga4gh-env/bin/activate
-    (ga4gh-env) $ ga4gh_client readgroupsets-search --datasetIds XXXX http://localhost:8000/current
+    (ga4gh-env) $ ga4gh_client readgroupsets-search --dataset_ids XXXX http://localhost:8000/current
```
jeromekelleher: I think these command line arguments might be hyphenated now.  I think argparse might change these on the fly. Can you check if this works now? I don't have a strong opinion about which way it should be for CLI stuff, but I'd like to check if these examples still work.

ga4gh/backend.py
```diff
@@ -70,12 +70,13 @@ def __init__(self, request, containerIdMap):
...
         self._searchAnchor = None
         self._distanceFromAnchor = None
-        if request.pageToken is None:
+        if request.page_token is None:
             self._initialiseIteration()
         else:
             # Set the search start point and the number of records to skip from
             # the page token.
-            searchAnchor, objectsToSkip = _parsePageToken(request.pageToken, 2)
+            searchAnchor, objectsToSkip = \
```
jeromekelleher: Hanging indent is preferable to splitting with a backslash here. From PEP8:

>The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

So, here we would have
```python
searchAnchor, objectsToSkip = _parsePageToken(
    request.page_token, 2)
```

ga4gh/backend.py
```diff
@@ -287,28 +289,28 @@ def listReferenceBases(self, id_, requestArgs):
             except ValueError:
                 raise exceptions.BadRequestIntegerException(
                     'end', endString)
-        if 'pageToken' in requestArgs:
-            pageTokenStr = requestArgs['pageToken']
-            start = _parsePageToken(pageTokenStr, 1)[0]
+        if 'page_token' in requestArgs:
+            page_tokenStr = requestArgs['page_token']
```
jeromekelleher: `page_token_str`?

ga4gh/cli.py
```diff
@@ -644,14 +644,14 @@ def addMd5ChecksumsArgument(parser):
 
 def addPageSizeArgument(parser):
     parser.add_argument(
-        "--pageSize", "-m", default=100, type=int,
+        "--page_size", "-m", default=100, type=int,
         help="The maximum number of results returned in one response.")
 
 
-def addDatasetIdsArgument(parser):
+def adddataset_idsArgument(parser):
```
jeromekelleher: This has gone weird here --- I would back out this change until later.

ga4gh/datamodel/__init__.py
```diff
@@ -70,9 +70,10 @@ def sanitizeVariantFileFetch(cls, contig=None, start=None, stop=None):
 
     @classmethod
     def sanitizeAlignmentFileFetch(
-            cls, referenceName=None, start=None, end=None):
-        if referenceName is not None:
-            referenceName = cls.sanitizeString(referenceName, 'referenceName')
+            cls, reference_name=None, start=None, end=None):
+        if reference_name is not None:
+            reference_name = cls.sanitizeString(reference_name,
```
jeromekelleher: I usually prefer hanging indent to vertical alignment as it's more robust to code changes. Here for example, we'll need to realign the arguments after we snake_case the rest of the code, but with a hanging indent we won't need to touch the arguments.

jeromekelleher:
Great work, thanks @hjellinek! This is very useful, and I think we should merge this much onto the snake_case branch as soon as the minor kinks pointed out above are worked out. There are two directions we can go from here:

1) Leave this as it is for now, and try to keep it up to date by merging back in develop periodically.
2) Go for the full conversion now, and try to merge back in changes to develop as they happen.

I'm leaning towards (1), as what we have now should be enough to do a good protobuf evaluation.

docs/source/demo.rst
dcolligan: In general I think command line arguments should be hyphenated instead of underscored... that seems more the UNIX way...

jeromekelleher: I agree, but this should be something we deal with in a follow-up though; if this works as it is, I'd leave it in the interest of keeping this (mammoth!) patch as simple as we can.

dcolligan: +1 for incremental development on global renaming pushes

hjellinek: I agree that command line arguments should be hyphenated - using underscores looks weird - but I was concerned that the parameter names (sans leading '--') might be used as object attribute names.  Danny, does your +1 mean you agree with Jerome (deal with it in a follow-up), or the opposite?

ga4gh/backend.py
hjellinek: Fixed.  Will be in my next push.

ga4gh/backend.py
hjellinek: That one got away from me!  Fixed.  Will be in my next push.

ga4gh/cli.py
hjellinek: Fixed. Will push.

ga4gh/datamodel/__init__.py
hjellinek: Fixed it.  Will be in my next push.

docs/source/demo.rst
dcolligan: Sorry for the ambiguity; it means I agree with Jerome

hjellinek: Thanks - I thought that's what you meant, but wanted to be 100% sure.

hjellinek: In reply to Jerome's original comment/question about args possibly being hyphenated now, I switched to the develop branch and tried

```
$ python client_dev.py variantsets-search --datasetIds dataset1 http://localhost:8000/current
```

which worked, unsurprisingly, and

```
$ python client_dev.py variantsets-search --dataset-ids dataset1 http://localhost:8000/current
```

which failed.  So, if I understood your note correctly and am not over-generalizing from a single example, hyphenated args (`dataset-ids` for `datasetIds`) do not work.

jeromekelleher: Sorry, for confusing things. There are two points here I guess:

1) Does `readgroupsets-search --dataset_ids XXXX` work on this branch? I was concerned that argparse might have automatically changed these underscores to hyphens.
2) Should we change `--dataset_ids` to ` --dataset-ids` for POSIX compliance?

I think the answer to (2) is a clear, "yes, but later". I'm still not sure about (1) though. Do these examples work on your branch @hjellinek?

jeromekelleher:
Excellent stuff @hjellinek. I think modulo my question about the CLI arguments still working, we should squash these commits and merge.

+1.

dcolligan:
+1 after the sqash

what is the plan for the `snake_case` branch?

jeromekelleher:
The plan for the `snake_case` branch currently is to do a decent evaluation of protobuf based on this. If we do changeover to protobuf, this will be the basis of a full snake_case conversion, which will ultimately get mapped back to develop.

docs/source/demo.rst
hjellinek: Yes, the `--dataset_ids` parameter works when running under the `snake_case` branch:
```
$ python client_dev.py variantsets-search --dataset_ids dataset1 http://localhost:8000/current
1kg-phase1
1kg-phase3
$
```

jeromekelleher:
Merging; thanks for the surgical precision @hjellinek!

PR #477, opened 2015-07-07
Build docs as part of CI and run_tests.py

dcolligan:
Also fix error in docs wrt red hat setup

Strangely, the `make` command doesn't fail when the error was present, but it does emit the following:
```
...
reading sources... [100%] introduction
/Users/dannyc/ga4gh/server/docs/source/demo.rst:34: ERROR: Error in "code-block" directive:
maximum 1 argument(s) allowed, 7 supplied.

.. code-block:: bash
    $sudo dnf install python-devel python-virtualenv zlib-devel
looking for now-outdated files... none found
...
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,3 +23,5 @@ script:
   - coverage erase
   - nosetests --with-coverage --cover-package ga4gh
               --cover-inclusive --cover-min-percentage 85
+  - make clean -C docs/source/
+  - make -C docs/source/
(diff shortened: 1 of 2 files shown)


jeromekelleher:
That's weird --- I thought that CI would fail when we had bad docs inserted. If it's not we should make an issue and fix it, as it's very tedious keeping track of docs as they go in otherwise.

Good catch @dcolligan, +1.

PR #485, opened 2015-07-09
Allow config to set a custom backend

danvk:
I have a bunch of BAM files sitting on a file system and I wanted to serve them using the GA4GH API without reorganizing everything to fit the directory structure it expects. My BAM files may not have a `.bam` extension, and they may appear after the server starts up. So the plain vanilla `FileSystemBackend` isn't a great fit.

I only care about the `/reads/search` endpoint, so I hacked up a [new backend][1] and wanted to plug it in. This was the simplest way I could come up with. Any other ideas?

Thanks!

[1]: https://gist.github.com/danvk/0aa59b96de00213b5a27

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -225,6 +225,8 @@ def configure(configFile=None, baseConfig="ProductionConfig",
             numReferenceSets, numReferencesPerReferenceSet, numAlignments)
     elif dataSource == "__EMPTY__":
         theBackend = backend.EmptyBackend()
+    elif isinstance(dataSource, backend.AbstractBackend):
+        theBackend = dataSource
     else:
         theBackend = backend.FileSystemBackend(dataSource)
     theBackend.setRequestValidation(app.config["REQUEST_VALIDATION"])


jeromekelleher:
This is very interesting @danvk! I'm sure we can come up with a solution that'll meet your requirements.

One suggestion I would have for the first part, is that it might be simplest to create some symlinks to the files you want to serve, that would follow the input requirements we currently have. This would involve some scripting on your part, but would work pretty well and have minimal overhead in terms of extra disk space used.

The second part is trickier, and would need some new code. It shouldn't be particularly difficult to do.  We would need to precisely spec out what we want the server to do, however, as things get a little tricky when we have mutable data.

danvk:
If the second part (picking up new files) happens, we might be able to make the proposed solution (generate symlinks) work.

My use case is viewing alignments in the browser. I'm currently using [pileup.js][1] to view BAM files, which it fetches piecemeal using HTTP range requests. The thought was that I could run `samtools` on the backend, to do all the seeking and parsing on a faster machine that's closer to the disks. That's essentially what the GA4GH server is.

My hack makes it a drop-in replacement for the existing file system access over HTTP. Otherwise I'll have to require users to put symlinks in place whenever they create a new BAM that they'd like to interact with.

Maybe I should be building a wrapper around the server rather than adding a hook to plug in a custom backend?

[1]: https://github.com/hammerlab/pileup.js

jeromekelleher:
This is very interesting @danvk. I'm not against the idea of adding your hook in to allow people to use their own backends, and I'm sure we can find a way to update the server to suit your use case. I'm still not quite clear on how you want to use it though --- would you mind talking me through a typical user session, what they want to do and how they would go about doing it?

jeromekelleher:
Can we close this PR @danvk, or is this still something you would like to incorporate?

danvk:
Feel free to close it. I found a way to embed the ga4gh server in my app
instead.

jeromekelleher:
OK, thanks @danvk. Any pointers to your app? I'd be very curious to see what you've done!

danvk:
Yep, check out https://github.com/hammerlab/cycledash/pull/833

Issue #487, opened 2015-07-10
Requests with invalid references produce Internal Server Error


danvk:
For example, if you're trying to request alignments from a BAM file with `chr20` as the reference, not `20`, you get:

```
$ curl .../v0.5.1/reads/search --data '{"readGroupIds":["foo:bar"],"referenceName":"20","start":144700,"end":144800}' | jq .
{
  "errorCode": 3424984409,
  "message": "Internal Server Error"
}
```

You have to look at the server logs to see the real error:

```
Exception on /v0.5.1/reads/search [POST]
--------------------------------------------------------------------------------
Traceback (most recent call last):
...
  File "pysam/calignmentfile.pyx", line 641, in pysam.calignmentfile.AlignmentFile._parseRegion (pysam/calignmentfile.c:8888)
ValueError: invalid reference `20`
```

It would be better to include the real error in the response:

```
$ curl .../v0.5.1/reads/search --data '{"readGroupIds":["foo:bar"],"referenceName":"20","start":144700,"end":144800}' | jq .
{
  "errorCode": 3424984409,
  "message": "invalid reference `20`"
}
```

(somewhat related to #486)

jeromekelleher:
This is true. @dcolligan, is there some interface in pysam to allow us to test if a reference is valid or not? We should try this before sending the request down to htslib as part of our sanitisation steps, as this will allow us to make a good error message.

dcolligan:
Thanks for the bug report, @danvk ; I'm looking into it.  I was able to reproduce this bug with our example data with this call:
```
curl http://localhost:8000/v0.5.1/reads/search --header 'Content-Type: application/json' --data '{"readGroupIds":["low-coverage:HG00096.mapped.ILLUMINA.bwa.GBR.low_coverage.20120522"],"referenceName":"chr20","start":144700,"end":144800}'
```

dcolligan:
A fix has been merged into the `release-0.1` branch and will be included in an upcoming release.

dcolligan:
We've released a new version; closing...

PR #490, opened 2015-07-13
Improve error on read query with invalid refName

dcolligan:
Issue #487 

Behavior after this patch:
```
$ curl http://localhost:8000/v0.5.1/reads/search --header 'Content-Type: application/json' --data '{"readGroupIds":["low-coverage:HG00096.mapped.ILLUMINA.bwa.GBR.low_coverage.20120522"],"referenceName":"chr20","start":144700,"end":144800}'
{"errorCode": 1344119189, "message": "referenceName 'chr20' not found in readGroupId 'low-coverage:HG00096.mapped.ILLUMINA.bwa.GBR.low_coverage.20120522'"}
```

### DIFF
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -235,7 +235,11 @@ class HtslibReadGroup(datamodel.PysamSanitizer, AbstractReadGroup):
             referenceName = self._samFile.getrname(referenceId)
         referenceName, start, end = self.sanitizeAlignmentFileFetch(
             referenceName, start, end)
-        # TODO deal with errors from htslib
+        if referenceName is not None:
+            if self._samFile.gettid(referenceName) == -1:
+                raise exceptions.ReferenceNameNotFoundException(
+                    referenceName, self.getId())
+        # TODO deal with other errors from htslib
         readAlignments = self._samFile.fetch(referenceName, start, end)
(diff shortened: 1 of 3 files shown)


dcolligan:
Er, this is another PR against `stable`, so don't merge it in!  Should we start another release branch?

jeromekelleher:
Here's what I'd suggest:

1) Reopen this PR against `develop`. All new things (unless they don't apply to `develop` at all) should be made as PRs to develop.
2) When we decide we want to make a new release, we cherry pick the commits in question back to `release-0.1`.
3) After we've made the release, we merge `release-0.1` on to stable. 

The patch looks perfect by the way @dcolligan, nice work!

dcolligan:
This bug isn't present in develop since the API has changed.  Only a `referenceId` can be submitted, no longer a `referenceName`.  If `referenceId` is outside the range [0, 85], there is a `DatamodelValidationException` thrown.  If it is inside that range, but there are no reads for that reference, pysam returns an empty list.

Should we put this change into the `stable` branch at some point and do a bugfix release?

jeromekelleher:
Ah, I see. Well, we should just make the PR against release-0.1 then, and this change will ready for when we are ready to cut a new release. I think we should never make changes to stable except to merge in a just-made release.

Issue #496, opened 2015-07-15
Correctly advertise lack of unmapped reads support


jeromekelleher:
We currently don't support unmapped reads, but instead of throwing an error we silently omit them. See issue #287 for a full discussion.

As a temporary fix to at least identify our lack of support, we should raise a `NotImplemented` exception when a query is made without specifying a reference.

dcolligan:
Fixed in #497

Issue #492, opened 2015-07-13
TypeError: 'NoneType' object is not iterable


danvk:
(forked from [this comment][1])

```
python client_dev.py -vv reads-search http://localhost:8000/current --readGroupIds=dataset2:danvk-bug-report:chr17.1-250
```

```
--------------------------------------------------------------------------------
ERROR in app [/home/jk/.local/lib/python2.7/site-packages/flask/app.py:1423]:
Exception on /current/reads/search [POST]
--------------------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jk/.local/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/jk/.local/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/jk/work/github/server/ga4gh/frontend.py", line 521, in searchReads
    version, flask.request, app.backend.searchReads)
  File "/home/jk/work/github/server/ga4gh/frontend.py", line 431, in handleFlaskPostRequest
    return handleHttpPost(flaskRequest, endpoint)
  File "/home/jk/work/github/server/ga4gh/frontend.py", line 297, in handleHttpPost
    responseStr = endpoint(request.get_data())
  File "/home/jk/work/github/server/ga4gh/backend.py", line 404, in searchReads
    self.readsGenerator)
  File "/home/jk/work/github/server/ga4gh/backend.py", line 376, in runSearchRequest
    for obj, nextPageToken in objectGenerator(request):
  File "/home/jk/work/github/server/ga4gh/backend.py", line 148, in next
    self._nextObject = next(self._searchIterator, None)
  File "/home/jk/work/github/server/ga4gh/datamodel/reads.py", line 246, in getReadAlignments
    yield self.convertReadAlignment(readAlignment)
  File "/home/jk/work/github/server/ga4gh/datamodel/reads.py", line 255, in convertReadAlignment
    ret.alignedQuality = list(read.query_qualities)
TypeError: 'NoneType' object is not iterable
127.0.0.1 - - [13/Jul/2015 11:25:52] "POST /current/reads/search?key=invalid HTTP/1.1" 500 -
```

[1]: https://github.com/ga4gh/server/issues/488#issuecomment-120886728

dcolligan:
Fixed in #497

PR #497, opened 2015-07-15
Fix various things with reads/search

dcolligan:
Issue #492
Issue #496


ga4gh/datamodel/reads.py
```diff
@@ -232,15 +233,15 @@ def getReadAlignments(self, referenceId=None, start=None, end=None):
...
-            referenceName = samFile.getrname(referenceId)
-        referenceName, start, end = self.sanitizeAlignmentFileFetch(
-            referenceName, start, end)
+        referenceId, start, end = self.sanitizeAlignmentFileFetch(
+            referenceId, start, end)
         # TODO deal with errors from htslib
-        readAlignments = samFile.fetch(referenceName, start, end)
+        try:
+            readAlignments = samFile.fetch(referenceId, start, end)
+        except ValueError:
```
jeromekelleher: Hmmm --- is ReferenceNotFound the _only_ reason we could ever get a ValueError here? I think all errors coming up from Pysam are ValueErrors, so this doesn't really distinguish. Isn't there some way we can test if the referenceId exists before we run the query? This way we know what the problem is and can raise the correct exception.

tests/end_to_end/test_remote_fetch.py
```diff
@@ -42,6 +45,11 @@ def _assertLogsWritten(self):
         # client should have gotten some reads back
         self.assertEqual(len(self.client.getOutLines()), 10)
 
+    def tearDown(self):
+        # clean up downloaded local index file
+        for indexFile in glob.glob("*.bai"):
```
jeromekelleher: Where do these get written? Shouldn't this be managed by the `cleanupHtslibMess` in datamodel?

tests/unit/test_client.py
```diff
@@ -167,12 +166,33 @@ def testRunSearchRequest(self):
...
             # verify requests.post called correctly
+            httpMethod = 'POST'
             url = "http://example.com/referencesets/search"
-            jsonString = protocolRequest.toJsonString()
+            data = protocolRequest.toJsonString()
             headers = {"Content-type": "application/json"}
-            httpMethod = 'POST'
-            mockPost.assert_called_twice_with(
-                httpMethod, url, jsonString, headers=headers, verify=False)
+            params = {u'key': u'KEY'}
```
jeromekelleher: This new block and the next one look almost identical. Would it be worth making a helper method to do this checking?

jeromekelleher:
Looks great, thanks @dcolligan. I think the error thing is important, but the rest of the comments are just minor nits.

tests/end_to_end/test_remote_fetch.py
dcolligan: they get written in the current directory.  let's put this in as a stop-gap and then file another issue to take care of this for real.

ga4gh/datamodel/reads.py
dcolligan: yeah, good point.  the remedy will involve moving some pysam-like logic up into our code, but I guess the alternative (catching a generic ValueError) is worse.

tests/end_to_end/test_remote_fetch.py
dcolligan: I've filed #500 ... I'll take these out in the mean time

tests/unit/test_client.py
dcolligan: Sure.  FYI, these lines needed to be changed after I did a library update.  Something with the interaction of requests / mock / something else?

dcolligan:
@jeromekelleher pushed an updated commit

tests/unit/test_client.py
jeromekelleher: Ah right, now I see. Maybe we should start pinning versions by default for libs --- this sort of thing is very disruptive.

jeromekelleher:
Excellent stuff, thanks @dcolligan. +1

dcolligan:
@jeromekelleher cool, why don't we merge this now since develop is broken without it (due to the library version issue)

PR #499, opened 2015-07-18
Authorisation support

jbu:
Not ready for merging yet, but posting to start discussion.

Authentication can be managed through serverconfig.py (see example).

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -282,6 +282,7 @@ def configure(configFile=None, baseConfig="ProductionConfig",
                 redirect_uris=[redirectUri],
                 verify_ssl=False)
             app.oidcClient.store_registration_info(response)
+        app.permissions = app.config['PERMISSIONS']
 
 
 def getFlaskResponse(responseString, httpStatus=200):
@@ -359,7 +360,7 @@ def startLogin():
     args = {
         "client_id": app.oidcClient.client_id,
         "response_type": "code",
-        "scope": ["openid", "profile"],
+        "scope": ["openid", "profile", "email"],
         "nonce": flask.session["nonce"],
         "redirect_uri": app.oidcClient.redirect_uris[0],
         "state": flask.session["state"]
@@ -376,6 +377,8 @@ def checkAuthentication():
     The request will have a parameter 'key' if it came from the command line
     client, or have a session key of 'key' if it's the browser.
     If the token is not found, start the login process.
+    If we are requesting particular datasets, then check that the user
+    has access to those datasets. If not, raise an exception.
 
     If there is no oidcClient, we are running naked and we don't check.
     If we're being redirected to the oidcCallback we don't check.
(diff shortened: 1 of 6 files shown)


ga4gh/serverconfig.py
```diff
@@ -44,7 +46,9 @@ class LocalOidConfig(DevelopmentConfig):
     """
     Configuration used for developing against a local OIDC server
     """
-    OIDC_PROVIDER = "https://localhost:8443"
+    OIDC_PROVIDER = "https://{}:8443".format(socket.gethostname())
+    PERMISSIONS = {'diana@example.org': ['d1'],
```
jeromekelleher: I think `PERMISSIONS` must be empty by default. It's fine having TestConfig and (say) DevelopmentConfig having some hard-coded permissions to make our lives easier, but in general we need to think of users actually filling out a config file.

ga4gh/frontend.py
```diff
@@ -397,6 +400,17 @@ def checkAuthentication():
         else:
             return startLogin()
 
+    jsonBody = flask.request.get_json() or {}
```
jeromekelleher: This is a definite break in layering, which is a bad idea. What we have here is fine for a proof of concept, but it's too easy to break this for real use. I'll outline a different approach that we might use below though.

jeromekelleher:
This is very interesting @jbu! I like the outline. I have a basic issue with how we're doing the actual authorisation though: I think it should be done in the backend, rather than the frontend. I think we should split this out into a separate module, called ga4gh.authz, which has a class called (say) `PermissionsManager`. In `frontend.config`, we do something like
```
permissionsManager = authz.PermissionsManager(app.config["PERMISSIONS"])
theBackend.setPermissionsManager(permissionsManager)
```
Then, in the backend, we check for authorisation at each of the entry points for the various methods. For example, in getVariantSet we might have:
```
def getVariantSet(self, id_, userId):
     dataset = self._getDatasetFromCompoundId(id_)
     self._permissionsManager.checkPermissions(userId, dataset)
     return self.runGetRequest(dataset.getVariantSetIdMap(), id_)
```
The trouble is, there's a bunch of different ways in which we can get the dataset for different queries. The only way we can be sure of getting the authorisation done properly is by doing it after we've parsed the incoming requests and we know its context. I think the backend is probably the right place for this.

diekhans:
Authorization often needs to be done a different levels of the
data hierarchy depending on the project.  Authorization at only
the DataSet level would not work for TCGA, where different files
have different access restrictions and this even changed over
the course of the project.

One can image cases where access restrictions could apply to
ranges of BAM files.  Say access to certain genes is not as
restricted as non-coding regions.

IMHO, authorization will have to permeate down into the depths
of back end eventually.

jeromekelleher:
I agree @diekhans --- I think the approach I've outlined above should provide the flexibility we need. Part of idea of farming the real work out to the new authz module is to make room for these more complex situations.  By making the backend delegate the actual checking to the authz module, hopefully we'll be able to keep things nice and clean, and also make it all testable.

jeromekelleher:
Closing this PR due to inactivity.

Issue #484, opened 2015-07-09
Docker problem: libxml extension failing


bwalsh:
Attempted to build docker image on macos

```
    copying src/lxml/isoschematron/resources/xsl/iso-schematron-xslt1/readme.txt -> build/lib.linux-x86_64-2.7/lxml/isoschematron/resources/xsl/iso-schematron-xslt1
    running build_ext
    building 'lxml.etree' extension
    creating build/temp.linux-x86_64-2.7
    creating build/temp.linux-x86_64-2.7/src
    creating build/temp.linux-x86_64-2.7/src/lxml
    x86_64-linux-gnu-gcc -pthread -fno-strict-aliasing -g -O2 -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -I/tmp/pip-build-cWMb1G/lxml/src/lxml/includes -I/usr/local/python/include/python2.7 -c src/lxml/lxml.etree.c -o build/temp.linux-x86_64-2.7/src/lxml/lxml.etree.o -w
    In file included from src/lxml/lxml.etree.c:239:0:
    /tmp/pip-build-cWMb1G/lxml/src/lxml/includes/etree_defs.h:14:31: fatal error: libxml/xmlversion.h: No such file or directory
     #include "libxml/xmlversion.h"
                                   ^
    compilation terminated.
    error: command 'x86_64-linux-gnu-gcc' failed with exit status 1

    ----------------------------------------
Command "/usr/local/python/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip-build-cWMb1G/lxml/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-QxYF6g-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-build-cWMb1G/lxml
The command '/bin/sh -c mod_wsgi-docker-build' returned a non-zero code: 1
```

jeromekelleher:
Sorry, I can't help here @bwalsh, I don't have access to a mac. Any apple fans able to reproduce this?

bwalsh:
Hi.  Thanks for the follow-up.  I'm not sure it is a macos problem.

I experimented a bit more and found that if I created a image based on this Dockerfile  ...

```
FROM grahamdumpleton/mod-wsgi-docker:python-2.7-onbuild
RUN apt-get update
[...]

PR #504, opened 2015-07-21
Updated docs for new lib requirement.

jeromekelleher:
Since bringing some extra requirements for the OIC extension, we need some extra low-level libraries to make it compile. We've hit this issue already in #484.

This updates the documentation to include the required package for ubuntu. I'll open an issue to update the docker configs as well.


dcolligan:
+1

Issue #503, opened 2015-07-21
test_oidc.py spams a lot


jeromekelleher:
Running test_oidc with output turned on results in a _lot_ of spam. This should be suppressed somehow.

To reproduce:
```
$ nosetests -vs tests/end_to_end/test_oidc.py:TestOidc
```

dcolligan:
Fixed in #508

PR #508, opened 2015-07-22
Remove test_oidc spam

dcolligan:
Issue #503 

Now looks normal:
```
$ nosetests -s -v tests/end_to_end/test_oidc.py
testMultipleOidcClients (tests.end_to_end.test_oidc.TestOidc) ... ok
testOidc (tests.end_to_end.test_oidc.TestOidc) ... ok
testOidcBadLoginKey (tests.end_to_end.test_oidc.TestOidc) ... ok
testOidcBadLoginPassword (tests.end_to_end.test_oidc.TestOidc) ... ok

Ran 4 tests in 20.339s

OK
```

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -14,6 +14,8 @@ import unittest
 import unittest.loader
 import unittest.suite
 
+import requests
+
 import ga4gh.client as client
 import ga4gh.protocol as protocol
 import ga4gh.converters as converters
@@ -236,7 +238,7 @@ def addBinaryOutputArgument(parser):
 ##############################################################################
 
 
-def addGlobalOptions(parser):
+def addServerOptions(parser):
     parser.add_argument(
         "--port", "-P", default=8000, type=int,
         help="The port to listen on")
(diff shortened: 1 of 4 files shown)


ga4gh/frontend.py
```diff
@@ -194,6 +194,7 @@ def configure(configFile=None, baseConfig="ProductionConfig",
     TODO Document this critical function! What does it do? What does
     it assume?
     """
+    requests.packages.urllib3.disable_warnings()
```
jeromekelleher: Oooo, not sure about this. The user should most definitely be warned if they're using dodgy certs. Blanket suppressing this warning is a definite security no-no. Can we just do this during tests? Isn't there a Flask thing we can check to see if we're in testing mode? I guess this is tricky here because we're actually running the server in a subprocess, not running it in test mode. We could add an option to do this in the server CLI I suppose.

jeromekelleher:
Looks good @dcolligan --- we need to do something sensible with the warnings from requests though. A CLI option for the server probably isn't a terrible idea.

ga4gh/frontend.py
dcolligan: Right, we need to suppress the warnings that come from the server subprocess, so we can't do it via the test file.

We could wrap this line in a `if app.config['TESTING']:` statement.  Which is better, but then one might deploy to prod and get a bunch of warnings one hasn't seen before... is this acceptable?

jeromekelleher: But we're not actually in testing mode when we do this though (I think), so it won't do any good. I think the simplest option is to add a server CLI option `--disable-urllib-warnings` and just call `` requests.packages.urllib3.disable_warnings()`` directly from `cli.py` if it's present. It's a hack, but it's generally handy to be able to turn off this nagging when you're developing. We might add it to the client too, come to think of it.

dcolligan:

ga4gh/cli.py
```diff
@@ -255,15 +255,20 @@ def addGlobalOptions(parser):
...
 
 def server_main(parser=None):
     if parser is None:
         parser = argparse.ArgumentParser(
             description="GA4GH reference server")
-    addGlobalOptions(parser)
+    addServerOptions(parser)
     args = parser.parse_args()
-    frontend.configure(args.config_file, args.config, args.port)
+    frontend.configure(
```
jeromekelleher: Why pass this in to the configure function? I think we should just call `` requests.packages.urllib3.disable_warnings()`` directly here rather than within the server configuration. This way, it's explicitly something we only do using the development CLI. If we make it an option for the configure function then people could abuse it in production. Disabling these warnings on a production deployment is a bad idea, and we shouldn't facilitate it.

dcolligan:

jeromekelleher:
Great, thanks @dcolligan. +1.

jeromekelleher:
Looks like we've picked up a merge conflict here @dcolligan --- can you rebase please?

dcolligan:
@jeromekelleher rebased and good to go

Issue #480, opened 2015-07-08
Frontend module cannot be cleanly reset


jbu:
Tests include a `reload(frontend)` call because modules are otherwise only loaded (and initialised) once, but tests require frontend to be re-initialized with different configurations. It would be better if frontend had a proper __init__ or something.

dcolligan:
Fixed in #510

PR #510, opened 2015-07-22
Get rid of reload(fronend) calls in tests

dcolligan:
Issue #480

### DIFF
--- a/ga4gh/serverconfig.py
+++ b/ga4gh/serverconfig.py
@@ -8,6 +8,8 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import datetime
+
 
 class BaseConfig(object):
     """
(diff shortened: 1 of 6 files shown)


ga4gh/serverconfig.py
```diff
@@ -84,3 +86,29 @@ class TestOidcConfig(TestConfig):
     OIDC_AUTHZ_ENDPOINT = "https://accounts.example.com/auth"
     OIDC_TOKEN_ENDPOINT = "https://accounts.example.com/token"
     OIDC_TOKEN_REV_ENDPOINT = "https://accounts.example.com/revoke"
+
+
+class FlaskDefaultConfig(object):
```
jeromekelleher: This is a bit ugly --- is there any other way to reset to a Flask default? If we keep this we should add some comments explaining that it's only used in testing.

dcolligan: I don't think so.  "Modifying" flask for us usually means modifying the config, which is just, essentially, a dictionary.  Reseting means getting this dictionary back to the defaults.  The Flask API does not seem to provide any method to do this for us: http://flask.pocoo.org/docs/0.10/api/#flask.Config

jeromekelleher: OK, the comments at least make it clear that this is what we're doing. Flask configuration really isn't all that amazing...

jeromekelleher:
+1

PR #513, opened 2015-07-23
test to validate mimetypes config, a dependency of oidc-provider

bwalsh:
Purpose of this PR is an initial validation of the github workflow as well as a simple test of oidc-provider mimetypes dependency.

```
-----------------------------------------------------------
TOTAL                          3928    352    91%
Ran 290 tests in 67.713s
```


dcolligan:
Looks great, but I'm sure this is going to fail CI because of style violations.  Use `scripts/run_tests.py` to run the CI locally (including flake, which checks style).

bwalsh:
` because of style violations.`  perfect, I hoped the PR would smoke out issues like that. Fixed.

dcolligan:
Awesome.  I'm ready to +1 this after the commits are squashed.  See http://ga4gh-reference-implementation.readthedocs.org/en/develop/development.html#github-workflow

bwalsh:
Thanks.  Looks like commits squashed ok.

dcolligan:
+1

jeromekelleher:
This is a great test, thanks @bwalsh. There seems to be some issue with the PR though, as the commit for the g2p stuff has come in here as well. Also, this PR is against the g2p branch, and something like this should be against the develop branch so we can use it immediately.

dcolligan:
agree with jerome -- strip out the test and file it against `develop` in its own PR and we'll be able to merge it

bwalsh:
Thanks guys.  Letting the g2p stuff slip in was an oversight.

-b

-Brian Walsh

Issue #498, opened 2015-07-17
Install problem of GA4GH API demo


hltdym:
hello：
 I   try to install the API demo following the command: "pip install --pre ga4gh". But I got Error：

subprocess.CalledProcessError: Command '['wget', 'https://pypi.python.org/packages/source/s/setuptools/setuptools-0.7.zip', '--quiet', '--output-document', '/srv/ga4gh/ga4gh-server-env/build/ga4gh/setuptools-0.7.zip']' returned non-zero exit status 8

----------------------------------------
Command python setup.py egg_info failed with error code 1 in /srv/ga4gh/ga4gh-server-env/build/ga4gh
Storing complete log in /root/.pip/pip.log

WHO can tell me WHY？
Thanks a lot!

jeromekelleher:
Can you give us a little more detail on your system please @hltdym? What's the OS and the version of Python?

Exit status 8 from wget seems to indicate a server error, so this might just be a transient pypi problem.

hltdym:
The version of Python is Python 2.7.3.
I think it is because of  the wget path "https://pypi.python.org/packages/source/s/setuptools/setuptools-0.7.zip" is not found. 
When I run:
wget https://pypi.python.org/packages/source/s/setuptools/setuptools-0.7.zip
--2015-07-17 16:26:33--  https://pypi.python.org/packages/source/s/setuptools/setuptools-0.7.zip
Resolving pypi.python.org (pypi.python.org)... 103.245.222.223
Connecting to pypi.python.org (pypi.python.org)|103.245.222.223|:443... connected.
HTTP request sent, awaiting response...
404 Not Found
2015-07-17 16:26:34 ERROR 404: Not Found.

But I do not known how to deal with it ?

jeromekelleher:
OK, this looks like a bug in our setup script. Basically, we're looking for setuptools and version 0.7 is the earliest version we can use. However, it seems that version 0.7 has disappeared. If you install setuptools manually, you should be able to continue and run the demo.

Try running ``pip install setuptools``. If this doesn't work, there are instructions on https://pypi.python.org/pypi/setuptools to help bootstrap this.

Sorry for the hassle, and thanks for the bug report!

dcolligan:
Should be fixed after #517

PR #517, opened 2015-07-29
Removed mandatory ez_setup usage.

jeromekelleher:
Issue #498

Hopefully this approach will be more robust than requiring a specific version of setuptools.


dcolligan:
+1

Issue #528, opened 2015-07-30
Add support for GET /variants/id


dcolligan:

jeromekelleher:
Fixed in #555.

PR #530, opened 2015-07-30
Implement GET /variants/{id}

dcolligan:
Issue #528


ga4gh/backend.py
```diff
@@ -413,6 +428,24 @@ def getVariantSet(self, id_):
...
+            raise exceptions.BadIdentifierException(id_)
+        try:
+            variantSetId = '{}:{}'.format(splits[0], splits[1])
+        except KeyError:
+            raise exceptions.BadIdentifierException(id_)
+        dataset = self._getDatasetFromCompoundId(id_)
+        variantSet = _safeMapQuery(
+            dataset.getVariantSetIdMap(), variantSetId)
+        variantId = ':'.join(splits[2:])
+        raise Exception(variantSet + variantId)  # TODO
```
dcolligan: @jeromekelleher what does the variant's "id" mean in our realm?  In `datamodel/variants.py` we return `variantSetId:record.contig:record.pos` as the id.  However, the VCF format has an "ID" for each record (that may or may not be there).  Which are we using, or is there some other concept of id?  Is this id guaranteed to be unique?  

And as far as accessing the VCF file goes, is the algorithm just going to be iterate and stop when the first matching variant is seen and return it?

jeromekelleher: Good questions @dcolligan! The ID field in the VCF corresponds to the `name` field in the GA4GH objects, and it's basically useless as an identifier.  As for our ID, I think we'll have to do something like ```datasetId:variantSetId:chrom:pos:md5(REF + ALT)``. The last part is necessary because we can have many variants with the same sort coordinate, and the only way to be sure it's _this_ exact variant  is to use the REF and ALT sequences. I think it's a good idea to use md5 here (or some other hash) because our IDs could become arbitrarily long otherwise.

In terms of finding the variant in question, we can use pysam to seek to the first record overlapping (chrom,pos) and then iterate from there until we hit the VCF record we're interested in. This will be fast the majority of the time.

ga4gh/datamodel/variants.py
```diff
@@ -415,6 +427,28 @@ def convertVariant(self, record, callSetIds):
...
+    def getVariant(self, compoundId):
+        if compoundId.chrom in self._chromFileMap:
+            varFileName = self._chromFileMap[compoundId.chrom]
+        else:
+            raise exceptions.ObjectNotFoundException(compoundId)
+        pos = int(compoundId.pos)
+        referenceName, startPosition, endPosition = \
+            self.sanitizeVariantFileFetch(
+                compoundId.chrom, pos, None)
+        # TODO vcf.fetch's startPosition arg seems to be exclusive,
```
dcolligan: @jeromekelleher we're not taking this into account in our interval iteration algorithm, are we?

dcolligan:
@jeromekelleher this is ready for review... writing tests now...

ga4gh/backend.py
```diff
@@ -413,6 +429,18 @@ def getVariantSet(self, id_):
         dataset = self._getDatasetFromCompoundId(id_)
         return self.runGetRequest(dataset.getVariantSetIdMap(), id_)
 
+    def getVariant(self, id_):
+        """
+        Returns a variant with the given id
+        """
+        compoundId = variants.CompoundVariantId(id_)
+        dataset = self._getDatasetFromCompoundId(compoundId.datasetId)
+        variantSet = _safeMapQuery(
```
jeromekelleher: Why not call ``dataset.getVariantSet(variantSetId)``, and push this functionality down into the dataset? It feels like we're moving a lot of maps around where we could be using an interface (and keeping the implementation flexible)?

ga4gh/client.py
```diff
@@ -99,8 +103,11 @@ def _deserializeResponse(self, response, protocolResponseClass):
         jsonResponseString = response.text
         self._updateBytesRead(jsonResponseString)
         self._debugResponse(jsonResponseString)
-        responseObject = protocolResponseClass.fromJsonString(
-            jsonResponseString)
+        try:
+            responseObject = protocolResponseClass.fromJsonString(
+                jsonResponseString)
+        except protocol.UnableToDecodeJsonException:
```
jeromekelleher: Are these always the same thing? What happens if the server sends back malformed JSON? I don't really see how re-raising a different exception is helping us here.

ga4gh/datamodel/__init__.py
```diff
@@ -27,6 +27,44 @@ def _cleanupHtslibsMess(indexDir):
...
 
+class CompoundId(object):
+    """
+    Base class for compoundIds
+    """
+    seperator = ':'
+
+    def __init__(self, compoundIdStr):
+        try:
+            splits = compoundIdStr.split(self.seperator)
```
jeromekelleher: `split` raises an `AttributeError` when? This is confusing

ga4gh/datamodel/__init__.py
```diff
@@ -27,6 +27,44 @@ def _cleanupHtslibsMess(indexDir):
...
+        except AttributeError:
+            raise exceptions.BadIdentifierException(
+                compoundIdStr, self._getParseErrorMessage())
+        if len(splits) != len(self.fields):
```
jeromekelleher: Where does ``self.fields`` come from? If it's an attribute, shouldn't we be passing it in to the constructor?

jeromekelleher: <looks down further > --- I see, disregard this

ga4gh/datamodel/variants.py
```diff
@@ -47,6 +48,17 @@ def convertVCFGenotype(vcfGenotype, vcfPhaseset):
     return genotype, phaseset
 
 
+class CompoundVariantId(datamodel.CompoundId):
+    """
+    The compound id for a variant
+    """
+    fields = ['datasetId', 'vsId', 'chrom', 'pos', 'md5']
```
jeromekelleher: We should follow the ga4gh names for this: ``variantSetId``, ``referenceName``, etc.

ga4gh/datamodel/variants.py
jeromekelleher: I don't know --- perhaps this is because VCFs have a 1-based coordinate system rather than 0 based? We should have hit this already if it wasn't being managed.

ga4gh/datamodel/variants.py
```diff
@@ -415,6 +427,28 @@ def convertVariant(self, record, callSetIds):
...
+                compoundId.chrom, pos, None)
+        # TODO vcf.fetch's startPosition arg seems to be exclusive,
+        # not inclusive!!!
+        startPosition = startPosition - 0.1
+        cursor = self.getFileHandle(varFileName).fetch(
+            referenceName, startPosition, endPosition)
+        for record in cursor:
+            if pos != record.pos:
+                continue
+            digest = hashlib.md5(record.ref + str(record.alts)).hexdigest()
```
jeromekelleher: We should calculate the digest in one place --- a method on the relevant class would seem like the best option.

ga4gh/exceptions.py
```diff
@@ -123,6 +123,13 @@ class BadPageTokenException(BadRequestException):
     message = "Request page token invalid"
 
 
+class BadIdentifierException(BadRequestException):
+    def __init__(self, id_, msg=None):
+        self.message = "The identifier provided is invalid: '{}' ".format(id_)
+        if msg:
```
jeromekelleher: ``if msg is not None``

ga4gh/protocol.py
```diff
@@ -26,6 +26,17 @@ def convertDatetime(t):
     return int(millis)
 
 
+class ProtocolException(Exception):
+    """
+    Something went wrong in the protocol module
+    """
+
+
+class UnableToDecodeJsonException(Exception):
```
jeromekelleher: Why isn't this in the exceptions module?

jeromekelleher:
Looks good @dcolligan, thanks. I'm not totally blown away by the current organisation --- it seems a bit brittle and piecemeal somehow.  Is there any way we can push the ID generation and parsing into the ``DataObject`` hierarchy, and keep all of the ID processing logic in one place? The CompoundId abstraction is clever, but I think ID management is a fundamental part of the responsibilty of the DataObjects and it would be great if we could keep it all in one place.

I don't have a strong notion of how to do this though, so perhaps the best approach is to continue with the way you're doing things and perhaps refactor afterwards.

ga4gh/backend.py
```diff
@@ -229,8 +245,13 @@ def _getDatasetFromVariantsRequest(self, request):
         return dataset
 
     def _getDatasetFromCompoundId(self, compoundId):
+        if compoundId is None:
+            raise exceptions.BadIdentifierException(compoundId)
+        if not hasattr(compoundId, 'split'):
```
jeromekelleher: This is a bad sign! If we're testing for attribute existence for control flow, we should be doing things differently (except in places where the protocol is stupid, and there's no other option).

dcolligan: this is just testing compoundId is a string, basically, but I agree it could be better phrased

ga4gh/protocol.py
dcolligan: It's not necessarily a server error.  In fact both of these were written with client errors in mind.  Should they go in `exceptions.py` (and just not inherit from BaseServerException) or in a separate module?

ga4gh/backend.py
dcolligan: We could, but then we'd have to either have something like `_safeMapQuery` accessible in `datasets.py` as well, or have a try/catch block here, the later of which I don't like much.  Where should `_safeMapQuery` live, do you think?  It's a very general method...

ga4gh/datamodel/__init__.py
dcolligan: when a non-basestring subclass is passed in... I'll change the phrasing

ga4gh/datamodel/variants.py
dcolligan: but the variantSetId is actually a combination of these fields (see `variantSetId` in the `comboFields` attr)

ga4gh/backend.py
jeromekelleher: Why don't you like having a try/catch in datasets.py? Seems simple enough to me:
```python
def getVariantSet(variantSetId):
    if variantSetId not in self._variantSetIdMap:
         raise VariantSetNotFoundException(variantSetId)
    return self._variantSetIdMap[variantSetId]
```

ga4gh/protocol.py
jeromekelleher: I would put it in the ``exceptions`` module, and just not have it inherit from BaseServerException. It's useful to keep all the keep all the exception in one place; we don't have to remember which module what exception comes from then.

ga4gh/backend.py
jeromekelleher: OK, reading through the rest of the code there would be a lot of work to move over to an getter based approach. Let's do it this way for now, and try to refactor away from all this passing around map references at some point.

ga4gh/client.py
```diff
@@ -99,8 +103,14 @@ def _deserializeResponse(self, response, protocolResponseClass):
         jsonResponseString = response.text
         self._updateBytesRead(jsonResponseString)
         self._debugResponse(jsonResponseString)
-        responseObject = protocolResponseClass.fromJsonString(
-            jsonResponseString)
+        try:
+            responseObject = protocolResponseClass.fromJsonString(
+                jsonResponseString)
+        except protocol.UnableToDecodeJsonException as exception:
```
jeromekelleher: This seems even more complicated! Is this necessary for the getVariant functionality? If not, I think we should put it into a separate PR.

ga4gh/datamodel/__init__.py
```diff
@@ -27,6 +27,45 @@ def _cleanupHtslibsMess(indexDir):
         shutil.rmtree(indexDir)
 
 
+class CompoundId(object):
+    """
+    Base class for compoundIds
```
jeromekelleher: This comment doesn't tell me much. What's a compoundId? How is it used?

ga4gh/datamodel/variants.py
jeromekelleher: OK, `variantSetId` is a special case, but `chrom` and `pos` should be `referenceName` and `start`.

ga4gh/datamodel/variants.py
```diff
@@ -415,6 +436,27 @@ def convertVariant(self, record, callSetIds):
             sampleIterator += 1  # REMOVAL
         return variant
 
+    def getVariant(self, compoundId):
+        if compoundId.chrom in self._chromFileMap:
+            varFileName = self._chromFileMap[compoundId.chrom]
+        else:
+            raise exceptions.ObjectNotFoundException(compoundId)
+        pos = int(compoundId.pos)
```
jeromekelleher: Why are we using pos here? Is this equal to the ``start`` value (0 based)? If not, we should just change the ``variant.id`` definition in convertVariant as indicated in the comments there.

ga4gh/datamodel/variants.py
```diff
@@ -415,6 +436,27 @@ def convertVariant(self, record, callSetIds):
...
+        # fetch pos is exclusive, not inclusive
```
jeromekelleher: We should need to do this --- we use 0-based ``start`` elsewhere and it works. I think we just need to change the ``variant.id`` definition. (Which we must do anyway, to add in the hash, right?)

jeromekelleher:
This is good progress @dcolligan, thanks. I'd like to see some systematic and data driven tests here. Are we generating the correct ID for variants? If we get some variants by a search query, and send back get requests, do we get exactly the same variant?

ga4gh/client.py
dcolligan: Okay, filed this #536

dcolligan:
@jeromekelleher pushed an update... 

I think the CompoundId abstraction is a good one.  We may want to create or parse ids outside of the context of a data object, so I think keeping it has not part of the data object hierarchy makes sense.  I would propose a refactor that used CompoundIds across the code, getting rid of all the `id.split(':')` and `':'.join(idSplits)` and `'{}:{}'.format(idPart1, idPart2)` statements scattered around the code.

Also, I'm a bit confused about when the vcf records' start attr is supposed to be 1-indexed... do we want it to be 1-indexed in our ids?  (see failing tests)

dcolligan:
@jeromekelleher any thoughts on this?

(ps I also sent you an email)

ga4gh/datamodel/variants.py
```diff
@@ -222,8 +243,13 @@ def generateVariant(self, variantSetId, referenceName, position,
         variant = self._createGaVariant()
         variant.names = []
         variant.referenceName = referenceName
-        variant.id = "{0}:{1}:{2}".format(
-            variant.variantSetId, referenceName, position)
+        md5 = 'notApplicable'
```
jeromekelleher: We should compute the compound ID of the variant in the `AbstractVariantSet` class, and use the values from a GA object to compute the hash. This will allow us to do this in one place. For example, we might have
```
```
We can then reuse this for the simulated and the Htslib derived variants.

ga4gh/backend.py
```diff
@@ -229,8 +245,13 @@ def _getDatasetFromVariantsRequest(self, request):
         return dataset
 
     def _getDatasetFromCompoundId(self, compoundId):
```
jeromekelleher: We're getting a bit of a proliferation of methods to get a dataset, and having ``compoundId`` refer to both a string and an object is very confusing. How about this:

1. As soon as we have an ID that we want to decompose, we allocate a CompoundId instance, and use this object to get attributes we want.
2. Then, use this to get the values we want. For example,

```python
def getVariantSet(self, id_):   
     compoundId = variants.VariantSetCompoundId(id_) # This takes care of the parsing, and throw exception if malformed.
     dataset = self.getDataset(compoundId.dataset)
     return self.runGetRequest(dataset.getVariantSetIdMap(), id_) 

def getVariant(self, id_):
     compoundId =  variants.VariantCompoundId(id_)
     dataset = self.getDataset(compoundId.dataset)
     variantSet = _safeMapQuery(  
          dataset.getVariantSetIdMap(), compoundId.variantSetId)  
      variant = variantSet.getVariant(compoundId)
      return variant.toJsonString()
```
Does this make sense? 

I've just realised that the backend's API really isn't very helpful at the moment. I'm going to attempt a quick refactor to see if it clears things up a bit.

jeromekelleher:
Looking at some other issues, they're all sort of tied in with this idea of the compound ID now. I think it would be good if we could pull the basic compound ID stuff out of this, and make it it's own PR (possibly on top of #545). Hopefully we can start making progress on several fronts then, and not get tied up with implementing getVariants (which is the most difficult, I think).

dcolligan:
Ok, good idea.  I'll merge #545 and then do a compoundId refactor.

dcolligan:
Closing... submitting in another PR...

Issue #527, opened 2015-07-30
Add support for GET /callsets/id


dcolligan:

dcolligan:
@jeromekelleher so here we have another what-is-the-id-actually question to decide...

We need the variantSet to look up a callset with a certain id.  So should the id of the callset that we pass in this request be `datasetId:vsId:callsetId`?  Otherwise we'd have to pass additional info in the URL.

dcolligan:
CallSetIds seem to be formatted as such now: e.g. `'dataset1:1kg-phase1.HG00533'` ... we can turn the period into a colon, I hope?

jeromekelleher:
Yeah, this is the way to do it (``datasetId:vsId:callsetId``). We need to make all these IDs consistent throughout --- anything else should be considered a bug. (Note, we want to see IDs of this form coming from the simulated datasets too...)

diekhans:
This raises a very important issue.  Many of us would like to see ids being globally unique and computed by hashing the data.  Also, having server-specific ids complicates writing protable tests cases.

jeromekelleher:
We can't do GUIDs right now @diekhans, this would take serious thought and work to do properly. I agree GUIDs are something we need, but it's not something we can implement here without widespread agreement about how it should be done and rigorous specification. Test cases will just have to discover the data using the existing API methods.

diekhans:
no argument @jeromekelleher ; this is a thorny issue and one GA4GH needs to stop avoiding.  UUIDs or sha1 are not efficient to index due to their random nature.

Something worth bringing up with the refvar group, as they are specifying UUID in the graph

dcolligan:
Fixed in #542

hjellinek:
What should `GET /callsets/some-ID-not-naming-a-real-callset` return?
right now it returns a `CallSet` with all of its fields set to default values, which seems wrong.  I'll open an issue about that if you like.

jeromekelleher:
This is wrong all right; any `GET /objects/id-that-doesn't-exist` should return a 404. I would be good to open an issue on this.

PR #542, opened 2015-08-05
Implement GET /callsets/{id}

dcolligan:
First pass.  Still need to add tests.  Feedback welcome.

```
$ python client_dev.py -O callsets-get dataset1:1kg-phase1:HG00533 http://localhost:8000/current
dataset1:1kg-phase1:HG00533
```

Issue #527

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -15,8 +15,8 @@ import ga4gh.datamodel.references as references
 import ga4gh.exceptions as exceptions
 import ga4gh.datamodel as datamodel
 import ga4gh.datamodel.reads as reads
-import ga4gh.datamodel.variants as variants
 import ga4gh.datamodel.datasets as datasets
+import ga4gh.datamodel.variants as variants
 
 
 def _parsePageToken(pageToken, numValues):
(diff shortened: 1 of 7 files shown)


ga4gh/backend.py
```diff
@@ -37,8 +38,8 @@ def _parsePageToken(pageToken, numValues):
     return values
 
 
-def _getVariantSet(request, variantSetIdMap):
-    variantSetId = request.variantSetId
+def _getVariantSet(requestOrCompoundId, variantSetIdMap):
```
jeromekelleher: I don't think this parameter overloading. This will get complicated quickly. Why not make a method ``_getVariantSetFromCompountId`` which parallels the ``_getDatasetFromCompoundId``?

jeromekelleher:
Looks good so far (except for the ``somethingOrSomethingElse`` parameter).

ga4gh/backend.py
dcolligan: Yeah, whatever choice we make we're in an intermediate state where something can be overloaded.

'compoundId' can refer to either an:id:that:looks:like:this or a instance that derives from the new type CompoundId

A refactor/cleanup will be needed after these routes are finished.

jeromekelleher: Maybe ``compoundIdStr`` and ``compoundId`` is a good convention to adopt?

dcolligan:

jeromekelleher:
+1. Merging.

I'm about to land a fairly big refactor, so it would be good if we could leave the rest of these until afterwards.

PR #550, opened 2015-08-06
Changed FASTA library to pyfaidx...

macieksmuga:
...and introduced an optional parameter on `SearchSequenceRequest`  that causes a  string to be given for every base returned.

### DIFF
--- a/ga4gh/_protocol_definitions.py
+++ b/ga4gh/_protocol_definitions.py
@@ -4156,6 +4156,7 @@ class SearchSequencesRequest(SearchRequest):
 "SearchSequencesRequest", "fields": [{"default": null, "doc": "",
 "type": ["null", "string"], "name": "referenceSetId"}, {"default":
 null, "doc": "", "type": ["null", "string"], "name": "variantSetId"},
+{"default": false, "doc": "", "type": "boolean", "name": "listBases"},
 {"default": null, "doc": "", "type": ["null", "int"], "name":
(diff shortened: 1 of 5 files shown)


macieksmuga:
and changed it back once it became clear that pyfaidx was causing more problems than it was solving. Sigh.

bioinformed:
What was the problem?  Also, have you tried the native index FASTA support
in pysam?

macieksmuga:
I believe the issue was with the string conversion: Pyfaidx returns bases in a custom type, that I then had to convert explicitly to string (so that it could be embedded into the JSON output stream). I saw no facilities to enable pyfaidx to emit the bases as a string to begin with.

I had initially experimented with alternate FASTA libraries due to poor performance of the graph server, but have since found two unrelated bugs that accounted for most of the problem. Thus I reverted to the library I had been using previously.

Independently of the above, I'd be happy to explore FASTA support in pysam, so long as it delivers similar or better performance to the current benchmark.

bioinformed:
The pysam Fastafile object is a pretty basic wrapper around the htslib
code.  See http://pysam.readthedocs.org/en/latest/api.html#pysam.FastaFile.

mdshw5:
I'm probably way too late on this but pyfaidx supports the `as_raw` argument which will just return a Python string. Depending on your usage, this will be much faster than either samtools or pysam, without depending on a C compiler.

PR #551, opened 2015-08-06
Straw-man outline of ID obfuscation.

jeromekelleher:
We use our IDs as a breadcrumb trail to allow us find objects easily without needing to maintain reverse mapping tables. However, this ID structure is not part of the protocol, and should not be depended on (or used at all) by clients as we might change this at any time. Also, other implementers should not receive the impression that this is the official or required structure for GA4GH IDs.

Therefore, I think we should obfuscate our ID values, so that the underlying structure at least isn't obvious. This should not be seen as a security measure, as it's not intended to make anything secret.

This is an initial straw man proposal where we base64 encode the IDs. For example, we have
an ID like ``dataset1:1kg-phase3`` becoming ``ZGF0YXNldDE6MWtnLXBoYXNlMw==``. Base64 seems like a good option as it's cheap and easily reversible, but I'm sure there are other options.

This breaks lots of tests, as we use this structure quite a lot in our own test cases, so I haven't bothered updating these until we have a consensus on whether this a good idea or not.


dcolligan:
I guess I'm -0 on this.  I don't really see the need.

jeromekelleher:
@diekhans, any thoughts on this? I'm worried that people will start writing conformance tests around our ID format, and this seems like a good way to ensure that clients don't start writing unportable code based on incorrect assumptions.

diekhans:
Seems like a great stop-gap measure until GA4GH gets somewhere
with identification.  We seem to have the mentality of
implementation by imitation rather than specification, so I
think this could making switching to GUIDs less of a shock.

jeromekelleher:
Closed in favour of #568.

PR #553, opened 2015-08-06
Add _safeMapQuery to backend

dcolligan:
Also adding test_compound_ids.py because it weirdly did not
get committed in the compoundId refactor

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -41,11 +41,26 @@ def _parsePageToken(pageToken, numValues):
 
 def _getVariantSet(request, variantSetIdMap):
     variantSetId = request.variantSetId
-    try:
-        variantSet = variantSetIdMap[variantSetId]
-    except KeyError:
-        raise exceptions.VariantSetNotFoundException(variantSetId)
+    variantSet = _safeMapQuery(
+        variantSetIdMap, variantSetId,
+        exceptionClass=exceptions.VariantSetNotFoundException)
     return variantSet
+
+
+def _safeMapQuery(idMap, id_, exceptionClass=None, idErrorString=None):
+    """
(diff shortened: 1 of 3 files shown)


ga4gh/backend.py
```diff
@@ -41,11 +41,27 @@ def _parsePageToken(pageToken, numValues):
...
+        variantSetIdMap, variantSetId,
+        exceptionClass=exceptions.VariantSetNotFoundException)
+    return variantSet
+
+
+def _safeMapQuery(idMap, id_, exceptionClass=None, idErrorString=None):
+    """
+    Attempt to retrieve a value from a map, throw an appropriate error
+    if the key is not present
+    (works for both dict and list types)
```
jeromekelleher: I don't follow this last sentence... What do you mean about list types?

jeromekelleher:
+1 modulo the comment above on the documentation.

ga4gh/backend.py
dcolligan: Since slicing syntax is applicable to both dict and list types, _safeMapQuery can be used on either:

`_safeMapQuery({'a':'b'}, 'a')` or `_safeMapQuery(['a'], 0)`

jeromekelleher: Yes, but we're explicitly saying it's a map, and passing a key, so this is just confusing. Anyway, a list doesn't throw a KeyError if an index doesn't exist (an IndexError) so this wouldn't work as advertised.

Just delete the comment --- we don't need it to support lists.

dcolligan:
@jeromekelleher I took out that comment... could you merge?

jeromekelleher:
Thanks @dcolligan, merging.

Issue #535, opened 2015-08-03
datasetId field not set in read group set objects returned by searchReadGroupSets


hjellinek:
I'm using data from the server unit tests as the server's data.  The files are arranged like so:
```
dataset1/
    reads/
        1kg-low-coverage/
            ... BAM files ...
        wgBam/
            ... BAM files ...
    variants/
```

**Assumption**: each subdirectory in `reads` corresponds to a read group set.

I create a search request with `datasetId` = `dataset1` and the other fields unset (default values).

**Expected**:
The call to `searchReadGroupSets` will return a `SearchReadGroupSetsResponse` with its `readGroupSets` field containing two read group sets, each with its `datasetId` field equal to `"dataset1"`.

**Actual**:
The call to `searchReadGroupSets` returns a `SearchReadGroupSetsResponse` with its `readGroupSets` field containing two read group sets.  Both have `datasetId` equal to `null`.

jeromekelleher:
Verified fixed in #559.

PR #554, opened 2015-08-06
Id cleanup

jeromekelleher:
This is a fairly widespread refactoring of the datamodel code. Changes:

1. All of the datamodel objects actually inherit from datamodel.DatamodelObject now. Each one has a CompoundId and a localId. 
2. Refactored the CompoundId concept a bit to fit into this.

The aim of this is to get a proper, linked up model in memory of our data tree. This should make issues like #535 simple to fix (should be fixed already, in fact; it would be better to verify this and add some test cases in a follow up PR though).

There's still some tests to fix, but I ran out of time. Hopefully I'll finish this PR later.

Comments/feedback?

### DIFF
--- a/ga4gh/datamodel/datasets.py
+++ b/ga4gh/datamodel/datasets.py
@@ -18,7 +18,10 @@ class AbstractDataset(datamodel.DatamodelObject):
     """
     The base class of datasets containing variants and reads
     """
-    def __init__(self):
+    compoundIdClass = datamodel.DatasetCompoundId
+
+    def __init__(self, localId):
+        super(AbstractDataset, self).__init__(None, localId)
         self._variantSetIds = []
         self._variantSetIdMap = {}
         self._readGroupSetIds = []
@@ -30,18 +33,6 @@ class AbstractDataset(datamodel.DatamodelObject):
         dataset = protocol.Dataset()
         dataset.id = self.getId()
         return dataset
-
-    def getId(self):
-        """
-        Return the id of the dataset
-        """
-        return self._id
-
-    def getDirectory(self):
-        """
-        Return the path of the directory of this dataset
-        """
-        return self._datasetDir
 
     def getVariantSetIds(self):
         """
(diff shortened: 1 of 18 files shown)


jeromekelleher:
This is ready for review. I haven't squashed the commits yet as I can't decide where the best place for the CompoundId classes is. It seems like they belong in the specific modules (which is where they are now), but his creates a circular import from datasets->reads/variants->datasets. I've worked around this issue fairly crudely and had to remove the snakefood test for circular imports. It works fine, but it's a bit ugly. The other approach is to put all the CompoundId classes into datamodel, which avoids this problem but seems confusing. 

Anyway, comments appreciated.

dcolligan:
Circular imports should be avoided unless there is a really, really good reason and this is not one of them.  We should prefer the solution that avoids circular imports.

dcolligan:
Okay, in general I like this refactor, although it does tie us strongly into hierarchical IDs.  My previous classes were a bit more flexible in that regard.  I suppose this is more appropriate for the data model we have now, however.

jeromekelleher:
Excellent, glad to hear it. I'm happy to change the circular import stuff. My original approach was to have all of the CompoundId classes in the `datamodel` module, which avoided the circular imports. See https://github.com/jeromekelleher/server/commit/ee0eb42d55b6733c86211554de9d8a9572f6ceba to get an idea of the changes. Are you happy reverting to something like this, or is there a better option?

ga4gh/datamodel/datasets.py
```diff
@@ -97,30 +94,32 @@ class SimulatedDataset(AbstractDataset):
...
+            self, localId, randomSeed=1, numCalls=1,
+            variantDensity=1, numVariantSets=1, numAlignments=1):
+        super(SimulatedDataset, self).__init__(localId)
         self._randomSeed = randomSeed
         self._randomGenerator = random.Random()
         self._randomGenerator.seed(self._randomSeed)
 
+        # We do the imports here because we have circular dependencies
+        # otherwise.
+        import ga4gh.datamodel.variants as variants
```
dcolligan: ew ew ew ew ew ew!  Let's move all the CompoundId subclasses to datamodel/__init__.py

ga4gh/datamodel/datasets.py
```diff
@@ -131,34 +130,33 @@ class FileSystemDataset(AbstractDataset):
...
-        super(FileSystemDataset, self).__init__()
-        self._id = os.path.basename(os.path.normpath(datasetDir))
-        self._datasetDir = datasetDir
+    def __init__(self, dataDir):
+        localId = os.path.basename(os.path.normpath(dataDir))
+        super(FileSystemDataset, self).__init__(localId)
+
+        # We do the imports here because we have circular dependencies
+        # otherwise.
+        import ga4gh.datamodel.variants as variants
```
dcolligan: same here

ga4gh/datamodel/reads.py
```diff
@@ -11,37 +11,34 @@
...
 
-import ga4gh.protocol as protocol
 import ga4gh.datamodel as datamodel
+import ga4gh.datamodel.datasets as datasets
 import ga4gh.exceptions as exceptions
+import ga4gh.protocol as protocol
 
 
-class CompoundReadGroupSetId(datamodel.CompoundId):
+class ReadGroupSetCompoundId(datasets.DatasetCompoundId):
```
dcolligan: name change to have the 'compound' part after the class type is probably a good idea

ga4gh/datamodel/reads.py
```diff
@@ -11,37 +11,34 @@
...
     """
     The compound id for a read group set
     """
-    fields = ['datasetId', 'rgsId']
+    fields = datasets.DatasetCompoundId.fields + ['readGroupSet']
```
dcolligan: composing the fields from the parent class' field attr is a good idea

tests/unit/test_imports.py
```diff
@@ -35,10 +35,6 @@ def setUpClass(cls):
         snakefoodScanner = SnakefoodScanner()
         cls.graph = snakefoodScanner.scan()
 
-    def testNoCycles(self):
```
dcolligan: reenable this

dcolligan:
yeah, put CompoundId classes in the datamodel module

jeromekelleher:
@dcolligan --- ickiness removed. Hopefully ready to land?

dcolligan:
+1

PR #555, opened 2015-08-06
Implement GET /variants/{id}

dcolligan:
Half-finished; will apply comments in #530 later

Issue #528

### DIFF
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -8,6 +8,7 @@ from __future__ import unicode_literals
 
 import datetime
 import random
+import hashlib
 
 import pysam
 
@@ -161,8 +162,10 @@ class AbstractVariantSet(datamodel.DatamodelObject):
         Returns an ID string suitable for the specified GA Variant
         object in this variant set.
         """
+        md5 = self.hashVariant(gaVariant)
         compoundId = datamodel.VariantCompoundId(
-            self.getCompoundId(), gaVariant.referenceName, gaVariant.start)
+            self.getCompoundId(), gaVariant.referenceName,
+            gaVariant.start, md5)
         return str(compoundId)
 
     def getCallSetId(self, sampleName):
@@ -173,6 +176,16 @@ class AbstractVariantSet(datamodel.DatamodelObject):
         compoundId = datamodel.CallSetCompoundId(
             self.getCompoundId(), sampleName)
         return str(compoundId)
+
+    @classmethod
+    def hashVariant(cls, gaVariant):
+        """
+        Produces an MD5 hash of the ga variant object to uniquely
+        identify it
+        """
+        return hashlib.md5(
+            gaVariant.referenceBases +
+            str(tuple(gaVariant.alternateBases))).hexdigest()
 
 
 class SimulatedVariantSet(AbstractVariantSet):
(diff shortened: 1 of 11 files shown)


dcolligan:

ga4gh/datamodel/variants.py
```diff
@@ -410,6 +445,26 @@ def convertVariant(self, record, callSetIds):
         variant.id = self.getVariantId(variant)
         return variant
 
+    def getVariant(self, compoundId):
+        if compoundId.referenceName in self._chromFileMap:
+            varFileName = self._chromFileMap[compoundId.referenceName]
+        else:
+            raise exceptions.ObjectNotFoundException(compoundId)
+        start = int(compoundId.start)
+        # fetch start is exclusive, not inclusive
```
jeromekelleher: Does this comment still hold? You don't seem to be doing anything about it if it does...

ga4gh/datamodel/variants.py
```diff
@@ -410,6 +445,26 @@ def convertVariant(self, record, callSetIds):
...
+        referenceName, startPosition, endPosition = \
+            self.sanitizeVariantFileFetch(
+                compoundId.referenceName, start, None)
+        cursor = self.getFileHandle(varFileName).fetch(
+            referenceName, startPosition, endPosition)
+        for record in cursor:
+            if start != record.start:
```
jeromekelleher: What about
```python
variant = self.convertVariant(record, [])
if record.start == start and compoundId.md5 == self.hashVariant(variant):
    return variant
elif record.start > start:
    raise ObjectNotFoundException()
```
Most of the time, we'll hit the variant first go and so there's no point in trying to save a few cycles doing the convert variant process. This way we have only one method to get the hash. We do need to guard against the variant not existing, otherwise we could iterate over millions of variants before exiting the loop. Another way to do this I suppose is to put in an end coodinate. Is there any reason we can't use ``start + 1`` there?

tests/datadriven/test_variants.py
```diff
@@ -346,3 +347,22 @@ def convertPyvcfNumber(number):
...
+        variantSet = self._gaObject
+        for referenceName in self._referenceNames:
+            refnameVariants = self._getPyvcfVariants(referenceName)
+            for variant in refnameVariants:
+                md5 = self._hashVariant(variant)
+                compoundId = datamodel.VariantCompoundId(
+                    variantSet.getCompoundId(), referenceName,
+                    variant.start, md5)
+                gotVariant = variantSet.getVariant(compoundId)
+                self.assertEqual(str(compoundId), gotVariant.id)
```
jeromekelleher: We need some negative tests here --- what happens when we try to get some variants that are very like this one? Off by one coords, slightly different ref and alt?

jeromekelleher:
Looks good @dcolligan, thanks. We need to think about what happens a bit more when variants are missing, but otherwise this is spot on.

dcolligan:

jeromekelleher:
Looks @dcolligan, thanks. 

+1 once the CI has completed (it failed with what looked like a pip error, so I kicked it off again).

dcolligan:
no that was my bad, I resubmitted with a corrected branch

jeromekelleher:
OK, merging. Thanks. That list of unimplemented URLS is getting small!

dcolligan:
I want to get it down to 0, but that's assigned to someone else!

Issue #557, opened 2015-08-10
Eliminate hardcoded object names in tests (and elsewhere?)


dcolligan:
Lots of tests rely on hardcoded object names in the simulator, etc.  It would be good to refactor these out to a common module so we didn't have hardcoded names scattered throughout different places in the code, and just call methods/attrs in that module when we need an example valid object name.

jeromekelleher:
Yes please! We might not need a module to do it, just using some methods in the backend object to get a variantset (say) and then storing the ids as instance variables on the test case should be fine. Ideally this should be done along with ID obfuscation, as this will break a lot of test cases and we might as well do it all in one go.

dcolligan:
Fixed in #558

PR #558, opened 2015-08-10
Consolidate hardcoded test values

dcolligan:
Issue #557

### DIFF
--- a/tests/end_to_end/test_gestalt.py
+++ b/tests/end_to_end/test_gestalt.py
@@ -21,6 +21,10 @@ class TestGestalt(server_test.ServerTest):
     An end-to-end test of the client and server
     """
     def testEndToEnd(self):
+        self.simulatedVariantSetId = "simulatedDataset1:simVs0"
+        self.simulatedReadGroupId = "simulatedDataset1:aReadGroupSet:one"
+        self.simulatedReferenceSetId = "referenceSet0"
+        self.simulatedReferenceId = "referenceSet0:srs0"
         self.client = client.ClientForTesting(self.server.getUrl())
         self.runVariantsRequest()
         self.assertLogsWritten()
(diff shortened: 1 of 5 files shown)


jeromekelleher:
I'm afraid I don't really see the point of this --- we should be adding methods to get these values in the places we're using them, and generalising away from using hardcoded values. This just centralises all the hardcoded values, and doesn't address the underlying bad practices. 

I've refactored out the hard coded values from ``test_simulated_stack.py``, and it wouldn't be hard to do the same for the others I think.

diekhans:
I dislike `utils.Hardcoded' (anything called utils or common doesn't
aid in the understanding of code.  Call a class Hardcoded kind of just
makes it sounds bad.  It's really static configuration, right?

What is a better approach?  While I agree having static
configuration scattered around the code isn't that great, I have
also seen people go to such great lengths to avoid being accused
of hardcoding stuff that they make things much worth.  Things
like multiple level of indirection to get variables out of
non-code text files that are tightly bound to the code are no
better than `hardcoding'.

The best luck I have had is having functions to get
configuration values, a lot of just static configuration values.

jeromekelleher:
These are really just ID values that we can discover at run time. I think it'll be easy to instantiate a backend, and find the values we want in the majority of cases.

diekhans:
As long as this doesn't create the danger of missing test because
test data is missing, sounds like a very sane approach ..

ok, I am going back to the peanut gallery ..

jeromekelleher:
Yeah, good point --- always check ``len(testObjects) > 0`` before starting the test!

dcolligan:
@jeromekelleher so what interface would you favor for this?  A classmethod on every simulated class to generate an id (given a counting variable which defaults to 0)?  Or just one object in `datasets/__init__` which generates ids given a class?  etc.

jeromekelleher:
I think we can just do something like
```python
backend = self._app.backend
dataset = backend.getDatasets()[0]
variantSet  = dataset.getVariantSets()[0]
self._datasetId = dataset.getId()
self._variantSetId = variantSet.getId()
```
in setup and then use ``self._variantSetId`` etc where we need ID values. Something like this will definitely work for `test_views.py`, which is currently the worst offender. It's a little more complicated for the end-to-end tests, I agree, but these can be dealt with too I'm sure.

dcolligan:
@jeromekelleher alright, pushed an update... just consolidated the ids for the end to end tests, but took your advice on the rest

jeromekelleher:
Superb, thanks for this @dcolligan! I was just about to hit off the hard-coded IDs in ``test_views`` and you saved me.

+1

Issue #531, opened 2015-07-31
Dataset ids should be set for VariantSets


jeromekelleher:
Currently, the datasetId field in variantSet is `NotImplemented`. This should be set. Also any other objects that have a datasetId should be checked and fixed also.

jeromekelleher:
Verified fixed in #559.

PR #559, opened 2015-08-10
Refactor reference handling and some tests.

jeromekelleher:
- This changes reference handling to be symmetrical with the way datasets
are managed (fully hierarchical).

- Fixed a missing part of the CLI and API where referenceSets were not
  correctly passed to searchReferences.

- Provide getters for all datamodel objects from the backend so that
tests can access the data through a uniform interface.

- Revert a recent change to make the empty string a valid identifier
again (so we correctly 404 when someone provides a non-existant ID as a
parameter, not a different error).

Also, verifies that issues #531 and #535 have been resolved.


dcolligan:
+1

PR #560, opened 2015-08-10
Keep PyCharm/IntelliJ module files out of the repo.

hjellinek:
The PyCharm/IntelliJ IDEA IDE creates *.iml files to describe projects in the same way it uses the .idea directory, which is already listed in the .gitignore.


jeromekelleher:
+1.

PR #563, opened 2015-08-11
Search filters

jeromekelleher:
This PR fixes some miscellaneous issues relating to CallSets, and updates the top level object search code so that we always 404 when a missing ID is provided as an argument.

See the commit messages for the issues involved. I've left it as two commits as these deal with different issues.


dcolligan:
looks good +1

Issue #534, opened 2015-08-03
SearchReadGroupSetsRequest "name" field ignored by searchReadGroupSets?


hjellinek:
I'm using data from the server unit tests as the server's data.  The files are arranged like so:
```
dataset1/
    reads/
        1kg-low-coverage/
            ... BAM files ...
        wgBam/
            ... BAM files ...
    variants/
```

**Assumption**: each subdirectory in `reads` corresponds to a read group set.

Per the `SearchReadGroupSetsRequest` docs: "_Only return read group sets with this name (case-sensitive, exact match)._"  I create a search request with `datasetId` = `dataset1`; `name` = `"wgBam"`.

**Expected**:
The call to `searchReadGroupSets` will return a `SearchReadGroupSetsResponse` with its `readGroupSets` field containing a single read group set corresponding to `wgBam`.

**Actual**:
The call to `searchReadGroupSets` returns a `SearchReadGroupSetsResponse` with its `readGroupSets` field containing two read group sets, one apparently for `1kg-low-coverage` and one for `wgBam`.

jeromekelleher:
Thanks for this @hjellinek, we missed this one. @dcolligan, looks we'll need something a bit more complicated for the ``readGroupSetsGenerator`` (paging is so tedious...).

jeromekelleher:
An option here if this is too complicated to hit by the milestone is to raise a ``NotImplemented`  if someone specifies the name.

macieksmuga:
@jeromekelleher, @dcolligan, @diekhans:  after discussing this with @adamnovak, @hjellinek and @hannes-ucsc, I see three, maybe four choices:

1. **Mark problem areas with `NotImplemented`** as you suggest above and move on. I'm very reluctant to adopt this solution, though, as it just kicks the can down the road. The overall purpose of baseline, after all, was to arrive at a set of schemas that we **do** have a full implementation of.
2. **Revise the baseline schedule** to give us time to deal properly with these issues. This is really a non-starter, considering that the baselining task has become a bottleneck for all other development, but I needed to list it anyway.
3. (no I'm not about to make a joke) **Remove paging from the schema methods**. Some form of paging could still be implemented by servers & clients, of course, as could streaming or some other mix of technical solutions to ensure the reliable transmission of large datasets. I'm just not convinced that it should be embedded into the schemas proper, where it gets in the way of the reference implementation (and presumably other implementations, that may wish to take a different approach to resumable transmission).

[...]

PR #564, opened 2015-08-11
Search by name support in readGroupSets + refactor.

jeromekelleher:
- Refactored to make ID search fully hierarchical
- Added support for searching ReadGroupSets by name (issue #534)
- Generalised the simulator initialisation structures
- Removed multiple locations for initialising internal maps, preferring
  the use of addDataset, addVariantSet, etc.
- Fixed up some test cases that depended on brittle interfaces.

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -8,7 +8,6 @@ from __future__ import unicode_literals
 
 import os
 import json
-import random
 
 import ga4gh.datamodel as datamodel
 import ga4gh.datamodel.datasets as datasets
@@ -218,14 +217,30 @@ class AbstractBackend(object):
     This class provides methods for all of the GA4GH protocol end points.
     """
     def __init__(self):
-        self._referenceSetIdMap = {}
-        self._referenceSetIds = []
         self._requestValidation = False
         self._responseValidation = False
         self._defaultPageSize = 100
         self._maxResponseLength = 2**20  # 1 MiB
         self._datasetIdMap = {}
         self._datasetIds = []
+        self._referenceSetIdMap = {}
+        self._referenceSetIds = []
+
+    def addDataset(self, dataset):
+        """
+        Adds the specified dataset to this backend.
+        """
+        id_ = dataset.getId()
+        self._datasetIdMap[id_] = dataset
+        self._datasetIds.append(id_)
+
+    def addReferenceSet(self, referenceSet):
+        """
+        Adds the specified reference set to this backend.
+        """
+        id_ = referenceSet.getId()
+        self._referenceSetIdMap[id_] = referenceSet
+        self._referenceSetIds.append(id_)
 
     def setRequestValidation(self, requestValidation):
         """
@@ -339,6 +354,13 @@ class AbstractBackend(object):
                 nextPageToken = str(currentIndex)
             yield object_.toProtocolElement(), nextPageToken
 
+    def _singleObjectGenerator(self, datamodelObject):
+        """
+        Returns a generator suitable for a search method in which the
+        result set is a single object.
+        """
+        yield (datamodelObject.toProtocolElement(), None)
+
     def datasetsGenerator(self, request):
         """
         Returns a generator over the (dataset, nextPageToken) pairs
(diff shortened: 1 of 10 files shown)


ga4gh/backend.py
```diff
@@ -339,6 +354,13 @@ def _topLevelObjectGenerator(self, request, idMap, idList):
                 nextPageToken = str(currentIndex)
             yield object_.toProtocolElement(), nextPageToken
 
+    def _singleObjectGenerator(self, datamodelObject):
+        """
+        Returns a generator suitable for a search method in which the
+        result set is a single object.
+        """
+        return [(datamodelObject.toProtocolElement(), None)]
```
dcolligan: this doesn't look like it returns a generator

jeromekelleher: We can wrap an iterator around it if you like to force it to be a generator. Once it's an iterable it doesn't matter. Maybe I should change the comment to "returns an iterable..."?

dcolligan: yeah, I wasn't concerned about semantics, more about clarity... if a comment says a method returns a generator and it doesn't, it's confusing

ga4gh/backend.py
```diff
@@ -740,5 +765,4 @@ def __init__(self, dataDir):
             directory != referencesDirName]
         for datasetDir in datasetDirs:
             dataset = datasets.FileSystemDataset(datasetDir)
-            self._datasetIdMap[dataset.getId()] = dataset
-        self._datasetIds = sorted(self._datasetIdMap.keys())
```
dcolligan: Are we throwing away the assumption that the datasets are in some kind of order here?  We control the dataset names in the simulated backend, but not in the file system backend.  Should we care?

ga4gh/datamodel/datasets.py
```diff
@@ -130,9 +147,7 @@ def __init__(self, dataDir):
             if os.path.isdir(relativePath):
                 variantSet = variants.HtslibVariantSet(
                     self, localId, relativePath)
-                self._variantSetIdMap[variantSet.getId()] = variantSet
-        self._variantSetIds = sorted(self._variantSetIdMap.keys())
```
dcolligan: same question about sorting here

ga4gh/datamodel/datasets.py
```diff
@@ -140,8 +155,4 @@ def __init__(self, dataDir):
             if os.path.isdir(relativePath):
                 readGroupSet = reads.HtslibReadGroupSet(
                     self, localId, relativePath)
-                self._readGroupSetIdMap[readGroupSet.getId()] = readGroupSet
-                for readGroup in readGroupSet.getReadGroups():
-                    self._readGroupIdMap[readGroup.getId()] = readGroup
-        self._readGroupSetIds = sorted(self._readGroupSetIdMap.keys())
-        self._readGroupIds = sorted(self._readGroupIdMap.keys())
```
dcolligan: and here

ga4gh/backend.py
jeromekelleher: I don't think it really matters. There's no requirement in the protocol that these objects be returned in any particular order, so we're not constrained by that. The only issue I can think of potentially causing problems is with paging over multiple processes. If (somehow) two different processes have these lists in different orders, and subsequent page requests get routed to different processes we'll have a bad time.

This seems quite unlikely, but perhaps it's possible. One option is to use the bisect module to insert the ids into the list in sorted order. Another would be to abstract out the file-system walking code and to make sure that files/directories are visited in sorted order. What do you think, is this worth worrying about?

jeromekelleher:
@dcolligan, pushed an update. I'm not sure what to do about the sortedness requirement --- we can put it in now, or delay thinking about it properly to another day with an issue.

dcolligan:
If what we should do isn't clear, than I have no problem +1'ing this.

Another consideration is if we want to have the ordering stable across server restarts.  I think that's a good property, but I also don't think it's a guarantee that the OS will return the directory names to us in the same order every time.  Something to think about when it becomes an issue...

Issue #532, opened 2015-07-31
Fix the callSetId semantics of getVariants for the current API version


jeromekelleher:
The getVariants method in ``ga4gh/datamodel/variants.py`` currently has a TODO item to fix the semantics of ``callSetIds`` parameter for v0.6. This should be fixed, and the 0.5 semantics removed.

jeromekelleher:
Fixed in #566.

PR #566, opened 2015-08-12
Change getVariant behavior to conform to 0.6

dcolligan:
Issue #532

### DIFF
--- a/tests/datadriven/test_variants.py
+++ b/tests/datadriven/test_variants.py
@@ -189,13 +189,10 @@ class VariantSetTest(datadriven.DataDrivenTest):
 
     def testSearchCallSetIdsSystematic(self):
         for sampleIds in utils.powerset(self.vcfSamples, maxSets=10):
-            # TODO remove this for protocol 0.6
-            if len(sampleIds) == 0:
-                sampleIds = self.vcfSamples
             self._verifyVariantsCallSetIds(None, list(sampleIds))
 
     def testVariantsValid(self):
-        end = 2**30  # TODO This is arbitrary, and pysam can choke. FIX!
+        end = datamodel.PysamDatamodelMixin.vcfMax
         for referenceName in self._referenceNames:
             iterator = self._gaObject.getVariants(
                 referenceName, 0, end)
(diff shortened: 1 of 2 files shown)


tests/datadriven/test_variants.py
```diff
@@ -187,15 +186,8 @@ def _verifyVariantsCallSetIds(self, searchVariants, searchsampleIds):
     def testSearchAllVariants(self):
         self._verifyVariantsCallSetIds(None, self.vcfSamples[:1])
 
-    def testSearchCallSetIdsSystematic(self):
```
jeromekelleher: Why remove this whole test case? I think the commend just refers to the ``len(callSetIds) == 0`` bit. The test case should still be good.

tests/datadriven/test_variants.py
```diff
@@ -349,6 +341,44 @@ def convertPyvcfNumber(number):
...
+        end = datamodel.PysamDatamodelMixin.vcfMax
+        callSetIds = variantSet.getCallSetIds()
+        someCallSetIds = callSetIds[0:3]
+        for referenceName in self._referenceNames:
+            # passing None as the callSetIds argument should be equivalent
+            # to passing all of the possible callSetIds as an argument
+            noneRecords = variantSet.getVariants(
+                referenceName, start, end, None, None)
+            allRecords = variantSet.getVariants(
+                referenceName, start, end, None, callSetIds)
```
jeromekelleher: need to check that ``len(noneRecords) == len(allRecords)`` here (zip will run to the shortest of the two iterators).

tests/datadriven/test_variants.py
```diff
@@ -349,6 +341,44 @@ def convertPyvcfNumber(number):
...
+            for record in emptyRecords:
+                self.assertEqual(len(record.calls), 0)
+
+            # passing some callSetIds as the callSetIds argument should
+            # return only those calls
+            someRecords = variantSet.getVariants(
+                referenceName, start, end, None, someCallSetIds)
+            for record in someRecords:
+                for call in record.calls:
+                    compoundId = datamodel.CallSetCompoundId(
```
jeromekelleher: I think we can zip(record.calls, someCallSetIds) and assert the ID of the call is equal to the specified ID. The protocol requires that the calls be returned in the same order that they are specified in, so this is safe.

jeromekelleher:
Looks good @dcolligan, thanks. +1 minor modulo the minor comments above.

dcolligan:

tests/datadriven/test_variants.py
```diff
@@ -349,6 +346,45 @@ def convertPyvcfNumber(number):
...
+                referenceName, start, end, None, [])
+            for record in emptyRecords:
+                self.assertEqual(len(record.calls), 0)
+
+            # passing some callSetIds as the callSetIds argument should
+            # return only those calls
+            someRecords = variantSet.getVariants(
+                referenceName, start, end, None, someCallSetIds)
+            for record in someRecords:
+                for call, someId in zip(record.calls, someCallSetIds):
```
jeromekelleher: Still need to check the ``len``s of these lists: they could be unequal.

tests/datadriven/test_variants.py
```diff
@@ -349,6 +346,45 @@ def convertPyvcfNumber(number):
...
+                    compoundId = datamodel.CallSetCompoundId(
+                        variantSetCompoundId, call.callSetName)
```
jeromekelleher: Why bother with the compound ID here? Why not just
```
self.assertEqual(call.id, someId)
```

dcolligan:

jeromekelleher:
Thanks @dcolligan, merging.

PR #568, opened 2015-08-12
ID obfuscation via base64 encoding.

jeromekelleher:
This obfuscates our hierarchical breadcrumb-trail ID format from clients. Clients should not depend on our internal ID values, which are private to a given server instance.  This discourages them from doing so.

### DIFF
--- a/ga4gh/datamodel/__init__.py
+++ b/ga4gh/datamodel/__init__.py
@@ -6,13 +6,14 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import glob
-import os
-import json
-import tempfile
-import shutil
 import atexit
+import base64
 import collections
+import glob
+import json
+import os
+import shutil
+import tempfile
 
 import ga4gh.exceptions as exceptions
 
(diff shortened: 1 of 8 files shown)


ga4gh/datamodel/__init__.py
```diff
@@ -177,11 +177,39 @@ def parse(cls, compoundIdStr):
...
+        Mildly obfuscates the specified ID string in an easily reversible
+        fashion. This is not intended for security purposes, but rather to
+        dissuade users from depending on our internal ID structures.
+        """
+        return base64.b64encode(idStr)
+
+    @classmethod
+    def deobfuscate(cls, idStr):
+        """
+        Reverses the obfuscation done by the :meth:`obfuscate` method.
```
dcolligan: are the colons and backticks sphnix syntax?

tests/datadriven/test_variants.py
```diff
@@ -97,7 +97,6 @@ def _verifyVariantCallEqual(self, gaCall, pyvcfCall):
         genotype, phaseset = variants.convertVCFGenotype(
             pyvcfCall.data.GT, pyvcfCall.phased)
         # callSetId information is not available in pyvcf.model._Call
-        self.assertIn(pyvcfCall.sample, gaCall.callSetId)
```
dcolligan: why delete this line?

ga4gh/datamodel/__init__.py
jeromekelleher: Yep.

tests/unit/test_backends.py
```diff
@@ -190,47 +194,8 @@ class TestFileSystemBackend(TestAbstractBackend):
     """
     def setUp(self):
         self._dataDir = os.path.join("tests", "data")
-        self._referencesDir = os.path.join(self._dataDir, "references")
```
dcolligan: what's up with all the deletions in this file?

tests/datadriven/test_variants.py
jeromekelleher: There's no longer a simple relationship between the sample name and the callSetId. There's no reason there should be, so I removed the check.

tests/unit/test_backends.py
jeromekelleher: We're not using any of these --- it's confusing to have a bunch of code lying around that's not being used for anything.

dcolligan:
+1

jeromekelleher:
@dcolligan merge me!

Issue #312, opened 2015-03-31
Script required to generate `ga4gh-example-data` from 1000G sources


jeromekelleher:
The current ga4gh-example-data tarball contains a subset of the variants for 1000G phase1 and phase3. To make things managable, we also use only 3 samples, and throw away the others. We include the low coverage reads for these samples also.

Generating this dataset manually is tedious and error prone, and we should seek to automate the process using a script. The script should do the following:

1. Open remote connections to the 1000G VCF files at either the EBI or NCBI mirror (it should be command line option which one we use. Probably default to NCBI) using PySAM.
2. For each VCF file, extract the first *n* records and write these out to our example data directory. We should also only write out sample information for a small subset of the available samples. The samples to use should also be a configuration option. Record the positions of the first and last variant we see.
3. For each sample we are interested in, open a remote connection to the relevant BAM file, and retrieve all reads in the regions that intersect with the variants we extracted in (2). Save these reads into the `reads` section of our example data tree.
4. Generate the corresponding references by download the regions we are interested and saving these to FASTA files.

This script should allow us to generate a manageable subset of the 1000G data which we can use for development and demonstration purposes.

We require VCF write functionality in pysam for this, which is coming soon. I'm marking this issue as blocked until this is released.

jeromekelleher:
I've added this to the Sync with schemas milestone as we need better example data, and this is the simplest way to do it. If pysam VCF write support doesn't arrive soon we should make a shell script to do this using Unix tools.

jeromekelleher:
Thinking about this again, we don't actually need VCF write support in pysam. We can use the Tabix interface to obtain the raw VCF records, and output these directly.

dcolligan:
Partially addressed in #570

dcolligan:
Fixed in #581, etc.

Issue #521, opened 2015-07-30
Raise duplicate sample id problem with htslib


jeromekelleher:
The  `testInstantiation` in `tests.unit.test_faultydata.TestDuplicateCallSetId` is currently disabled because it provokes an error that results in htslib aborting. This is a bug in htslib. We should:

(a) track down where in htslib this happens, and confirm the problem still exists; and 
(b) open an issue with htslib to get it fixed.

david4096:
Verified by turning back on the test. `htslib` @jmarshall crashed with:

The VCF in question:  https://github.com/ga4gh/server/tree/master/tests/faultydata/variants/duplicated_sampleid

```
(env)➜  server-1 git:(1253_codecov) ✗ nosetests tests.unit.test_faultydata
[E::bcf_hdr_add_sample] Duplicated sample name 'S1'
[1]    39488 abort      nosetests tests.unit.test_faultydata
```

jmarshall:
Yes, that's because it considers this malformed `#CHROM` header line terrifying enough that it prints the error message and aborts.  Did you ever file an htslib issue about this?  Then we wouldn't forget about it and would probably calm `bcf_hdr_add_sample()` down.

david4096:
Issue opened, we'll leave this here to help track progress on the htslib side.

PR #570, opened 2015-08-12
Script to download data from NCBI, etc.

dcolligan:
Still a WIP; not very well factored right now

This current code will download the index files for, write, compress, and index all the VCFs from a given site (only NCBI supported right now) in about 3 minutes.  Unfortunately, it takes a few tries since htslib segfaults a lot...

Question for @jeromekelleher ... which (read) samples do we want to download?  There's quite a lot of them and I expect we want to download a subset...

Issue #312

### DIFF
--- a/scripts/download_data.py
+++ b/scripts/download_data.py
@@ -1,0 +1,294 @@
+"""
+Constructs a data source for the ga4gh server by downloading data from
+authoritative remote servers.
+"""
+# TODO
+# - need some kind of checkpoint functionality to resume process
+#   where it left off since getting a clean run is so rare...
+# - references support
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+import glob
+import gzip
+import os
+import tempfile
+import urllib2
+
+import pysam
+
+import utils
+
+
+def mkdirAndChdirList(dirNameList):
+    """
+    For each entry in dirNameList, make a directory if it does not exist and
+    cd into it
+    """
+    for directory in dirNameList:
+        mkdirAndChdir(directory)
+
+
+def mkdirAndChdir(dirName):
+    """
+    Make a directory if it does not exist and cd into it
+    """
+    if not os.path.exists(dirName):
+        os.mkdir(dirName)
+    os.chdir(dirName)
+
+
+def cleanDir():
+    """
+    Removes genomic files in current directory
+    """
+    cwd = os.getcwd()
+    utils.log("Cleaning out directory '{}'".format(cwd))
+    globs = [
+        "*.tbi", "*.vcf", "*.vcf.gz", "*.bam", "*.bam.bai", "*.fasta.gz",
+        "*.fasta", "*.fasta.fai"]
+    for fileGlob in globs:
+        fileNames = glob.glob(fileGlob)
+        for fileName in fileNames:
+            os.remove(fileName)
+
+
+def escapeDir(levels=4):
+    # back to orig dir
+    for _ in range(levels):
+        os.chdir('..')
+
+
+class ChromMinMax(object):
+    """
+    A container class for storing the min and max position seen
+    for every chromosome
+    """
+    defaultMinPos = 2**30
+    defaultMaxPos = 0
+
+    class MinMax(object):
+        def __init__(self):
+            self.minPos = ChromMinMax.defaultMinPos
+            self.maxPos = ChromMinMax.defaultMaxPos
+
+    def __init__(self):
+        self.chromMap = {}
+
+    def addPos(self, chrom, position):
+        if chrom not in self.chromMap:
+            self.chromMap[chrom] = self.MinMax()
+        minMax = self.chromMap[chrom]
+        if position > minMax.maxPos:
+            minMax.maxPos = position
+        if position < minMax.minPos:
+            minMax.minPos = position
+
+    def getMinPos(self, chrom):
+        minMax = self.chromMap[chrom]
+        return minMax.minPos
+
+    def getMaxPos(self, chrom):
+        minMax = self.chromMap[chrom]
+        return minMax.maxPos
+
+
+class AbstractFileDownloader(object):
+    """
+    Base class for individual site genome file downloaders
+    """
+    def __init__(self, args):
+        self.args = args
+        self.datasetName = 'dataset1'
+        self.variantSetName = self.args.source
+        self.chromMinMax = ChromMinMax()
+        self.chromosomes = self.args.chromosomes.split(',')
+
+    def _getVcfFilenames(self):
+        baseFileName = (
+            "ALL.chr{}.phase3_shapeit2_mvncall_integrated_v5a"
+            ".20130502.genotypes.vcf.gz")
+        chrNames = [
+            str(i) for i in range(1, 23)
+            if str(i) in self.chromosomes]
+        fileNames = [baseFileName.format(chrName) for chrName in chrNames]
+        # the X and Y files use a different filename prefix
+        if 'X' in self.chromosomes:
+            fileNames.append(
+                'ALL.chrX.phase3_shapeit2_mvncall_integrated_v1a.'
+                '20130502.genotypes.vcf.gz')
+        if 'Y' in self.chromosomes:
+            fileNames.append(
+                'ALL.chrY.phase3_integrated_v1a.'
+                '20130502.genotypes.vcf.gz')
+        return fileNames
+
+    def getVcfBaseUrl(self):
+        return os.path.join(self.getBaseUrl(), 'ftp/release/20130502/')
+
+    def getBamBaseUrl(self):
+        return os.path.join(self.getBaseUrl(), 'ftp/phase3/data/')
+
+    def _prepareDir(self):
+        dirList = [
+            self.args.dir_name, self.datasetName, 'variants',
+            self.variantSetName]
+        mkdirAndChdirList(dirList)
+        cleanDir()
+
+    def _updatePositions(self, fileName):
+        localVariantFile = pysam.VariantFile(fileName)
+        localIterator = localVariantFile.fetch()
+        for record in localIterator:
+            self.chromMinMax.addPos(record.chrom, record.start)
+        localIterator = None
+        localVariantFile.close()
+        utils.log('chrom: {}, maxPos: {}, minPos: {}'.format(
+            record.chrom, self.chromMinMax.getMaxPos(record.chrom),
+            self.chromMinMax.getMinPos(record.chrom)))
+
+    def _processFileName(self, fileName):
+        url = os.path.join(self.getVcfBaseUrl(), fileName)
+        utils.log("Downloading '{}'".format(url))
+        response = urllib2.urlopen(url)
+        megabyte = 1024 * 1024
+        data = response.read(megabyte)
+        lineCountQuota = 1000
+        localFileName, _ = os.path.splitext(fileName)
+        utils.log("Writing '{}'".format(localFileName))
+        with tempfile.NamedTemporaryFile() as binaryFile:
+            binaryFile.write(data)
+            binaryFile.flush()
+            gzipFile = gzip.open(binaryFile.name, "r")
+            outputFile = open(localFileName, "w")
+            lineCount = 0
+            for line in gzipFile:
+                outputFile.write(line)
+                if not line.startswith("#"):
+                    lineCount += 1
+                if lineCount >= lineCountQuota:
+                    break
+            assert lineCount == lineCountQuota
+            outputFile.close()
+            gzipFile.close()
+        utils.log("Compressing '{}'".format(localFileName))
+        utils.runCommand('bgzip -f {}'.format(localFileName))
+        utils.log("Indexing '{}'".format(fileName))
+        utils.runCommand('tabix {}'.format(fileName))
+        self._updatePositions(fileName)
+
+    def downloadVcfs(self):
+        self._prepareDir()
+        fileNames = self._getVcfFilenames()
+        for fileName in fileNames:
+            self._processFileName(fileName)
+        escapeDir()
+
+    def downloadBams(self):
+        dirList = [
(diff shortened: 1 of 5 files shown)


jeromekelleher:
Looks good @dcolligan, thanks. Sorry to hear about the htslib 'error handling' issues.

The samples we're already using is probably the best place to start (HG00096, HG00533 and HG00534).

jmarshall:
@dcolligan: Want to give us a hint what the circumstances of these alleged segfaults are?  (Perhaps you are referring to the GIGO in #521?)

dcolligan:
@jmarshall Sure, they seem to be a bit nondeterministic; sometimes the script finishes fine, other times not.  Here's one example: https://gist.github.com/dcolligan/552b95cf7e6b41be9b3f I'll report others as I see them.

jmarshall:
That's an `Illegal instruction`, not a segfault.  I'll investigate further in htslib, but at first glance encountering non-determistic illegal instructions points more to a problem with the low-level Python wrapping in pysam rather than to htslib itself.

jeromekelleher:
Thanks for looking into this @jmarshall, I'll see if I hit the problem and get a gdb trace if possible.

@dcolligan, I wonder if we need to do the tabix file fetching at all actually. All we really need is the first (say) 1000 variants from each file, plus the headers. It should be possible to do this just opening a standard http connection and writing the results to an output  file. Then, once we have these files we can get the coordinate of the last variant in each file, and use this to define the subset of the reference we want to download, as well as the reads we want to retrieve (which we'll need pysam for, definitely).

Does this make sense?

dcolligan:
Yes, but that would require manually parsing the VCF file, which has generally been something we have avoided.

jeromekelleher:
Well, just finding the end of the header isn't that big a deal. I think we'll have to do that anyway with the tabix method, as there's no headers in the VCFs currently being downloaded. It's up to you though, whatever works out simplest.

dcolligan:
@jeromekelleher pushed update with reads downloads

@jmarshall two additional illegal instruction errors, with this version of the code: 
https://gist.github.com/dcolligan/850cb6f9e17a142ebc27
https://gist.github.com/dcolligan/76e25651c74b48ee4755

(I realize that illegal instruction errors and segfaults are different; I was getting segfaults occasionally as well; I will report those when I see them.)

scripts/download_data.py
```diff
@@ -0,0 +1,254 @@
...
+        baseUrl = (
+            'ftp://ftp-trace.ncbi.nih.gov/1000genomes/ftp/phase3/data/')
+        samples = self.args.samples.split(',')
+        for sample in samples:
+            samplePath = '{}/alignment/'.format(sample)
+            study = studyMap[sample]
+            fileName = (
+                '{}.mapped.ILLUMINA.bwa.{}.'
+                'low_coverage.20120522.bam'.format(sample, study))
+            sampleUrl = baseUrl + samplePath + fileName
```
jeromekelleher: `os.path.join` is useful for joining URLs too.

jeromekelleher:
I've hit a nondeterministic buffer overflow and abort on this version of the code also @dcolligan --- running it in gdb until I can reproduce and get a stacktrace.

jeromekelleher:
@jmarshall I reproduced the error, and got a stacktrace [here](https://gist.github.com/jeromekelleher/074ac7f7e0f0c95a7be4).

dcolligan:
@jeromekelleher pushed an update that downloads the reference

jeromekelleher:
Thanks @dcolligan. There's a couple of issues here:

- On my machine the header data from `remoteVariantFile.header` is corrupted, and this is causing later stuff to fail. I really think it would be easier to just skip this TabixFile phase and just iterate over the lines in the file. Here's what I've come up with:

```python
    def _processFileName(self, fileName):
        url = os.path.join(self._getBaseUrl(), fileName)
        response = urllib2.urlopen(url)
        megabyte = 1024 * 1024
        data = response.read(megabyte)
        # Write this to a temporary file, and filter it into a gzip.
        with tempfile.NamedTemporaryFile() as binaryFile:
            binaryFile.write(data)
            binaryFile.flush()
            gzipFile = gzip.open(binaryFile.name, "r")
            outputFile = open(fileName, "w")
            lineCount = 0
            for line in gzipFile:                
                outputFile.write(line)
                if not line.startswith("#"):
                    lineCount += 1
                if lineCount >= 1000:
                    break
            outputFile.close()
            gzipFile.close()
```
This just downloads the first megabyte of the file, and then un-gzips it to get an uncompressed tabix. We can then read this file in using pysam, and get the coordinate of the last variant using that. (Call it ``maxCoord`` for the purpose of this discussion.)

- We don't want the whole reference. I was hoping that we would be able to connect to the remote FASTAs and get the sequence for 0 to ``maxCoord`` + 1000 (say, we want a bit of a buffer on the other side of our reference), and save these into our local, mini reference. However, I haven't been able to get ``pysam.FastaFile`` to work with the remote references here. Maybe we should try other tools, or other sites?

jeromekelleher:
OK, I think there's a better approach for references. We can use the [ENA web apis]( http://www.ebi.ac.uk/ena/browse/data-retrieval-rest#sequence_region_fasta) to get the sequences we want, and then store these in FASTA format. 

So, I think we should get this script to the point where it can:

1) Download the first n records of VCFs, as above, and get the maximum coordinate for each chromsome.
2) Download the reads covering this interval for the samples we're interested in.

Then, hopefully we can merge this much and add the reference downloads afterwards.

jmarshall:
The salient part of @jeromekelleher's stack trace is `__fdelt_chk (d=-1)` / `socket_wait (fd=-1, …)`, which is down to knetfile.c trying to read from a non-existent socket.

I suspect what is going on here is that the FTP server is intermittently overloaded and refuses the connection, and with Heng's traditional level of error-checking knetfile.c carries on regardless rather than reporting EBUSY or similar.  This may account for the illegal instruction crashes too.

I haven't yet tested this at a heavy traffic time of day and observed it myself, or implemented better response checking in knetfile.c (which is not something I'm super-excited about doing).

jeromekelleher:
Thanks @jmarshall, this is good to know. Is this code still used with the libcurl based URL handling (i.e., has it been replaced in more recent version of htslib)? If not, then is there anything we can do to make people excited about fixing the error handling in htslib?

jmarshall:
This knetfile.c code (which is technically part of [klib](https://github.com/attractivechaos/klib) rather than htslib) is indeed not used by the libcurl-based network file access code, which checks for errors very carefully.  As you probably suspected, this is a major cause of the lack of excitement…

jeromekelleher:
@jmarshall Ah, good, that's reassuring. Roll on the libcurl based code!

dcolligan:
@jeromekelleher pushed an update: improved references handling by 1) unzipping data in memory on the fly as the file is downloaded 2) only downloading up to a certain specified number of bases

jeromekelleher:
That's good @dcolligan, but I don't think it really solves the problem. We need to be able to get the reference sequence from different chromosomes, so just downloading the first n bases doesn't solve that. We'd end up having to download the whole file, which is slow and tedious. The ENA web APIs look a lot simpler.

As I said above, I think it would be good to just get the VCFs and BAMs for now, and worry about the references in a follow-up.

dcolligan:
@jeromekelleher pushed a version that implements your strategy for the VCFs and removes references support.  It runs in ~30 mins and downloads ~7G of data.  Log of a full run is here: https://gist.github.com/dcolligan/1eefe10984ed245dd25f

jeromekelleher:
@dcolligan --- this looks great, thanks! I think all we need to do now is put a limit on the number of reads we download and we're good to go. There's no point in downloading 7G of data. It would be nice if we could randomly sample the reads covering this interval (like in ``samtools view -s FRACTION``), but just taking the first n will do fine if this functionality isn't available through pysam. (Or, it might turn out to be simpler to run samtools in a subprocess using the calculated intervals.)

dcolligan:
Ok, this new version finishes in 3 minutes, downloads ~100MB

scripts/download_data.py
```diff
@@ -0,0 +1,256 @@
...
+        # the X and Y files use a different filename prefix
+        fileNames.append(
+            'ALL.chrX.phase3_shapeit2_mvncall_integrated_v1a'
+            '.20130502.genotypes.vcf.gz')
+        fileNames.append(
+            'ALL.chrY.phase3_integrated_v1a.20130502.genotypes.vcf.gz')
+        return fileNames
+
+    def getVcfBaseUrl(self):
+        return self.getBaseUrl() + '/ftp/release/20130502/'
```
jeromekelleher: Use ``os.path.join`` here -- this is currently giving a ``//`` in the middle of the URL.

scripts/download_data.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def getBamBaseUrl(self):
+        return self.getBaseUrl() + '/ftp/phase3/data/'
```
jeromekelleher: Ditto here.

scripts/download_data.py
```diff
@@ -0,0 +1,256 @@
...
+            header = remoteFile.header
+            utils.log("Writing '{}'".format(fileName))
+            localFile = pysam.AlignmentFile(
+                fileName, 'wb', header=header)
+            for reference in remoteFile.references:
+                utils.log("reference {}".format(reference))
+                iterator = remoteFile.fetch(
+                    reference, start=self.minPos, end=self.maxPos)
+                numRecords = 0
+                for record in iterator:
```
jeromekelleher: Using 
```python
for index, record in enumerate(iterator):
     localFile.write(record)
     if index > args.num_reads:
         break
```
would be slightly more compact here. Doesn't make any real difference though, just a suggestion.

scripts/download_data.py
```diff
@@ -0,0 +1,256 @@
...
+                    if numRecords >= args.num_reads:
+                        break
+                utils.log("{} records written".format(numRecords))
+            iterator = None
+            remoteFile.close()
+            localFile.close()
+            baiFileName = fileName + '.bai'
+            os.remove(baiFileName)
+            utils.log("Indexing '{}'".format(fileName))
+            # TODO pysam.index(fileName) gives unicode error;
```
jeromekelleher: Try ``filename.encode()`` to convert to bytes?

scripts/download_data.py
```diff
@@ -0,0 +1,256 @@
...
+            sampleUrl = os.path.join(baseUrl, samplePath, fileName)
+            utils.log("Downloading index for '{}'".format(sampleUrl))
+            remoteFile = pysam.AlignmentFile(sampleUrl)
+            header = remoteFile.header
+            utils.log("Writing '{}'".format(fileName))
+            localFile = pysam.AlignmentFile(
+                fileName, 'wb', header=header)
+            for reference in remoteFile.references:
+                utils.log("reference {}".format(reference))
+                iterator = remoteFile.fetch(
```
jeromekelleher: Hold on --- what's `minPos` and `maxPos` here? These should be different for every chromosome.  We need to store the ``minPos`` and ``maxPos`` for every chromosome individually, not just take a global max and min. Also, there seems to be reads mapped to references ``GL000207.1`` (etc) as well as ``1``, ``2``, etc. We only want the reads mapped to ``1``, ``2``, etc.

scripts/utils.py
```diff
@@ -82,23 +77,83 @@ def _getFileNameDisplayString(self):
...
+        self.fileSize = int(response.headers['content-length'])
+        with open(self.path, 'wb') as outputFile:
+            for chunk in response.iter_content(chunk_size=self.chunkSize):
+                self.bytesReceived += self.chunkSize
+                self._updateDisplay()
+                outputFile.write(chunk)
+        self._cleanUp()
+
+
+class FtpFileDownloader(FileDownloader):
```
jeromekelleher: Are we using this now? If not we should remove it. It might be worth considering dropping the changes to this file entirely if we're not actually using them now .

jeromekelleher:
Looks good @dcolligan. I think we just need to fix some minor issues with the reads download, so we have a range specific to each chromosome, and we should merge then. We can add the references in a follow up.

Also, do we really need all of the chrs? Perhaps [1, 2, 3, X, Y] would do?

dcolligan:
@jeromekelleher ok, updated... down to sub-1 min runtime

scripts/download_data.py
```diff
@@ -0,0 +1,290 @@
...
+        return minMax.maxPos
+
+
+class AbstractFileDownloader(object):
+    """
+    Base class for individual site genome file downloaders
+    """
+    def __init__(self, args):
+        self.args = args
+        self.datasetId = 'dataset1'
```
jeromekelleher: I guess technically this should be ``datasetName`` now, as the IDs are server defined. Likewise for the variantSet and ReferenceSet.

scripts/download_data.py
```diff
@@ -0,0 +1,290 @@
...
+    parser.add_argument(
+        "--dir-name", default="ga4gh-downloaded-data",
+        help="the name of the directory that the data is downloaded to")
+    parser.add_argument(
+        "--samples", default='HG00096,HG00533,HG00534',
+        help="a comma-seperated list of samples to download")
+    parser.add_argument(
+        "--num-reads", default=1000,
+        help="the number of reads to download per reference")
+    parser.add_argument(
```
jeromekelleher: We should use this for the VCFs we download too: these are the chromosomes we're interested in, so we download the VCF, get the reads, and download the reference (later). So, I would change this to ``--chromosomes``. If the X and Y are awkward because of different file naming, then just forget about them

dcolligan:
@jeromekelleher alright, down to 25sec runtime, 100MB download, only downloading relevant chromosomes

jeromekelleher:
Superb @dcolligan! 

+1 after squashing. I'll open some issues on what needs to be done afterwards.

dcolligan:
@jeromekelleher squashed

dcolligan:
@jeromekelleher can be merged now

jeromekelleher:
Merged; thanks @dcolligan.

Issue #526, opened 2015-07-30
API versioning and Content-Type


jeromekelleher:
Currently, the schemas provide very little specification on the details of the protocol at the HTTP level. One aspect that is currently not well defined is how API versioning occurs and how the HTTP Content-Type is used. This  should, of course, be documented. In the mean time, since we are moving to using the reference server as the de-facto definition of how the protocol is supposed to behave we should agree here on what the correct behaviour is and implement it. Documentation can follow.

In the server, we currently use a HTTP Content-Type of  `application/json` and have a simple URL based API versioning scheme (see #161 for the initial discussion on this). The URL based scheme has proved to be quite awkward in practice, and semantically ambiguous from a REST perspective (are we serving version X of the object, or is this version X of the protocol? Fundamentally, the same URL should point to the same resource, and URL based API versioning breaks this.).  Also, we need to update our Content-Type if we are to serve JSON and a binary encoding of the data (i.e., protobuf).

I propose we follow the approach taken by the MatchMaker team and provide  API version negotiation via the HTTP Accept and Content-Type headers. See [here](https://github.com/ga4gh/mme-apis/blob/develop/search-api.md#versioning) for details on how the MatchMaker API has done this. We don't need to implement version negotiation on the server yet, but we should provide the framework within which it can be done later when the schemas settle down.

So, to be concrete, I would like to
1. Remove the API version prefix from our URLs;
2. Change the Content-Type to something like `application/ga4gh.v0.6+json` and `application/ga4gh.v0.6+proto`

Any thoughts/objections? @diekhans, what do you think of this? 

@calbach, how does Google manage this? Do you agree we should adopt a common approach to content type negotiation and API versioning?

diekhans:
I am adding a section on API version negotiation to the
streaming protocol specification.

diekhans:
One issue, which I am not finding any good guidance on is how to
specify version matching in accepts header.

So lets say we use a 3-part version number:
<major>.<minor>.<bugfix>.  Backwards and forwards compatibility
is assumed within a minor version but not between major ones.

So lets say we have releases version 1.1 and there is a bug fix
in 1.1 that is needed.2.  We want to be able to accept version
1.1.2 or greater, but not version 2.*, which might not be
compatible.

Any ideas on specifying this in request headers?  I am surprised
to not find this as a commonly address issue.

jeromekelleher:
[...]

PR #571, opened 2015-08-13
Removed URL versioning.

jeromekelleher:
This PR removes the URL based versioning mechanism, and is the first step towards a content-type negotiation schema discussed in #526.

Since URL versioning is not part of the schema, and we are planning on replacing this with a header-based approach, it is a good idea to remove this URL based approach now rather than later. Since this is the reference implementation, other implemetors might assume that URL based API versioning is required.

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -45,10 +45,10 @@ class NoConverter(werkzeug.routing.BaseConverter):
 
     This is needed because if there are e.g. two routes:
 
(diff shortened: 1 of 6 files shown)


diekhans:
+1

dcolligan:
+1 :cry:

jeromekelleher:
Thanks @dcolligan --- sorry to nuke all that work!

PR #572, opened 2015-08-13
Tidy-up reference handling.

jeromekelleher:
- Added general range checking for ListReferenceBases, and raise a HTTP
  416 when these are out of bounds.
- Updated the object models for ReferenceSet and References, and
  generally refactored reference.py.
- Updated the simulated object models, and tested the translation of
  these through the protocol.
- Added more tests for ListReferenceBases.

### DIFF
--- a/ga4gh/datamodel/__init__.py
+++ b/ga4gh/datamodel/__init__.py
@@ -381,18 +381,6 @@ class PysamDatamodelMixin(object):
         return referenceName, start, end
 
     @classmethod
-    def sanitizeFastaFileFetch(cls, start=None, end=None):
-        if start is not None:
-            start = cls.sanitizeInt(
-                start, cls.fastaMin, cls.fastaMax, 'start')
-        if end is not None:
-            end = cls.sanitizeInt(
-                end, cls.fastaMin, cls.fastaMax, 'end')
-        if start is not None and end is not None:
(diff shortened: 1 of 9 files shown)


ga4gh/datamodel/__init__.py
```diff
@@ -381,18 +381,6 @@ def sanitizeAlignmentFileFetch(
         return referenceName, start, end
 
     @classmethod
-    def sanitizeFastaFileFetch(cls, start=None, end=None):
```
dcolligan: why are we removing this?  it looks like we're still calling into fastaFile.fetch ...

dcolligan:
looks good, generally

ga4gh/datamodel/__init__.py
jeromekelleher: There didn't seem to be much point to it --- ``checkQueryRange`` makes sure the values make sense and are in the correct range, so we'd just be doing the same thing again with the sanitize call. The values are already guaranteed to be ints, and they're in the correct range, so we're as safe as we're going to get.

jeromekelleher:
@dcolligan --- I've pushed an update on this to include some test cases for ListReferenceBases paging.

dcolligan:
+1

jeromekelleher:
@dcolligan, merge me!

PR #575, opened 2015-08-14
Complete refactor to explicit getter methods.

jeromekelleher:
This PR completes the process of refactoring client-facing getter methods to within their actual containers. This is safer than passing around maps, as there is a single location where we must raise the correct exception. This also gives a better interface, as we can change the internal indexing structures without affecting client code.

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -52,30 +52,6 @@ def _parsePageToken(pageToken, numValues):
     return values
 
 
-def _getVariantSet(request, variantSetIdMap):
-    variantSetId = request.variantSetId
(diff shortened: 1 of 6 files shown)


ga4gh/backend.py
```diff
@@ -202,19 +165,18 @@ def _getStart(cls, readAlignment):
 
     @classmethod
     def _getEnd(cls, readAlignment):
-        return cls._getStart(readAlignment) + \
-            len(readAlignment.alignedSequence)
+        return (
```
dcolligan: kitten was saved here, phew

dcolligan:
+1

PR #577, opened 2015-08-17
Remove external map access

jeromekelleher:
This removes the need for passing around references to internal
structures, and allows us to program against an interface rather than an
internal representation.

Depends on PR  #575, which should be reviewed first.


dcolligan:
could you rebase this on top of `develop` so only the new changes appear in the diff?

jeromekelleher:
@dcolligan --- done.

dcolligan:
+1

PR #584, opened 2015-08-20
Temporary fix to cleanup htslib-downloaded indexes

### LINKED ISSUE #500: Local index files from remote fetch not getting cleaned up property
On the current `develop` branch:
```
$ git status -s
$ nosetests tests/end_to_end/test_remote_fetch.py
..
Ran 2 tests in 3.429s

OK
$ git status -s
?? chr1.vcf.gz.tbi
?? wgEncodeUwRepliSeqBg02esG1bAlnRep1_sample.bam.bai
```
We should investigate why these downloaded index files are not getting cleaned up properly.

dcolligan:
Does not resolve #500 ... that's blocked because pysam does not yet have the hooks into the htslib methods we need to write remote index files where we want them to go.

### DIFF
--- a/tests/end_to_end/test_remote_fetch.py
+++ b/tests/end_to_end/test_remote_fetch.py
@@ -5,6 +5,9 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import glob
+import os
+
 import client as client
 import server as server
 import server_test as server_test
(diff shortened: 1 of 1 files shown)


tests/end_to_end/test_remote_fetch.py
```diff
@@ -42,6 +45,12 @@ def _assertLogsWritten(self):
         # client should have gotten some reads back
         self.assertEqual(len(self.client.getOutLines()), 10)
 
+    # TODO take out when #500 is fixed
+    def tearDown(self):
+        # clean up downloaded local index file
```
jeromekelleher: Can we spew something out into the log here maybe? I'm worried we'll forget about this and there'll be some unexpected nastiness...

jeromekelleher:
Yeah, this is a good temporary fix.

tests/end_to_end/test_remote_fetch.py
dcolligan: Which log?  nose will only show us output / logs for this test if it fails.

jeromekelleher: Could just do a ``print("Removing index files; see issue #500")``. This will show up when we do a ``nosetests -vs`` (which is how I usually use it anyway).

dcolligan:
@jeromekelleher updated with print statements

jeromekelleher:
Thanks @dcolligan, +1.

Merging.

Issue #585, opened 2015-08-24
De-obfuscate ids on html page


dcolligan:
Currently, with base64-obfuscated IDs, our html page looks like this:
```
Data

ReferenceSets

DataSets

ZGF0YXNldDE=

VariantSets

ZGF0YXNldDE6MWtnLXBoYXNlMQ==
ZGF0YXNldDE6MWtnLXBoYXNlMw==
ZGF0YXNldDE6dW1k
ReadGroupSets

ZGF0YXNldDE6cmVtb3Rl	
ZGF0YXNldDE6cmVtb3RlOndnRW5jb2RlVXdSZXBsaVNlcUJnMDJlc0cxYkFsblJlcDFfc2FtcGxl
ZGF0YXNldDE6ZGFudms=	
ZGF0YXNldDE6ZGFudms6Y2hyMTcuMS0yNTA=
ZGF0YXNldDE6bG93LWNvdmVyYWdl	
ZGF0YXNldDE6bG93LWNvdmVyYWdlOkhHMDAwOTYubWFwcGVkLklMTFVNSU5BLmJ3YS5HQlIubG93X2NvdmVyYWdlLjIwMTIwNTIy
ZGF0YXNldDE6bG93LWNvdmVyYWdlOkhHMDA1MzMubWFwcGVkLklMTFVNSU5BLmJ3YS5DSFMubG93X2NvdmVyYWdlLjIwMTIwNTIy
ZGF0YXNldDE6bG93LWNvdmVyYWdlOkhHMDA1MzQubWFwcGVkLklMTFVNSU5BLmJ3YS5DSFMubG93X2NvdmVyYWdlLjIwMTIwNTIy
```
I get that we don't want people relying on an ad-hoc convention, but perhaps there is a way to make this page more useful as a sanity check?  For instance, displaying the de-obfuscated ids alongside the obfuscated ones?  Thoughts?

jeromekelleher:
How about we use the names as well to make it more clear? So, everywhere we have an ID now, we'd put a pair (name, id) to make it more readable.

dcolligan:
Fixed in #586

PR #586, opened 2015-08-24
Add deobfuscated ids to html page

dcolligan:
Issue #585

### DIFF
--- a/ga4gh/templates/index.html
+++ b/ga4gh/templates/index.html
@@ -2,8 +2,11 @@
     <head>
         <title>GA4GH reference server {{ info.getServerVersion()}}</title>
         <link rel="shortcut icon" href="{{ url_for('static', filename='favicon.ico') }}">
+        <link rel=stylesheet type=text/css href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.min.css">
+        <script src="https://code.jquery.com/jquery-2.1.4.min.js"></script>
+        <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/js/bootstrap.min.js"></script>
     </head>
-    <body>
+    <body><div class="container">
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Why are we deobfuscating the IDs rather than just using (e.g) ``dataset.getName()``? The name is the human readable identifier, and the hierarchy should be clear from the display, so I don't see why the user needs to know what's inside our IDs (and, I think there's a clear opportunity for them to get confused by this). Perhaps we should have an explicit table with headers and borders, with **name** and **id** for each object?

dcolligan:
There is no `getName` method.  Are you suggesting adding one (perhaps on DatamodelObject) that returns the deobfuscated id?

jeromekelleher:
Sorry, that should have been ``getLocalId()`` --- this is what we're mapping to the ``name`` concept at the moment.

dcolligan:
@jeromekelleher updated

jeromekelleher:
This looks soooo much better! I really like this @dcolligan, thanks. It's easy to see how we can build this page into a really useful summary of the data from here.

There's one potentially sticky issue though: licensing-wise, are we OK shoving in the jquery source? It looks like jquery is MIT licensed, so I think we're OK. I'm no license lawyer though, so it would be nice to get some other opinions on this.

dcolligan:
Yeah, I'm not sure either.  One option is to source the js / css from CDNs but that can be annoying if developing locally, so I'd rather not do that.

jeromekelleher:
Any ideas who we should ask about licencing related issues? Anyone?

dcolligan:
Just to record this, it was resolved on the call that we should fetch the libs from a CDN rather than check them into our source code tree.

dcolligan:
@jeromekelleher js/css now loaded from CDN

jeromekelleher:
Thanks @dcolligan --- merging.

PR #589, opened 2015-08-26
588 readgroup stats

dcolligan:


ga4gh/datamodel/reads.py
```diff
@@ -251,6 +262,14 @@ def getSamFilePath(self):
         """
         return self._samFilePath
 
+    def getNumAlignedReads(self):
```
jeromekelleher: We should really have some docstring comment here for these methods. It is a bad idea having duplicated documentation in both subclasses, so I think the best thing to do is to define this methods in the superclass, along with appropriate docstring comments, and for them to raise a `NotImplementedError`. See `AbstractVariantSet.getNumVariants` for an example.

This is a good way of being explicit about the superclasses interface too I think.

jeromekelleher:
This looks good @dcolligan. We will have to modify this once we correct our ReadGroup interpretation, as we're not going to be able to get these values from pysam any more, and will have to read them from a pre-computed metadata file.

dcolligan:
@jeromekelleher updated.  I'd rather merge these minor issues before tackling the big refactor.

jeromekelleher:
Thanks @dcolligan, +1.

Merging.

Issue #587, opened 2015-08-25
Program fields in ReadGroup should be filled from @PG tags


jeromekelleher:
The ``ReadGroup.programs`` field is currently not being populated. This should be filled in using the SAM @PG tags, if present.

macieksmuga:
Considering that `ReadGroup` doesn't represent a BAM, whereas a `ReadGroupSet` does, shouldn't the `.programs` field be moved to the latter? (see #574)

jeromekelleher:
@macieksmuga --- yes, probably. This would be a  schemas issue though.

dcolligan:
Fixed in #590

PR #590, opened 2015-08-26
Add programs field to read groups

dcolligan:
Issue #587

### DIFF
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -186,7 +186,14 @@ class AbstractReadGroup(datamodel.DatamodelObject):
         stats.unalignedReadCount = self.getNumUnalignedReads()
         stats.baseCount = None  # TODO requires iterating through all reads
         readGroup.stats = stats
+        readGroup.programs = self.getPrograms()
         return readGroup
+
+    def getPrograms(self):
+        """
+        Returns an array of Programs used to generate this read group
+        """
+        raise NotImplementedError()
 
     def getReadAlignmentId(self, gaAlignment):
         """
@@ -227,6 +234,9 @@ class SimulatedReadGroup(AbstractReadGroup):
 
     def getNumUnalignedReads(self):
         return 0
+
+    def getPrograms(self):
+        return []
 
     def _createReadAlignment(self, i):
         # TODO fill out a bit more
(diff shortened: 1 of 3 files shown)


ga4gh/datamodel/reads.py
```diff
@@ -251,6 +255,22 @@ def getSamFilePath(self):
         """
         return self._samFilePath
 
+    def getPrograms(self):
```
jeromekelleher: This should be documented --- see the comments in #589 for the recommended approach (i.e., define and document in the superclass).

jeromekelleher:
Looks good @dcolligan --- we could do with some data driven tests though to make sure it's all sensible.

dcolligan:
@jeromekelleher updated

tests/datadriven/test_reads.py
```diff
@@ -67,6 +71,25 @@ def getDataModelClass(self):
     def getProtocolClass(self):
         return protocol.ReadGroupSet
 
+    def testPrograms(self):
+        # test that program info is set correctly
+        readGroupSet = self._gaObject
+        for readGroup in readGroupSet.getReadGroups():
+            readGroupInfo = self._readGroupInfos[readGroup.getSamFilePath()]
+            gaPrograms = readGroup.getPrograms()
+            htslibPrograms = readGroupInfo.programs
```
jeromekelleher: Need to check the lengths here. If (say) gaPrograms was empty here, this would still pass happily as zip doesn't care if the lists are different lengths.

jeromekelleher:
Looks good, thanks @dcolligan. +1 modulo the minor quibble in the tests.

tests/datadriven/test_reads.py
dcolligan: Gah, always forget about that.  It would probably be good to implement a `zipSafe` or something which throws an error if the lengths are not the same.

dcolligan:
@jeromekelleher updated

tests/utils.py
```diff
@@ -18,6 +18,13 @@
 packageName = 'ga4gh'
 
 
+def zipSameLength(*iterables):
+    length = len(iterables[0])
+    for iterable in iterables[1:]:
+        assert len(iterable) == length
```
jeromekelleher: Technically, this is assuming that the iterable is a list/tuple/etc as you can't can't call `len` on a generator. See 
https://docs.python.org/3/library/itertools.html#itertools.zip_longest for clues on how to do it generally.

dcolligan: I think that's fine.  We're not dealing with large enough sequences that we can't call `list(generator)` beforehand anyways.  In fact, we often do just this in tests.

jeromekelleher: Maybe be more explicit then, and call it ``zipLists()``? I'm sure it'll lead to confusing bugs if we don't make it clear it's expecting lists. A docstring explaining the intent would be good too.

dcolligan:
@jeromekelleher updated

jeromekelleher:
Perfect --- thanks @dcolligan. Merging.

Issue #591, opened 2015-08-27
SearchReads takes reference name, not ID


macieksmuga:
SearchReads should take a referenceID (per schemas), but uses reference name instead. Thus, the following query (against my local server instance, yours may take different names & ID's) fails:

    ga4gh_client -vv reads-search http://localhost:8000 --readGroupId="YnJjYTE6MWtnLWxvdy1jb3ZlcmFnZTpCUkNBMV9IRzAwMDk2" --start=0 --end=200 --referenceId="aGczNzpoczM3ZDVfY2hyMTdfNDExOTYzMTItNDEyNzc1MDA="

while this one works:

    ga4gh_client reads-search http://localhost:8000 --readGroupId="YnJjYTE6MWtnLWxvdy1jb3ZlcmFnZTpCUkNBMV9IRzAwMDk2" --start=0 --end=200 --referenceId="ref_brca1"

jeromekelleher:
This is a bug, thanks @macieksmuga.

dcolligan:
Fixed in #597

PR #597, opened 2015-08-31
Search reads now takes reference ID

dcolligan:
Issue #591


jeromekelleher:
This is a bit short-term: we'll should really be getting the reference object in ``backend.py``, pass this through as an argument to the query. However, we need other stuff to fit together for this, so I'm happy to merge this quick fix if you'd prefer to deal with the full solution later.

dcolligan:
This fixes the bug.  We can commit this and then you can file a new issue about the internal refactor you'd like to see.

PR #603, opened 2015-09-02
Added reference metadata + updated scripts.

jeromekelleher:
Work in progress.

Adds metadata to the reference object by associating a JSON file with it.

Datadriven tests need to be fixed and extended.


scripts/download_data.py
```diff
@@ -280,14 +293,36 @@ def downloadFastas(self):
                 'display': 'fasta',
                 'range': '{}-{}'.format(0, maxPos)})
             url = '{}%26{}'.format(path, args)
+            tempFileName = '{}.fa.temp'.format(chromosome)
             fileName = '{}.fa'.format(chromosome)
-            downloader = utils.HttpFileDownloader(url, fileName)
+            downloader = utils.HttpFileDownloader(url, tempFileName)
             downloader.download()
+            # We need to replace the header on the downloaded FASTA
+            with open(tempFileName, "r") as inFasta,\
```
dcolligan: ooo I didn't know one could have multiple clauses in a with statement like this (but a puppy died!)

scripts/download_data.py
```diff
@@ -280,14 +293,36 @@ def downloadFastas(self):
...
+                    open(fileName, "w") as outFasta:
+                # Write the new header
+                print(">{}".format(chromosome), file=outFasta)
```
dcolligan: can we use `outFasta.write` instead of a `print` call?

scripts/download_data.py
```diff
@@ -280,14 +293,36 @@ def downloadFastas(self):
...
+                inFasta.readline()
+                for line in inFasta:
+                    print(line, file=outFasta, end="")
```
dcolligan: same here

scripts/generate_fasta.py
```diff
@@ -6,79 +6,103 @@
...
+        self.bases = ""
+
+    def writeFasta(self):
+        """
+        Write the random fasta file
+        """
+        utils.log("writing {} bases to {} ...".format(
+            self.numBases, self.fastaFileName))
+        with open(self.fastaFileName, 'w') as fastaFile:
+            print(">", self.referenceId, sep="", file=fastaFile)
```
dcolligan: same

scripts/generate_fasta.py
```diff
@@ -6,79 +6,103 @@
...
+            for i in range(numLines):
+                if basesRemaining < basesPerLine:
+                    basesToWrite = basesRemaining
+                else:
+                    basesToWrite = basesPerLine
+                bases = ''.join(
+                    [random.choice(baseChoices) for _ in range(basesToWrite)])
+                line = "{}".format(bases)
+                self.bases += line
+                print(line, file=fastaFile)
```
dcolligan: and here

dcolligan:
Is there a reason we are using json instead of yml?  yml is a bit more human readable/writable, which is what these config files are, right?

scripts/download_data.py
jeromekelleher: Why? I prefer using print, as I don't have to remember to put "\n" into the string. Feels like I'm writing C doing that!

scripts/download_data.py
jeromekelleher: Haha, I knew you were going to bring the puppies into this! It turns out that this is the one example in PEP8 where using backslashes is recommended. You can't use brackets in this context for some reason.

scripts/download_data.py
dcolligan: I suppose it's no big deal... print statements just look ugly to me since they're usually not the right thing to do in systems code.  "print" seems like it should be printing something to the screen, not a file.  This is a script tho so... I guess it's alright.

jeromekelleher:
I went with JSON over YML, as I'm starting to think of these files as something machine generated rather than user config files. The overall vision for the future is that we write a data importer program which will layout the files in the right way. Since we're using JSON a lot, this seems like a good option for this.

jeromekelleher:
Pushed an update @dcolligan, I think it's ready for review now. This will cause some problems as it now requires the existance of the JSON files. It seemed like more trouble to optionally have them present, and to figure out how to handle md5s in this case.

scripts/download_data.py
jeromekelleher: I can change it if you like, but as you say, it's a script so I wouldn't get too worried about it.

ga4gh/exceptions.py
```diff
@@ -412,12 +412,31 @@ class NotExactlyOneReferenceException(MalformedException):
...
+    def __init__(self, fileName, numReferences):
         self.message = (
             "FASTA files must have one and only one reference.  "
-            "File {} has {} references.".format(id_, numReferences))
+            "File {} has {} references.".format(fileName, numReferences))
 
 
+class InconsistentReferenceNameException(MalformedException):
+    """
+    A FASTA file has a reference name not equal to it's file name.
```
dcolligan: typo: should be "its"

dcolligan:
If this is going to have the server choke on any data dir without json files, then we should do either one of two things:
1) update the example data with json files
2) have a server config option like "mandate reference metadata json files" that is off by default
I think just doing 1 is fine, since presumably there is no one using this version that isn't a dev.

ga4gh/exceptions.py
jeromekelleher: Fixed, thanks.

jeromekelleher:
I agree we'll need to update the example data, but we should wait until your readgroup stuff comes through first. The only people this will really annoy are the compliance suite people; they should get plenty notice this is in the pipeline, and perhaps a tool to generate the files.

jeromekelleher:
So, maybe we should wait for the readgroupset stuff to land before merging this?

jeromekelleher:
@dcolligan --- this has been updated, and should be mergable. I think we should merge this much and take on the remaining reference stuff afterwards, as this PR is big enough all ready.

dcolligan:
+1

Issue #523, opened 2015-07-30
Add support for  GET `/datasets/id`


jeromekelleher:
This can be based on the definition for `getVariantSet` in `frontend.py`.

pcingola:
This is my first PR to this repo, so I'd appreciate some feedback / code review

Run example:

    $ python server/client_dev.py --verbose datasets-get MTAwMEc= http://localhost:8000/
    INFO:ga4gh.client:GET http://localhost:8000/datasets/MTAwMEc=
    INFO:requests.packages.urllib3.connectionpool:Starting new HTTP connection (1): localhost
    MTAwMEc=

dcolligan:
The code review happens on the pull request page.  Everyone subscribing to the repo is automatically notified with a new PR is made.

dcolligan:
Fixed in #617

Issue #595, opened 2015-08-28
Documentation needs to be updated.


jeromekelleher:
As part of  a new release of the server (version 0.2.x?), we need to update the documentation to make sure all the changes since 0.1 have been included.

PR #604, opened 2015-09-02
Added support for GET '/datasets/{id}' Issue #523

pcingola:
Added support for GET '/datasets/{id}' Issue #523

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -955,8 +955,8 @@ def client_main(parser=None):
     addReadGroupsGetParser(subparsers)
     addCallsetsGetParser(subparsers)
     addVariantsGetParser(subparsers)
-    addReferencesBasesListParser(subparsers)
     addDatasetsGetParser(subparsers)
+    addReferencesBasesListParser(subparsers)
 
     args = parser.parse_args()
     if "runner" not in args:
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -557,6 +557,17 @@ class GetCallsetRunner(AbstractGetRunner):
         self._run(self._httpClient.getCallset)
 
 
+class GetDatasetsRunner(AbstractGetRunner):
+    """
+    Runner class for the datasets/{id} method
+    """
+    def __init__(self, args):
+        super(GetDatasetsRunner, self).__init__(args)
+
+    def run(self):
+        self._run(self._httpClient.getDataset)
+
+
 class GetVariantRunner(AbstractGetRunner):
     """
     Runner class for the variants/{id} method
@@ -825,6 +836,15 @@ def addReadsSearchParser(subparsers):
     return parser
 
 
+def addDatasetsGetParser(subparsers):
+    parser = subparsers.add_parser(
+        "datasets-get",
+        description="Get a dataset",
+        help="Get a dataset")
+    parser.set_defaults(runner=GetDatasetsRunner)
+    addGetArguments(parser)
+
+
 def addDatasetsSearchParser(subparsers):
     parser = subparsers.add_parser(
         "datasets-search",
@@ -936,6 +956,7 @@ def client_main(parser=None):
     addCallsetsGetParser(subparsers)
     addVariantsGetParser(subparsers)
     addReferencesBasesListParser(subparsers)
+    addDatasetsGetParser(subparsers)
 
     args = parser.parse_args()
     if "runner" not in args:
--- a/tests/unit/test_views.py
+++ b/tests/unit/test_views.py
@@ -133,6 +133,13 @@ class TestFrontend(unittest.TestCase):
         response = self.sendGetRequest(path)
         return response
 
+    def sendGetDataset(self, id_=None):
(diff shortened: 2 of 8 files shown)


ga4gh/cli.py
```diff
@@ -936,6 +956,7 @@ def client_main(parser=None):
     addCallsetsGetParser(subparsers)
     addVariantsGetParser(subparsers)
     addReferencesBasesListParser(subparsers)
+    addDatasetsGetParser(subparsers)
```
dcolligan: switch this line with the line above, so we have all the add*GetParser calls together, just for clarity

dcolligan:
@pcingola this is a good start.  The tests are failing, however.  Run `python script/run_tests.py` to run the Travis tests locally.  You should also add a test that tests that the route you added works... see the other tests in the same file in which the current failing test is located for examples.

pcingola:
OK, apparently installing "virtualenv" is a must for the tests to pass. I somehow I missed it in the documentation (it appeared to be optional).
Also, a module called 'funcsigs' is also necessary but it is either not in the requirements.txt or for some reason is missing on my computer....
After that, the tests are passing.

How can I update the documentation? There are several examples that simply do not work, which is a complication when you are trying to learn how things work...(or even how to install a development environment).

pcingola:
Added test case 'tests.unit.test_views.TestFrontend.testGetDataset'

dcolligan:
@pcingola yes, the documentation is out of date.  There is an issue to remedy that: #595

I doubt there is a module required for the project that is not in `requirements.txt` (or the packages that the packages in `requirements.txt` depend on, which are automatically installed with a normal installation), otherwise the tests would fail.

The docs are in the `docs` subdirectory; they are tracked just like any other file in our project.  We use sphinx for our docs.  `run_tests` will build the docs as part of its run, but you can do the same manually with:

```
make clean -C docs/source/ && make -C docs/source/
```

tests/unit/test_views.py
```diff
@@ -133,6 +133,13 @@ def sendGetVariantSet(self, id_=None):
         response = self.sendGetRequest(path)
         return response
 
+    def sendGetDataset(self, id_=None):
+        if id_ is None:
+            id_ = self.variantSetId
```
dcolligan: why is this `self.variantSetId`?  shouldn't it be a dataset id?

tests/unit/test_views.py
```diff
@@ -358,7 +373,6 @@ def testNotImplementedPaths(self):
         pathsNotImplementedPost = [
         ]
         pathsNotImplementedGet = [
-            '/datasets/<id>',
```
dcolligan: we can just delete this whole test now, since it doesn't do anything anymore

dcolligan:
This is progressing nicely, but I would like to see two more tests, one for the client and one for the cli.  Do something in `tests/unit/test_cli.py` and `tests/unit/test_client.py` analogous to what I did in 2f6180120db855109b9baff24e132b5531d70a8c for the get callset endpoint.

tests/unit/test_views.py
```diff
@@ -290,6 +297,14 @@ def testVariantSetsSearch(self):
             response.data)
         self.assertEqual(len(responseData.variantSets), 1)
 
+    def testGetDataset(self):
+        response = self.sendDatasetsSearch()
+        responseData = protocol.SearchDatasetsResponse.fromJsonString(
+            response.data)
+        datasetId = responseData.datasets[0].id
+        response = self.sendGetDataset(datasetId)
+        self.assertEqual(200, response.status_code)
```
jeromekelleher: We need to test the 'not found' condition here too, which should give us a 404 HTTP response. See the second half of testGetVariantSet for a model to follow.

jeromekelleher:
This looks great @pcingola. I agree with the comments @dcolligan makes above.

I'd also like to see two more tests:

1. Add a line to ``testCors`` in test_views.py.
2. Add a ``testGetDataset()`` in tests/unit/test_simulated_stack.py. Use ``testGetReferenceSet`` as a model.

This is quite a few different tests, but we're touching a lot of different parts of the codebase here, on both the client and server side.

pcingola:
Added your suggestions and tests.
Thank you for your feedback!

jeromekelleher:
Excellent, thanks @pcingola! A few more very minor things and I'm ready to +1 and merge:

1. We still need the 404 check in testGetDataset in tests_views. You can pretty much copy the second half of testGetVariantSet.
2 The commented out code for ``testNotImplementedPaths`` should be deleted.
3. Once that's done, the commits need to be squashed down to one commit, and rebased on upstream/develop. I think there's a section on this in the develop docs.

pcingola:
OK, added the 404 check and deleted testNotImplementedPaths.

Created a new PR #611 with all commits squashed.

dcolligan:
Closing this PR since there is a newer one open.

Btw @pcingola usually the best practice is to keep the same PR open and force push to the same branch that you are working on if you have modified the commits.  That way the comments are not split across two PRs for the same change.

Issue #574, opened 2015-08-14
Data hierarchy for ReadGroupSets and ReadGroups is wrong


jeromekelleher:
As discussed in ga4gh/compliance#34, we have misinterpreted the mapping between SAM files and readgroups. Currently, we map a SAM file to a ReadGroup but it should map to a ReadGroupSet.

jeromekelleher:
Copying a comment made by @lh3 here for reference: 

> One read group typically corresponds to one lane. For old data, one BAM usually contains multiple read groups. For typical resequencing, we use one BAM for one sample, though there are exceptions at times. A readgroupset usually corresponds to one BAM.

Thanks for the clarification @lh3!

diekhans:
While a given SAM file can have multiple read groups, it is
often desirable to have a read group set split across multiple
BAM files.

This is a huge problem for TCGA.  Both the data distribution
model and many of the tools assume that all readgroups for a
given sample were in a single BAM.  However Broad's data
production model didn't generate data in this manner.  Often
additional readgroups would be sequenced at disjoint times when
more sequence capacity was available.

The TCGA model required new `top-off' BAMs be created with all
readgroups, generating hundreds of terabytes of data churn.

One of the advantages of API access to data is to remove all of
the assumptions around file organization and instead think in
terms of the actually immutable data production units.  In the
case of sequencing, this is readgroup, not readgroupset

I would encourage support a readgroupset being stored in one or
more BAMs.

jeromekelleher:
(copying @lh3's comment here from  ga4gh/compliance#34 )
> All combinations happen in practice, but I think we should just implement the common practice (one bam per sample containing one or several read groups) and throw an error in the uncommon cases for now.

I agree --- let's assume the common case for now. This is just our data input layer, we can generalise later when the problem arises.

dglazer:
I agree with @jeromekelleher that the ref server can start by assuming the common case, and adding complexity later if and when the problem arises.

(FYI, as I mentioned elsewhere, Google's import code supports the common case by default, but also lets you import data with other mappings -- see our [understanding readgroupset assignment](https://cloud.google.com/genomics/v1beta2/managing-reads?hl=en#readgroupsetids) doc for details.)

macieksmuga:
[...]

PR #605, opened 2015-09-03
Map BAM files to ReadGroupSets, not ReadGroups

dcolligan:
Issue #574
Issue #612


dcolligan:
@jeromekelleher this ended up being easier than expected (so far), but I wanted your advice about how to modify `test_reads.py` ... there's a lot of indirection there

ga4gh/datamodel/reads.py
```diff
@@ -122,29 +139,81 @@ class SimulatedReadGroupSet(AbstractReadGroupSet):
...
     Class representing a logical collection ReadGroups.
     """
-    def __init__(self, parentContainer, localId, dataDir):
+    def __init__(self, parentContainer, localId, dataFile):
         super(HtslibReadGroupSet, self).__init__(parentContainer, localId)
-        self._dataDir = dataDir
-        self._setAccessTimes(dataDir)
-        self._scanDataFiles(dataDir, ["*.bam"])
+        self._samFilePath = dataFile
+        samFile = self.getFileHandle(self._samFilePath)
```
jeromekelleher: We should close this samFile after we're done with it.

jeromekelleher: Ooops! No, we most certainly shouldn't close the file handle, my bad.

ga4gh/datamodel/reads.py
```diff
@@ -182,19 +252,15 @@ def toProtocolElement(self):
         readGroup.referenceSetId = None
         readGroup.sampleId = None
         stats = protocol.ReadStats()
-        stats.alignedReadCount = self.getNumAlignedReads()
-        stats.unalignedReadCount = self.getNumUnalignedReads()
+        stats.alignedReadCount = \
+            self._parentContainer.getNumAlignedReads()
```
jeromekelleher: Is this true? Shouldn't the number of aligned and unaligned reads be calculated per-readgroup? This will make the number of reads equal for all readgroups, which isn't the intention I think.

ga4gh/datamodel/reads.py
```diff
@@ -223,21 +277,12 @@ class SimulatedReadGroup(AbstractReadGroup):
     """
     def __init__(self, parentContainer, localId, randomSeed, numAlignments=2):
         super(SimulatedReadGroup, self).__init__(parentContainer, localId)
-        self._numAlignments = numAlignments
+        self._parentContainer = parentContainer
```
jeromekelleher: This is set in the superclass, isn't it?

ga4gh/datamodel/reads.py
```diff
@@ -223,21 +277,12 @@ class SimulatedReadGroup(AbstractReadGroup):
...
 
     def getReadAlignments(self, referenceId=None, start=None, end=None):
-        for i in range(self._numAlignments):
+        for i in range(self._parentContainer.getNumAlignedReads()):
```
jeromekelleher: I think we should keep this as the per-readgroup numAlignments, not inheriting it from the parent.

ga4gh/datamodel/reads.py
```diff
@@ -271,42 +316,12 @@ class HtslibReadGroup(datamodel.PysamDatamodelMixin, AbstractReadGroup):
...
-        for htslibProgram in htslibPrograms:
-            program = protocol.Program()
-            program.id = htslibProgram['ID']
-            program.commandLine = htslibProgram.get('CL', None)
-            program.name = htslibProgram.get('PN', None)
-            program.prevProgramId = htslibProgram.get('PP', None)
-            program.version = htslibProgram.get('VN', None)
-            programs.append(program)
-        return programs
+        self._parentContainer = parentContainer
```
jeromekelleher: This should already be set in the superclass, as we're passing it as a parameter to ``__init__``

ga4gh/datamodel/reads.py
```diff
@@ -271,42 +316,12 @@ class HtslibReadGroup(datamodel.PysamDatamodelMixin, AbstractReadGroup):
...
+        self._parentSamFilePath = parentContainer.getSamFilePath()
+        self._filterReads = not parentContainer.isDefaultReadGroup()
+        self._localId = localId
```
jeromekelleher: Ditto for localId --- should already be set.

ga4gh/datamodel/reads.py
```diff
@@ -325,8 +343,17 @@ def getReadAlignments(self, referenceId=None, start=None, end=None):
         referenceId, start, end = self.sanitizeAlignmentFileFetch(
             referenceId, start, end)
         readAlignments = samFile.fetch(referenceId, start, end)
-        for readAlignment in readAlignments:
-            yield self.convertReadAlignment(readAlignment)
+        if self._filterReads:
+            for readAlignment in readAlignments:
+                for tag in readAlignment.tags:
```
jeromekelleher: What is ``readAlignment.tags``? If it's a ``dict`` we should just be able to say 
```
if 'RG' in readAlignment.tags:
    if readAlignment.tags['RG'] == self.getLocalId():
          yield self.convertReadAlignment(readAlignment)
```

ga4gh/datamodel/reads.py
```diff
@@ -345,7 +372,8 @@ def convertReadAlignment(self, read):
         ret.alignment.mappingQuality = read.mapping_quality
         ret.alignment.position = protocol.Position()
         self.sanitizeGetRName(read.reference_id)
-        samFile = self.getFileHandle(self._samFilePath)
+        samFile = self._parentContainer.getFileHandle(
```
jeromekelleher: Maybe we should pass ``samFile`` as an argument to ``convertReadAlignment`` to avoid hitting the handle cache again?

ga4gh/datamodel/reads.py
```diff
@@ -122,29 +139,81 @@ class SimulatedReadGroupSet(AbstractReadGroupSet):
...
+    def openFile(self, dataFile):
+        return pysam.AlignmentFile(dataFile)
+
+    def getSamFilePath(self):
+        """
+        Returns the file path of the sam file
+        """
+        return self._samFilePath
+
+    def isDefaultReadGroup(self):
```
jeromekelleher: I find this slightly confusing (it implies that the ReadGroupSet itself is a default read group). Maybe a more transparent flag would be ``hasReadGroupTags()``? This would make the client code easier to understand I think.

jeromekelleher:
Looks good so far @dcolligan, it's all pretty clear. I think we can probably take a similar approach in the ``datadriven/test_reads.py``: the ``ReadGroupInfo`` takes a samFile and a readGroupName as a parameters, and we just filter out the reads that apply to this readgroup. The readGroupInfos are then indexed by readGroupName instead of the samfile name. I think most of the tests should just work then.

We'll need to rearrange the test data, and make sure that we have at least one bam in there that has no RG tags.

ga4gh/datamodel/reads.py
dcolligan: True, but that requires iterating through all the reads in the read group.  What should we do here?  We could put some dummy value here.  Maybe we could even change the schema to remove readGroup.stats ...

ga4gh/datamodel/reads.py
dcolligan: it's not... it's a list of tag, value 2-tuples

ga4gh/datamodel/reads.py
jeromekelleher: I think we'll have to put -1 in here for now, and look at generating some metadata JSON files like we do (or, soon will do) for references. There's really no other way to do it.

ga4gh/datamodel/reads.py
jeromekelleher: It's tempting to just make a dict out of this... This is potentially a performance bottleneck, but it might not matter. If it does, we can always change it back to something like you have now. So, specifically, I'd put:
```
# This may be a performance bottleneck; pysam returns a list of tuples not a dict.
tags = dict(readAlignment.tags)
if 'RG' in tags:
    # etc
```

ga4gh/datamodel/reads.py
dcolligan: let's not micromanage the cache until we have to

dcolligan:
@jeromekelleher updated... all tests are passing except for remote read fetch (since searching for urls.json is broken now for reads)

jeromekelleher:
Looks good, thanks @dcolligan.

In terms of urls.json, I'm happy to disable this stuff for a while if this is the quickest option. It can't be used in reality anyway, after the issues we've encountered with the download script. We have to wait until libcurl based htslib is available to us before we can seriously tackle this feature.

Also, we should update the download script to it puts files in the right structure. Might as well do this here.

I'm +1 once the tests are passing anyway.

dcolligan:
The `download_data` script seems to have other problems as well (I'll file an issue shortly)... we can fix those in another PR

dcolligan:
@jeromekelleher updated

jeromekelleher:
Looks great @dcolligan, +1.

PR #607, opened 2015-09-04
G2P '/genotypephenotype/search' endpoint

bwalsh:
## Summary
We extended the GA4GH Reference server to include a the '/genotypephenotype/search' endpoint.  This document describes the work to date  

## Approach
We based our work on the model captured in [ga4gh/schemas commit of Jul 30, 2015](https://github.com/ga4gh/schemas/tree/be171b00a5f164836dfd40ea5ae75ea56924d316).  This version of the schema predates the [separated genotype to phenotype files from baseline](https://github.com/ga4gh/schemas/commit/846b711fdcf544bf889cc7dbab19c6c48e9a9428).

The code was based on a [branch](https://github.com/ga4gh/server/tree/g2p) setup for this purpose by the server team.
No major refactoring of the server was needed, additional code was added to ga4gh/backend.py,ga4gh/frontend.py and test/unit/test_views.py.  A new model was created in datamodel/genotype_phenotype.py

## Data
The cancer genome database [Clinical Genomics Knowledge Base](http://nif-crawler.neuinfo.org/monarch/ttl/cgd.ttl) published by the Monarch project was the source of Evidence.   

image

## API
The GA4GH schemas define a single endpoint `/genotypephenotype/search` which accepts a POST of a [request body](https://github.com/ga4gh/schemas/blob/be171b00a5f164836dfd40ea5ae75ea56924d316/src/main/resources/avro/genotypephenotypemethods.avdl#L102)  containing one or more of [Feature](https://github.com/ga4gh/schemas/blob/be171b00a5f164836dfd40ea5ae75ea56924d316/src/main/resources/avro/sequenceAnnotations.avdl#L105), [PhenotypeInstance](https://github.com/ga4gh/schemas/blob/be171b00a5f164836dfd40ea5ae75ea56924d316/src/main/resources/avro/genotypephenotype.avdl#L77), [EnvironmentalContext](https://github.com/ga4gh/schemas/blob/be171b00a5f164836dfd40ea5ae75ea56924d316/src/main/resources/avro/genotypephenotype.avdl#L49), and [Evidence](https://github.com/ga4gh/schemas/blob/be171b00a5f164836dfd40ea5ae75ea56924d316/src/main/resources/avro/genotypephenotype.avdl#L113) which are combined as a logical AND to query the underlying datastore.  Missing types are treated as a `wildcard` returning all data.  Responses of matching data are returned as a list of  [FeaturePhenotypeAssociation](https://github.com/ga4gh/schemas/blob/be171b00a5f164836dfd40ea5ae75ea56924d316/src/main/resources/avro/genotypephenotype.avdl#L132).  All types rely heavily on [OntologyTerm](https://github.com/ga4gh/schemas/blob/be171b00a5f164836dfd40ea5ae75ea56924d316/src/main/resources/avro/ontologies.avdl#L10)

### Request
image
http://yuml.me/edit/bf06b90a

### Response
image

http://yuml.me/edit/25343da1

----
## Implementation

image

http://yuml.me/c97fada2

----
### Query by example

There are four datatypes types for each entity [string, external identifier, ontology identifier and 'entity'].
Currently the implementation handles queries of [string, external identifier and ontology identifier].  

The  'entity' query is a type of query-by-example has been deferred.  Challenges that arose: 

* schema constraints:  there are several fields within the schemas that are defined as non-null. This may be fine when creating an entity from a data store, however, they are problematic when creating an entity to be used in a query.

* additional discussions needed to determine what properties from an existing entity will be used for the query and which will be ignored.  For example [a Feature](https://github.com/ga4gh/schemas/blob/6725fb6653eb5642c0dc4884c82016db1c24c5c3/src/main/resources/avro/sequenceAnnotations.avdl#L105) has [id,parentIds, featureSetId, referenceName, start,end, strand, featureType, attributes]  we need to specify exactly what the query's expectations are.  

*Recommendation:* Leave the schema definitions as-is.  However, leave the entity query-by-example unimplemented.  Implement when demand exists with sufficient use case details.

 
### Name collision (SearchFeaturesResponse)

That schema contains two definitions of the class `[SearchFeaturesRequest,SearchFeaturesResponse]`.  These conflict in the generated code with other classes of the same name.

The schema project the current server is based on is  `version = '0.6.be171b00'`
Snippets from this commit follow

* One in the file genotypephenotypemethods.avdl,  protocol GenotypePhenotypeMethods
```
/** This is the response from `POST /genotypephenotype/search` expressed as JSON. */
record SearchFeaturesResponse {
  /**
  The list of matching FeaturePhenotypeAssociation.
  */
  array<org.ga4gh.models.FeaturePhenotypeAssociation> associations = [];

  ...
```

* The second one is found in sequenceAnnotationmethods.avdl 
```
  /** This is the response from `POST /features/search` expressed as JSON. */
  record SearchFeaturesResponse {
    /**
    The list of matching annotations, sorted by start position. Annotations which
    share a start position are returned in a deterministic order.
    */
    array<org.ga4gh.models.Feature> features = [];

    ... 
```

* The generated code only has the class associated with `sequenceAnnotationmethods.avdl`
 

Both sequenceAnnotationmethods.avdl and genotypephenotypemethods.avdl share the same namespace `@namespace("org.ga4gh.methods")` each file defines an enclosing `protocol`.

In the [names section of the spec](https://avro.apache.org/docs/1.7.7/spec.html#Names)

```
A name only is specified, i.e., a name that contains no dots. In this case the namespace is taken from the most tightly enclosing schema or protocol. For example, if "name": "X" is specified, and this occurs within a field of the record definition of org.foo.Y, then the fullname is org.foo.X. If there is no enclosing namespace then the null namespace is used.
```

The schemas pass validation.
```
A schema or protocol may not contain multiple definitions of a fullname. Further, a name must be defined before it is used ("before" in the depth-first, left-to-right traversal of the JSON parse tree, where the types attribute of a protocol is always deemed to come "before" the messages attribute.)
```

*Recommendation:*  Rename the GenotypePhenotypeMethods [SearchFeaturesRequest,SearchFeaturesResponse] to [SearchGenotypePhenotypeRequest,SearchGenotypePhenotypeResponse]


jeromekelleher:
This is superb @bwalsh --- great overview, and the code looks great too. I'll get back with some specific comments next week: there's a lot to take in here!

adamnovak:
As Benedict just said, can you please add corresponding PRs against ga4gh/schemas and ga4gh/compliance?

The compliance PR may have to be based on the compliance_redux branch.

Once that's done, we can merge all 3 PRs at once, with discussion probably happening mostly here.

tests/unit/test_views.py
```diff
@@ -27,7 +27,7 @@ def setUpClass(cls):
             "SIMULATED_BACKEND_NUM_CALLS": 1,
             "SIMULATED_BACKEND_VARIANT_DENSITY": 1.0,
             "SIMULATED_BACKEND_NUM_VARIANT_SETS": 1,
-            # "DEBUG" : True
+            "DEBUG": True
```
adamnovak: Do you want to leave this in?

adamnovak:
So on the server side it looks like this backends into RDF queried with sparql.

How is the RDF data indexed? Will queries be efficient from disk, or will the server have to keep the whole g2p graph in memory?

Also, does this PR include the recommended changes to resolve the name collision? (That would probably actually belong in the corresponding PR against the schemas repo).

Where are the .avdl files that correspond to this server?

tests/unit/test_views.py
bwalsh: Good catch.  I'll remove it

bwalsh:
Thanks for the detailed feedback.  I'll be attending the meeting & Hackathon in NYC in October.  My goal is to support this PR and hopefully be ready for new areas (metadata).


>Also, does this PR include the recommended changes to resolve the name collision? (That would probably actually belong in the corresponding PR against the schemas repo).

It does not.  As suggested, I'll create a PR against the schema.

>Where are the .avdl files that correspond to this server?

The version of the schema these changes were based on:  https://github.com/ga4gh/schemas/commit/846b711fdcf544bf889cc7dbab19c6c48e9a9428


>As Benedict just said on the call, can you please add corresponding PRs against ga4gh/schemas and ga4gh/compliance?  The compliance PR may have to be based on the compliance_redux branch.


Did not realize there was a call.  Will do.  This may get tricky as  I saw the following in the ga4gh/compliance  docs.  Is there a contact over on that project available for questions?


```
This version of the Schema is not tracking changes to the real v0.5.1 Schema!

Nor, of course, does it track to your own version of the schema if you were changing that. So you may need to convert the schemas module to be a git submodule which is tracking the version/branch of Schema you care about. If you do this (or otherwise edit the schema) you'll need to run the assertions generator so your test assertThat() functions match your new schema classes
```


> Will queries be efficient from disk, or will the server have to keep the whole g2p graph in memory?

Regarding the runtime performance,  my goal was to create an initial implementation; something self contained and easy to implement and understand ( no extra servers and minimum dependencies).

As-is, the server uses memory, there are [other stores](http://rdflib.readthedocs.org/en/latest/persistence.html#stores-currently-shipped-with-core-rdflib) that ship with RDFLib.  My preference is to roadmap adjustments or optimizations after this (set of) PR

hjellinek:
Hi @bwalsh - The comment

> This version of the Schema is not tracking changes to the real v0.5.1 Schema!

is obsolete.  I will fix that document.

It is correct to say that the compliance tool kit does not automatically sync its schemas with an outside repo or authority.  If you're testing against modified schemas, you need to copy those modifications into the compliance test kit's `ctk-schemas` module, add appropriate endpoint URLs, etc.

ga4gh/backend.py
```diff
@@ -10,11 +10,13 @@
 import json
 import random
 
+
```
jeromekelleher: Is this extra blank line intentional? We don't have this in any of the other files, so I would remove it in the interest of consistency (unless there's a good argument for it).

ga4gh/backend.py
```diff
@@ -470,7 +472,15 @@ def searchDatasets(self, request):
             protocol.SearchDatasetsResponse,
             self.datasetsGenerator)
 
-    # Iterators over the data hieararchy
```
jeromekelleher: This was an intentional separator between the different groups of methods. I'd like to keep this unless there's a good argument for removing it.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+from rdflib import Graph
+
+
+class G2PDataset:
+    """
+    an rdf object store
+    """
+
+    # keep the graph as a static variable
+    g = None
```
jeromekelleher: Single letter variables are generally frowned upon around here. They're only really considered OK when used in things like list comprehensions or lambda expressions. Definitely not good as a static class variable though. Something like 
```python
rdfGraph = rdflib.Graph()
```
is much easier to folllow.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
+"""
+Module responsible for translating g2p data into GA4GH native
+objects.
+"""
+
+import ga4gh.protocol as protocol
+import ga4gh.exceptions as exceptions
+
+import rdflib
+from rdflib import URIRef
```
jeromekelleher: We usually  avoid importing specific functions or classe, and prefer using the long version, .e.g.,
use ``` rdflib.Graph()`` below. The reason for this is the reader can tell immediately where a particular class is coming from, and doesn't have to search back to the header to see where the class was imported from.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+    """
+
+    # keep the graph as a static variable
+    g = None
+    # overwrite tests/unit/cgd-test.ttl
+    record_test = False
+    #
+    g2 = Graph()
+
+    def __init__(self, is_prod):
```
jeromekelleher: What does this boolean condition signify? It should be camelcased, and more explicit about what it means.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+
+        self.search_query = """
```
jeromekelleher: We tend to make instance variables protected by default. So, this should be ``self._search_query``.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+                    ?s  OBAN:association_has_object  ?d .
+                    ?d  rdfs:label ?disease_label  .
+                    ?d rdf:type ?disease .
+                    ?s  OBAN:association_has_object_property  ?evidence .
+                    OPTIONAL {  ?evidence  rdfs:label ?evidence_label } .
+              %FILTER%
+            }
+            """
+
+        if (G2PDataset.g):
```
jeromekelleher: Two issues here. Brackets are unnecessary, and also we follow PEP8's guidance and always compare with  ``None`` explicitly, ie.
```python
if self.rdfGraph is None:
    return
```
However, the intention of this is also confusing. Why do we need a class variable rather than an instance variable? There should only be one instance of this class in use, so what's the point of the class variable? This would be much simplified I think if we just had an instance variable ``self._rdfGraph = rdflib.Graph()``.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+            return
+        G2PDataset.g = rdflib.Graph()
+
+        if (is_prod):
```
jeromekelleher: Remove brackets.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+            # larger data set for exploration
+            # "http://nif-crawler.neuinfo.org/monarch/ttl/cgd.ttl"
+            G2PDataset.g.parse('tests/unit/cgd.ttl', format="n3")
```
jeromekelleher: We shouldn't be hard coding these values in here. Pass the parameters we need to instantiate the graph as parameters to the class. This will remove the need for the awkward ``isProduction`` flag, and simplify this constructor.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+            G2PDataset.g.parse('https://raw.githubusercontent.com/oborel/'
+                               'obo-relations/master/ro.owl',
+                               format="xml")
+
+        else:
+            # small data set for unit Tests
+            G2PDataset.g.parse('tests/unit/cgd-test.ttl', format="n3")
+
+    def _search(self, request):
+        # from IPython.core.debugger import Pdb
```
jeromekelleher: These should be deleted.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+        # Pdb().set_trace()
+        offset = 0
+        if request.pageToken:
+            print("pageToken " + request.pageToken)
+            offset = request.pageToken.split(':')[1]
+            print("offset " + offset)
+
+        associations = self.query_labels(request.feature,
```
jeromekelleher: This is a subjective style issue, so feel free to disregard if you don't agree. I find vertical alignment like this takes up a lot of vertical space and is brittle to code changes (for example, changing this to ``queryLabels`` means the arguments all need to be realigned). Hanging indent is easier, and is PEP8 compliant:
```
associations = self.queryLabels(
    request.feature,  request.evidence, request.phenotype, request.pageSize, offset)
```

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+                                         request.evidence,
+                                         request.phenotype,
+                                         request.pageSize,
+                                         offset
+                                         )
+        # from IPython.core.debugger import Pdb ;        Pdb().set_trace()
+
+        self.associations_length = len(associations)
```
jeromekelleher: All instance variables should be declared in the constructor, be camelCase and be protected by default, i.e.
``self._associationsLength``.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+        else:
+            # small data set for unit Tests
+            G2PDataset.g.parse('tests/unit/cgd-test.ttl', format="n3")
+
+    def _search(self, request):
+        # from IPython.core.debugger import Pdb
+        # Pdb().set_trace()
+        offset = 0
+        if request.pageToken:
+            print("pageToken " + request.pageToken)
```
jeromekelleher: Remove prints.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+
+        else:
+            # small data set for unit Tests
+            G2PDataset.g.parse('tests/unit/cgd-test.ttl', format="n3")
+
+    def _search(self, request):
+        # from IPython.core.debugger import Pdb
+        # Pdb().set_trace()
+        offset = 0
+        if request.pageToken:
```
jeromekelleher:
```python
```

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+            print("pageToken " + request.pageToken)
+            offset = request.pageToken.split(':')[1]
```
jeromekelleher: This is a layering violation --- page token parsing should be done in ``backend.py``.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+                                         request.pageSize,
+                                         offset
+                                         )
+        # from IPython.core.debugger import Pdb ;        Pdb().set_trace()
+
+        self.associations_length = len(associations)
+        for association in associations:
+            yield association
+
+    def query_labels(self, location=None, drug=None,
```
jeromekelleher: This should be ``queryLabels``. All identifiers should be camelCase, so I'll stop commenting on specific instances now.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+                     disease=None, pageSize=None, offset=0):
+        """
```
jeromekelleher: This docstring comment isn't hugely helpful, and appears to be identical to the documentation for ``queryCounts``.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+        count = rows[0]['count'].toPython()
+
+        return count
+
+    def format_query(self, location=None, drug=None, disease=None):
+        """
+        Generate a formatted sparql query with appropriate filters
+        """
+        query = self.search_query
+        if (location is None and drug is None and disease is None):
```
jeromekelleher: Lots of brackets around ``if``s here. These should all be removed.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+            raise exceptions.NotImplementedException(
+               "At least one of [location, drug, disease] must be specified")
+        filters = []
+
+        if (location and isinstance(location, basestring)):
```
jeromekelleher: Always check for ``None`` explicitly. Plus, this is redundant, isn't it? I don't think None can be an instance of basestring.

How can location be either a string or a dict here? It's not obvious how this can happen reading the rest of the code.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+            filters.append('regex(?location_label, "%s")' % location)
```
jeromekelleher: I realise this might be a bit unpopular, but we've been consistently using Python's new style string formatting in the rest of the code base. This would become 
```python
 filters.append('regex(?location_label, "{}")'.format(location))
```
It would be nice if we could maintain this consistency...

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+        if (disease and isinstance(disease, basestring)):
+            filters.append('regex(?disease_label, "%s")' % disease)
+
+        location_clause = ""
+        if (location and isinstance(location, dict)):
+            locations = []
+            for id in location['ids']:
+                    locations.append('?location = <%s> ' %
+                                     (id['database'] + id['identifier']))
+            location_clause = "(" + " || ".join(locations) + ")"
```
jeromekelleher: Again, it would be good if we could be consistent in the string formatting operations and use 
```
locationClause = "({})".format("||".join(locations))
```

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+        Return a list of dictionaries.
+        Each dict is an [annotation](http://www.w3.org/ns/oa#Annotation)
+        All keys in the dict are predicates of the annotation.
+        All cells in the dict are predicate->object converted to strings.
+
+        If an annotation has a <http://purl.org/oban/association_has_subject>
+        predicate that class is appended to the annotation dict in the
+        'location' property
+        """
+        #
```
jeromekelleher: stray #.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+          OPTIONAL {?o  rdfs:label ?label .} .
+        %FILTER%
+        }
+        """
+
+        annotation_query = annotation_query.replace("%FILTER%", filter)
+        # print(annotation_query)
+        results = G2PDataset.g.query(annotation_query)
+
+        if (G2PDataset.record_test):
```
jeromekelleher: Not sure about this... what's the intention here? We shouldn't be polluting the production code with test code --- these should be fully separated.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+            self.test_data(results)
+        rows = [row.asdict() for row in results]
```
jeromekelleher: This seems quite complicated --- what's the intention here?

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+        from urlparse import urlparse, urlunsplit
+        o = urlparse(_url)
+        _id = o.path.split(_separator)[-1]
+        _ontologySource = urlunsplit([o[0],
+                                      o[1],
+                                      o[2].replace(_id, ''),
+                                      o[4], ''])
+        return(_id, _ontologySource)
+
+    def test_data(self, results):
```
jeromekelleher: I don't like these two functions: test code should be in the test modules. We're complicating the production code here with tests.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+            if (row['p'] == 'http://purl.org/oban/association_has_subject'):
+                location = row['o']
+                location_query = location_query.replace("%FILTER%",
+                                                        "FILTER (?s = <" +
+                                                        location + ">)")
+                # print(location_query)
+                results = G2PDataset.g.query(location_query)
+                if (G2PDataset.record_test):
+                    self.test_data(results)
+                location_rows = [row.asdict() for row in results]
```
jeromekelleher: Why do we need to do this ``toPython`` step every time?

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+                    for k in row:
+                        row[k] = row[k].toPython()
+        self.record_more(rows)
+
+        annotation = self.flatten(rows)
+        location = self.flatten(location_rows)
+        annotation['location'] = location
+        return annotation
+
+    def flatten(self, dict):
```
jeromekelleher: Why do we need to do this? Is there any way we can move directly from the results that we get to the GA4GH objects without all the intermediate dicts? I'm finding it hard to follow what's going on here.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+        """
+
+        fpa = None
+
+        # annotation keys
+        # hasSubject = 'http://purl.org/oban/association_has_subject'
+        # hasPredicate = 'http://purl.obolibrary.org/obo/RO_has_environment'
+        source = 'http://purl.org/dc/elements/1.1/source'
+        location = 'location'
+        _evidence = 'http://purl.obolibrary.org/obo/RO_0002558'
```
jeromekelleher: Why ``_evidence`` and not ``evidence``?

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+        # _location = 'http://biohackathon.org/resource/faldo#location'
+        # GENO_results_in_amino_acid_change = 'http://purl.obolibrary.org/obo/'
+        # 'GENO_results_in_amino_acid_change'
+        # RO_0002200 = 'http://purl.obolibrary.org/obo/RO_0002200'
+        # RO_0002205 = 'http://purl.obolibrary.org/obo/RO_0002205'
+        # label = 'http://www.w3.org/2000/01/rdf-schema#label'
+
+        location = annotation['location']
+        if(GENO_0000408 in location):
+            (_id, _ontologySource) = self.namespace_split(
```
jeromekelleher: The brackets aren't needed here. Why the leading underscores? For things like ``id`` we follow the common Python convention of ``id_`` to avoid clobbering the global function.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+                                        location[GENO_0000408]['val'])
+            name = location[GENO_0000408]['label']
+        else:
+            (_id, _ontologySource) = self.namespace_split(location['id'])
+            name = location['id']
+
+        d = {
```
jeromekelleher: This should be removed. Outgoing object validation is a feature we can turn on and off using the config file. These should be checked in testing also, but not inlined as part of the main production code.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+                "id": "",
+                "type": {"ontologySource": _ontologySource,
+                         "id": _id,
+                         "name": annotation[hasObject]['label']}
+                },
+            "id": annotation['id'],
+            "description": "null"
+        }
+
+        fpa = protocol.FeaturePhenotypeAssociation.fromJsonDict(d)
```
jeromekelleher: I think this should be replace with something like:
```python
fpa = protocol.FeaturePhenotypeAssociation()
fpa.id = annotation['id']
fpa.features = []
fpa.description = None
# etc
```
I think if we get rid of the intermediate dictionaries and just update the GA object _directly_ this will simplify the code a lot.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+        fpa.features.append(f)
+
+        #  ECO or OBI is recommended
+        if (source in annotation):
+            if (annotation[source].__class__.__name__ != "list"):
```
jeromekelleher: ``isinstance(annotation[source], list)``

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+                annotation[source] = [annotation[source]]
+            for s in annotation[source]:
```
jeromekelleher: ``s`` is a poor variable name here.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,467 @@
...
+            raise exceptions.RequestValidationFailureException(
+                fpa.toJsonDict(), protocol.FeaturePhenotypeAssociation)
+
+        return fpa
+
+    def namespace_split(self, _url, _separator='/'):
+        """
+        given a url return the id of the resource and the ontology source
+        """
+        from urlparse import urlparse, urlunsplit
```
jeromekelleher: We should import `urlparse` in the header.

tests/unit/cgd-test.ttl
```diff
@@ -0,0 +1,137 @@
+@prefix ns1: <http://purl.org/oban/> .
```
jeromekelleher: Test data is great, but it should be put in the ``tests/data`` directory. The next file seems rather large --- do we really need a 27K line file of test data?

tests/unit/test_oidc.py
```diff
@@ -10,6 +10,7 @@
 import mock
 import urlparse
 import logging
+import mimetypes
```
jeromekelleher: This seems to have come in from another PR?

tests/unit/test_views.py
```diff
@@ -346,3 +345,158 @@ def testNoAuthentication(self):
...
+        self.assertEqual(400, response.status_code)
+
+    def testGenotypePhenotypeSearchFeature(self):
+        """
+        Search for evidence on a genomic feature given feature name
+        """
+        # simple string regexp
+        request = protocol.SearchFeaturesRequest()
+
+        request.feature = "KIT *wild"
```
jeromekelleher: These tests are great, but they are really data driven tests, right? This depends on the server having a specific data set which it serves and we can test against. This module avoids this, and makes sure that all the tests are data agnostic, and really is only intended to test the basic properties of the http frontend. So, I think the tests in this module should be changed to use a simulated backend (more on this later), and also just check the basic HTTP stuff like the rest of the tests here.

jeromekelleher:
Having gone through this in a bit of detail, I think it's a great starting point. There is a bit of work to be done I think in the overall layout, and particularly in how we run the queries and do the testing. However, I don't think these are major problems and we should be able to iterate towards something that could be merged quickly. Here are some high-level issues:

**Data representation and input**

Where does the G2P `data set` live within the current data hierarchy? Is there an ``AssociationSet`` object (or whatever) which we can use to model the container for a given set of associations? Without this, it's difficult to fit the G2P model into the current outline, which expects a container. 

Assuming this is the case, and that an ``AssociationSet`` belongs to a ``Dataset``, we can then make a new directory that is akin to ``variants`` and ``reads`` in the data file system hierarchy. We then read files from this hierarchy, which defines the source for a particular association set. We should then be able to make a ``AbstractAssociationSet`` which extends ``DatamodelObject``, and fit in to the existing infrastructure much more neatly.

** Testing **

Currently, test code is mixed in with production code, which we should avoid. The main reason for this I think is the lack of clarity on what constitutes a distinct source of data. Ideally, we would like

1. Data driven tests, following the examples in ``tests/datadriven`` which take in specific data files stored in ``tests/data/`` and verify that the GA objects we produce from this intput are valid and correctly transformed into the GA4GH format.
2. Random data tests. Randomly generated data is used extensively to test the API, and basically allows us to test everything that doesn't involve translating data (which is checked using the data driven tests outlined above). A method of generating random G2P data is therefore essential. The output doesn't have to be a realistic model in any way, it just has to represent a reasonable spread of possible output values and allow us to exercise the API without having huge data files lying around.

** Client code **

There is no client API provided. This should be added and tested, so we can run queries from the command line without using curl or whatever.

These is my take in any case. I'm afraid I don't understand the G2P model very well, so I apologise for any stupid points I've made above. As I say, I think this is a great start and we should have a good implementation ready to merge soon.

ga4gh/backend.py
```diff
@@ -708,3 +729,90 @@ def __init__(self, dataDir):
             dataset = datasets.FileSystemDataset(datasetDir)
             self._datasetIdMap[dataset.getId()] = dataset
         self._datasetIds = sorted(self._datasetIdMap.keys())
+
+
+class GeneotypePhenotypeBackend(AbstractBackend):
```
jeromekelleher: I don't think we should view this as a seperate backend, but rather as a part of the data model. See my earier comments about AssociationSets, etc.

ga4gh/backend.py
```diff
@@ -708,3 +729,90 @@ def __init__(self, dataDir):
...
+    """
+    an rdf object store
+    """
+    # keep the graph as a static variable
+    def __init__(self, is_prod):
+        super(GeneotypePhenotypeBackend, self).__init__()
+        self.dataset = genotype_phenotype.G2PDataset(is_prod)
+
+
+class GeneotypePhenotypeIterator(IntervalIterator):
```
jeromekelleher: I don't think we need this class. The IntervalIterator infrastructure is needed for reads and variants as they are both defined over genomic intervals, and we can only query using start and end coordinates. We therefore have to do complicated things to make sure we don't return the same result several times. For G2P associations, we just need an offset into the query, right?

ga4gh/backend.py
```diff
@@ -570,6 +580,15 @@ def callSetsGenerator(self, request):
             request, variantSet.getCallSetIdMap(),
             variantSet.getCallSetIds())
 
+    def genotypePhenotypeGenerator(self, request):
+        """
+        Returns a generator over the (genotypePhenotype, nextPageToken)
+        pairs defined by the specified request.
+        """
+        intervalIterator = GeneotypePhenotypeIterator(
```
jeromekelleher: We should do more work in here, and remove the GeneotypePhenotypeIterator. All you really need is to get the offset from the pageToken, and then pass this on to the query method in the data model object.

ga4gh/datamodel/genotype_phenotype.py
bwalsh: without the toPython() what is returned is a `rdflib.term.Literal(u'sorafenib')`.  With toPython() what is returned is a unicode string

ga4gh/backend.py
bwalsh: @jeromekelleher  Hi.  I'm looking at making this change. As you point out, the GeneotypePhenotypeIterator  class is large compared to what it actually needs to do.  The tradeoff seems to be between  keeping the class and allowing G2P requests to run via the existing `runSearchRequest` vs creating my own implementation of runSearchRequest.  I'm completely open to that possibility, just wanted to verify that it made sense to you.

ga4gh/backend.py
bwalsh: @jeromekelleher Hi.   Thanks again for the thoughtful feedback.  On this item, I've looked into removing GeneotypePhenotypeIterator.  As you mentioned, it is heavyweight for the actual work it needs to do.  As far as I can tell, if I remove this class, it will mean implementing an equivalent of `runSearchRequest` in `searchGenotypePhenotype`.  I have no problem doing this as the g2p request is quite different than the reads/referenceSets/etc.  I just wanted to see if there were any objections to this approach.

bwalsh:
@hjellinek, @adamnovak - following up on the request to add tests to compliance_redux

As a first step, I wanted to ensure that compliance_redux works as expected.

I've forked and build the compliance tests and started the test process given the instructions below.  I'm using the dataset from wget http://www.well.ox.ac.uk/~jk/ga4gh-example-data.tar   ( with directory changes per the github issue )

Also, I saw in the notes `the ID given below is specific to the GA4GH reference server running on the compliance dataset`.  **Where would I find that dataset?**

Also,  the work I'm doing here is based on the g2p branch of the server.  **Are the compliance_redux tests compatible with that branch?**



`[TESTLOG] Overall: Tests run: 100, Failures: 12, Errors: 76, Skipped: 12, Time elapsed: 1.300 sec`


```
Before running any tests, you may want to set an environment variable to record the server URL you’re testing against:

 > export CTK_TGT_URLROOT=“<your GA4GH-compatible server URL>"
You may also wish to set the dataset ID for the test dataset as an environment variable. Note that the ID given below is specific to the GA4GH reference server running on the compliance dataset, and will likely be different on another server implementation.

 > export CTK_TGT_DATASET_ID=“YnJjYTE="
```

ga4gh/backend.py
jeromekelleher: This sounds fine to me @bwalsh; go for it.

bwalsh:
All: we've put in some additional work:
* created a PR for the schema
* rebased our fork server/g2p from ga4gh/server/develop
Nose tests, etc are fine.

My git kung-fu may be lacking, but I believe that if we rebase ga4gh/server/g2p from ga4gh/server/develop the commit noise above will go away?

We remain a bit stuck on getting going on the compliance tests.  Specifically, the mvn project builds fine, but I remain stuck on getting compliance data into the server.

I used the script from  https://github.com/ga4gh/server/pull/656 which worked.  However, when I launch the server, I get the following

```
# python server_dev.py
    ga4gh.cli.server_main()
  File "/home/ga4gh/tmp/rebase/server/ga4gh/cli.py", line 71, in server_main
  File "/home/ga4gh/tmp/rebase/server/ga4gh/frontend.py", line 200, in configure
    theBackend = backend.FileSystemBackend(dataSource)
  File "/home/ga4gh/tmp/rebase/server/ga4gh/backend.py", line 836, in __init__
    referenceSetName, relativePath, self)
  File "/home/ga4gh/tmp/rebase/server/ga4gh/datamodel/references.py", line 366, in __init__
    self._setMetadata()
  File "/home/ga4gh/tmp/rebase/server/ga4gh/datamodel/references.py", line 371, in _setMetadata
    with open(metadataFileName) as metadataFile:
IOError: [Errno 2] No such file or directory: u'ga4gh-example-data/references/hg37.json'

```

The contents of the test directory is:

```
 server]# find ga4gh-example-data -print
ga4gh-example-data
ga4gh-example-data/brca1
ga4gh-example-data/brca1/reads
ga4gh-example-data/brca1/reads/BRCA1_HG00096.bam
ga4gh-example-data/brca1/reads/BRCA1_HG00096.bam.bai
ga4gh-example-data/brca1/reads/BRCA1_HG00099.bam
ga4gh-example-data/brca1/reads/BRCA1_HG00099.bam.bai
ga4gh-example-data/brca1/reads/BRCA1_HG00101.bam
ga4gh-example-data/brca1/reads/BRCA1_HG00101.bam.bai
ga4gh-example-data/brca1/variants
ga4gh-example-data/brca1/variants/1kgPhase3
ga4gh-example-data/brca1/variants/1kgPhase3/1000g_offset.HG00096_HG00099_HG00101.2010502.17_41196312-41277500.vcf.gz
ga4gh-example-data/brca1/variants/1kgPhase3/1000g_offset.HG00096_HG00099_HG00101.2010502.17_41196312-41277500.vcf.gz.tbi
ga4gh-example-data/references
ga4gh-example-data/references/hg37
ga4gh-example-data/references/hg37/ref_brca1.fa.gz
ga4gh-example-data/references/hg37/ref_brca1.fa.gz.fai
ga4gh-example-data/references/hg37/ref_brca1.fa.gz.gzi
ga4gh-example-data/references/hg37/ref_brca1.json
```

jeromekelleher:
Thanks @bwalsh. I don't have time to review the update right now, but here's some responses to your questsions:

- Yes, rebasing on top of upstream/develop should make the extra commits go away. Ultimately, we like to see one (or a handful of logically self-contained) commit(s) which apply cleanly to develop.
- This is unfortunate; we also need the JSON file to go with a reference set now. If you copy one of the JSON files for the references to `references/hg37.json` you should be able to get it going (with a little tweaking, possibly). You should raise this problem on PR #656 though if it's producing data input that doesn't work.

bwalsh:
Added sibling pull request in compliance: https://github.com/ga4gh/compliance/pull/115

bwalsh:
>Yes, rebasing on top of upstream/develop should make the extra commits go
away. Ultimately, we like to see one (or a handful of logically
self-contained) commit(s) which apply cleanly to develop.

I've done some research on this.  As far as I can tell `rebase
upstream/develop`
needs to be applied to `server/g2p`.   I'm willing to do this, however, I
don't have access.

-Brian Walsh

jeromekelleher:
Yes, that's true. @dcolligan, what's the simplest way to bring the g2p branch up to date with 'develop'?

dcolligan:
Someone should (I can, if you want) rebase g2p on top of develop.  It's easy since it's a fast-forward merge... there's no conflicts (that is, nothing has been committed to the g2p branch yet that has not been committed to develop).

bwalsh:
@dcolligan - if you could, that would be wonderful.  I would, but I don't
have write access to the ga4gh/server repo.

Much appreciated,
-b

dcolligan:
@bwalsh done

bwalsh:
@dcolligan  perfect.  thanks!

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
+"""
+Module responsible for translating g2p data into GA4GH native
+objects.
+"""
```
jeromekelleher: We're missing the standard future imports here. Copy these from any existing file.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
...
+
+import ga4gh.protocol as protocol
+import ga4gh.exceptions as exceptions
```
jeromekelleher: Imports are grouped differently elsewhere.

1. futures
2. standard lib
3. third party
4. local imports

See frontend.py as a good example. This grouping helps us know where things come from.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
...
+from urlparse import urlparse, urlunsplit
+import rdflib
```
jeromekelleher: We need to add rdflib to requirements.txt

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
+"""
+Module responsible for translating g2p data into GA4GH native
+objects.
+"""
+
+import ga4gh.protocol as protocol
+import ga4gh.exceptions as exceptions
+from urlparse import urlparse, urlunsplit
```
jeromekelleher: Elsewhere we use 
```
import urlparse
```
and call ``urlparse.urlparse()`` and ``urlparse.urlunsplit()`` in the code. This helps us see where functions are being imported from without having to hunt back up to the header (I agree this is pretty marginal benefit for something like urlparse.urlparse(), but consistency is nice...)

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
...
+                    ?s  OBAN:association_has_object  ?d .
+                    ?d  rdfs:label ?disease_label  .
+                    ?d rdf:type ?disease .
+                    ?s  OBAN:association_has_object_property  ?evidence .
+                    OPTIONAL {  ?evidence  rdfs:label ?evidence_label } .
+              %FILTER%
+            }
+            """
+
+        if G2PDataset.rdfGraph is not None:
```
jeromekelleher: I find this confusing ---why to we have the rdfGraph as a static variable? When will we ever have more than one instance of this class, and why would we want all instances to share the same rdfGraph?

This would be much simpler if we just have 
```
self._rdfGraph = rdflib.Graph()
```
and got rid of this logic.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
...
+
+import ga4gh.protocol as protocol
+import ga4gh.exceptions as exceptions
+from urlparse import urlparse, urlunsplit
+import rdflib
+
+
+class G2PDataset:
+    """
+    an rdf object store
```
jeromekelleher: We need better documentation here. What are we doing? We're translating data from (somewhere) in (some format) into GA4GH values, right? This should be spelled out here.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
...
+    """
+
+    # keep the graph as a static variable
+    rdfGraph = None
+
+    def __init__(self, sources):
+
```
jeromekelleher: Remove the whitespace above and below docstring. I'm not really any the wiser about what sources is though: what does it mean, and what does it contain?

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
...
+        # load with data
+        for source in sources:
+            if not source['format']:
+                G2PDataset.rdfGraph.parse(source['source'])
+            else:
+                G2PDataset.rdfGraph.parse(source['source'],
+                                          format=source['format'])
+        # TODO is this necessary?
+        self.associationsLength = 0
```
jeromekelleher: We should keep all instance variables as private by default. Change to ``self._associationsLength = 0``.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
...
+        # from IPython.core.debugger import Pdb ;        Pdb().set_trace()
+
+        self.associationsLength = len(associations)
+        for association in associations:
+            yield association
+
+    def queryLabels(
+         self, location=None, drug=None, disease=None, pageSize=None,
+         offset=0):
+
```
jeromekelleher: Remove whitespace here before docstring (and elsewhere?)

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
...
+        """
+        Thus query is the main search mechanism.
```
jeromekelleher: s/Thus/This

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
...
+        It queries graph for annotations that match wild card search of
```
jeromekelleher: Some typos in this comment.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
...
+        source = 'http://purl.org/dc/elements/1.1/source'
+        location = 'location'
+        evidenceURI = 'http://purl.obolibrary.org/obo/RO_0002558'
+        hasObject = 'http://purl.org/oban/association_has_object'
+        # location keys
+        GENO_0000408 = 'http://purl.obolibrary.org/obo/GENO_0000408'
+
+        location = annotation['location']
+        if GENO_0000408 in location:
+            id_, ontologySource_ = self.namespaceSplit(
```
jeromekelleher: This id_ here is to prevent the local variable ``id`` from clashing with the global ``id`` function. This is a fairly common Python idiom. No need for ontologySource to end with _.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,338 @@
...
+        """
+        given a url return the id of the resource and the ontology source
+        """
+        o = urlparse(_url)
+        _id = o.path.split(_separator)[-1]
+        _ontologySource = urlunsplit([o[0],
+                                      o[1],
+                                      o[2].replace(_id, ''),
+                                      o[4], ''])
+        return(_id, _ontologySource)
```
jeromekelleher: No need for brackets here,
```python
return id_, ontologySource
```
works fine.

Also, lots of unnecessary underscores here. We just need to suffix id because of clash with built in function. All the other variable here should be without underscores.

ga4gh/frontend.py
```diff
@@ -497,6 +497,12 @@ def searchDatasets():
         flask.request, app.backend.runSearchDatasets)
```
jeromekelleher: Fronend.py seems to have acquired execute permissions. Please change this back. Ditto for _protocoldefinitions.py.

tests/unit/test_views.py
```diff
@@ -377,6 +383,161 @@ def testDatasetsSearch(self):
...
+        """
+        request = protocol.SearchGenotypePhenotypeRequest()
+        response = self.sendPostRequest('/genotypephenotype/search', request)
+        self.assertEqual(501, response.status_code)
+
     def testNoAuthentication(self):
         path = '/oauth2callback'
         self.assertEqual(501, self.app.get(path).status_code)
+
+    def testGenotypePhenotypeSearchFeature(self):
```
jeromekelleher: We should break these tests into a seperate module, tests/unit/test_g2p.py. This module is really just for top-level HTTP stuff.

jeromekelleher:
Looks great @bwalsh, thank for the updates. I've made some more style-type comments above. Some high-level stuff that I think needs to be addressed before we merge:

1. We have to make the g2p dataset parameterisable. We should be able to find it in the data tree just like ``referenceSets`` and ``dataSets``.  
2. We need to break the tests for g2p out into their own module. Do we really need the cgd.ttl file to be included in our test data? Isn't this an example file that should be provided as example data, and not shipped around with our source code? In general, users are going to install the code from pypi, so we can't make any assumptions at all about file system paths. So, I suggest deleting this file, unless it's absolutely necessary we keep it for _testing_ purposes.
3. I'm pretty sure we can simplify the paging stuff. Let's talk about it at the hackathon.

ga4gh/backend.py
```diff
@@ -822,3 +844,102 @@ def __init__(self, dataDir):
...
+            [
+                {'source': 'tests/data/cgd-test.ttl', 'format': "n3"},
+                {'source':
+                 'https://raw.githubusercontent.com/oborel/obo-relations'
+                 '/master/ro.owl',
+                 'format': "xml"}
+            ])
+
+
+class GenotypePhenotypeBackend(AbstractBackend):
```
jeromekelleher: This class doesn't really do anything, just remove it.

ga4gh/backend.py
```diff
@@ -822,3 +844,102 @@ def __init__(self, dataDir):
         for datasetDir in datasetDirs:
             dataset = datasets.FileSystemDataset(datasetDir, self)
             self.addDataset(dataset)
+
+        self._g2p_backend = GenotypePhenotypeBackend(
```
jeromekelleher: Change this to ``self._g2pDataset = genotype_phenotype.G2PDataset(sources)``. There is only on backend object, and it gathers together all the disparate data sources.

jeromekelleher:
Also, it would be good to rebase this against the latest develop so we get the new data input format code.

bwalsh:
Guys,  I was working through these latest comments.  All was fine until I tried to rebase again.  
I wont get into the minutia, but to check my sanity, I cloned https://github.com/ga4gh/server and ran tests on that.    

```
======================================================================
ERROR: test suite for <class 'tests.end_to_end.test_oidc.TestOidc'>
  File "/root/.pyenv/versions/2.7.10/lib/python2.7/site-packages/nose/suite.py", line 209, in run
    self.setUp()
  File "/root/.pyenv/versions/2.7.10/lib/python2.7/site-packages/nose/suite.py", line 292, in setUp
    self.setupContext(ancestor)
  File "/root/.pyenv/versions/2.7.10/lib/python2.7/site-packages/nose/suite.py", line 315, in setupContext
    try_run(context, names)
  File "/root/.pyenv/versions/2.7.10/lib/python2.7/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/home/ga4gh/testupstream/server/tests/end_to_end/server_test.py", line 70, in setUpClass
    cls.otherSetup()
  File "/home/ga4gh/testupstream/server/tests/end_to_end/test_oidc.py", line 58, in otherSetup
    cls.opServer.start()
  File "/home/ga4gh/testupstream/server/tests/end_to_end/server.py", line 243, in start
    super(OidcOpServerForTesting, self).start()
  File "/home/ga4gh/testupstream/server/tests/end_to_end/server.py", line 59, in start
    assert not self.isRunning(), "Another server is running"
AssertionError: Another server is running

======================================================================
ERROR: testSamConversion (tests.unit.test_converters.TestSamConverter)
  File "/home/ga4gh/testupstream/server/tests/unit/test_converters.py", line 69, in testSamConversion
    self.verifyFullConversion(readGroupSet, readGroup, reference)
  File "/home/ga4gh/testupstream/server/tests/unit/test_converters.py", line 43, in verifyFullConversion
    converter.convert()
  File "/home/ga4gh/testupstream/server/ga4gh/converters.py", line 71, in convert
    alignedSegment = SamLine.toAlignedSegment(read, targetIds)
  File "/home/ga4gh/testupstream/server/ga4gh/converters.py", line 129, in toAlignedSegment
    ret.flag = cls.toSamFlag(read)
  File "/home/ga4gh/testupstream/server/ga4gh/converters.py", line 163, in toSamFlag
    if read.nextMatePosition.strand == protocol.Strand.NEG_STRAND:
AttributeError: 'NoneType' object has no attribute 'strand'

```

jeromekelleher:
The first problem is due to a server running in a background process left over from some old tests. If you kill this it should go away.

The second problem is doesn't happen for me when I run the tests. Do you have some other read data in the ``tests/data`` directory? What's happening is you've provoked issue #705, but this shouldn't happen with the way that particular test is set up at the moment. Or, perhaps tests are run slightly differently on a whatever platform you're on... You're using Fedora, right? It'd be surprising if there was a difference between this and Debian.

Issue #606, opened 2015-09-03
``experiment`` field in ReadGroup needs to be filled


jeromekelleher:
This is a similar situation to the ``programs`` field. The RG tag in SAM contains information about libraries, sequencing technology, etc, and this should be mapped to the Experiment object.

jeromekelleher:
Fixed in #608.

PR #608, opened 2015-09-04
Fill out read group experiment field

dcolligan:
Issue #606

### DIFF
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -172,6 +172,7 @@ class HtslibReadGroupSet(datamodel.PysamDatamodelMixin, AbstractReadGroupSet):
         super(HtslibReadGroupSet, self).__init__(parentContainer, localId)
         self._samFilePath = dataFile
         samFile = self.getFileHandle(self._samFilePath)
+        self._setHeaderFields(samFile)
         if 'RG' not in samFile.header or len(samFile.header['RG']) == 0:
             self._defaultReadGroup = True
             readGroup = HtslibReadGroup(self, 'default')
@@ -182,6 +183,20 @@ class HtslibReadGroupSet(datamodel.PysamDatamodelMixin, AbstractReadGroupSet):
                 readGroup = HtslibReadGroup(
                     self, readGroupHeader['ID'], readGroupHeader)
                 self.addReadGroup(readGroup)
+
+    def _setHeaderFields(self, samFile):
+        programs = []
+        if 'PG' in samFile.header:
+            htslibPrograms = samFile.header['PG']
+            for htslibProgram in htslibPrograms:
+                program = protocol.Program()
+                program.id = htslibProgram['ID']
+                program.commandLine = htslibProgram.get('CL', None)
+                program.name = htslibProgram.get('PN', None)
+                program.prevProgramId = htslibProgram.get('PP', None)
+                program.version = htslibProgram.get('VN', None)
+                programs.append(program)
+        self._programs = programs
 
     def openFile(self, dataFile):
         return pysam.AlignmentFile(dataFile)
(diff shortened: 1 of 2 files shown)


ga4gh/datamodel/reads.py
```diff
@@ -187,6 +187,13 @@ def toProtocolElement(self):
         stats.baseCount = None  # TODO requires iterating through all reads
         readGroup.stats = stats
         readGroup.programs = self.getPrograms()
+        experiment = protocol.Experiment()
+        experiment.id = 'TODO'
```
dcolligan: where are these values supposed to come from?  the sam header does not have most of them... metadata, presumably?

jeromekelleher: Here's my take. In the RG tag we have

CN -> sequencingCentre
DT -> runTime
LB -> library

etc. See the SAM spec on page 3 for this. In the 1000G data we have

LB:2845856850   
DS:SRP001294    
PI:206  
CN:WUGSC
PL:ILLUMINA

We just shove these in as-is. For recordCreateTime just take the file creation time for the parent BAM file.

dcolligan:
@jeromekelleher obviously this is a bloody mess right now, but I think this is as much info we can squeeze out of the SAM header according to the spec (except for DT... I think that actually corresponds to recordCreateTime, not runTime)

jmarshall:
Here's my take on what the **RG** fields correspond to in the GA4GH schema:

 | rg | rg.experiment
---- | ---- | ----
ID | rg.id
CN | | rg.experiment.sequencingCenter
DS | rg.description | rg.experiment.description
DT | | rg.experiment.runTime
FO	| |
KS	| |
LB	| | rg.experiment.library
PG	|rg.programs
PI	| rg.predictedInsertSize
PL	| | rg.experiment.instrumentModel
PM	| |
PU	| | rg.experiment.platformUnit
SM	| rg.sampleId

It's not totally clear which ReadGroup fields have been siphoned off into Experiment; e.g., they both have a `description` field, so **RG-DS** might as well populate both of them.  **RG-DT** is the date(/time) that the sample was run on the sequencing machine, which corresponds to `experiment.runTime`; in the absence of other information (like BAM file timestamps) **RG-DT** could I guess be used to populate `experiment.recordCreateTime` too.

```python
# TODO change after merge
readGroupHeader = samFile.header['RG'][0]
```

BAM files typically contain several read groups, so I guess the TODO refers to how at some point this is going to have to pick out the right corresponding **RG** header?  So a BAM file corresponds to GA4GH's ReadGroupSet, and the existing code in `toProtocolElement` could be filling in a bunch of the `readGroup` fields from the corresponding **RG** header too.

ga4gh/datamodel/reads.py
```diff
@@ -189,6 +191,34 @@ def toProtocolElement(self):
         stats.baseCount = None  # TODO requires iterating through all reads
         readGroup.stats = stats
         readGroup.programs = self.getPrograms()
+        experiment = protocol.Experiment()
+        # required experiment fields
+        experiment.id = 'TODO'
```
jeromekelleher: Can we make a hierarchical ID for experiment? We can just tack and 'experiment' on the end to distinguish it from the ReadGroup id.

ga4gh/datamodel/reads.py
```diff
@@ -189,6 +191,34 @@ def toProtocolElement(self):
...
+        experiment.recordUpdateTime = 'TODO'
+        experiment.sequencingCenter = 'TODO'
+        samFile = self.getFileHandle(self._samFilePath)
+        if 'RG' in samFile.header:
+            # TODO change after merge
+            readGroupHeader = samFile.header['RG'][0]
+            experiment.description = readGroupHeader.get('DS', None)
+            experiment.id = 'TODO'
+            experiment.info = {}
+            experiment.instrumentDataFile = 'TODO'
```
jeromekelleher: If a field is nullable, we should set it to ``None`` rather than TODO. I'd guess that some of these will never be done.

ga4gh/datamodel/reads.py
```diff
@@ -189,6 +191,34 @@ def toProtocolElement(self):
         stats.baseCount = None  # TODO requires iterating through all reads
         readGroup.stats = stats
         readGroup.programs = self.getPrograms()
+        experiment = protocol.Experiment()
+        # required experiment fields
+        experiment.id = 'TODO'
+        experiment.instrumentModel = 'TODO'
+        experiment.recordCreateTime = 'TODO'
```
jeromekelleher: might as well set this to ``self._iso8601Time``.

ga4gh/datamodel/reads.py
```diff
@@ -189,6 +191,34 @@ def toProtocolElement(self):
...
+            experiment.instrumentModel = 'TODO'
+            experiment.library = readGroupHeader.get('LB', None)
+            experiment.libraryLayout = 'TODO'
+            experiment.molecule = 'TODO'
+            experiment.name = 'TODO'
+            experiment.platformUnit = readGroupHeader.get('PU', None)
+            experiment.recordCreateTime = self._iso8601
+            experiment.recordUpdateTime = self._iso8601
+            experiment.runTime = 'TODO'
+            experiment.selction = 'TODO'
```
jeromekelleher: Typo? selction?

jeromekelleher:
Thanks @jmarshall, this is super-useful!

dcolligan:
@jeromekelleher updated

jeromekelleher:
Looks good, +1. Merging.

PR #609, opened 2015-09-04
Implement read strand flags

### LINKED ISSUE #491: Strand is always "positive"
See `TODO` in [`server/reads.py`][1]. This surprised me when I was trying to debug a disagreement between `samtools` and the GA4GH server.

[1]: https://github.com/ga4gh/server/blob/40bdc60eee6e82981abb8f18e27f7be88c7193d9/ga4gh/datamodel/reads.py#L265

jmarshall:
I am not a Python programmer, and I have not attempted to feed this to a python.  Your mileage may vary.  But I do know what (SAM) plugs into what (GA4GH).  Fixes #491.

(Not too sure why there's this `SamFlags` class rather than using `is_reverse(read)` et al as supplied by pysam…?)

### DIFF
--- a/ga4gh/converters.py
+++ b/ga4gh/converters.py
@@ -11,6 +11,7 @@ import collections
 import pysam
 
 import ga4gh.datamodel.reads as reads
+import ga4gh.protocol as protocol
 
 
 class AbstractConverter(object):
@@ -144,6 +145,12 @@ class SamLine(object):
         if read.properPlacement:
             reads.SamFlags.setFlag(
                 flag, reads.SamFlags.PROPER_PLACEMENT)
+        if read.alignment.position.strand == protocol.Strand.NEG_STRAND:
+            reads.SamFlags.setFlag(
+                flag, reads.SamFlags.REVERSED)
+        if read.alignment.position.strand == protocol.Strand.NEG_STRAND:
+            reads.SamFlags.setFlag(
+                flag, reads.SamFlags.NEXT_MATE_REVERSED)
         if read.readNumber:
             reads.SamFlags.setFlag(
                 flag, reads.SamFlags.READ_NUMBER_ONE)
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -51,6 +51,8 @@ class SamFlags(object):
     """
     NUMBER_READS = 0x1
     PROPER_PLACEMENT = 0x2
+    REVERSED = 0x10
+    NEXT_MATE_REVERSED = 0x20
     READ_NUMBER_ONE = 0x40
     READ_NUMBER_TWO = 0x80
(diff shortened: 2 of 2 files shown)


jeromekelleher:
Thanks @jmarshall, this is great! Trivial for you, but head-scratchingly hard for us I'm afraid... I've added some notes above on how to make this pass the style checks. If you want to implement these, then we can check the code in, so you can get proper git-credit for this. If you'd prefer not, we can make another PR based on this easily enough.

You can run the CI checks locally by running
```
python scripts/run_tests.py
```

ga4gh/datamodel/reads.py
```diff
@@ -350,7 +352,8 @@ def convertReadAlignment(self, read):
             read.reference_id)
         ret.alignment.position.position = read.reference_start
         ret.alignment.position.strand = \
```
jeromekelleher: The ternary operator is a bit awkward here with the style requirements for line breaking. The following suggestion is easier style-wise, and is quite straightforward:
```python
ret.alignment.position.strand =  protocol.Strand.POS_STRAND
if SamFlags.isFlagSet(read.flag, SamFlags.REVERSED):
     ret.alignment.position.strand =  protocol.Strand.NEG_STRAND
```

ga4gh/converters.py
```diff
@@ -144,6 +145,12 @@ def toSamFlag(cls, read):
         if read.properPlacement:
             reads.SamFlags.setFlag(
                 flag, reads.SamFlags.PROPER_PLACEMENT)
+        if read.alignment.position.strand == protocol.Strand.NEG_STRAND
```
jeromekelleher: Need a colon after the ``if`` to designate the start of the block, i.e.:
```python
if read.alignment.position.strand == protocol.Strand.NEG_STRAND:
    # stuff
```

jmarshall: D'oh!

jmarshall:
Done.  Thanks for the kindergarten-level Python assistance…

jeromekelleher:
Awesome, thanks for this @jmarshall!

+1

@dcolligan, we should create an issue to make some data driven tests on this change after this is merged.

dcolligan:
@jmarshall thanks for this.  I want to echo Jerome wrt the "trivial" issue.  Just as you are not a python programmer, we are not biologists!  There is also a lot of tacit knowledge not in our (Jerome and my) brains wrt rationales for why the schema was developed a certain way, the current state of the art in bioinformatics, if certain warts in the schema were deliberate or an oversight or someone somewhere planned on changing in the future... in short, for a lot of questions of the type that you just asked wrt  `is_reverse` we are similarly scratching our heads.  The reads implementation is based on me just reading the SAM specification and trying to harmonize it with pysam / the ga4gh api to the extent that I could.  Expert advice is always welcome!

PR #617, opened 2015-09-10
Implement GET /datasets/{id}

dcolligan:
Issue #523


hjellinek:
+1
Looking forward to this!

jeromekelleher:
I think we should change the commit so that it has the original author's email address attached. Is this possible @dcolligan?

dcolligan:
@jeromekelleher okay, did that (the author has been changed, but not the committer, which is still me)

PR #622, opened 2015-09-14
Fix strand flag copy-pasto

jmarshall:
Alas, I was so busy getting the Python syntax right that I didn't adequately check the actual tests here…


dcolligan:
+1

jeromekelleher:
+1. Merging.

PR #624, opened 2015-09-14
619 variants bug

dcolligan:


jeromekelleher:
Looks good @dcolligan, thanks. We should add a test to confirm that we have implemented getVariants correctly though. test_simulated_stack is probably the best place for this. Looking at it now, we don't have any ``testGetVariant`` in there, so we should definitely fill this gap in the tests as part of this PR.

dcolligan:
@jeromekelleher updated

tests/unit/test_simulated_stack.py
```diff
@@ -334,6 +334,29 @@ def testGetVariantSet(self):
...
+        responseData = self.sendSearchRequest(
+            path, request, protocol.SearchVariantsResponse)
+        variant = responseData.variants[0]
+
+        # get 'the same' variant using the get method
+        path = '/variants'
+        responseObject = self.sendGetObject(
+            path, variant.id, protocol.Variant)
+        self.assertEqual(responseObject.referenceName, referenceName)
+        self.assertEqual(responseObject.start, start)
```
jeromekelleher: Why can't we do a deep equality test here? The simulated variant should be exactly the same whichever way we get it, that's why we seed the RNG with the position. We should be able to directly compare the objects, i.e., ``variant1 == variant2``. Also, testing just the first one seems a bit wimpy --- set a page size of 10 (say) and do a GET for all of them?

dcolligan: The ids don't come back the same, but I guess that means there's an issue with our random number generation... I'll look into it...

dcolligan:
@jeromekelleher fixed.  the issue was that in `getVariants` there was an extra call to the seeded generator before `generateVariant` was called, resulting in a different-state generator getting passed in and misaligning it with what is passed in to `generateVariant` via `getVariant`.

jeromekelleher:
Nice, well spotted. +1.

Issue #540, opened 2015-08-04
searchReferences only matches MD5 checksums when SearchReferencesRequest.md5checksums contains "TODO"


hjellinek:
(The point of this issue is that the ref server needs to implement the MD5-related `Reference` search and retrieval operations.)

**Background**: I'm using data from the server unit tests as the server's data.  The reference data looks like this:
```
references/
    example_1/
        simple.fa.gz
        simple.fa.gz.fai
        simple.fa.gz.gzi
```

**Action**: I create a `SearchReferencesRequest` and set the value of `md5checksums` to contain the string `"TODO"` and pass it to `searchReferences`.

**Expected**:
The call to `searchReferences` will return a `SearchReferencesResponse` with its `references` containing zero results, since the `"TODO"` in the request is not a legal MD5 value.

**Actual**:
The call to `SearchReferencesResponse` returns a `SearchReferencesResponse` containing a single `Reference`.  The `Reference`'s `md5checksum` field contains the string __`"TODO"`__.

jeromekelleher:
The issue here is that we haven't settled on an import format for references. We have the FASTA files with the data, but we need something else to go along with it to hold the metadata. This must also contain the md5 that we want. We can't compute the md5 on server startup, as references are large.

I think the best thing to do here is to make the md5 field an empty string, and raise a NotImplemented error when someone tries to search by md5 values. We should also create an issue for proper handling of references.

jeromekelleher:
Fixed in #626.

Issue #539, opened 2015-08-04
searchReferences only returns References with md5checksum == "TODO"


hjellinek:
**Background**: I'm using data from the server unit tests as the server's data.  The reference data looks like this:
```
references/
    example_1/
        simple.fa.gz
        simple.fa.gz.fai
        simple.fa.gz.gzi
```

**Action**: I create a `SearchReferencesRequest` without setting the value of `md5checksums` (meaning it's equal to the empty list) and pass it to `searchReferences`.

**Expected**:
The call to `searchReferences` will return a `SearchReferencesResponse` with its `references`  containing a single `Reference` object with a valid `md5checksum` field (32 characters long, all hex, and ideally _correct_, though I have no means for testing the value's validity yet).

**Actual**:
The call to `SearchReferencesResponse` returns a `SearchReferencesResponse` containing a single `Reference`.  The `Reference`'s `md5checksum` field contains the string __`"TODO"`__.

jeromekelleher:
This is related to #540. We can't compute MD5 values on the fly because references are large (several GB). Until we figure out how to properly import reference data, we need a filler value here. Is it better to have the empty string or something that looks like an md5 (say, 32 zeros)?

jeromekelleher:
Fixed in #626.

PR #626, opened 2015-09-15
Adds reference search filters.

jeromekelleher:
Also brings cli up to date with reference schema changes.

We add filtering to reference and reference sets search in the most straightforward way I could think of. It is disappointing that we have to build a list of all the objects, but this avoids a lot of tricky issues. The number of references and reference sets is likely to be rather small, so I doubt this will be an issue in practice.

The tests aren't entirely exhaustive, but they should exercise the code well enough.

Issues #539 and #540.

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -361,6 +361,14 @@ class AbstractBackend(object):
                 nextPageToken = str(currentIndex)
             yield object_.toProtocolElement(), nextPageToken
 
+    def _objectListGenerator(self, request, objectList):
+        """
+        Returns a generator over the objects in the specified list using
+        _topLevelObjectGenerator to generate page tokens.
+        """
+        return self._topLevelObjectGenerator(
+            request, len(objectList), lambda index: objectList[index])
+
     def _singleObjectGenerator(self, datamodelObject):
         """
         Returns a generator suitable for a search method in which the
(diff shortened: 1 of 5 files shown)


ga4gh/backend.py
```diff
@@ -395,8 +403,21 @@ def referenceSetsGenerator(self, request):
...
         """
-        return self._topLevelObjectGenerator(
-            request, self.getNumReferenceSets(), self.getReferenceSetByIndex)
+        results = []
+        for obj in self.getReferenceSets():
+            include = True
+            if request.md5checksum is not None:
+                if request.md5checksum != obj.getMd5Checksum():
+                    include = False
+            if request.accession is not None:
```
dcolligan: if include is false, we can bail on the subsequent if clauses... how about something here like `if include is True and request.accession is not None`

dcolligan: or maybe there's already enough control flow logic here and avoiding a small time penalty isn't worth further cluttering the code...

dcolligan:
some of the cli tests are failing... besides that and my nit I'm +1

ga4gh/backend.py
jeromekelleher: I think it's easier to read as it is --- the time penalty will be trivial I think. I can add
```python
if not include and request.accession is not None:
    # stuff
``` 
if you prefer it though, I don't have a strong opinion.

jeromekelleher:
CLI tests fixed @dcolligan.

dcolligan:
+1

Issue #287, opened 2015-03-25
Support returning unmapped reads


dcolligan:
pysam's `AlignmentFile.fetch`, which we use to implement the `reads/search` endpoint, does not return unmapped reads.  Presumably clients would be interested in receiving unmapped reads, however.

bioinformed:
@dcolligan: There is nuance here.  AlignmentFile.fetch will return an unmapped reads if another segment is mapped (see section 2, item 4.1 of the SAM specification).  If all segments in the template are unmapped, those reads are obtained by fetching from tid=-1.

dcolligan:
@bioinformed thanks for the heads up.  Your input is always appreciated.

jeromekelleher:
What is the correct behaviour here according to the spec? What are the conditions under which we're supposed to return unmapped reads?

dcolligan:
In the schemas repo, SearchReadsRequest.referenceId has this comment:
```
The reference to query. Leaving blank returns results from all
references, including unmapped reads - this could be very large.
```
and SearchReadsResponse.alignments has this comment:
```
The list of matching alignment records, sorted by position.
Unmapped reads, which have no position, are returned last.
```
If/when we support unmapped reads, this will introduce some complexity for our paging algorithm since all unmapped reads have POS = 0 and are not returned by the regular `fetch` call.

jeromekelleher:
OK, so at the moment, we're not quite implementing the protocol as defined? Perhaps we should insist on having a reference, and throw a `NotImplemented("Unmapped reads not currently supported")` if someone makes a query without specifying a reference? That way at least we are correctly implementing the subset of the schema that we support and are explicit about not supporting the rest of it.

dcolligan:
That sounds like a good stopgap measure.  Could you file an issue?

Also, would there be any situation in which a client would want only unmapped reads?  If so, the protocol does not provide for this -- we only have (as specified) mapped + unmapped (that is, all reads) and mapped for a particular reference.

diekhans:
Getting only unmapped reads is an important use cases (search
for viral DNA associated with cancer for example).

I seem to remember this being discussed, but don't know if there
is an issues for it.

Great example of why one has to implement an API for it to be
real.

jeromekelleher:
@dcolligan --- done, see issue #496.

@diekhans --- this searchReads query is very complex as it is. Perhaps a separate API call to get unmapped reads would be better than trying to funnel everything through this one interface?

PR #634, opened 2015-09-16
Reads/references glue

jeromekelleher:
This PR glues together reads and references so that we know what ReferenceSet a ReadGroupSet is referring to. It's currently a WIP, so feedback is welcome.

The basic idea is to equate @SQ.AS SAM tag with the ReferenceSet.name, and use this as the connection point between the two parts of the API. There's lots of details to be resolved, but this seems to be the simplest approach available.

@dcolligan, there's a few bits here I'm a bit iffy on, particularly the new ``backend`` parameter to all the HtslibXSet classes. I can't see any other way of giving these objects access to the backend context without resorting to global singletons (which is evil).

I've also jiggled the file names in the download data script so that our IDs aren't so absurdly long.


jeromekelleher:
@dcolligan --- updated. I think this brings us in sync date, modulo some documentation.

We should also create some issues about our lack of support for unmapped reads.

ga4gh/datamodel/reads.py
```diff
@@ -10,9 +10,26 @@
...
+import ga4gh.datamodel.references as references
 import ga4gh.exceptions as exceptions
+import ga4gh.protocol as protocol
+
+
+def parseMalformedBamHeader(headerDict):
+    """
+    Parses the (probably) intended values out of the specified
+    BAM header dictionary, which is incompletely parsed by pysam.
+    This is cause by some tools incorrectly using spaces instead
```
dcolligan: typo: "this is caused"

dcolligan:
For unmapped reads, this is open: https://github.com/ga4gh/server/issues/287

dcolligan:
One thing that has been mentioned previously to avoid passing lots of objects around in constructors is some kind of IOC pattern.  We already do have one global singleton, the file handle cache (maybe two counting the app itself).  Maybe it's time to look into something like that before we push this change.

jeromekelleher:
I don't think the file handle cache is a global singleton @dcolligan --- it's a _module_ singleton, only referenced within one specific module, and only referenced in one method. If it gets referenced any more widely, then I would say it should be brought into the backend as an instance variable.

For the Flask app, we have no choice but to follow Flask's design pattern. We have isolated its use to strictly within the frontend module, so it doesn't pollute the rest of our code. I think singletons create a lot of problems, and they should be avoided where possible.

Do you have something specific in mind  for the IOC pattern?

dcolligan:
I've thought about it a bit, and concluded that what you already have is IOC.  But getting beyond the jargon, I don't have a better alternative.  Other things that we could do would be:

- Creating some annotation like `@injectBackend` which adds the backend to the `__dict__` of the method's instance
- Add some class (BackendView) that implements a method called `getBackend` that then can be (multiply) inherited from 
- Add some global broker class which keeps track of singletons like backend and can return them upon request, e.g. `broker.getService(Backend)`

The issue with all of these is that they create a circular dependency.  `backend` already depends on `datamodel` since it needs to create objects like `Dataset`.  So we can't have `datamodel` classes depending on `backend`... unless we want to start adding CDs to our code, which I would be generally against.  The way your solution avoids this by making `backend` an argument that is passed into the `datamodel` classes but -- crucially -- is created elsewhere so `datamodel` doesn't have to import it.

jeromekelleher:
Sounds good @dcolligan ---- I think what we have here is probably simpler than the other options, and works for now. We can revisit this design if we think of a better approach.  Are you happy to merge this, or are there outstanding issues?

dcolligan:
+1

PR #640, opened 2015-09-18
Set Docker to install dependencies, and list all dependencies

adamnovak:
This updates the graph server's dockerfile to install the dependencies using pip, and adds in the missing pyfasta dependency.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,9 @@ MAINTAINER Alastair Firth
 ADD deploy/config.py /app/ga4gh/docker_config.py
 RUN cp --no-clobber /app/ga4gh/docker_config.py /app/ga4gh/config.py
 
+# Install dependencies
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Should we be installing dependencies for the main docker images also @adamnovak, or is this only required for the graph branch? 

Some tests are failing, but since this is on the graph branch the usual rules don't apply, so you're free to merge if @macieksmuga is happy with it.

macieksmuga:
@jeromekelleher the graph branch has separate dependencies, and unfortunately the tests haven't been updated to deal with these correctly. So I'll merge this shortly.

adamnovak:
I would recommend a corresponding change to install the requirements.txt
dependencies on the main branch also, just to make sure that if a new
dependency is added that's not in the base image, it doesn't break the
docker build.

PR #641, opened 2015-09-21
Removed Google API workarounds on client side.

jeromekelleher:
This PR removes workarounds put in to allow our client work with the v0.5 Google API. Since we no longer support v0.5, it doesn't make sense to keep these complications lying around in our code base. We should assume that Google will fully implement the next API version, and so we won't need these workarounds to interoperate.


dcolligan:
+1

PR #643, opened 2015-09-21
CLI refactor

jeromekelleher:
This is a WIP, so comments welcome. The PR depends on #641 and #634, so has a pretty large diff right now. This should look more sensible once these are merged and this is rebased.

The basic idea here is to refactor the CLI and client library so that we can make the client application smarter. If we give ``variant-sets-search`` no dataset arguments, we run the search over all datasets. Similar logic applies to all the other requests: if we don't specify enough information, the client application will query the server and do the right thing. We also add JSON output, which I think is quite useful and might persuade people to stop using CURL to query the server. These changes are all motivated by the upcoming documentation updates, which are going to be hard to write with any clarity using the current CLI.

The major refactoring here is that we change the client methods like ``searchVariants`` to take the arguments directly as parameters, and not the request objects. We therefore insulate the user from this detail of the protocol, and allow them to use the methods in a relatively high-level way.

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -7,7 +7,6 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import time
 import argparse
 import logging
 import unittest
@@ -17,7 +16,6 @@ import unittest.suite
 import requests
 
 import ga4gh.client as client
-import ga4gh.protocol as protocol
 import ga4gh.converters as converters
 import ga4gh.frontend as frontend
 import ga4gh.configtest as configtest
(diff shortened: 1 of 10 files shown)


dcolligan:
I support the idea of a more usable and documentable CLI, but I'll hold off on diff comments until the other changes are merged so the diff is manageable.

dcolligan:
@jeromekelleher I'll review this after you rebase this (both PRs it depends on have been merged)

jeromekelleher:
Should be ready for review @dcolligan. See the commit message for a summary of the changes.

ga4gh/cli.py
```diff
@@ -34,173 +32,6 @@
 AVRO_LONG_MAX = 2**31 - 1
```
dcolligan: This is a separate issue, but this should probably be moved to another file, for generality... like `avro_constants` or even in `protocol`

ga4gh/cli.py
```diff
@@ -933,6 +828,137 @@ def client_main(parser=None):
...
+
+class Ga2VcfRunner(SearchVariantsRunner):
+    """
+    Runner class for the ga2vcf
+    """
+    def __init__(self, args):
+        super(Ga2VcfRunner, self).__init__(args)
+        self._outputFile = args.outputFile
+        self._binaryOutput = False
+        if args.output_format in "bcf":
```
dcolligan: shouldn't this be `args.output_format == "bcf"`?  similar comment for a number of other places where `output_format` is used

dcolligan:
Thoughts:
- the new client API is more convenient for users, although less forwards-compatible... it will have to change every time the protocol is modified in a way that adds/deletes (especially required) args from the request (on the other hand, one could argue that we are currently only pushing the breakage onto other layers since the request object changes)
- the filling-in-arguments idea with the CLI, again, makes this change more user friendly, but less explicit... I suppose we don't really care about maintaining command line compatibility, so this seems generally good
- removing dead code = win
- json output format is convenient 
- I'm not thrilled about the reordering of the ID and URL CLI args (previously the url was always at the end), but hopefully we will have a config file for this soon that makes typing out the URL unnecessary

ga4gh/cli.py
jeromekelleher: It's a bit of a wart that we have this at all --- the protocol should be able to state that end=null means everything on the reference. Fixing the protocol would be the preferred way to get rid of this.

ga4gh/cli.py
jeromekelleher: The idea was to allow partial matching. However, this is a bit complicated (and I got it wrong in different places) so I've gone with the simpler exact match. Update pushed.

jeromekelleher:
Thanks for the feedback @dcolligan. Replies below:

> the new client API is more convenient for users, although less forwards-compatible... it will have to change every time the protocol is modified in a way that adds/deletes (especially required) args from the request (on the other hand, one could argue that we are currently only pushing the breakage onto other layers since the request object changes)

I think this is probably OK --- we need to start worrying about forward compatability at some point. Using keyword arguments makes this slightly easier to deal with, and we can always add some compatibility options to the client library later, say, warning when a deprecated argument is used or whatever.

> the filling-in-arguments idea with the CLI, again, makes this change more user friendly, but less explicit... I suppose we don't really care about maintaining command line compatibility, so this seems generally good

Hopefully this will take some of the tedium of copying and pasting arguments. In most cases the intention when we provide no arguments is fairly clear. In reality, the client CLI is just a way of demonstrating the API, so making it easy for people to get started with should be a high priority.

> I'm not thrilled about the reordering of the ID and URL CLI args (previously the url was always at the end), but hopefully we will have a config file for this soon that makes typing out the URL unnecessary

Sorry about bringing this change in now, I should have spotted it earlier and discussed it then. For me, having the URL argument consistently at position 1 is important for usablility. If the URL is sometimes arg1 and other times arg2, then I have to think about what type of command I'm running. Since we always have to specify a URL, and we only sometimes need an ID, I think we must have the URL consistently as the first argument.

dcolligan:
Ok, as long as we consistently have the URL as first always or last always; we should stick to whatever convention we establish.

+1

Issue #647, opened 2015-09-23
Searching by non-existent name should not 404


jeromekelleher:
Currently when we run a search with a name specified which does not exist, the server returns a NOT FOUND error. For example,

```
$ python client_dev.py readgroupsets-search http://localhost:8000 --name=SDF
ERROR:ga4gh.client:404 {"errorCode": 983484777, "message": "ReadGroupSet with name 'SDF' not found"}
```

This should return an empty set of results. I think ``readGroupSetGenerator`` and ``callSetGenerator`` are affected by this in ``backend.py``.

@hjellinek, what do the compliance tests expect here?

dcolligan:
That's not the error I get for that command on develop:
```
Traceback (most recent call last):
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/dannyc/ga4gh/server/ga4gh/frontend.py", line 461, in searchReadGroupSets
    flask.request, app.backend.runSearchReadGroupSets)
  File "/Users/dannyc/ga4gh/server/ga4gh/frontend.py", line 391, in handleFlaskPostRequest
    return handleHttpPost(flaskRequest, endpoint)
  File "/Users/dannyc/ga4gh/server/ga4gh/frontend.py", line 268, in handleHttpPost
    responseStr = endpoint(request.get_data())
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 681, in runSearchReadGroupSets
    self.readGroupSetsGenerator)
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 556, in runSearchRequest
    for obj, nextPageToken in objectGenerator(request):
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 405, in readGroupSetsGenerator
    dataset = self.getDataset(request.datasetId)
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 271, in getDataset
    raise exceptions.DatasetNotFoundException(id_)
DatasetNotFoundException: The requested dataset 'None' was not found
```
Isn't our usual MO to return a 404 if any part of the query doesn't exist?  Why would we do things differently here?

[...]

PR #650, opened 2015-09-23
Prevent read group set requests with...

dcolligan:
...nonexistant names from issuing 404s

Issue #647

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -389,6 +389,12 @@ class AbstractBackend(object):
         """
         yield (datamodelObject.toProtocolElement(), None)
 
+    def _noObjectGenerator(self):
+        """
+        Returns a generator yielding no results
+        """
+        return iter([])
+
     def datasetsGenerator(self, request):
         """
         Returns a generator over the (dataset, nextPageToken) pairs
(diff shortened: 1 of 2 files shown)


ga4gh/backend.py
```diff
@@ -389,6 +389,13 @@ def _singleObjectGenerator(self, datamodelObject):
         """
         yield (datamodelObject.toProtocolElement(), None)
 
+    def _noObjectGenerator(self):
+        """
+        Returns a generator yielding no results
+        """
+        return
```
jeromekelleher: This looks a bit weird --- how about 
```
return iter([])
```

jeromekelleher:
Looks good @dcolligan, but we really should add some test cases to cover this. ``test_simulated_stack.py`` is probably the best place to do this. In fact, the existing test ``testCallSetsSearch`` should be failing now, as it's asserting the old 404 behaviour; we should get to the bottom of this.

dcolligan:
@jeromekelleher updated.  I don't see why testCallSetsSearch would be affected.  This change only affects the path for searching for read group sets.

jeromekelleher:
Ah yes, my bad. We do need to fix CallSets too, so we shouldn't close #647 until that's done.

+1.

Issue #648, opened 2015-09-23
Dataset and VariantSet name not set.


jeromekelleher:
The Dataset name is not set. Running
```
$ python client_dev.py datasets-search http://localhost:8000
```
gives 
```
MWtnLXAzLXN1YnNldA==    None
``` 
The expected output is 
```
MWtnLXAzLXN1YnNldA==  1kg-p3-subset
```

Similarly for VariantSets --- the name should be the directory name.

jeromekelleher:
Again, this example is using the updated client tools in #643.

dcolligan:
Fixed in #651

PR #651, opened 2015-09-24
Set VariantSet and Dataset names

dcolligan:
Issue #648


jeromekelleher:
Looks good, but we do need some tests. We should be able to test this easily enough in ``test_simulated_stack.py``.

dcolligan:
@jeromekelleher updated

jeromekelleher:
+1. Merging.

PR #654, opened 2015-09-24
Made pageSize an attribute of httpClient.

jeromekelleher:
Having a pageSize argument on all of the  searchX methods in the httpClient is confusing. This is a low-level network utilisation knob, which doesn't affect the output of the methods in any way. It therefore makes much more sense to make this an attribute of the httpClient rather than to pass it in as a parameter to each search call.


dcolligan:
makes sense +1

jeromekelleher:
@dcolligan --- merge me!

PR #656, opened 2015-09-24
Added python script to wrangle compliance data into server format.

macieksmuga:


dcolligan:
I only agreed to a bash script because, without the bash syntax in front of me, I forgot how truly unintuitive and ugly it is... you're sure you can't do this in python, the language that everyone on this project understands and is comfortable with?

(Not to mention doing it in python has advantages... the command line parsing library will allow you to enforce correct invocation.  As of right now, it looks like the script will run (and do ???) if not provided any arguments.)

jeromekelleher:
I tend to agree with @dcolligan --- I think this would probably take about 20 minutes to port to Python, plus the data could be downloaded directly from github, rather than requiring  a synced compliance data dir. However, any script is better than no script, so I won't push too hard on it.

One thing we do need to fix though is the lack of ReferenceSet metadata. Since #657 was merged, we have to have a JSON file next to the reference set directory telling us about it. This will need to be added somehow.

macieksmuga:
The JSON sidecar file for the `ReferenceSet` was added to compliance_redux in that repo's PR 99 (https://github.com/ga4gh/compliance/pull/99/files)

macieksmuga:
Good point @jeromekelleher about sourcing the compliance dataset from GitHub. I'll add this when I rewrite this script in Python (when I get the chance).

jeromekelleher:
What do you think @dcolligan, merge and make an issue for translation to Python or wait for Python version? I don't mind.

dcolligan:
I think waiting for the python version would be a good idea.

macieksmuga:
@jeromekelleher @dcolligan this is just a trivial convenience script - are you guys feeling the lack of it, or can you load the latest compliance data into the ref server without too much pain?

dcolligan:
I guess that's your decision -- if it's helpful enough to commit to the repo, it's helpful enough to write in python!

bwalsh:
We remain a bit stuck on getting going on the compliance tests.  Specifically, the mvn project builds fine, but I remain stuck on getting compliance data into the server.

I used the script from  https://github.com/ga4gh/server/pull/656 which worked.  However, when I launch the server, I get the following

```
# python server_dev.py
    ga4gh.cli.server_main()
  File "/home/ga4gh/tmp/rebase/server/ga4gh/cli.py", line 71, in server_main
  File "/home/ga4gh/tmp/rebase/server/ga4gh/frontend.py", line 200, in configure
    theBackend = backend.FileSystemBackend(dataSource)
  File "/home/ga4gh/tmp/rebase/server/ga4gh/backend.py", line 836, in __init__
    referenceSetName, relativePath, self)
  File "/home/ga4gh/tmp/rebase/server/ga4gh/datamodel/references.py", line 366, in __init__
    self._setMetadata()
  File "/home/ga4gh/tmp/rebase/server/ga4gh/datamodel/references.py", line 371, in _setMetadata
    with open(metadataFileName) as metadataFile:
IOError: [Errno 2] No such file or directory: u'ga4gh-example-data/references/hg37.json'

```

The contents of the test directory is:

```
 server]# find ga4gh-example-data -print
ga4gh-example-data
ga4gh-example-data/brca1
ga4gh-example-data/brca1/reads
ga4gh-example-data/brca1/reads/BRCA1_HG00096.bam
ga4gh-example-data/brca1/reads/BRCA1_HG00096.bam.bai
ga4gh-example-data/brca1/reads/BRCA1_HG00099.bam
ga4gh-example-data/brca1/reads/BRCA1_HG00099.bam.bai
ga4gh-example-data/brca1/reads/BRCA1_HG00101.bam
ga4gh-example-data/brca1/reads/BRCA1_HG00101.bam.bai
ga4gh-example-data/brca1/variants
ga4gh-example-data/brca1/variants/1kgPhase3
ga4gh-example-data/brca1/variants/1kgPhase3/1000g_offset.HG00096_HG00099_HG00101.2010502.17_41196312-41277500.vcf.gz
ga4gh-example-data/brca1/variants/1kgPhase3/1000g_offset.HG00096_HG00099_HG00101.2010502.17_41196312-41277500.vcf.gz.tbi
ga4gh-example-data/references
ga4gh-example-data/references/hg37
ga4gh-example-data/references/hg37/ref_brca1.fa.gz
ga4gh-example-data/references/hg37/ref_brca1.fa.gz.fai
ga4gh-example-data/references/hg37/ref_brca1.fa.gz.gzi
ga4gh-example-data/references/hg37/ref_brca1.json
```

macieksmuga:
The script is now in python and is far more general in that it can format arbitrary dataset inputs. Please re-review this now-completely reformulated PR, @jeromekelleher  @dcolligan.

macieksmuga:
Also note that I switched to doing all the compression & indexing of the data files using appropriate pysam commands (some, uh, better documented than others).

scripts/prepare_compliance_data.py
```diff
@@ -0,0 +1,155 @@
...
+        self.datasetDirs = [path.join(self.outputDirectory, ds)
+                            for ds in self.datasets]
+
+    def run(self):
+        if not path.exists(self.outputDirectory):
+            os.makedirs(self.outputDirectory)
+
+        # Clean out, make and re-populate references directory
+        # For now, assume a single, statically-named referenceSet
+        print("Converting references...", file=sys.stderr)
```
dcolligan: why does this go to sys.stderr instead of sys.stdout?

scripts/prepare_compliance_data.py
```diff
@@ -0,0 +1,155 @@
...
+        "--outputDirectory", "-o", default="ga4gh-compliance-data",
+        help="The directory to output the server-ready data bundle to.")
+    parser.add_argument(
+        "--inputDirectory", "-i",
+        help="Path to local directory containing compliance dataset.",
+        default='.')
+    parser.add_argument('--verbose', '-v', action='count', default=0)
+    args = parser.parse_args()
+    cdm = ComplianceDataMunger(args)
+    try:
```
dcolligan: I don't see why the try/finally is necessary.  The only cleanup is printing that it's done, which shouldn't be printed if the process fails.

scripts/prepare_compliance_data.py
```diff
@@ -0,0 +1,155 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import shutil
+import argparse
+import os
+import sys
+from os import path
+from glob import glob
```
dcolligan: We prefer the import Y instead of the from Y import X so it's clear when reading the code which calls are external library calls... see https://github.com/ga4gh/server/blob/develop/STYLE.md#imports

scripts/prepare_compliance_data.py
```diff
@@ -0,0 +1,155 @@
...
+                                destFile)
+                    # Pysam's tabix_index automatically compresses the file
+                    # in place, creates a tabix index.
+                    pysam.tabix_index(destFile, preset="vcf")
+
+    def cleanup(self):
+        print("done converting compliance data.", file=sys.stderr)
+
+
+def main():
```
dcolligan: I like putting the utils.Timed decorator on the main function of scripts, just to see how long they take to execute; it's often helpful.

scripts/prepare_compliance_data.py
```diff
@@ -0,0 +1,155 @@
+"""
+A script that takes the compliance dataset (the released version
+of which is at https://github.com/ga4gh/compliance/tree/master/test-data)
```
dcolligan: If this is always in a well known place, why doesn't the script download it?

diekhans: it could default to pull from get, but needs to be parameterized,
since it might come from a fork or locally modified data.,

scripts/prepare_compliance_data.py
diekhans: Normally with unix programs, verbose messages go to stderr, with
stdout being used for piping data.  Note that
logging.handlers.StreamHandler defaults to stderr.

Although messages like this are not that useful.  If it needs to
talk to me, I would rather have it tell me after that fact that
it important so many records, just to give a sense that it did
something.

scripts/prepare_compliance_data.py
```diff
@@ -0,0 +1,155 @@
...
+class ComplianceDataMunger(object):
+    def __init__(self, args):
+        self.inputDirectory = args.inputDirectory
+        self.outputDirectory = args.outputDirectory
+
+        # get all the reference files (they'll be the ones with .fa extension)
+        self.referenceFiles = map(path.basename,
+                                  glob(path.join(self.inputDirectory, "*.fa")))
+
+        self.refsetsDirectory = path.join(self.outputDirectory,
```
jeromekelleher: As @dcolligan pointed out above, we tend to use ``os.path.join`` and ``glob.glob`` elsewhere.

scripts/prepare_compliance_data.py
```diff
@@ -0,0 +1,155 @@
...
+                                          "referenceSets")
+        self.hg37Directory = path.join(self.refsetsDirectory, "hg37")
+
+        # datasets
+        self.datasetsDirectory = path.join(self.outputDirectory, "datasets")
+        self.readFiles = map(path.basename,
+                             (glob(path.join(self.inputDirectory, "*.sam"))))
+        self.variantFiles = map(path.basename,
+                                (glob(path.join(self.inputDirectory,
```
jeromekelleher: I dont' think this first bracket is necessary. I think these lines would be easier to manage if you used hanging indents rather than visual indent, i.e.
```python
 self.variantFiles = map(
        os.path.basename,  glob.glob(os.path.join(self.inputDirectory, "*.vcf"))
```
I tend to prefer hanging indents anyway because they're a lot more robust to code changes (for example, if we changed to self._variantFiles, we'd have to reindent all following lines).

scripts/prepare_compliance_data.py
```diff
@@ -0,0 +1,155 @@
...
+        for ds in self.datasets:
+            dsdir = path.join(self.datasetsDirectory, ds)
+            os.makedirs(dsdir)
+
+            # Reads
+            print("Converting reads...", file=sys.stderr)
+            dsReadsdir = path.join(dsdir, "reads")
+            os.makedirs(dsReadsdir)
+            for readFile in self.datasetReads[ds]:
+                destFile = path.join(dsReadsdir,
```
jeromekelleher: Definitely hanging indent here...

jeromekelleher:
Looks great, thanks @macieksmuga! I have some minor style comments, and I agree with @dcolligan's points. I don't really care if we print to stderr or stdout for this script to be honest, I don't think it's all that important for a utility script like this.

scripts/prepare_compliance_data.py
macieksmuga: @dcolligan : It's not downloaded from there because one of its principal uses is by GA4GH developers to develop NEW compliance tests with new data: Thus, as they work on a dataset/tests, they can point the script at their local working directory version.

scripts/prepare_compliance_data.py
macieksmuga: I understood that `from os import path` and `from glob import glob` were pretty idiomatic. `glob.glob` doesn't really tell you anything you didn't already know, after all. Note I constrained it to these two well-known, supported cases, and used the default `import x` in all others. If you feel strongly about it, I'll change it, though, although I suspect it will hurt, not help readability.

scripts/prepare_compliance_data.py
macieksmuga: Sure, ok. When in Rome...

scripts/prepare_compliance_data.py
macieksmuga: Reason is, because the underlying pysam library is already spewing a cryptic message out to stderr. So I needed to put in some obvious "user friendly" messages around it, to confirm that the program is indeed working as suggested, at @hjellinek's (very good) suggestion.

scripts/prepare_compliance_data.py
macieksmuga: good catch, thanks!

macieksmuga:
I hope I addressed all outstanding issues identified by @dcolligan and @jeromekelleher - thank you both for going through this!

dcolligan:
+1

jeromekelleher:
+1. Thanks @macieksmuga, this is very useful.

PR #658, opened 2015-09-28
Searching by non-existent callset name...

dcolligan:
... now returns no results instead of a 404

Issue #647

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -515,9 +515,10 @@ class AbstractBackend(object):
                 request, variantSet.getNumCallSets(),
                 variantSet.getCallSetByIndex)
         else:
-            # Since names are unique within a callSet, we either have one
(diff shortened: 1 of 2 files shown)


jeromekelleher:
+1

macieksmuga:
How does this square with compliance tests @hjellinek? Do we need to modify them to suit this?

hjellinek:
The tests are all right.

The test `ReadGroupSetsGetByNameIT.testSearchForBogusNameReturnsEmptyList` actually does what its name implies: the `SearchReadGroupSetsResponse` for a nonexistent name is expected to contain an empty `readGroupSets` list.

In other words, fixing this bug will increase the Ref Server's compliance score.

hjellinek:
p.s. I had already updated the docs in the `documentation` branch of ga4gh/schemas to describe this behavior.

PR #672, opened 2015-10-06
Changed the input data format for consistency.

jeromekelleher:
After starting to document the input data format, I realised that it was very confusing and inconsistent. This PR makes datasets and referencesets symmetrical and treated in the same way.

The input data directory must now be in the following format:

```
dataDir/
    datasets/
        dataset1
            /variants
            /reads
        dataset2
            #etc
    referenceSets
        referenceSet1/
        referenceSet2/
```
Datasets and ReferenceSets are now treated symmetrically (other than their annoying spelling difference...).

This is a breaking change, and will affect all users. It is better to get this change out of the way before we start the release process though.

PR #673, opened 2015-10-06
Remaining docs update

jeromekelleher:
Update the docs for the data input format. Assumes #672 is accepted.

Needs some rebasing once other PRs are merged.

### DIFF
--- a/docs/demo.rst
+++ b/docs/demo.rst
@@ -59,8 +59,8 @@ Now we can download some example data, which we'll use for our demo:
 
 .. code-block:: bash
 
-    (ga4gh-env) $ wget http://www.well.ox.ac.uk/~jk/ga4gh-example-data-v2.1.tar
-    (ga4gh-env) $ tar -xvf ga4gh-example-data-v2.1.tar
+    (ga4gh-env) $ wget http://www.well.ox.ac.uk/~jk/ga4gh-example-data-v3.0.tar
+    (ga4gh-env) $ tar -xvf ga4gh-example-data-v3.0.tar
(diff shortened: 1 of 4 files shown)


jeromekelleher:
Any thoughts @dcolligan?

docs/demo.rst
```diff
@@ -59,8 +59,8 @@ Now we can download some example data, which we'll use for our demo:
 
 .. code-block:: bash
 
-    (ga4gh-env) $ wget http://www.well.ox.ac.uk/~jk/ga4gh-example-data-v2.1.tar
-    (ga4gh-env) $ tar -xvf ga4gh-example-data-v2.1.tar
+    (ga4gh-env) $ wget http://www.well.ox.ac.uk/~jk/ga4gh-example-data-v3.0.tar
```
dcolligan: sphinx probably has some notion of variables, right, so we don't need to change this in two places going forward?

jeromekelleher: Excellent idea @dcolligan. It turns out we can add some rst substitutions easily enough using the ``rst_prolog`` in conf.py. These work perfectly well in the main document. However, they don't work in code blocks, which is the only place we actually want them. We can work around this by using parsed-literal blocks:

http://stackoverflow.com/questions/27392934/sphinx-variable-substitution-in-code-blocks

However, these look totally different to the rest of the code-blocks and the result is very jarring visually. It would be great to solve this problem properly at some point, but I think we'll have to leave it as it is for now.

dcolligan:
+1

PR #699, opened 2015-10-09
Updated ga2sam and improved tests.

jeromekelleher:
This change allows us to directly test the sam->ga->sam conversion process by comparing the converted BAM file to the original BAM file.

This is only a start, and it needs lots of work. However, the framework should be clear and it will provide a good source of work for hackathon issues.


dcolligan:
+1

jeromekelleher:
@dcolligan, merge me!

PR #710, opened 2015-10-11
Fix doc errors

nouyang-curoverse:
* `requirements.txt` only available after checking out develop branch
* need `libdb-dev` or else wormtable complains (at least with `python setup.py install --user` -- forgot to uninstall `libdb-dev` before testing this version of the instructions where we checkout `develop` branch first)

     _wormtablemodule.c:22:16: fatal error: db.h: No such file or directory
     #include <db.h>
                     ^
    compilation terminated.

Let me know if I should go through the full fork-PR process on the docs -- just did this in-browser quickly, before running off to bus to nyc


jeromekelleher:
There must be something wrong here @nouyang-curoverse --- we got rid of the wormtable dependency ages ago, so we shouldn't need libdb at all.

I'm sure we'll sort it out tomorrow.

dcolligan:
Closing this... this is apparently what happens when one tries to run on the master branch from a long time ago!

Issue #701, opened 2015-10-09
Find unused exception classes and remove


jeromekelleher:
The ga4gh/exceptions.py module is the central repository for exceptions, which makes importing them and keeping track of the inheritance hierarchy easy. However, it can be hard to keep track of which ones are being used. 

Find the exceptions which are currently not being used in the code base and delete them. 
For example,

```bash
$ grep -IR BadReadsSearchRequestBothRefs ga4gh
ga4gh/exceptions.py:class BadReadsSearchRequestBothRefs(BadRequestException):
```

Since the only match we have is the exceptions.py file, this exception isn't being used anywhere and can be removed.

Try using the [vulture](https://pypi.python.org/pypi/vulture) package to find other unused exceptions. The output of vulture contains lots of false positives, so beware. However, it does find some other unused exceptions.

Can we build an automated way we can include in our Travis setup to detect unused exceptions using vulture?

denis-yuen:
python neophyte here, will take a look at this

dcolligan:
Fixed in #729

PR #729, opened 2015-10-12
Feature/feature 701 unused exceptions

denis-yuen:
This removes the two unused exceptions and adds a ~/.travis.yml step to automatically check for them in the future

See https://travis-ci.org/denis-yuen/server/builds/84943690 for proof that the tests pass

### DIFF
--- a/ga4gh/exceptions.py
+++ b/ga4gh/exceptions.py
@@ -201,11 +201,6 @@ class DatasetNotFoundException(NotFoundException):
             datasetId)
 
 
-class ReadGroupSetNotFoundException(ObjectNotFoundException):
-    def __init__(self, readGroupSetId):
-        self.message = "readGroupSetId '{}' not found".format(readGroupSetId)
-
-
 class ReadGroupNotFoundException(ObjectNotFoundException):
     def __init__(self, readGroupId):
         self.message = "readGroupId '{}' not found".format(readGroupId)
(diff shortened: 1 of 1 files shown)


dcolligan:
A number of things:

1) You don't need to post proof that the CI build passes... the Travis CI build will kick off when you submit a PR ... this should give the same result as `python scripts/run_tests.py`

2) For us to accept a PR, we usually like to have it in one commit, so these commits need to be squashed into one

3) When I run `vulture .` on the project I get this error: `RuntimeError: maximum recursion depth exceeded`

4) Anything that is in `.travis.yml` gets run every time on travis and also by `run_tests.py` so we only want to put in there what we want to run on every test run... I'm not sure if this qualifies as is.  Vulture seems like a useful tool (if it works) to run once in a while.

5) I would be up for accepting just the changes in `ga4gh/exceptions.py`.  If you resubmit the branch with those changes alone, I think that was more what the original issue was asking for.  (er, now that I look at the issue, it seems like the issue *was* asking for incorporating vulture into CI, but let's not do that for now)

6) We usually like the issue number in the commit message: `Issue #701` for this issue, for example.

jeromekelleher:
Looks like vulture might be a bit tricky to incorporate without more evaluation. I think we should trim this PR back to just the changes to exceptions.py, just in case we introduce some problems for people. Thanks for the check though, hopefully we can integrate with something like this later!

dcolligan:
+1

jeromekelleher:
+1.

PR #732, opened 2015-10-12
Output FASTA for reference-bases-list command

marghoob:
Make ListReferenceBasesRunner be able to output in FASTA or text depending on the option

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -447,14 +447,21 @@ class ListReferenceBasesRunner(AbstractQueryRunner):
         self._referenceId = args.id
         self._start = args.start
         self._end = args.end
+        self._outputFormat = args.outputFormat
 
     def run(self):
-        iterator = self._client.listReferenceBases(
+        sequence = self._client.listReferenceBases(
             self._referenceId, self._start, self._end)
-        # TODO add support for FASTA output.
-        for segment in iterator:
-            print(segment, end="")
-        print()
+        if self._outputFormat == "text":
+            print(sequence)
+        else:
+            start = self._start if self._start else ""
+            end = self._end if self._end else ""
+            print(">{}:{}-{}".format(self._referenceId, start, end))
+
+            textWidth = 70
+            for index in xrange(0, len(sequence), textWidth):
+                print(sequence[index: index+textWidth])
 
 
 # Runners for the various GET methods.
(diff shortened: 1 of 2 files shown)


dcolligan:
looks good, but in the code base we prefer the `.format` function to `%` for string formatting

see here for an example: http://stackoverflow.com/questions/5082452/python-string-formatting-vs-format

marghoob:
Thanks. Let me fix to use format.

ga4gh/cli.py
```diff
@@ -414,14 +415,19 @@ def __init__(self, args):
...
+        sequence = self._client.listReferenceBases(
             self._referenceId, self._start, self._end)
-        # TODO add support for FASTA output.
-        for segment in iterator:
-            print(segment, end="")
-        print()
+        if self._outputFormat == "text":
+            print(sequence)
+        else:
+            fastaId = "{}".format(self._referenceId) + \
```
jeromekelleher: We try to avoid splitting lines and string concatenation where we can. How about we try something like
```python
start = start if self._start is not None else ""
end = end if self._end is not None else ""
fastaId = "{}:{}-{}".format(self._referenceId, start, end)
```

marghoob: Sold and put in the changes.

dcolligan:
when you want to commit this, squash the commits

also, shouldn't there be a test for this as well? @jeromekelleher

ga4gh/cli.py
```diff
@@ -414,14 +415,18 @@ def __init__(self, args):
...
-        for segment in iterator:
-            print(segment, end="")
-        print()
+        if self._outputFormat == "text":
+            print(sequence)
+        else:
+            start = self._start if self._start else ""
+            end = self._end if self._end else ""
+            print(">{}:{}-{}".format(self._referenceId, start, end))
+            print(textwrap.fill(">{}".format(sequence)), sep="")
```
jeromekelleher: Unfortunately, it turns out this is _very_ slow, so we have to do the splitting by hand. Something like
```
          width = 70
            chunks = len(sequence) // width
            for j in range(chunks):
                print(sequence[j * width : (j + 1) * width])
            if len(sequence) % width != 0:
                print(sequence[chunks * width:])
```
should work.

Of course, we really should add some tests for this. Can we add an issue stating that we need some tests for the FASTA sequence output?

ga4gh/cli.py
```diff
@@ -414,14 +414,22 @@ def __init__(self, args):
...
-        print()
+        if self._outputFormat == "text":
+            print(sequence)
+        else:
+            start = self._start if self._start else ""
+            end = self._end if self._end else ""
+            print(">{}:{}-{}".format(self._referenceId, start, end))
+
+            textWidth = 70
+            print(">{}".format(sequence[:textWidth-1]))
```
jeromekelleher: Is this correct? I thought FASTAs just had the first line (with the ID) prefixed by '>'?

marghoob: Nope, this is definitely incorrect.

jeromekelleher: I'm pretty sure only the header line has a > prefix... https://en.wikipedia.org/wiki/FASTA_format

jeromekelleher:
Looks great, thanks @marghoob. I hope you can find the time to send us some more PRs!

+1

marghoob:
Oh yeah. It was great meeting you guys there. I am definitely interested in
contributing more to the project.

PR #735, opened 2015-10-12
Datasets now have metadata. Fixes #679

### LINKED ISSUE #679: Dataset.description not filled in
Running
```
python client_dev.py datasets-search http://localhost:8000 -O json | jq .
```
currently gives
```json
{
  "description": null,
  "name": "1kg-p3-subset",
  "id": "MWtnLXAzLXN1YnNldA=="
}
```
The description field on Dataset is currently not filled in as there is no way to provide it. This should be provided by an optional json file that lives next to the dataSet directory, following the way this is done for ReferenceSet. We also need to add a description for the SimulatedDataSet.

- Open ga4gh/datamodel/datasets.py and add a new instance variable ``self._description = None`` in the constructor for AbstractDataset. Add a suitably documented getDescription method to this class (follow patterns elsewhere for getter functions).
- Set ``self._description`` to a suitable value in the constructor for SimulatedDataset.
- Open tests/unit/test_simulated_stack.py and update ``verifyDatasetsEqual`` to include
```
self.assertEqual(gaDataset.description, dataset.getDescription()) 
```
Run the test using 
```
nosetests -vs tests/unit/test_simulated_stack.py:TestSimulatedStack.testGetDataset
```
This should fail. 
- Fix the failing test by updating ``toProtocolElement`` in AbstractDataset.

- Once we know the basic plumbing works, update FileSystemDataset to read in a JSON file called ``$localId.json$`` following the pattern in ``ga4gh.datamodel.references.HtslibReferenceSet``. This file should be *optional* for now, so that description is still null if the JSON file isn't provided. Only one value is expected in the JSON file "description". If this is not present, raise an ``exceptions.MissingDatasetMetadataException``. Define this exception in ga4gh/exceptions.lpy following the pattern for ``exceptions.MissingReferenceSetMetadata`` .

- Try this out by adding a file ``ga4gh-example-data/datasets/1kg-p3-subset.json``, and restarting your server.

andrewyatz:
Includes an additional test not specified in issue #679 to test bad metadata


dcolligan:
this looks great!  I'm +1 if the tests pass

@jeromekelleher , can you review this as well?

jeromekelleher:
Yes, this is awesome!

+1.

dcolligan:
quite deserving of the finest British pastries, if I do say so myself!

Issue #734, opened 2015-10-12
process_schemas.py should fail more obviously when pointed to the wrong directory


dcolligan:
Running `process_schemas.py` pointing to the wrong directory e.g.
```
python scripts/process_schemas.py  -i ../schemas/ --avro-tools-jar ~/Downloads/avro-tools-1.7.7.jar 0.6.4568e6f6
```
produces a `_protocol_definitions.py` that is unusable, but should probably throw an error.

dcolligan:
Fixed in #736

PR #736, opened 2015-10-12
Make process_schemas.py fail more obviously...

dcolligan:
... when the input directory has no avdl files

Issue #734


jeromekelleher:
+1

Issue #702, opened 2015-10-09
Better default behaviour for CLI reads-search command


jeromekelleher:
The client command line interface provides an easy to use interface to the server functions. For example, we can use 
```
$ python client_dev.py datasets-search http://localhost:8000 -O json
{"description": null, "name": "1kg-p3-subset", "id": "MWtnLXAzLXN1YnNldA=="}
```
to run the datasets/search command. To make this as user-friendly as possible we try to do the obvious thing when arguments are not provided. For example, when we run the variants-search without arguments like
```
$ python client_dev.py variants-search http://localhost:8000
```
we find all ``VariantSets`` on the server, and run the variants-search command on each one. This makes quick exploration of the data on the server much faster, as we don't need to manually copy IDs to get a quick look at the data.

However, the ``reads-search`` command does not have this useful feature.

Edit ``ga4gh/cli.py`` and find SearchReadsRunner. 

- Defined a method ``getAllReadGroups`` that returns all ReadGroups on the server using ``getAllRedGroups``.
- Then, in SearchReadsRunner.run(), If ``readGroupIds`` has not been specified, iterate over all readGroups and search for all reads in each readGroup. See SearchVariantsRunner for an example of how this can be done.
- While processing a given readGroup, if the referenceId has been specified, use this reference. If not, find the ReferenceSet that the readGroup refers to, and iterate over all references in this referenceSet.

The [client API docs](http://ga4gh-reference-implementation.readthedocs.org/en/develop/api.html) will be useful for this.

pkrusche:
For the last part, the ReadGroup objects only seem to have a referenceSetId, what we'd need is a referenceId. I am guessing this referenceId will come from each read, so this would need to be done in the backend?

dcolligan:
Fixed in #738

PR #738, opened 2015-10-12
Added support for finding all read groups when no read group specified

pkrusche:
Fixes issue #702

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -192,6 +192,18 @@ class AbstractSearchRunner(FormattedOutputRunner):
             for readGroupSet in iterator:
                 yield readGroupSet
 
+    def getAllReadGroups(self):
+        """
+        Get all read groups in a read group set
+        """
+        for dataset in self.getAllDatasets():
+            iterator = self._client.searchReadGroupSets(
+                datasetId=dataset.id)
+            for readGroupSet in iterator:
+                readGroupSet = self._client.getReadGroupSet(readGroupSet.id)
+                for readGroup in readGroupSet.readGroups:
+                    yield readGroup.id
+
     def getAllReferenceSets(self):
         """
         Returns all reference sets on the server.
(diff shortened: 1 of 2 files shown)


ga4gh/cli.py
```diff
@@ -192,6 +192,18 @@ def getAllReadGroupSets(self):
...
 
+    def getAllReadGroups(self):
+        """
+        Get all read groups in a read group set
+        """
+        for dataset in self.getAllDatasets():
+            iterator = self._client.searchReadGroupSets(
+                datasetId=dataset.id)
+            for readGroupSet in iterator:
+                rgs = self._client.getReadGroupSet(readGroupSet.id)
```
dcolligan: as a stylistic convention, we try and avoid abbreviations like this... use `readGroupSet` as a variable name instead

ga4gh/cli.py
```diff
@@ -192,6 +192,18 @@ def getAllReadGroupSets(self):
...
+                for rg in rgs.readGroups:
```
dcolligan: same style comment: `rg` -> `readGroup`

ga4gh/cli.py
```diff
@@ -401,10 +434,10 @@ def _textOutput(self, gaObjects):
             # TODO add in some more useful output here.
             print(read.id)
 
-
 # ListReferenceBases is an oddball, and doesn't fit either get or
 # search patterns.
 
+
```
dcolligan: the commit diffs should be as small as possible... try to avoid changing lines unnecessarily

ga4gh/cli.py
```diff
@@ -385,13 +397,34 @@ def __init__(self, args):
         if args.readGroupIds is not None:
             self._readGroupIds = args.readGroupIds.split(",")
 
+    def _run(self, rgid, refid=None):
+        """
+        automatically guess reference id if not passed
+        """
+        # check if we can get reference id from rg
+        if not refid:
```
dcolligan: checks for equivalence to None should read `if x is None`

ga4gh/cli.py
```diff
@@ -385,13 +397,34 @@ def __init__(self, args):
...
+            refid = self._referenceId
+        if not refid:
```
dcolligan: same comment here

tests/end_to_end/test_client_json.py
```diff
@@ -73,3 +73,7 @@ def testGetDataset(self):
         for dataset in self._client.searchDatasets():
             self.verifyParsedOutputsEqual(
                 [dataset], "datasets-get", dataset.id)
+
+    def testSearchAllReadGroups(self):
+        cliOutput = self.captureJsonOutput("reads-search")
+        self.assertGreater(len(cliOutput), 0)
```
dcolligan: there's probably a more rigorous test that we can write here... if the output is not what we expect, this test still passes

ga4gh/cli.py
```diff
@@ -192,6 +192,18 @@ def getAllReadGroupSets(self):
...
+        """
+        Get all read groups in a read group set
+        """
+        for dataset in self.getAllDatasets():
+            iterator = self._client.searchReadGroupSets(
+                datasetId=dataset.id)
+            for readGroupSet in iterator:
+                rgs = self._client.getReadGroupSet(readGroupSet.id)
+                for rg in rgs.readGroups:
+                    yield rg.id
```
jeromekelleher: We should yield the readgroup object here rather than the ID to keep symmetry with the other methods.

dcolligan:
good start but still needs some work

jeromekelleher:
Looks great @pkrusche, +1 modulo the stylistic nits above.

ga4gh/cli.py
pkrusche: I inserted this to make flake8 pass.

dcolligan: hm, interesting... I wonder why this wasn't failing before, then...

make sure you're using the same version of flake that is specified in `requirements.txt`, if you aren't already...

pkrusche: I installed via pip install -r requirements.txt, and used a virtualenv, so the version should be the same.

```
(server-ve)$:~/workspace/server$ flake8 --version
2.3.0 (pep8: 1.6.2, mccabe: 0.3.1, pyflakes: 1.0.0) CPython 2.7.9 on Linux
(server-ve)$:~/workspace/server$ cat requirements.txt | grep flake8
flake8==2.3.0  # Set due to conflict with pep8 in version 2.4.0
```

dcolligan:
cool.  +1

Issue #690, opened 2015-10-08
Unit tests for Variants


jeromekelleher:
We need unit tests for variants, where we test the basic properties of objects defined in ga4gh/datamodel/variants.py outside of the context of input data. The tests created here should be similar to the new tests outlined in #683, #688, #687.

Open ``tests/unit/test_variants.py``, and add the following class:

```python
class TestAbstractVariantSet(unittest.TestCase):
    """
    Unit tests for the abstract variant set.
    """
    def setUp(self):
        self._backend = backend.AbstractBackend()
        self._dataset = datasets.AbstractDataset(self._backend)
        self._variantSet = variants.AbstractVariantSet(
            self._dataset, "testVariantSet")

    def testAddOneCallSet(self):
        self.assertEqual(self._variantSet.getNumCallSets(), 0)
        callSetName = "callSetName"
        self._variantSet.addCallSet(callSetName)
        self.assertEqual(self._variantSet.getNumCallSets(), 1)
        callSet = self._variantSet.getCallSetByIndex(0)
        self.assertEqual(
            self._variantSet.getCallSetByName(callSetName), callSet)
        self.assertEqual(                                         
            self._variantSet.getCallSet(callSet.getId()), callSet)   
        self.assertEqual(self._variantSet.getCallSets(), [callSet])
```

Run the tests by running ``nosetests -vs tests/unit/test_variants.py:TestAbstractVariantSet``

- Add another test ``testAddMultipleCallSets`` which is similar.

- Add tests to check that calling methods like getCallSet and getCallSetByName raise the correct exceptions (see an example in #683).

- Add a test for the hashVariant method.

jsh2134:
Working on this!

dcolligan:
Fixed in #739

PR #739, opened 2015-10-12
Tests for AbstractVariantSet - Issue #690

jsh2134:
Applies to Issue #690 

Generates some a SimulatedVariantSet for use in testing methods that require a variant see https://github.com/jsh2134/server/commit/341ad1c19ee490c608174c510ce2fba7f4ec8136#diff-7858b4c3a3f24acfcfa6d89486e1544cR167 and https://github.com/jsh2134/server/commit/341ad1c19ee490c608174c510ce2fba7f4ec8136#diff-7858b4c3a3f24acfcfa6d89486e1544cR175

### DIFF
--- a/tests/unit/test_variants.py
+++ b/tests/unit/test_variants.py
@@ -8,7 +8,10 @@ from __future__ import unicode_literals
 
 import unittest
 
+import ga4gh.exceptions as exceptions
+import ga4gh.backend as backend
 import ga4gh.datamodel.variants as variants
+import ga4gh.datamodel.datasets as datasets
 
 
 class TestGenotypes(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


tests/unit/test_variants.py
```diff
@@ -8,7 +8,10 @@
 
 import unittest
 
+from ga4gh import exceptions
```
dcolligan: this import line should like like the following import lines

tests/unit/test_variants.py
```diff
@@ -70,3 +73,115 @@ def testPhasesetZero(self):
...
+        self._variantSet = variants.AbstractVariantSet(
+            self._dataset, self._vsname)
+
+        # create some simulated variants with dummy data
+        self._simulatedVariantSet = variants.SimulatedVariantSet(
+            self._dataset, self._vsname, randomSeed=0,
+            numCalls=1, variantDensity=1)
+        self._simulatedVariants = self._simulatedVariantSet.getVariants(
+            'ref', 100, 110, ["nope"])
+        self._real_variant = list(self._simulatedVariants)[0]
```
dcolligan: use camelCase, not snake_case

tests/unit/test_variants.py
```diff
@@ -70,3 +73,115 @@ def testPhasesetZero(self):
...
+        self.assertEqual(
+            self._variantSet.getCallSetByName(callSetName), callSet)
+        self.assertEqual(
+            self._variantSet.getCallSet(callSet.getId()), callSet)
+        self.assertEqual(self._variantSet.getCallSets(), [callSet])
+
+    def testAddMultipleCallSets(self):
+        # disable maxdiff value in assertEqual because list
+        # of 10 variants was too large
+        self.assertEqual.__self__.maxDiff = None
```
dcolligan: does this affect the state of `self.assertEqual` outside of this particular test?  if so, this entire test should be surrounded with a try/finally block and the original state restored at the end of the test

tests/unit/test_variants.py
```diff
@@ -70,3 +73,115 @@ def testPhasesetZero(self):
...
+        self.assertEqual(self._variantSet.getNumCallSets(), 0)
+        callSetName = "callSetName{}"
+        callSetCount = 10
+        allCallSets = []
+        for x in range(0, callSetCount):
```
dcolligan: use `i` as the counting variable

use `range(callSetCount)` -- the 0 argument is implied

jsh2134:
thanks @dcolligan - updated

tests/unit/test_variants.py
```diff
@@ -70,3 +73,112 @@ def testPhasesetZero(self):
...
     def testGenotypeHaploid(self):
         self.verifyGenotypeConversion("1", "376", [1], None)
+
+
+class TestAbstractVariantSet(unittest.TestCase):
+    """
+    Unit tests for the abstract variant set.
+    """
+    def setUp(self):
+        self._vsname = "testVariantSet"
```
jeromekelleher: We tend to avoid abbreviations (although it can be tedious, I agree). Prefer ``self._variantSetName`` here.

tests/unit/test_variants.py
```diff
@@ -70,3 +73,112 @@ def testPhasesetZero(self):
...
+        self._backend = backend.AbstractBackend()
+        self._dataset = datasets.AbstractDataset(self._backend)
+        self._variantSet = variants.AbstractVariantSet(
+            self._dataset, self._vsname)
+
```
jeromekelleher: Why do we want a simulated variant set here too?

tests/unit/test_variants.py
```diff
@@ -70,3 +73,112 @@ def testPhasesetZero(self):
...
+                self._variantSet.getCallSet(callSet.getId()), callSet)
+            allCallSets.append(callSet)
+
+        self.assertEqual(self._variantSet.getNumCallSets(), callSetCount)
+        self.assertEqual(self._variantSet.getCallSets(), allCallSets)
+
+    def testCallSetMethods(self):
+        # test invalid lookups
+        self.assertRaises(IndexError, self._variantSet.getCallSetByIndex, 10)
+
```
jeromekelleher: We don't usually have this much whitespace, I think we can trim out the gaps between the lines here.

tests/unit/test_variants.py
```diff
@@ -70,3 +73,112 @@ def testPhasesetZero(self):
...
+        self.assertRaises(exceptions.CallSetNotFoundException,
+                          self._variantSet.getCallSet, None)
+
+        self.assertRaises(TypeError, self._variantSet.addCallSet,
+                          ['a list of', 2])
+
+        self.assertRaises(NotImplementedError,
+                          self._variantSet.getNumVariants)
+
+    def testGetVariantId(self):
```
jeromekelleher: I think the next two tests are aimed at the HtslibVariantSet class, and are tested elsewhere. Here, we really just want to focus on AbstractVariantSet, so we can remove them.

tests/unit/test_variants.py
```diff
@@ -70,3 +73,112 @@ def testPhasesetZero(self):
...
+    def testHashVariant(self):
+        self.assertRaises(AttributeError,
+                          self._variantSet.hashVariant, None)
+        self.assertRaises(AttributeError,
+                          self._variantSet.hashVariant, "hi")
+
+        vhash = self._variantSet.hashVariant(self._realVariant)
+        self.assertEquals(type(vhash), str)
+
+    def testVariantSetProtocolElement(self):
```
jeromekelleher: This is tested extensively elsewhere, so I think we can remove it.

jeromekelleher:
Nice tests @jsh2134, I like it. I think we're spreading out a bit beyond what we need here which is complicating things a bit. If we remove the tests indicated (and simplify the setUp method) I think we're good to go.

tests/unit/test_variants.py
jsh2134: I needed an actual variant to pass through to some methods (see PR description) because `hashVariant(v)` and `toProtocolElement()` just die because certain values are `None` - i suppose if we only want to test the methods as is then validating method failure is a fine enough test. let me know and ill remove these parts (or move to simulatorTest) 
https://github.com/jsh2134/server/commit/341ad1c19ee490c608174c510ce2fba7f4ec8136#diff-7858b4c3a3f24acfcfa6d89486e1544cR161 and https://github.com/jsh2134/server/commit/341ad1c19ee490c608174c510ce2fba7f4ec8136#diff-7858b4c3a3f24acfcfa6d89486e1544cR170

jeromekelleher: Ah, I see. These are good tests, but the features are heavily covered elsewhere. I think we can remove them, see the comments below.

jsh2134:
squashed :)

jeromekelleher:
thanks @jsh2134, +1.

PR #740, opened 2015-10-12
688

### LINKED ISSUE #688: Unit tests for references
We need unit tests for references, where we test the basic properties of objects defined in ``ga4gh/datamodel/references.py`` outside of the context of input data. The tests created here should be similar to the new tests outlined in #683.

- Create a new file ``tests/unit/test_references.py`` modeled after ``tests/unit/test_variants.py``, making sure to update the comments appropriately.

- Start with this test: 

```python
class TestAbstractReferenceSet(unittest.TestCase):
    """
    Unit tests for the abstract reference set.
    """
    def setUp(self):
        self._backend = backend.AbstractBackend()
        self._referenceSet = references.AbstractReferenceSet(
            self._backend)

    def testAddOneReference(self):
        self.assertEqual(self._referenceSet.getNumReferences(), 0)
        referenceName = "ref"
        reference = references.AbstractReference(
            self._referenceSet, referenceName)
        self._referenceSet.addReference(reference)
        self.assertEqual(self._referenceSet.getNumReferences(), 1)
        self.assertEqual(
            self._referenceSet.getReferenceByIndex(0), reference)
        self.assertEqual(
            self._referenceSet.getReferenceByName(referenceName), reference)
        self.assertEqual(
            self._referenceSet.getReference(reference.getId()), reference)
        self.assertEqual(self._referenceSet.getReferences(), [reference])
```

Run the tests with ``nosetests -vs tests/unit/test_references.py``

- Add a test case ``testMultipleReferences`` which is similar.

- Add test cases to check that the appropriate exceptions are raised when we try to find references with IDs and names that don't exist (see #683 for an example).

- Write a test to check that the md5sum is correctly computed for a set of references. (We will probably want a local ``DummyReference(AbstractReference)`` which we can easily create instances of and assign MD5 values. See the [schemas
(https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/references.avdl) for the definition of how the MD5 should be calculated.

- Write some test cases for Reference objects. One thing we should test is ``Reference.checkQueryRange`` to ensure that it throws the correct exception.

jercoh:
Fix #688

### DIFF
--- a/tests/unit/test_references.py
+++ b/tests/unit/test_references.py
@@ -1,0 +1,91 @@
+"""
+Unit tests for reference objects. This is used for all tests
+that can be performed in isolation from input data.
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import unittest
+
+import ga4gh.backend as backend
+import ga4gh.datamodel.references as references
+import ga4gh.exceptions as exceptions
+
+
+class TestAbstractReferenceSet(unittest.TestCase):
+    """
+    Unit tests for the abstract reference set.
+    """
+    def setUp(self):
+        self._backend = backend.AbstractBackend()
+        self._referenceSet = references.AbstractReferenceSet(
+            self._backend)
+
+    def testAddOneReference(self):
+        self.assertEqual(self._referenceSet.getNumReferences(), 0)
+        referenceName = "ref"
+        reference = references.AbstractReference(
+            self._referenceSet, referenceName)
+        self._referenceSet.addReference(reference)
+        self.assertEqual(self._referenceSet.getNumReferences(), 1)
+        self.assertEqual(
+            self._referenceSet.getReferenceByIndex(0), reference)
+        self.assertEqual(
+            self._referenceSet.getReferenceByName(referenceName), reference)
+        self.assertEqual(
+            self._referenceSet.getReference(reference.getId()), reference)
+        self.assertEqual(self._referenceSet.getReferences(), [reference])
+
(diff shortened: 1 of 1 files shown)


dcolligan:
looks pretty good so far... the original issue calls for a md5 test, which I don't see here

tests/unit/test_references.py
```diff
@@ -0,0 +1,91 @@
...
+        self.assertEqual(
+            self._referenceSet.getReferenceByName(referenceName), reference)
+        self.assertEqual(
+            self._referenceSet.getReference(reference.getId()), reference)
+        self.assertEqual(self._referenceSet.getReferences(), [reference])
+
+    def testAddMultipleReference(self):
+        referenceList = []
+        referenceBaseName = "ref"
+        callSetCount = 10
```
jeromekelleher: Should probably be ``referenceCount``?

jeromekelleher:
Yep, looks good, thanks @jercoh! We should add some tests for the MD5's all right though.

jercoh:
What about `testMd5checksums` in `tests/datadriven/test_references.py`?
https://github.com/ga4gh/server/blob/develop/tests/datadriven/test_references.py#L161-L173

jeromekelleher:
Yeah, I guess this is covered already. We should fix the variable naming issue above though. (+ squash the commits)

jeromekelleher:
Looks great, thanks @jercoh --- I hope you can send us more PRs in the future!

+1

dcolligan:
+1

PR #745, opened 2015-10-12
Set a name for ReferenceSet. Fixes #676.

### LINKED ISSUE #676: ReferenceSet name not set
If we run 
```
$ python client_dev.py referencesets-search http://localhost:8000
R1JDaDM4LXN1YnNldA==    None
```
This output shows the ID and name of an object. We expect the name of a reference set to be non-null, so this is a bug.

Edit ga4gh/datamodel/references.py. Find the AbstractReferenceSet.toProtocolElement method and insert the line
```
ret.name = self.getLocalId()
```
After saving the file and reloading the server, we should now see

```
$ python client_dev.py referencesets-search http://localhost:8000
R1JDaDM4LXN1YnNldA==    GRCh38-subset
```

Bug fixed!

To make sure the bug stays fixed, we must add some test cases.

Open ``tests/unit/test_simulated_stack.py`` and update the ``verifyReferenceSetsEqual method``. Verify the tests work by running ``nosetests -vs tests/unit/test_simulated_stack.py``. (To make double sure that the tests  work correctly, comment out the line you added above, and check that the test now fails.)

andrewyatz:


dcolligan:
+1

jeromekelleher:
+1. Merging.

PR #747, opened 2015-10-12
Provide a useful error if server appears to already be running during the server end to end test

adamnovak:
I was poking around for a couple minutes and couldn't find the offending port, or an open terminal. This makes it easier.


jeromekelleher:
+1.

dcolligan:
+1

Issue #717, opened 2015-10-11
Improve Simulated reads.


jeromekelleher:
Simulated data is very important for testing, as discussed in #715. The simulator for reads is currently very crude. Some improvements we can make

- Add some simulated programs. See the [schemas](https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/reads.avdl) for definitions of what they are. These don't need to be particularly realistic, we just want something to put in there to exercise the translation machinery.

- Refactor the AbstractReadGroupSet, SimulatedReadGroupSet, and HtslibReadGroupSet classes so that there is one definition of ``getPrograms`` in AbstractReadGroupSet.

- Test these by updating the ``verifyReadGroupSetsEqual`` in tests/unit/test_simulated_stack.py. Run the tests using ``nosetests -vs tests/unit/test_simulated_stack.py``.

- Add a new method ``verifyReadsEqual`` in ``tests/unit/test_simulated_stack.py`` and call this from testReads. Add as many checks as can be reasonably done in the available time.

- Ideally, we'd like to update the simulated reads to be much more sophisticated. However, this is quite an open ended task. Any improvements would be great though!

marghoob:
I am interested in this one.

adamnovak:
I am sort of working on this.

PR #749, opened 2015-10-12
Use URLs to specify data sources

adamnovak:
Instead of `__EMPTY__` we use `empty://`. Instead of `__SIMULATED__` we use things like `simulated://?numCalls=5&numReferencesPerReferenceSet=1` (and pull the un-specified values from the config). Currently-used file paths get automatically turned into `file://` URLs, and `file://` URLs can be accepted.

Unless your config requested `__EMPTY__` or `__SIMULATED__`, this is backwards-compatible. It also potentially will let us do stuff like `mysql://` in the future if we get more different backends. Maybe one for `graphdatabase://`?

I sort of thought this might help with #717, keeping with the theme of improving the simulated backend.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,6 +26,7 @@ Some general rules to follow:
 * Create a branch for each update that you're working on. These branches 
   are often called "feature" or "topic" branches. Any changes that you 
   push to your feature branch will automatically be shown in the pull request.
+* Test your code (by running `nosetests`) before creating a pull request.
 * Keep your pull requests as small as possible. Large pull requests are 
   hard to review. Try to break up your changes into self-contained and 
   incremental pull requests.
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -172,12 +172,13 @@ def configure(configFile=None, baseConfig="ProductionConfig",
     cors.CORS(app, allow_headers='Content-Type')
     app.serverStatus = ServerStatus()
     # Allocate the backend
-    # TODO is this a good way to determine what type of backend we should
-    # instantiate? We should think carefully about this. The approach of
-    # using the special strings __SIMULATED__ and __EMPTY__ seems OK for
-    # now, but is certainly not ideal.
-    dataSource = app.config["DATA_SOURCE"]
-    if dataSource == "__SIMULATED__":
+    # We use URLs to specify the backend. Currently we have file:// URLs (or
+    # URLs with no scheme) for the FileSystemBackend, and special empty:// and
+    # simulated:// URLs for empty or simulated data sources.
+    dataSource = urlparse.urlparse(app.config["DATA_SOURCE"], "file")
+
+    if dataSource.scheme == "simulated":
+        # Ignore the query string
         randomSeed = app.config["SIMULATED_BACKEND_RANDOM_SEED"]
         numCalls = app.config["SIMULATED_BACKEND_NUM_CALLS"]
         variantDensity = app.config["SIMULATED_BACKEND_VARIANT_DENSITY"]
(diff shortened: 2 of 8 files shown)


dcolligan:
hm, interesting idea.  I do agree it is more forwards-compatible.  I don't have any strong objections.  @jeromekelleher  ?

dcolligan:
although, smashing together all of the simulated variables into a URL loses some clarity

jeromekelleher:
Nice idea @adamnovak, nice idea. Not sure I like the idea of adding the variables as URL parameters, but the prefix idea is good.

Also, I think the empty backend is superfluous and we should just get rid of it.

adamnovak:
If you don't want to smash the variables together, you can just use the config options. It seems like it would be nice if the URL could encode all the config, but maybe something with as many options as the simulator backend (especially if you want to override them all) would be better served by using the config variables.

jeromekelleher:
I think this is a good approach. We'll need to revisit our config if/when we move to CherryPy, so there's no point in putting too much effort in right now. However, the URL approach is a definite improvement, so is a good addition.

adamnovak:
I've changed it up a bit; I'm using parameters in the URLs when there are only a couple non-default parameters. and using the config value syntax where there are more. I think this looks cleaner.

adamnovak:
Alright, I've finally satisfied PEP8 and made this vaguely robust.

jmarshall:
If _empty_ isn't superfluous, it could be spelt in URL form as `about:blank` :smile:

adamnovak:
Possibly.

I (as the person who wrote this) think we should merge this and then worry
about dropping the empty backend later.

-Adam

jeromekelleher:
I'm still not crazy about having the parameters for the simulated backend embedded in the string. We're making an already horribly complicated function even more complicated for a very limited use-case (creating simulated backends for testing in a couple of lines less code). I hope we'll be able to dump all of this Flask-config horribleness and replace it with something sane and hierarchical at some point, so I don't see much value in making what we have more elaborate (and difficult to maintain). 

100% enthusiastic about the ``file://``, ``simulated://`` and ``empty://`` prefixes though (we can drop empty soon).

adamnovak:
So you want to see just "simulated://", with all the parameters going
through the config file variables?

I guess it makes sense not to have multiple ways to do things.

jeromekelleher:
>So you want to see just "simulated://", with all the parameters going
> through the config file variables?

That's the idea.

adamnovak:
OK I've changed back to the config parameters instead of query parameters. I also had to tweak the dependency versions to get the tests to run on my Ubuntu 12.04 machine with the venerable Python 2.7.3, and (as I had forgotten how to actually run the tests, and I didn't see it documented anywhere) I made a note in CONTRIBUTING.md that you should test with `nosetests`.

jeromekelleher:
Looks good @adamnovak, thanks. +1 on the latter two commits.

One final question on the first one: what happens when no prefix is specified? If we throw an error then we'll need to go through the docs and the docker images to make sure they are updated. If not, +1.

jeromekelleher:
How's this looking @adamnovak? Will ordinary file paths work without the "file://" prefix?

adamnovak:
Yes, ordinary paths are given the file prefix by default. See https://github.com/ga4gh/server/pull/749/files#diff-0ffa05ceeab404c46544eb9c7d18e46aR178

Note that you can do `file:///absolute/path/to/file` or `file:relative/path/to/file` or even `relative/path` or `/absolute/path`, but `file://first-path-component/rest/of/path` is out of spec for a file URL (and refers to a remote host named `first-path-component`).

jeromekelleher:
Great, thanks @adamnovak. Merging.

Issue #748, opened 2015-10-12
Conversion to SAM produces flags that are all 0


mp15:
Conversion to SAM produces flags that are all 0, this is manifestly incorrect.

```
python ga2sam_dev.py http://localhost:8000 MWtnLXAzLXN1YnNldDpIRzAwMDk2OlNSUjA2MjYzNQ== --referenceId=R1JDaDM4LXN1YnNldDoz
```

mp15:
Resolved

Issue #707, opened 2015-10-09
Update ga2sam tests to check all fields


jeromekelleher:
(This issue is part of a trio of closely related issues #705, #706 and #707 that would be well suited to a group of people working together on.)

The ga2sam utility is tested by converting a specific readGroupSet to SAM format and comparing with the original BAM file. This is done in `tests/unit/test_converters.py`. Run the tests by running `nosetests -vs tests/unit/test_converters.py`.

Update the verifySamRecordsEqual function to compare the full record (currently the check is very incomplete). Ideally, we'd like to convert both records to SAM text format and compare these. The conversion code is quite raw, so there may well be bugs in there.

If #705 has been resolved, update the testSamConversion method to check all readGroups and references on the server. (These fill fail currently because of the bug in #705.)

If #706 has been resolved, add tests for the headers to check that these have been converted correctly.

david4096:
Closing, for continued development see https://github.com/BD2KGenomics/ga4gh-integration/issues/67

PR #750, opened 2015-10-12
Routine to set sam flags mistakenly assumes integer is passed by ref

mp15:
Make this routine return the value instead.

### DIFF
--- a/ga4gh/converters.py
+++ b/ga4gh/converters.py
@@ -158,34 +158,40 @@ class SamLine(object):
     def toSamFlag(cls, read):
         flag = 0
         if read.numberReads:
-            reads.SamFlags.setFlag(
+            flag = reads.SamFlags.setFlag(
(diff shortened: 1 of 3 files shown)


dcolligan:
oh man, how did we miss that.  lol 

good catch

+1

mp15:
just realised there's one more thing that needs changing line 175 is wrong

dcolligan:
this fixes #748, I presume?

jeromekelleher:
Great catch. I think it's pretty clear the test coverage is poor on this code!

+1.

macieksmuga:
Added test case for header in `test_converters.py` also partly fixes #707

dcolligan:
the CI is failing because of syntax errors

mp15:
Fixed them.

dcolligan:
+1

PR #752, opened 2015-10-12
Add integer parsing tests. Fixes #682

### LINKED ISSUE #682: Improve test for backend string parsing functions.
In ga4gh/backend.py, the ``_parseIntegerArgument`` function is responsible for parsing integer values passed as arguments in GET requests.

Add some tests for this in ``tests/unit/test_backends.py``. Add a  test method to the TestPrivateBackendMethods class to test positive results (i.e., do we correctly return the right value?) and another to test negative results (i.e., do we correctly raise a BadRequestIntegerException when called with non-integer values?).

We also want to improve the tests for the _parsePageToken function. We want to include tests for all negative results, as well as improved tests for the positive results. (The existing test is good, but we'd like to check more possibilities, like varying numbers of colon separated values)

andrewyatz:
Ronseal

### DIFF
--- a/tests/unit/test_backends.py
+++ b/tests/unit/test_backends.py
@@ -242,3 +242,15 @@ class TestPrivateBackendMethods(unittest.TestCase):
         goodPageToken = "12:34:567:8:9000"
         parsedToken = backend._parsePageToken(goodPageToken, 5)
         self.assertEqual(parsedToken[2], 567)
+
+    def testParseIntegerArgument(self):
+        good = {"one": "1", "minusone": "-1"}
+        expected = {"one": 1, "minusone": -1}
+        bad = {"string": "A", "float": "0.98"}
+        self.assertEqual(backend._parseIntegerArgument({}, "missing", 0), 0)
+        for key in good:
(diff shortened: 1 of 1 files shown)


dcolligan:
+1

tests/unit/test_backends.py
```diff
@@ -242,3 +242,14 @@ def testParsePageToken(self):
         goodPageToken = "12:34:567:8:9000"
         parsedToken = backend._parsePageToken(goodPageToken, 5)
         self.assertEqual(parsedToken[2], 567)
+
+    def testParseIntegerArgument(self):
+        args = {"good": "1", "negative": "-1", "bad": "A", "badfloat": "0.98"}
```
jeromekelleher: Why not split the success and failure cases into two lists rather than using the keys here? Using hard-coded strings for control flow is always tricky...

andrewyatz: Fair point will refactor

jeromekelleher:
+1. Merging.

Thanks @andrewyatz, looking forward to more PRs!

Issue #719, opened 2015-10-11
Update the download_data.py to download correct reference


jeromekelleher:
The ``scripts/download_data.py`` is used to create the ga4gh-example-data file. Reads and variant data are download from 1000 genomes, and works by pulling down a small number of variants, and the subset of reads and the reference sequence covering those variants. 

Getting access to the correct reference data has proved to be tricky though. See issue #635 for a discussion of the basic issue. Basically, we need a web service where we can get the GRCh37 sequence for a specific interval. FASTA output isn't essential as we can easily reformat it ourselves.

Update script to use this web service. Run the script using ``python scripts/download_data.py``.

PR #754, opened 2015-10-12
closes #635: added ga4gh.utils.seqfetcher to fetch remote sequences by accession and interval

### LINKED ISSUE #635: Downloaded references are GRCh38 but reads are mapped to h37
It seems that the ENA API that we're using only supports h38, so we need to find another way to programatically download a subset of the human references. Any ideas?

reece:
NOTE: I interpreted this issue (#635) as adding the sequence fetching functionality. Will work on #719 next to use this code to update test data.


ga4gh/utils/seqfetcher.py
```diff
@@ -0,0 +1,207 @@
...
+    RuntimeError: No sequence available for NM_9.9
+
+    >>> fetch_seq('QQ01234')
+    Traceback (most recent call last):
+       ...
+    RuntimeError: No sequence fetcher for QQ01234
+
+    """
+
+    ac_dispatch = [
```
dcolligan: we use camelCase convention, not snake_case

ga4gh/utils/seqfetcher.py
```diff
@@ -0,0 +1,207 @@
...
+
+def _fetch_seq_ncbi(ac, start_i=None, end_i=None):
+    """Fetch sequences from NCBI using the eutils interface.
+
+    An interbase interval may be optionally provided with start_i and
+    end_i. NCBI eutils will return just the requested subsequence,
+    which might greatly reduce payload sizes (especially with
+    chromosome-scale sequences).
+
+    >>> len(_fetch_seq_ncbi('NP_056374.2'))
```
dcolligan: the comments in all of these files seem a bit excessive...

tests/unit/test_imports.py
```diff
@@ -152,25 +152,34 @@ class ImportGraphLayerChecker(object):
...
+            ],
+        'libraries': [
+            'ga4gh/configtest.py',
+            'ga4gh/converters.py',
+            ],
+        'protocol': [
+            'ga4gh/_protocol_definitions.py',
+            'ga4gh/protocol.py',
+            ],
+        'utils': ['ga4gh/utils/seqfetcher.py'],
```
dcolligan: not sure if we want to add a 'utils' submodule to the main code... names should be more descriptive than that...

tests/utils_seqfetcher.py
```diff
@@ -0,0 +1,51 @@
...
+
+from __future__ import unicode_literals
+
+import unittest
+
+from ga4gh.utils.seqfetcher import fetch_seq
+
+
+class TestSeqFetcher(unittest.TestCase):
+    def test_ensembl_full(self):
```
dcolligan: again, wrong code convention: camelCase, not snake_case

dcolligan:
This is useful code, but I have difficulty seeing it merged with minor modifications... this functionality is probably only relevant in the `scripts` directory (especially in `download_data.py` which is the module which is responsible for downloads at present), i.e. not as part of the main ga4gh package

ga4gh/utils/seqfetcher.py
```diff
@@ -0,0 +1,207 @@
...
+        {'re': re.compile('^ENS[TP]\d+'),
+         'fetcher': _fetch_seq_ensembl},
+        ]
+
+    eligible_fetchers = [dr['fetcher']
+                         for dr in ac_dispatch
+                         if dr['re'].match(ac)]
+
+    if len(eligible_fetchers) == 0:
+        raise RuntimeError("No sequence fetcher for {ac}".format(ac=ac))
```
jeromekelleher: Is RuntimeError the right choice here? I think this is more like a ValueError: the user provided a value as input that we don't like, right? Looking at the docs for RuntimeError, it seems a bit obscure.

ga4gh/utils/seqfetcher.py
```diff
@@ -0,0 +1,207 @@
...
+        logger.debug("Multiple sequence fetchers found for "
+                     "{ac}; using first".format(ac=ac))
+        fetcher = eligible_fetchers[0]
+
+    logger.debug("fetching {ac} with {f}".format(
+        ac=ac, f=fetcher))
+
+    try:
+        return fetcher(ac, start_i, end_i)
+    except requests.HTTPError:
```
jeromekelleher: It would be nice to keep the requests error here, as we're absorbing a whole class of errors into (essentially) reporting 404s here. Even adding the str() of the original exception would help debug.

ga4gh/utils/seqfetcher.py
jeromekelleher: I tend to agree with @dcolligan here. While the doctest comments are useful, are they likely to be fully maintained in the future? I guess we'd need to actually run the doctests as part of CI to ensure this. However, we'll then probably have random CI failures when NCBI is overloaded and so on... It's a tricky one!

jeromekelleher:
This code is super useful, and I like it a lot. The style is clear, and I have no trouble following it. However, it doesn't agree with some of the stylistic conventions that we've adopted and worked hard to maintain, so I'd be uncomfortable about merging the code in as it is. There's also some new additions in terms of doctests, which we'd need to integrate into CI to be truly useful. Some specific points:

- Since we only need the code for one script, I suggest we just absorb all the code into this script. We can always migrate the code into the central code base if we need to fetch sequences in the future. I'm not quite as concerned about maintaining uniformity of style in the scripts.

- As we are planning a conversion to snake case in the near future, I could be persuaded to make an exception and include it in to ``download_data.py``.

What do you think @dcolligan? I'm not quite sure where to put the unit tests and what to do with the doctests.

Anyway, thanks for this @reece, it is a great addition and we'll definitely use some version of this code.

dcolligan:
Some of the functionality in `scripts` is getting really built out, and could use tests of its own... this contribution definitely fits under that heading.  Whether we want to run those tests every time as part of the core test suite is debatable; I'd lean towards no.  That goes double if we relax the requirements of the test suite to be able to call out over the network.

reece:
I agree with Danny's comments and that the code doesn't belong in server proper. Please reject this PR.

Re: doctests: Since this code is not core and unlikely to change, the doctests (as invoked with python seqfetcher.py) are useful for development. For the failure reason's Jerome mentioned, they should not be run with package testing.

I did take a look at download_data.py, but couldn't grok what it was doing on the first pass and therefore haven't attempted a second take at these issues.

Options: Sketch what you think should happen, and I'll take a stab. Or, feel free to reassign the issues to yourselves and use this code (or not). Your choice.

ga4gh/utils/seqfetcher.py
reece: This code is snakecase because it was pulled from another code base. Imposing non-standard formatting requirements on such packages creates barriers. If this were an external package, this code would look completely normal. (And, I completely get the desire for json/avro symbol parity. I've wrestled with this myself.)

In other words, it's worth reconsidering when/where to impose strict adherence to camelcase.

ga4gh/utils/seqfetcher.py
reece: There was minor gitter discussion on this. Adam Novak and I both use RuntimeError for general errors (for right or wrong). I think of ValueError as being the wrong *type* of input. But, I could get behind KeyError since this is essentially a key-based lookup. To be honest, I don't know what's "right" here.

jeromekelleher: Yeah, there is no 'right answer' here. RuntimeError is fine.

jeromekelleher:
OK, I'll close this PR then. I'll play with the code here and create a new issue, with an eye to reusing as much as possible of this framework. I'll @mention you @reece, and you can have first refusal on tackling it. (It might take a few days to materialise, as I'm at a conference at the moment.)

Issue #711, opened 2015-10-11
Update server_benchmark.py


jeromekelleher:
The ``server_benchmark.py`` script is used to benchmark the server backend without the complications of HTTP overheads. It runs some queries, and uses guppy to profile where the time was spend in generating the responses.

This script has fallen out of date and does not work with the current  server.

- Change the UI so that the mandatory positional argument is dataDir.   Remove the callSetIds and variantSetId arguments.

- Update main so that we have

```python
    backend = backendClass(args.dataDir)
    dataset = backend.getDatasetByIndex(0)
    variantSet = dataset.getVariantSetByIndex(0)
    minTime = benchmarkOneQuery(
        _heavyQuery(variantSet.getId()), args.repeatLimit, args.pageLimit)
```

- Update the ``_heavyQuery`` function to suit. This is based on an older   version of the schema, so check the [schemas](https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/variantmethods.avdl) to find   the correct form of the request.

- Run the script on the ga4gh-example-data using the different profile options.   If this run is too short, run the ``scripts/download_data.py`` script to get   a larger example dataset (run python scripts/download_data.py --help for help    on how to run this script).

- Add a test case to make sure that this script does not become out of   date again.  We need to refactor the script a little to do this.

- Create a class called QueryBenchmarker, which takes a backend object   and a query as parameters. This has one method, benchmark(), which   corresponds to the current benchmarkOneQuery function.

- Use the following framework:

```python
def main(argList=None)
    parser = argparse.ArgumentParser(
        description="GA4GH reference server benchmark")
    # etc
    args = parser.parse_args(argList)
    # as before, 
    query = _heavyQuery(variantSet.getId())
    benchmarker = QueryBenchmarker(backend, query)
    benchmarker.benchmark(args.repeatLimit, args.pageLimit)
    # as before

if __name__ == "__main__":
    main()
```

- Once this is working, make a test case in a new file   ``tests/unit/test_dev_scripts.py``.  This should check to see if the  script at least runs when we point it at the test data, and the output is roughly what we expect. Use the ``utils.captureOutput`` function (discussed in #720) to capture stdout and stderr. The call will look something like

```python
    stdout, stderr = utils.captureOuput(server_benchmark.main, ["tests/data"])
```

PR #755, opened 2015-10-12
updated server_benchmark to new schemas

bwalsh:
https://github.com/ga4gh/server/issues/711#event-433016431


tests/unit/test_server_benchmark.py
```diff
@@ -0,0 +1,57 @@
...
+
+from StringIO import StringIO
+
+
+class TestServerBenchmark(unittest.TestCase):
+    """
+    Tests heap profile
+    """
+    def testHeapProfile(self):
+        # capture std out
```
jeromekelleher: We can use `utils.captureOutput` here now.

server_benchmark.py
```diff
@@ -113,12 +113,15 @@ def benchmarkOneQuery(request, repeatLimit=3, pageLimit=3):
     # return sum(times[2:])/len(times[2:])
     return min(times)
 
-if __name__ == '__main__':
+
+def main(args):
+    global __BACKEND__
```
jeromekelleher: Global variables can lead to nasty headaches, and don't scale well in terms of code complexity. It would be a bit nicer if we could use a class to encapsulate the state (the QueryBenchmarker class suggested in the issue). This would have an instance variable ``self._backend``, which keeps this information.

jeromekelleher:
It would be good to clear up the issues above @bwalsh, but generally this is a very nice update.

bwalsh:
closed this PR in favor of https://github.com/ga4gh/server/pull/760 in order to clean up and squash commits

PR #756, opened 2015-10-12
Generate random read sequences and qualities.

adamnovak:
In pursuit of #717, this PR provides simulated reads with random seeds derived from the overall simulator seed and the number of the read. (Node that you can still, I think, page through reads forever.)

The reads also generate themselves fragment IDs, lengths, and the appropriate number of quality scores and sequence bases.

### DIFF
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import datetime
+import random
 
 import pysam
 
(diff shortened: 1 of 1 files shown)


ga4gh/datamodel/reads.py
```diff
@@ -432,17 +433,31 @@ class SimulatedReadGroup(AbstractReadGroup):
     """
     def __init__(self, parentContainer, localId, randomSeed, numAlignments=2):
         super(SimulatedReadGroup, self).__init__(parentContainer, localId)
+        self._randomSeed = randomSeed
 
     def getReadAlignments(self, referenceId=None, start=None, end=None):
+        rng = random.Random(self._randomSeed)
+        if start is not None:
+            # Skip ahead to start id
+            for i in xrange(start):
```
jeromekelleher: A more efficient approach might be to have 
```
rng.seed(self._randomSeed + start)
```
this way we avoid having to iterate over (potentially) a very large range.

ga4gh/datamodel/reads.py
```diff
@@ -432,17 +433,31 @@ class SimulatedReadGroup(AbstractReadGroup):
...
         # TODO fill out a bit more
+        rng = random.Random(seed)
         alignment = protocol.ReadAlignment()
-        alignment.alignedQuality = [1, 2, 3]
-        alignment.alignedSequence = "ACT"
-        alignment.fragmentId = 'TODO'
+        alignment.fragmentLength = rng.randint(10, 100)
+        alignment.alignedQuality = []
+        alignment.alignedSequence = ""
+        for i in xrange(alignment.fragmentLength):
```
jeromekelleher: xrange is python 2 only. Just use range() and don't worry about the list-cost.

jeromekelleher:
Looks good, thanks @adamnovak. Not sure why the CI checks are failing. 

+1 once CI passes and the nits above are fixed.

ga4gh/datamodel/reads.py
adamnovak: The problem is that the random seeds don't just increment. The state derived from seed 2 isn't the state you get if you seed with 1 and then generate a value.

What we could do is generate each read from an individual seed of (base seed + read number), re-seeding for every read.

adamnovak:
OK I've fixed this up according to your comments. Maybe Travis didn't like it because I used xrange and it wants to test on Python 3? We'll see if it's happier now.

jeromekelleher:
Looks like you have trailing whitespace @adamnovak:

```
ga4gh/datamodel/reads.py:460:60: W291 trailing whitespace
ga4gh/datamodel/reads.py:471:1: W293 blank line contains whitespace
```

If you follow the "details" link next to the travis check you can see the output.

jeromekelleher:
Can we update this please @adamnovak? We want to merge in open PRs ASAP so we can change the default branch to master.

adamnovak:
How about this?

Is the tool we use to check whitespace `flake8`? Is there an easy way to
run that on the whole repository? Is that documented in the `README.md` or
`CONTRIBUTING.md` file? It looks like the Travis tests are all specced out
in the Travis file, but there's no easy way to replicate them locally.

I think my computer is just too slow for Travis; I click on the details
link and the log just loads forever, instead of appearing like it does on
other machines.

jeromekelleher:
Thanks @adamnovak, merging.

@dcolligan provided us with a very handy script to do just that: ``python scripts/run_tests.py``. This should be an exact replica of the travis run.

I've also add added git pre-commit hook to check for flake8 for when I forget to run this script (which is often):
```bash
exec flake8 *.py tests ga4gh scripts --exclude=ez_setup.py,ga4gh/proto
```

PR #757, opened 2015-10-13
Added header information to ga2sam

srblum:
Added header information to ga2sam.


jeromekelleher:
This is is good start, thanks @srblum. We can add an issue to put in tests and extend to use md5s and so on later.

+1

dcolligan:
+1

Issue #761, opened 2015-10-16
Process schemas scripts crashes....


pcingola:
Process schemas scripts crashes when:
  i) Schema entry has no comments (documentation) lines
  ii) Union has null in different order than expected (e.g. union<null, string> as opposed to union<string, null>)

dcolligan:
Fixed in #762

PR #762, opened 2015-10-16
Fixing two bug causing crashes in process_schepas.py. Issue #761

pcingola:


scripts/process_schemas.py
```diff
@@ -163,10 +167,12 @@ def writeConstructor(self, outputFile):
             self._writeWithIndent(string_, outputFile, 3)
             # Backtick quoted strings cause problems with Sphinx, so we
             # strip them out here.
-            doc = field.doc.replace('`', '')
-            self._writeWithIndent('"""', outputFile, 2)
-            self._writeWrappedWithIndent(doc, outputFile, 2)
-            self._writeWithIndent('"""', outputFile, 2)
+            if field.doc:
```
jeromekelleher: Change to ``if field.doc is not None`` please. This is discouraged in PEP8:

> Also, beware of writing if x when you really mean if x is not None -- e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

scripts/process_schemas.py
```diff
@@ -163,10 +167,12 @@ def writeConstructor(self, outputFile):
...
+                doc = field.doc.replace('`', '')
+                doc = field.doc.replace('`', '')
```
jeromekelleher: Line repeated here.

jeromekelleher:
Looks good @pcingola. Have you checked this against the current schemas? If this doesn't change the ``_schema_definitions.py`` then I'm happy with these changes (modulo nits above).

pcingola:
Fixed those minor issues.
Here is the test using current schema:

    $ python ./scripts/process_schemas.py --verbose --outputFile ../_protocol_definitions.py --avro-tools-jar ../avro-tools-1.7.7.jar --inputSchemasDirectory ../schemas/src/main/resources/avro/ 0.6.4568e6f6
    $ diff ../_protocol_definitions.py ga4gh/_protocol_definitions.py
    $

So nothing changes.
I'm squashing commits now....

pcingola:
Commits squashed!
I hope everything looks fine now.

pcingola:
OK, this still looks like 2 commits. 
Following the recommended procedure still shows as 2 commits 
and it's giving me the following error:

    $ git push
    To git@github.com:pcingola/server.git
     ! [rejected]        Issue_761_process_schema -> Issue_761_process_schema (non-fast-forward)
    error: failed to push some refs to 'git@github.com:pcingola/server.git'
    hint: Updates were rejected because the tip of your current branch is behind
    hint: its remote counterpart. Integrate the remote changes (e.g.
    hint: 'git pull ...') before pushing again.
    hint: See the 'Note about fast-forwards' in 'git push --help' for details.

The error recommends using 'git pull', but you guys told me to NEVER use "git pull".

This is ridiculous. I'm currently spending more time dealing with commit squashing 
than actual code writing. I suggest that we either consider accepting non-squash commits 
until we can find a bullet proof procedure to do so.

pcingola:
I think I (partially) solved the issue by doing

    git push --force

after the "git rebase -i develop" command, otherwise I get 
the error above. 

Please consider lowering the bureaucratic 'git' burden, it is 
a serious determent for contributors.

dcolligan:
@pcingola I have a pull request out to update some of the documentation to clarify some common git issues, such as why we advise people not to use git pull.  However, what you just did (using force to push to a topic branch in your own repo after modifying branch history) is indeed covered in our github workflow instructions: http://ga4gh-reference-implementation.readthedocs.org/en/develop/development.html#github-workflow

The reason we don't want to accept (unnecessary) merge commits into the project is because it would severely lower the utility of the git log.  The git log is supposed to be a clean, readable history that conveys what actually happened in the project.  Clogging it up with intermediate commits and merge commits would make it harder for everyone to use.

pcingola:
I followed the docs.
I did not use "git pull". 
Still errors show up and (apparently) the solution was adding an 
additional "git push --force" after the rebase (not before as I understood 
from the docs you mention).
I understand the reason for requesting this. Unfortunately I still think that 
until we have a reliable procedure you should consider accepting non-squashed 
commits from contributors and/or helping performing the squashing when errors 
pop up.

jeromekelleher:
Looks like we're down to 1 commit and ready to merge. 

+1.

I agree it's frustrating @pcingola, and squashing is messy. I think avoiding running ``git pull`` and ``git merge`` keeps a lot of the messiness away. Here's my simple-minded recipe:

```bash
$ git fetch upstream # fetch the latest code
$ git checkout upstream/develop # check out the tip of develop
$ git checkout -b 123-feature-branch # Create feature branch.
$ # Write code and make commits.
$ git fetch upstream # fetch the latest code
$ git rebase -i upstream/develop # Squash and rebase
$ git push origin 123-feature-branch # Push new branch
$ # get feedback, update the code and make some commits
$ git fetch upstream # fetch the latest code
$ git rebase -i upstream/develop # Squash and rebase
$ git push -f origin 123-feature-branch # force-push updated commits.
```

This is what I do 99% of the time, and it works pretty well. Occasionally I need to resolve some merge conflicts, but they're usually easy enough and nearly always my own fault for over-complicating things. 

I think this is more-or-less the same as what we recommend in the docs, except I don't bother maintaining a local copy of develop. I'm sure there's a million ways of doing it, I just want to share my method as it works for me.

PR #770, opened 2015-10-23
G2P '/genotypephenotype/search' endpoint

bwalsh:
A continuation of the conversation from  https://github.com/ga4gh/server/pull/607

* extended CLI genotypephenotype-search method to take json strings as arguments
* sparql perfomance improvements
* split g2p tests out of test_views.py to test_g2p.py; also fixed some style issues
* optimize and correct queries
* modified filesystem backend to expect g2pDatasets in ga4gh-example-data
* simplify by removing  GeneotypePhenotypeIterator

This PR is paired with sibling pull requests:
* compliance: https://github.com/ga4gh/compliance/pull/115
* schema: https://github.com/ga4gh/schemas/pull/432

Note:  the commit history is not as clean as I'd like.  Consecutive commits have been squashed, however the merges from develop intersperse our commit history.  
Edit: If necessary, I can drop and recreate the repo


jeromekelleher:
Nit: Some files have gained executable permissions --- please remove these.

ga4gh/backend.py
```diff
@@ -762,6 +770,69 @@ def runSearchDatasets(self, request):
...
+        if request.pageSize <= 0:
+            raise exceptions.BadPageSizeException(request.pageSize)
+        responseBuilder = protocol.SearchResponseBuilder(
+            responseClass, request.pageSize, self._maxResponseLength)
+        # ensure request has data we can query
+        if (request.evidence is None and
+                request.phenotype is None and
+                request.feature is None):
+            msg = "Error:One of evidence,phenotype or feature must be non-null"
+            raise exceptions.BadRequestException(msg)
```
jeromekelleher: This message isn't actually returned to the user. You probably need a more specific exception, or else change the plumbing of BadRequestException to take an optional argument.

ga4gh/cli.py
```diff
@@ -428,16 +429,50 @@ def run(self):
...
 
 
+class SearchGenotypePhenotypeRunner(AbstractSearchRunner):
+    """
+    Runner class for the genotypephenotype/search method.
+    """
+    def __init__(self, args):
+        super(SearchGenotypePhenotypeRunner, self).__init__(args)
+        argStore = vars(args)
+        for key, value in vars(args).iteritems():
```
jeromekelleher: I'm slightly confused by this --- is the idea to parse the argument for these values into a JSON dict? I'd try to be a bit more explicit about it
```python
def parseJson(value):
     ret = None
     if value is not None:
          ret = json.loads(value)
     return ret
self._feature = parseJson(args.feature)
# etc
```

ga4gh/client.py
```diff
@@ -471,6 +486,21 @@ def _runListReferenceBasesPageRequest(self, id_, request):
         return self._deserializeResponse(
             response.text, protocol.ListReferenceBasesResponse)
 
+    def searchGenotypePhenotype(self, feature=None, phenotype=None,
```
jeromekelleher: This isn't necessary here, it's already defined in the superclass.

ga4gh/backend.py
```diff
@@ -762,6 +770,69 @@ def runSearchDatasets(self, request):
...
+        self.validateResponse(responseString, responseClass)
+        self.endProfile()
+        return responseString
+
+        # # old ......
+        # return self.runSearchRequest(
+        #     request, protocol.SearchGenotypePhenotypeRequest,
+        #     protocol.SearchGenotypePhenotypeResponse,
+        #     self.genotypePhenotypeGenerator)
+
```
jeromekelleher: This is redoing quite a lot of things, and we're breaking layer a bit as well. I've played with this a bit, and here's what I suggest as a starting point:
```python
    def genotypePhenotypeGenerator(self, request):
        if (request.evidence is None and
                request.phenotype is None and
                request.feature is None):
            msg = "Error:One of evidence,phenotype or feature must be non-null"
            raise exceptions.BadRequestException(msg)
        annotationList = self._g2pDataset.queryLabels(
            request.feature, request.evidence, request.phenotype)
        return self._objectListGenerator(request, annotationList)
        
    def runSearchGenotypePhenotype(self, request):
        return self.runSearchRequest(
            request, protocol.SearchGenotypePhenotypeRequest,
            protocol.SearchGenotypePhenotypeResponse,
            self.genotypePhenotypeGenerator)
```

This implemenatation pawns off all the paging work to the ``_objectListGenerator``, which expects a list of objects that have a ``toProtocolElement`` method. If you return a list of Annotation objects from ``queryLabels`` then this should be easy.

This has major disadvantages, as we get the list of all associations every time we make a request. I wonder how much of a problem this will be in practise though. Do we expect to stream millions of annotations? Managing the paging properly without violating all sorts of layering assumptions is a massive headache. I think this approach is probably good enough for initial purposes, and it has the advantage of being correct.

ga4gh/backend.py
```diff
@@ -803,6 +874,12 @@ def __init__(
                 numAlignments=numAlignments)
             self.addDataset(dataset)
 
+        # g2pDatasets
+        self._g2pDataset = genotype_phenotype.\
+            G2PDataset('cgd',
```
jeromekelleher: This isn't empty...  We should really have a list of G2P datasets (or G2pAnnotationSets, or whatever we want to call them). The semantics of this sort of stuff is a lot easier then.

We can't have paths referring to the git structure in the production code, as these will all break when the package is installed.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,366 @@
...
+            }
+            """
+
+        # initialize graph
+        self._rdfGraph = rdflib.ConjunctiveGraph()
+
+        # load with data
+        for source in self._sources:
+            if source.endswith('.ttl'):
+                self._rdfGraph.parse(source,
```
jeromekelleher: Do we need to break this line?

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,366 @@
...
+        # now = time.time()
+        associations = self.queryLabels(
+            request.feature, request.evidence, request.phenotype,
+            request.pageSize, offset)
+        # responseTime = time.time()-now
+        # if responseTime > self.RESPONSETIME_LOGGING_THRESHOLD:
+        #     print('_search', responseTime)
+        #     print(request)
+
+        self.associationsLength = len(associations)
```
jeromekelleher: This doesn't seem to be used anywhere.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,366 @@
...
+        for association in associations:
+            yield association
+
+    def queryLabels(
+         self, location=None, drug=None, disease=None, pageSize=None,
```
jeromekelleher: This pageSize and offset stuff is horribly tricky. We can't just pass pageSize directly in to the LIMIT statement here because we have potential DOS attacks. We need to be able to limit the size of the output buffer, and this has to be done at the top level. The only way this works efficiently and cleanly for large result sets is if we can get a stream of objects from the rdflib graph, and be able to restart that stream from a particular offset. If we don't expect to stream large sets of results back, and realistically should be able to fit all results onto a single page, then we should forget about paging entirely at this level and just return a list of all the results.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,366 @@
...
+
+            if row['p'] in a:
+                a[row['p']].append(obj)
+            else:
+                a[row['p']] = obj
+
+            a['id'] = row['s']
+        return a
+
+    def toGA4GH(self, annotation):
```
jeromekelleher: It feels like we're munging a lot of stuff together in this class. Perhaps we could have another class called Annotation, which takes care of the process of converting one of these low-level annotations into a GA4GH object via the ``toProtocolElement`` method? This would make unit testing a lot easier and I think clarify the purpose of classes a bit more.

tests/unit/test_g2p.py
```diff
@@ -0,0 +1,208 @@
...
+import unittest
+import logging
+
+import ga4gh.frontend as frontend
+import ga4gh.protocol as protocol
+
+
+class TestGenotypePhenotypeSearch(unittest.TestCase):
+    """
+    Tests the basic routing and HTTP handling for
```
jeromekelleher: We don't really care about the HTTP level stuff here  as this is handled by other tests. What we really want to see are some unit tests of the classes in the G2P module. I wouldn't bother with the flask routing thing here, and instead try instantiating your backend directly. You could run some tests using the client API based on the data that you have in the tests directory.

jeromekelleher:
It's looking good @bwalsh, thanks for the updates. I still think it's worthwhile defining a simulator for G2P data. It can be a stupid as a box of rocks, and doesn't have to generate data that's in any way realistic. However, if we have simulated data this will make testing so, so much easier as we'll be able to fit in to the existing frameworks and not have to worry about the existence of the test data.

ga4gh/backend.py
bwalsh: Thanks.  I've added this change.  As you say, we will determine how much of a problem it will be in practice.    One future change in backend.py might be to separate out "total number of rows"  from "length of the collection in the request".

ga4gh/datamodel/genotype_phenotype.py
bwalsh: Unclear.  Break the _searchQuery multi-line string?

ga4gh/datamodel/genotype_phenotype.py
bwalsh: Agreed.  I'll remove it

ga4gh/datamodel/genotype_phenotype.py
bwalsh: I think we are ok with this now.

ga4gh/datamodel/genotype_phenotype.py
bwalsh: agreed

tests/unit/test_g2p.py
bwalsh: agreed

ga4gh/datamodel/genotype_phenotype.py
jeromekelleher: I was just wondering if we could put in 
```python
                   self._rdfGraph.parse(source, format='n3')
```
The line doesn't look that wide to me.

ga4gh/backend.py
bwalsh: adding
```
class BadRequestException(RuntimeException):
    """
    A request that we don't like was sent to the server.
    """
    httpStatus = 400
    message = "Bad request"

    def __init__(self, messageToDisplay):
        self.message = messageToDisplay

```

ga4gh/backend.py
bwalsh: done.

PR #772, opened 2015-10-23
768 merge seqfetcher

### LINKED ISSUE #719: Update the download_data.py to download correct reference

### LINKED ISSUE #768: Simplify seqfetcher code for specific needs in download_data.py
This is a follow on from PR #754, which is in turn adressing #635 and #719. Closing this issue should resolve both of these issues too.

The code in #754 is very powerful, but is really a lot more than we need. All we want is a simple method to download GRCh37 sequences, and no more. Therefore, adapt the code in #754 as follows:

1) Copy the function fetch_seq_ncbi to ``scripts/download_data.py``, and rename to fetchSequence. CamelCase the variables and rename to follow the local conventions (e.g, use accession, startIndex rather than acc, start_i).
2) Replace the code in downloadFastas where we make calls to the ENA apis to use this function instead.
3) Add  command line option ``--test`` that runs the doctests for the ``fetchSequence`` function.
4) Change the accessions around line 135 from (e.g.) CM000663.2 to CM000663.1.

@reece -- would you like to tackle this? All the hard work is done, and it's just a case of removing stuff. No worries if you don't want it.

reece:
This PR should close #768, #719, and #635.
When merged, we need to close #719 and #635 manually.
Please see comments regarding this commit in #768.

### DIFF
--- a/scripts/download_data.py
+++ b/scripts/download_data.py
@@ -15,8 +15,8 @@ import gzip
 import json
 import hashlib
 import os
+import requests
 import tempfile
-import urllib
 import urllib2
 
 import pysam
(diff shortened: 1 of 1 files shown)


scripts/download_data.py
```diff
@@ -116,6 +116,45 @@ def getMaxPos(self, chrom):
         return minMax.maxPos
 
 
+def fetchSequence(ac, startIndex=None, endIndex=None, as_list=False):
```
jeromekelleher: ``as_list`` -> ``asList``. We need to camelCase all identifiers I'm afraid. (I'm looking forward to the change to protobuf so I won't need to say this any more!)

scripts/download_data.py
```diff
@@ -116,6 +116,45 @@ def getMaxPos(self, chrom):
...
+
+    >>> fetchSequence('NP_056374.2')[0:10]
+    'MESRETLSSS'
+
+    >>> fetchSequence('NP_056374.2',0,10) == fetchSequence('NP_056374.2')[0:10]
+    True
+
+    """
+
+    url_fmt = ("http://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?"
```
jeromekelleher: camelCase

scripts/download_data.py
```diff
@@ -116,6 +116,45 @@ def getMaxPos(self, chrom):
...
+               "db=nucleotide&id={ac}&rettype=fasta")
+    if (startIndex is None or endIndex is None):
```
jeromekelleher: Delete the brackets here, as they aren't necessary.

scripts/download_data.py
jeromekelleher: Actually, why do we have the ``as_list`` argument here at all? We're only using this function in one place, and it uses the non-default value for this. It would seem simpler to just return a list in all situations, and use ``"".join(fetchSequence('NP_056374.2',0,10))`` for the doctests.

jeromekelleher:
Looks good, thanks @reece. I have some minor nits above. The ``asList`` argument does make it simpler to test the output in the doctests, but at the cost of making the client code a bit more complex (and, in a way, not actually tested at all).

reece:
Removed as_list entirely

url_fmt -> urlFmt (ugh -- an exemplar of fugly casing)

removed parens

RE: why have as_list? Because no one would expect a function called fetchSequence to return a list of sequence snippets.

jeromekelleher:
Great, thanks @reece! I think we're good to merge once the commits are squashed.

I agree on the aesthetic stuff --- I am looking forward to having a snake_case codebase sometime soon!

jeromekelleher:
Can we squash the commits please @reece? It would be nice to merge this as @dcolligan has another patch ready to land on the same code in #778.

reece:
@jeromekelleher: collapsed, committed, and checks passed.

jeromekelleher:
Excellent, thanks @reece. +1, merging.

Issue #742, opened 2015-10-12
Add test for FASTA output from reference-list-bases


marghoob:
The tests should make sure the FASTA output is as expected. Tests the feature added in #732 pull request (issue #712).

dcolligan:
Fixed in #773

PR #773, opened 2015-10-26
Tests for fasta, etc. cli output options

dcolligan:
Issue #742

### DIFF
--- a/tests/end_to_end/test_client_json.py
+++ b/tests/end_to_end/test_client_json.py
@@ -16,11 +16,9 @@ import ga4gh.cli as cli
 import tests.utils as utils
 
 
-class TestClientJson(unittest.TestCase):
+class TestClientOutput(unittest.TestCase):
     """
-    Tests that the JSON output by the client on the command line for
-    various options is equal to the values we find using the Python
-    client API.
+    Base class for client output tests
     """
     def setUp(self):
         self._dataDir = "tests/data"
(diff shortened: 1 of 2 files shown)


tests/unit/test_cli.py
```diff
@@ -211,3 +214,63 @@ def testReferenceBasesListArguments(self):
...
+    def _getRunPrintMethodCalls(self, runner):
+        printCalls = []
+        with mock.patch('__builtin__.print') as printMethod:
+            printMethod.side_effect = \
+                lambda *args, **kwargs: printCalls.append((args, kwargs))
+            runner.run()
+        return printCalls
+
+    def testListReferenceBasesFasta(self):
+        args = self.FakeArgs('fasta')
```
jeromekelleher: Mocks are neat, but I find them hard to follow and (particularly) maintain sometimes. Here we're violating some interface rules by accessing the ``_client`` attribute, and we also need to know the detailed interface of  the ``listReferenceBases`` method. Depending on the internal structures in the class makes it difficult to change things, as we don't know what tests we're going to break, or necessarily how to fix the tests that have been broken afterwards. I hit a lot of this in the last refactoring pass.

I would _much_ rather depend on the external, advertised interface for tests. Here for example, why not use a local simulated backend, and capture the output? We can then compare the fasta output with the truth that we get from the backend.

dcolligan: two reasons:
1) that's more integration test-y than unit test-y ... mocks make it easy to isolate a particular piece of code and not test any other coupled pieces of code (for instance, if the simulated backend breaks, your version of the test would fail)
2) it would be difficult to capture the output without invoking client_dev.py in another process, since run calls directly to print

jeromekelleher: OK, point (1) is well taken, and convincing.  We also need integrationy test for this though--- we should test the end-to-end correctness of FASTA output.

Point (2) isn't true any more though: we now have some rather snazzy machinery for running CLI commands and capturing the output, without forking a subprocess. See ``tests/end_to_end/test_client_json.py`` for examples.

dcolligan:
@jeromekelleher updated with end to end test

tests/end_to_end/test_client_json.py
```diff
@@ -28,17 +26,54 @@ def setUp(self):
...
+    Tests client FASTA output
+    """
+    def captureFastaOutput(self, command, arguments=""):
+        stdout = self.captureCliOutput(command, arguments, "fasta")
+        lines = stdout.split()
+        return lines
+
+    def testListReferenceBases(self):
+        referenceSetIterator = self._client.searchReferenceSets()
+        referenceSet = referenceSetIterator.next()
```
jeromekelleher: Calling ``.next`` is Python 2 only --- the Python 2/3 way to do it is to call ``next(referenceSetIterator)`` instead.

jeromekelleher:
Thanks @dcolligan, looks good. +1 modulo the  ``next`` thing above.

dcolligan:
@jeromekelleher updated

jeromekelleher:
Thanks @dcolligan, merging.

Issue #509, opened 2015-07-22
Update example data


danvk:
I'd like to run `ga4gh_server` from `origin/develop`, but I'm getting this error:

```
  File "/Users/danvk/github/ga4gh-server/ga4gh/datamodel/datasets.py", line 141, in __init__
    for variantSetId in os.listdir(variantSetDir):
OSError: [Errno 2] No such file or directory: 'ga4gh-example-data/reads/variants'
```

Has the expected data layout changed from what's provided at `http://www.well.ox.ac.uk/~jk/ga4gh-example-data.tar`? If so, it would be helpful to provide an updated version of the example data that does work.

jeromekelleher:
Yes, the layout has changed. To make this work on the new layout, use
```
cd ga4gh-example-data
mkdir dataset1
mv reads variants dataset1
```
I'm not sure what the right thing to do is re the tarball. Certainly the existing one shouldn't change until we update the released version. Perhaps we should make a new tarball with a versioned name, and update the docs for develop to point to this?

danvk:
Yes, you should version the sample data.

Moving `reads` and `variants` into a `dataset1` directory lets the server run, but it doesn't look like cross-linking of references is working (see my command on the linked issue).

danvk:
(probably because `ga4gh-example-data/references` is empty)

jeromekelleher:
References support is crude at the moment @danvk --- we're basically not checking if things link up at all, and we're not quite sure how to do this. See issue #489 and also #312. We definitely want to have some example data in which all of this stuff links up (reads, references and variants), but we don't have it at the moment.

jeromekelleher:
I've updated the example data to have the right format. It's at http://www.well.ox.ac.uk/~jk/ga4gh-example-data-v1.0.tar.

danvk:
Thanks. Just to be clear, the `references` directory is still empty?

jeromekelleher:
Yes --- sorry @danvk, I just don't have the time (or the knowledge) to make the example data link up correctly. We could drop in some arbitrary stuff in as the reference, but I think this would do more harm than good. Ideally, we want 

1. A small chunk of the 1000G reference
2. A 1kg-phase1 dataset containing some reads and variants over this part of the reference
3. A 1kg-phase3 dataset (likewise).

I'll push #489 up the priority list and see if we can get some volunteers to help, as we really are slightly crippled without good data. Sorry to keep you waiting on this.

macieksmuga:
[...]

PR #774, opened 2015-10-26
Changed links in README to point to develop branch.

jeromekelleher:
Changes the links for the version of the README in develop to point to the docs in develop. I guess when we create a release branch for the 0.2 version we can change these back to stable, so that we're pointing at the right version from PyPI. This is slightly error prone though, so if anyone has better ideas please shout.

Discussion on #509.


dcolligan:
Hm... tricky issue.  I guess if there is information like this that we want to remain different in certain branches it makes sense to have it somewhere else besides in the branche(s) proper since, as you point out, the state in a certain branch that we want to keep will get overwritten in a merge unless we are careful to manually specify it.  But if we want to keep the readme file as is, we're kind of stuck with that...

How about this as an alternative?  Link both the develop and stable branches in the README.  E.g.  "Here is the link for documentation about this line, here is the documentation for that line."  That is, the READMEs in both branches are the same, and we don't need to worry about not updating them as we go forward.

jeromekelleher:
This is a tricky one all right. Your idea is more maintainable, but I worry about the potential confusion on the [PyPI page](https://pypi.python.org/pypi/ga4gh). This should be our first point of contact with the majority of users as we become more mature, and it would be good to have a simple message to follow there.

Perhaps this is being caused by our current trick of using the README.rst as the landing page for both GitHub and PyPI. Maybe we should have two separate ones? One is focused on users and the other on devs?

dcolligan:
That's another idea.

Yet another idea would be to just send people to the readthedocs page, with no branch specified.  That will be applicable for all branches, if a bit ambiguous.

jeromekelleher:
>Yet another idea would be to just send people to the readthedocs page, with no branch specified. That will be applicable for all branches, if a bit ambiguous.

I'm not crazy about this one --- it's not at all obvious what branch you're looking at when you hit readthedocs. I this would lead to even more confusion than we have now.

I'll make a new PR with a proposal for the two README's approach.  Putting a -1 on this one until we've decided which one we prefer.

-1.

dcolligan:
The other PR seems better... closing...

PR #776, opened 2015-10-29
Split landing pages into dev and stable.

jeromekelleher:
Here is the counter proposal for #774. I prefer this; the landing pages for PyPI and GitHub can now evolve independently, as we don't have to compromise on the content that is suitable for both.


dcolligan:
+1

jeromekelleher:
@dcolligan, merge me!

PR #780, opened 2015-11-03
non-merge RNA API comment hunting PR

saupchurch:
This is a preview of the RNA API server code and not intended for merging.  Due to the size of the PR I wanted to get started on the review and comment phase while I finish up the last few TODO's.

Known issues:
TODOs concerning compoundId (backend.py) - currently in progress
_protocol_definitions.py - was manually edited, will generate with script for real PR
59K additions are in the data files for the tests - will trim these down for real PR


ga4gh/backend.py
```diff
@@ -531,6 +539,96 @@ def callSetsGenerator(self, request):
...
+            compoundId = datamodel.RnaQuantificationCompoundId.parse(request.rnaQuantificationId)
+            rnaQuantificationName = compoundId.rnaQuantification
+            dataset = self.getDataset(compoundId.datasetId)
+            try:
+                rnaQuant = dataset.getRnaQuantification(rnaQuantificationName)
+            except exceptions.RnaQuantificationNotFoundException:
+                return self._noObjectGenerator()
+            return self._singleObjectGenerator(rnaQuant)
+
+    def expressionLevelGenerator(self, request):
```
dcolligan: these are pretty long methods you're adding here... generally it's a good idea to keep methods to as few lines as possible so maintainers don't need to keep track of a lot of in-scope variables... for instance, in this method at least building the ExpressionLevel object could be moved to its own private method

ga4gh/backend.py
```diff
@@ -531,6 +539,96 @@ def callSetsGenerator(self, request):
...
+                yield expressionLevel, nextPageToken
+                expressionLevelData = nextExpressionLevelData
+
+    def featureGroupGenerator(self, request):
+        featureGroupId = request.featureGroupId
+        currentIndex = 0
+        #TODO; remove hard coded dataset ID
+        dataset = self.getDataset('MWtnLXAzLXN1YnNldA==')
+        if request.pageToken is not None:
+            currentIndex, = _parsePageToken(request.pageToken, 1)
```
dcolligan: the comma after currentIndex is easy to miss... change this line to be:
```
currentIndex = _parsePageToken(request.pageToken, 1)[0]
```

ga4gh/backend.py
```diff
@@ -531,6 +539,96 @@ def callSetsGenerator(self, request):
...
+        for rnaQuantId in rnaQuantificationIds:
+            rnaQuant = rnaQuantifications[rnaQuantId]
+            expressionLevelIterator = rnaQuant.getExpressionLevel(
+                expressionLevelId, featureGroupId)
+            expressionLevelData = next(expressionLevelIterator, None)
+            while (expressionLevelData is not None and
+                    currentIndex < self._defaultPageSize):
+                nextExpressionLevelData = next(expressionLevelIterator, None)
+                nextPageToken = None
+                if nextExpressionLevelData is not None:
```
dcolligan: is it not possible to eliminate some of this logic with some of our already-written utility functions in `backend`?  Like `_topLevelObjectGenerator` or a similar method?

ga4gh/backend.py
```diff
@@ -803,6 +949,10 @@ def __init__(
                 numAlignments=numAlignments)
             self.addDataset(dataset)
 
+        # Features
+        self._featureIdMap = {}
+        self._featureIds = sorted(self._featureIdMap.keys())
```
dcolligan: not really necessary to sort an empty data structure

ga4gh/datamodel/datasets.py
```diff
@@ -190,3 +256,4 @@ def _setMetadata(self):
                 except KeyError as err:
                     raise exceptions.MissingDatasetMetadataException(
                         metadataFileName, str(err))
+
```
dcolligan: flake will likely ding you for empty lines at the end of a file

ga4gh/datamodel/datasets.py
```diff
@@ -180,6 +226,26 @@ def __init__(self, localId, dataDir, backend):
...
                 self.addReadGroupSet(readGroupSet)
 
+        # Rna Quantification
+        rnaQuantDir = os.path.join(dataDir, "rnaQuant")
+        for rnaQuantId in os.listdir(rnaQuantDir):
+            relativePath = os.path.join(rnaQuantDir, rnaQuantId)
+            if os.path.isdir(relativePath):
+                rnaQuantification = rna_quantification.RNASeqResult(
+                    self, rnaQuantId, relativePath)
+                self._rnaQuantificationIdMap[rnaQuantId] = rnaQuantification
```
dcolligan: it would be good to have a `addRnaQuantification` method, a la the methods that are provided for other objects (see above in the file)... same goes for other objects you've added

ga4gh/datamodel/ontologies.py
```diff
@@ -0,0 +1,82 @@
...
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import ga4gh.protocol as protocol
+
+
+class OntologyTermSet(object):
+    """A set of related ontology terms.
```
dcolligan: We usually have the comment style with the triple quotes on their own lines
```
"""
like so
"""
```

ga4gh/datamodel/ontologies.py
```diff
@@ -0,0 +1,82 @@
...
+
+class OntologyTerm(object):
+    """A specific ontology term."""
+    def __init__(self, ontologySource, id, name):
+        self.__ontologySource = ontologySource
+        self.__id = id
+        self.__name = name
+
+    def __str__(self):
+        return self.ontologySource + "/" + self.id + "/" + str(self.name)
```
dcolligan: string formatting should be done with the `.format` method ... both as a matter of style and because the way you're doing it is slightly less efficient in creating some strings that are immediately discarded

ga4gh/datamodel/ontologies.py
```diff
@@ -0,0 +1,82 @@
...
+        "create a new ontology term"
+        self.add(OntologyTerm(self.ontologySource, id, name))
+
+
+class OntologyTerm(object):
+    """A specific ontology term."""
+    def __init__(self, ontologySource, id, name):
+        self.__ontologySource = ontologySource
+        self.__id = id
+        self.__name = name
```
dcolligan: We don't use double underscore prefixed attributes anywhere in the code, to my knowledge... we can probably just stick with single underscore ones

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class Feature(object):
+    """
+    Class representing a Feature annotations
+    """
+    def __init__(
+            self, id, parentIds, featureSetId,
+            referenceName, start, end, featureType, attributes):
+        self.__id = id
```
dcolligan: same comment here about double underscores

ga4gh/frontend.py
```diff
@@ -497,6 +497,30 @@ def searchDatasets():
         flask.request, app.backend.runSearchDatasets)
 
 
+@DisplayedRoute('/rnaquantification/search',postMethod=True)
```
dcolligan: space between the comma and `postMethod` so it's easier to read

ga4gh/gff3.py
```diff
@@ -0,0 +1,223 @@
+"""
+Object representation of GFF3 data and parser for GFF3 files.
+See: http://www.sequenceontology.org/gff3.shtml
+"""
+
```
dcolligan: you're missing the division future import here... we want to have all three imports in every python file we create

ga4gh/gff3.py
```diff
@@ -0,0 +1,223 @@
...
+import gzip
+import bz2
+import collections
+
+
+class GFF3Exception(Exception):
+    "exception associated with GFF3 data"
+    def __init__(self, message, fileName=None, lineNumber=None):
+        if fileName is not None:
+            message = str(fileName) + ":" + str(lineNumber) + ": " + message
```
dcolligan: use `format` for string formatting

ga4gh/gff3.py
```diff
@@ -0,0 +1,223 @@
...
+        Exception.__init__(self, message)
```
dcolligan: use `super` instead of hardcoding the superclass

ga4gh/gff3.py
```diff
@@ -0,0 +1,223 @@
...
+        self.parents = set()
+        self.children = set()
+
+    @staticmethod
+    def __dotIfNone(val):
+        return val if val is not None else '.'
+
+    def __attributeStr(self, name):
+        "return name=value for a single attributed"
+        return name + "=" + ",".join(urllib.quote(v)
```
dcolligan: use `format` for string formatting

saupchurch:
Working through these - keep em coming.  Thanks @dcolligan !

ga4gh/gff3.py
```diff
@@ -0,0 +1,223 @@
...
+            return open(self.fileName)
+
+    SPLIT_ATTR_RE = re.compile("^([a-zA-Z_]+)=(.+)$")  # parses `attr=val'
+
+    def __parseAttrVal(self, attrStr):
+        """returns tuple of tuple of (attr, value), multiple are returned to
+        handle multi-value attributes"""
+        m = self.SPLIT_ATTR_RE.match(attrStr)
+        if m is None:
+            raise GFF3Exception("can't parse attribute/value: '" + attrStr +
```
dcolligan: same comment about `format`

ga4gh/gff3.py
```diff
@@ -0,0 +1,223 @@
...
+                                    self.fileName, self.lineNumber)
+            attributes[name] = vals
+        return attributes
+
+    GFF3_NUM_COLS = 9
+
+    def __parseRecord(self, gff3Set, line):
+        row = line.split("\t")
+        if len(row) != self.GFF3_NUM_COLS:
+            raise GFF3Exception("Wrong number of columns, expected " +
```
dcolligan: I'll stop the `format` comments now... I think you get the point

scripts/encode2ga4gh.py
```diff
@@ -0,0 +1,296 @@
+#!/usr/bin/env python
+from __future__ import division
```
dcolligan: missing 2 future imports

scripts/encode2ga4gh.py
```diff
@@ -0,0 +1,296 @@
...
+
+import os
+import sys
+import json
+import errno
+import urllib2
+import optparse
+
+
+class Color:
```
dcolligan: make this a new-style class:
```
class Color(object):
```

scripts/encode2ga4gh.py
```diff
@@ -0,0 +1,296 @@
...
+    def yellow(astr):
+        return Color.YELLOW + str(astr) + Color.NORMAL
+
+    @staticmethod
+    def blue(astr):
+        return Color.BLUE + str(astr) + Color.NORMAL
+
+
+def log(message):
+    print >>sys.stderr,  "== " + Color.green("{}".format(message))
```
dcolligan: oh god no python3 incompatible print statements!

scripts/encode2ga4gh.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+def get_files_from_host(data, host, output_type, request=False, subset=None):
```
dcolligan: our style is camelCase not snake_case (at least for now...)

scripts/encode2ga4gh.py
```diff
@@ -0,0 +1,296 @@
...
+                       description, annotationId, subset=None):
+    req = urllib2.Request(url, headers=headers)
+    try:
+        response = urllib2.urlopen(req)
+    except urllib2.URLError as e:
+        if hasattr(e, 'reason'):
+            print 'We failed to reach a server.'
+            print 'Reason: ', e.reason
+        elif hasattr(e, 'code'):
+            print 'The server couldn\'t fulfill the request.'
```
dcolligan: Instead of using escapes, you can just surround the string with the quote marks you are not using:
```
"The server couldn't fulfill the request."
```

scripts/encode2ga4gh.py
```diff
@@ -0,0 +1,296 @@
...
+    print("subset size:    {}".format(Color.blue(subset)))
+    print("output folder:  {}".format(Color.blue(output_folder)))
+    get_data_from_host(url, headers, host, data_type, output_folder,
+                       options.description, options.annotationId, subset)
+
+    log("DONE")
+
+
+if __name__ == '__main__':
+    main(sys.argv)
```
dcolligan: a lot of snake_case crept into this file... please change all identifiers to camelCase

scripts/kallisto2ga.py
```diff
@@ -0,0 +1,135 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import os
+import sys
+import errno
+
+
+def makeDir(path):
```
dcolligan: if you find yourself using the same methods more than once, consider putting them in `scripts/utils.py`

scripts/kallisto2ga.py
```diff
@@ -0,0 +1,135 @@
...
+        # output table
+        rnaSeqTable = os.path.join(outputFolder, analysisId, "rnaseq.table")
+
+        # write rnaseq table
+        with open(rnaSeqTable, "w") as rnaQuantFile:
+            writeRNAQuant(rnaQuantFile, analysisId, description, annotationId)
+
+
+def writeExpressionTables(data, annotationId, outputFolder):
+    for (analysisId, quantfile) in data:
```
dcolligan: you shouldn't need the parentheses here

dcolligan:
I've finished a pass through, but obviously this is a mammoth PR (manipulating a lot of formats / objects I don't quite understand) and it's hard to catch everything.

That being said, I would recommend running flake8 on this code (this is included in the `scripts/run_tests.py` run) since I saw a lot of things in there that flake should catch.

Looks reasonable on a first pass.

ga4gh/datamodel/ontologies.py
```diff
@@ -0,0 +1,82 @@
...
+class OntologyTermSet(object):
+    """A set of related ontology terms.
+    """
+    def __init__(self, ontologySource):
+        self.ontologySource = ontologySource
+        self.byId = dict()
+        self.byName = dict()
+
+    def add(self, ontologyTerm):
+        "add an ontology term to the object"
```
jeromekelleher: docstrings need to be triple quotes
```python
"""
Add an ontotology term to the object.
"""
```

ga4gh/datamodel/ontologies.py
```diff
@@ -0,0 +1,82 @@
...
+from __future__ import unicode_literals
+
+import ga4gh.protocol as protocol
+
+
+class OntologyTermSet(object):
+    """A set of related ontology terms.
+    """
+    def __init__(self, ontologySource):
+        self.ontologySource = ontologySource
```
jeromekelleher: Instance variables should be private by default:
```
self._ontologySource = ontologySource
```

ga4gh/datamodel/ontologies.py
```diff
@@ -0,0 +1,82 @@
...
+        self.byId = dict()
```
jeromekelleher: It would be nice to follow the conventions used elsewhere for these maps. Something like ``self._ontologyTermIdMap`` (or just ``self._idMap`` if this is too verbose) would be good.

ga4gh/datamodel/ontologies.py
```diff
@@ -0,0 +1,82 @@
...
+to support BRCA changeling.
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import ga4gh.protocol as protocol
+
+
+class OntologyTermSet(object):
```
jeromekelleher: We should inherit from ``datamodel.DatamodelObject`` here, That way, we'll inherit a lot of useful functionality, particularly dealing with compound ids.

ga4gh/datamodel/ontologies.py
```diff
@@ -0,0 +1,82 @@
...
+                raise ValueError(
+                    "OntologyTerm idan already exists: "
+                    + str(ontologyTerm))
+
+    def create(self, id, name):
+        "create a new ontology term"
+        self.add(OntologyTerm(self.ontologySource, id, name))
+
+
+class OntologyTerm(object):
```
jeromekelleher: We should inherit from ``datamodel.DatamodelObject`` here too probably.

ga4gh/datamodel/ontologies.py
```diff
@@ -0,0 +1,82 @@
...
+        gaOntologyTerm.id = self.__id
+        gaOntologyTerm.name = self.__name
+        return gaOntologyTerm
+
+
+class SequenceOntologyTermSet(OntologyTermSet):
+    """TODO: tmp hack to encode the sequence ontology terms used by the BRCA
+    gene sets"""
+    def __init__(self):
+        OntologyTermSet.__init__(self, "http://www.sequenceontology.org/")
```
jeromekelleher: Use ``super(SequenceOntologyTermSet, self).__init__("http://www.sequenceontology.org/")``

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def getCharacterization(self, rnaQuantificationId):
+        """
+        input is tab file with no header.  Columns are:
+        analysisId, complexity, exonicFraction, fractionMapped,
+        intergenicFraction, intronicFraction
+        """
+        characterizationData = open(self._characterizationFile, "r")
+        quantCharacterization = characterizationData.readline()
+        fields = quantCharacterization.split('/t')
```
jeromekelleher: Shouldn't this be `\t`? I think it's usually better to just take the default `split()` on whitespace rather than insisting on tabs anyway.

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,273 @@
...
+
+        return readCharacterization
+
+    def getCharacterization(self, rnaQuantificationId):
+        """
+        input is tab file with no header.  Columns are:
+        analysisId, complexity, exonicFraction, fractionMapped,
+        intergenicFraction, intronicFraction
+        """
+        characterizationData = open(self._characterizationFile, "r")
```
jeromekelleher: Use a context manager when opening files like this:
```python
with open(self._characterizationFile, "r") as characterizationData:
    # Stuff
```

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def convertCharacterization(self, record):
+        readCharacterization = protocol.Characterization
+        readCharacterization.analysisId = record[0]
+        readCharacterization.complexity = float(record[1])
+        readCharacterization.exonicFraction = float(record[2])
+        readCharacterization.fractionMapped = float(record[3])
+        readCharacterization.intergenicFraction = float(record[4])
+        readCharacterization.intronicFraction = float(record[5])
+
```
jeromekelleher: I don't think this blank like before the return adds any readability --- I would suggest removing it.

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,273 @@
...
+        return readCount
+
+    def getReadCounts(self, rnaQuantificationId):
+        """
+        input is tab file with no header.  Columns are:
+        analysisId, multiCount, multiSpliceCount, totalReadCount, uniqueCount,
+        uniqueSpliceCount
+        """
+        readCountData = open(self._readCountFile, "r")
```
jeromekelleher: Ditto on the with block here.

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,273 @@
...
+
+        return rnaQuantification
+
+    def getRnaQuantification(self, rnaQuantificationId):
+        """
+        input is tab file with no header.  Columns are:
+        Id, annotations, description, name, readGroupId
+        where annotation is a comma separated list
+        """
+        rnaQuantificationData = open(self._rnaQuantificationFile, "r")
```
jeromekelleher: And here.

ga4gh/gff3.py
```diff
@@ -0,0 +1,223 @@
+"""
+Object representation of GFF3 data and parser for GFF3 files.
```
jeromekelleher: High level question here: do we need to implement our own GFF parser? Are there no GFF3 Python packages we can leverage? This is a lot of code we'll need to test pretty hard, so it would be nice if someone else had done it already...

scripts/encode2ga4gh.py
jeromekelleher: Ditto this.

scripts/encode2ga4gh.py
```diff
@@ -0,0 +1,296 @@
+#!/usr/bin/env python
```
jeromekelleher: What does this script do? A header comment would be good.

scripts/encode2ga4gh.py
jeromekelleher: Aren't there Python packages to do this for us?

scripts/encode2ga4gh.py
```diff
@@ -0,0 +1,296 @@
...
+def main(argv):
+
+    usage = "Usage: {} <data-folder>".format(argv[0])
+    if len(argv) != 2:
+        print(Color.red(usage))
+        sys.exit(1)
+
+    parser = makeParser(usage)
+    (options, args) = parser.parse_args(argv[1:])
+
```
jeromekelleher: There's a lot of blank lines here... I tend to agree with the PEP8 point of view

> Use blank lines in functions, sparingly, to indicate logical sections.

scripts/kallisto2ga.py
```diff
@@ -0,0 +1,135 @@
...
+        os.makedirs(path)
+    except OSError as exception:
+        if exception.errno != errno.EEXIST:
+            raise
+
+
+# TODO: placeholder values need to be calculated then removed
+def getCount(expressionId):
+    rawCount = 0
+    return "%d" % rawCount
```
jeromekelleher: We tend to use the formatting mini-language in favour of string interpolation. This would be 
```python
return "{}".format(rawCount)
```
It's a bit more flexible, and consistency is nice.

ga4gh/gff3.py
diekhans: I wrote the original version of this parser and couldn't find one. However now
I see:

      https://pypi.python.org/pypi/gff3/0.3.0

don't know how I missed it.  Worth taking a look at.

diekhans: Then again, https://pypi.python.org/pypi/gff3/0.3.0 has a single, >300 line
deeply nested function that implements all the parsing.  I am not sure I would
want to look at it any more.

jeromekelleher:
Thanks for this @saupchurch, it looks great! It's a huge lump of code, so I've only gone through it in a very shallow way. My initially impression is that it's quite close to being ready, and that it's obviously the result of a huge amount of work. 

It would be nice to see more systematic testing at the API level, by adding some tests to the `test_simulated_stack.py` file. We probably need some HTTP level testing in the `test_views.py` module also.

I'm slightly worried about the GFF parser. I'd much rather this was someone else's problem. What about 
[this](https://pypi.python.org/pypi/bcbio-gff/0.6.2) or [this](https://pypi.python.org/pypi/gff3/0.2.0)? There's a few other places where it feels like we're reinventing wheels, so it would be good to stand back and see what PyPI has to offer.

Other than that, I think you're clearly on the right track!

diekhans:
bcbio-gff is dependent on BioPython, which is a monster; I would
suggest not creating it as a dependency.

diekhans:
also, the tests in gff3 0.3.0 are empty. BCBio has no tests.

sometimes rolling your own leads to a better smoke

saupchurch:
Thank you for the review and comment @dcolligan and @jeromekelleher - will be working through the comments and get new version pushed shortly.

We were looking at gff3 earlier, but didn't have strong opinions other than "maybe we should use this".  Not encouraged by the report of empty tests in gff3.  Since parsing gff seems to be a more widespread issue, I'm getting the feeling that this might be a larger discussion.  Will think on this and see if there is a good way to present the issue to get a good discussion going.

saupchurch:
Updated push resolves many of the issues pointed out.  Still a work in progress.

saupchurch:
closing to switch updates to ran branch

PR #787, opened 2015-11-17
Variant annotaions

pcingola:
This PR was created mainly to be discussed in today's variant annotation call.
We are missing test cases and commits have not been squashed.
A proper PR will follow.

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -199,7 +199,7 @@ def configure(configFile=None, baseConfig="ProductionConfig",
         theBackend = backend.EmptyBackend()
     elif dataSource.scheme == "file":
         theBackend = backend.FileSystemBackend(os.path.join(
-            dataSource.netLoc, dataSource.path))
+            dataSource.netloc, dataSource.path))
     else:
         raise exceptions.ConfigurationException(
             "Unsupported data source scheme: " + dataSource.scheme)


jeromekelleher:
I've had a quick scan through this @pcingola --- it's looks great. I spotted a couple of minor issues, but otherwise I think it's in a really good state.

ga4gh/backend.py
```diff
@@ -208,6 +265,7 @@ def __init__(self):
         self._referenceSetIdMap = {}
         self._referenceSetNameMap = {}
         self._referenceSetIds = []
+        self._ontologies = {}
```
jeromekelleher: We should probably have _ontologyIdMap, _ontologyNameMap and _ontologyIds here for consistency.

ga4gh/backend.py
```diff
@@ -218,6 +276,12 @@ def addDataset(self, dataset):
         self._datasetNameMap[dataset.getLocalId()] = dataset
         self._datasetIds.append(id_)
 
+    def addOntologies(self, ontologies):
+        """
+        Adds ontologies to this backend.
```
jeromekelleher: Hmm, maybe I'm misunderstanding here though. What's an ontology? It would be nice to give a little bit of detail about what the parameter is here.

ga4gh/backend.py
```diff
@@ -192,6 +192,63 @@ def _getEnd(cls, variant):
...
 
+class VariantAnnotationsIntervalIterator(IntervalIterator):
+    """
+    An interval iterator for annotations
+    """
+
+    def __init__(self, request, parentContainer):
+        super(VariantAnnotationsIntervalIterator, self).__init__(
+            request, parentContainer)
+        self.featureIds = set(self._request.feature_ids)
```
jeromekelleher: We should have private variables here: ``self._featureIds`` and ``self._effects``.

ga4gh/backend.py
```diff
@@ -192,6 +192,63 @@ def _getEnd(cls, variant):
...
+    def next(self):
+        while True:
+            ret = super(VariantAnnotationsIntervalIterator, self).next()
+            vann = ret[0]
+            if self.filterVariantAnnotation(vann):
+                return ret
+        return None
+
+    def filterVariantAnnotation(self, vann):
+        if not(self.featureIds or self.effects):
```
jeromekelleher: I'm pretty critical of using implicit casting to boolean, as I think it's nearly always harder to read and can contain tricky exceptional cases that are hard to spot. PEP8 recommends against it also. It would be helpful to rephrase these conditionals in terms of explicit comparisons so it's clearer. For example, I would change the first to
```python
if len(self._featureIds) == 0 and len(self._effects) == 0:
    return True
```

ga4gh/cli.py
```diff
@@ -547,18 +632,44 @@ def addVariantSearchOptions(parser):
...
+
 def addReferenceNameArgument(parser):
     parser.add_argument(
         "--referenceName", "-r", default="1",
         help="Only return variants on this reference.")
 
 
+def addReferenceIdArgument(parser):
+    parser.add_argument(
+        "--referenceId", "-rid", default="",
```
jeromekelleher: Short form arguments must  be a single letter for POSIX --- just pick an unused one.

ga4gh/cli.py
```diff
@@ -568,6 +679,22 @@ def addCallSetIdsArgument(parser):
             """)
 
 
+def addFeatureIdsArgument(parser):
+    parser.add_argument(
+        "--featureIds", "-fids", default=[],
```
jeromekelleher: Same here, and below.

ga4gh/datamodel/datasets.py
```diff
@@ -190,3 +237,70 @@ def _setMetadata(self):
                 except KeyError as err:
                     raise exceptions.MissingDatasetMetadataException(
                         metadataFileName, str(err))
+
+
+class AbstractOntology:
```
jeromekelleher: should extend object here for consistency.

ga4gh/datamodel/datasets.py
```diff
@@ -190,3 +237,70 @@ def _setMetadata(self):
...
+    """
+    The base class of storing an ontology
+    """
+
+    def __init__(self):
+        self.nameById = dict()
```
jeromekelleher: These should be private variables. We should follow the usual conventions for these maps too, _[something]IdMap and _[something]NameMap. I guess [something] can be empty here?

ga4gh/datamodel/datasets.py
```diff
@@ -190,3 +237,70 @@ def _setMetadata(self):
...
+        return self.nameById[id_]
+
+
+class AbstractOntologies(object):
+    """
+    The base class of storing ontologies
+    """
+
+    def __init__(self):
+        self.ontologyByName = dict()
```
jeromekelleher: Same comments about variable naming here.

ga4gh/datamodel/datasets.py
```diff
@@ -190,3 +237,70 @@ def _setMetadata(self):
...
+        self.idByName[name] = id_
+
+    def getId(self, name):
+        return self.idByName[name]
+
+    def getName(self, id_):
+        return self.nameById[id_]
+
+
+class AbstractOntologies(object):
```
jeromekelleher: How does this relate to the AbstractOntology class above? One of these should probably be deleted.

ga4gh/datamodel/datasets.py
```diff
@@ -190,3 +237,70 @@ def _setMetadata(self):
...
+        # Find TXT files
+        for filename in os.listdir(dataDir):
+            if fnmatch.fnmatch(filename, '*.txt'):
+                ontologyName, _ = os.path.splitext(filename)
+                path = os.path.join(dataDir, filename)
+                ontology = self.readOntology(path)
+                self.add(ontologyName, ontology)
+
+    def readOntology(self, filename):
+        ontology = AbstractOntology()
```
jeromekelleher: OK, I see what's happening now. I'm confused about the plurality here though. Is one 'ontology' a single line in the text file, or is the the whole file? If one line is an ontology, then we should consider renaming the class to OntologySet, and trying to follow the design for the other XSet objects. If not, we should use OntologyTerm a bit more prominently and be clear that an ontology maps to set of ontology terms.

ga4gh/datamodel/variants.py
```diff
@@ -547,3 +596,162 @@ def buildMetadata(
...
+    Class representing a single variant annotation set backed by a directory of
+    indexed VCF or BCF files.
+    """
+    def __init__(self, parentContainer, localId, dataDir, backend):
+        super(HtslibVariantAnnotationSet, self).__init__(
+            parentContainer, localId, dataDir, backend)
+        self.compoundIdClass = datamodel.VariantAnnotationSetCompoundId
+        self._compoundId = datamodel.VariantAnnotationSetCompoundId(
+            self.getCompoundId(), 'variantannotations')
+        self._sequenceOntology = backend.getOntologies() \
```
jeromekelleher: We prefer the following line breaking:
```python
self._sequenceOntology = backend.getOntologies().get(
      'sequence_ontology')
```

ga4gh/datamodel/variants.py
```diff
@@ -547,3 +596,162 @@ def buildMetadata(
...
+                allLoc.referenceSequence = match.group(2)
+                allLoc.alternateSequence = match.group(3)
+                return allLoc
+        return None
+
+    def convertLocationHgvsP(self, hgvsp):
+        if hgvsp == '':
+            return None
+        match = re.match("p.(\D+)(\d+)(\D+)", hgvsp)
+        if match:
```
jeromekelleher: Again, implicit conversion to boolean is bad for readability. Would prefer
```python
if match is not None
```

ga4gh/datamodel/variants.py
```diff
@@ -547,3 +596,162 @@ def buildMetadata(
...
+            referenceName, startPosition, endPosition = \
+                self.sanitizeVariantFileFetch(
+                    referenceName, startPosition, endPosition)
+            cursor = self.getFileHandle(varFileName).fetch(
+                referenceName, startPosition, endPosition)
+            for record in cursor:
+                yield self.convertVariantAnnotation(record)
+
+    def convertLocation(self, pos):
+        if pos == '':
```
jeromekelleher: I worry a bit about comparing equality with the empty string. There's a potential Python2/Python3 issue here with bytes/str.  Perhaps ``if len(pos) == 0``? I'm not sure what the right thing to do is here, so feel free to ignore this point.

ga4gh/datamodel/variants.py
```diff
@@ -547,3 +596,162 @@ def buildMetadata(
...
+        """
+        md5 = self.hashVariant(gaVariant)
+        compoundId = datamodel.VariantAnnotationCompoundId(
+            self.getCompoundId(), gaVariant.referenceName,
+            gaVariant.start, md5)
+        return str(compoundId)
+
+    def getTranscriptEffectId(self, gaTranscriptEffect):
+        effs = [eff.name for eff in gaTranscriptEffect.effects]
+        return hashlib.md5(
```
jeromekelleher: Use the ``.format()`` method here rather than string concatenation. It's a bit more efficient and more robust, and we're using it everywhere else for string building.

ga4gh/datamodel/variants.py
```diff
@@ -547,3 +596,162 @@ def buildMetadata(
...
+            ).hexdigest()
+
+    def hashVariantAnnotation(cls, gaVariant, gaVariantAnnotation):
+        """
+        Produces an MD5 hash of the ga variant and gaVariantAnnotaiton objects
+        """
+        treffs = [treff.id for treff in gaVariantAnnotation.transcriptEffects]
+        return hashlib.md5(
+            gaVariant.referenceBases +
+            str(tuple(gaVariant.alternateBases)) +
```
jeromekelleher: Ditto here for .format. It saves you the trouble of calling ``str``, as it automatically converts to string.

pcingola:
Hi Jerome,
  Thank you for your feedback! I've updated the relevant lines.

A word about Ontology / Ontologies. For the time being, an "ontology" is a simple 
map between ontology_ID and ontology_Term (i.e. 'name').  So, at the moment 
an "ontology" is a simple map between 'ids' and 'names' which is stored in a 
simple TXT file ("id \t name \n").

Obviously this is far form ideal: i) it should be extended into objects that really capture 
the relationships between ontology terms; and ii) should be backed by a an OBO 
file. I can try to implement all these in the future, but at the moment I just implemented 
the "minimum viable object" needed for variant annotations.

diekhans:
there is an OntologyTerm in

   https://github.com/ga4gh/schemas/blob/metadata/src/main/resources/avro/metadata.avdl

Mark

pcingola:
Yes Mark, we can implement a similar data structure later.

jeromekelleher:
Thanks @pcingola. @dcolligan, could you take a quick look through this and see what you think please?

dcolligan:
Yeah, I agree this code looks really clean and well-factored.  

This code isn't actually being proposed to be merged into `develop`, though, right?  As the initial comment says, is there going to be another PR targeting a GA4GH server branch?  (since there is no concept of ontology or annotation in the schema yet, and presumably will not be until after a release)

pcingola:
Yes, the idea is to merge it into develop once you approve the code. Obviously 
the schemas for variant annotations should be merges as well. It would 
be nice to be able to release with the variant annotations functionality.
I hope this makes sense.

jeromekelleher:
I don't think we can merge the code into develop until the schema changes have been merged into master upstream @pcingola. My understanding is that we have three coordinated branches on the schemas, server and compliance tests, and changes to the schemas will only be accepted once there's a working implementation and tests in the branches on the server and compliance repos. I think this process is described at https://github.com/ga4gh/schemas/blob/master/CONTRIBUTING.md

So, the short version is, we can't accept pull requests to develop until there is a corresponding schema in the schemas/master branch. We also can't release a version of the server until the corresponding schemas have been released. However, we can make a topic branch here on the server, and I think your pull request would be accepted on to that branch quickly. 

Is there already a variant annotations branch? If not we'll make one (it should have the same name as the corresponding branch on the schemas repo).

pcingola:
The schema was accepted and merged until someone decided to split
it again. So having it merged back in should not be a problem
since it has already been reviewed and accepted. How can we
re-merge it and ensure nobody else decides to split it back?

The server is working, and you reviewed the code, so we should
not have any problems on that end either. Yes, there is corresponding
topic branch.

We are only missing the compliance test. I'll try to start working on
them this week.

jeromekelleher:
@benedictpaten, can you point towards the official docs on the schema development process please?

@pcingola, I'm happy to make the topic branch here for this PR. What should I call it?

diekhans:
Don't worry Pablo, it was not an arbitrary decision to split out
the work.  GA4GH DWG decide to do this to switch to the working
master model.  It was a decision made after a lot of discussion.

UCSC can provide guidance and assistance on the compliance
suite.  Just yell.  Maciej is your man.

Mark

PR #788, opened 2015-11-17
Add virtualenv to dependencies

dcolligan:
Needed for running OIDC tests, if nothing else

### DIFF
--- a/docs/demo.rst
+++ b/docs/demo.rst
@@ -27,13 +27,13 @@ packages using:
 
 .. code-block:: bash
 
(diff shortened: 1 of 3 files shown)


dcolligan:
Er, don't merge this yet... there's something screwy here.  On ubuntu, virtualenv and python-virtualenv are two different packages.  Not sure what the difference is.

dcolligan:
AFAICT, python-virtualenv should install the virtualenv program, so the problem in the OIDC test is some other config issue... closing...

Issue #783, opened 2015-11-13
testSamConversion fails on ubuntu 14.04 and ubuntu 15.05


IanRogers-LShift:
Running nosetests on latest server develop branch (commit  58b228bd8de761d110de897e50e36447553ba748) on ubuntu 14.04 fails with this error:

FAIL: testSamConversion (tests.unit.test_converters.TestSamConverter)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ian/workspace/ga4gh/ga4gh/server/tests/unit/test_converters.py", line 75, in testSamConversion
    self.verifyFullConversion(readGroupSet, readGroup, reference)
  File "/home/ian/workspace/ga4gh/ga4gh/server/tests/unit/test_converters.py", line 66, in verifyFullConversion
    self.verifySamRecordsEqual(sourceReads, convertedReads)
  File "/home/ian/workspace/ga4gh/ga4gh/server/tests/unit/test_converters.py", line 35, in verifySamRecordsEqual
    self.assertEqual(source.flag, converted.flag)
AssertionError: 73 != 65

Also confirmed by someone else to fail in ubuntu 15.04

Environment details (created by following http://ga4gh-reference-implementation.readthedocs.org/en/stable/demo.html )  :

pysam-0.8.3-py2.7 

$ uname -a
Linux mrchatterbox 3.13.0-66-generic #108-Ubuntu SMP Wed Oct 7 15:20:27 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
$ which python
/home/ian/workspace/ga4gh/ga4gh/server/ga4gh-env/bin/python
$ python --version
Python 2.7.6

jeromekelleher:
Thanks @IanRogers-LShift --- also confirmed on a fresh Ubuntu 15.10 installation.

dcolligan:
Could not reproduce on Ubuntu 15.4:
```
(cur)dannyc@dannyc-VirtualBox:~/ga4gh/server$ uname -a
Linux dannyc-VirtualBox 3.19.0-15-generic #15-Ubuntu SMP Thu Apr 16 23:32:37 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
(cur)dannyc@dannyc-VirtualBox:~/ga4gh/server$ which python
/home/dannyc/ga4gh/cur/bin/python
(cur)dannyc@dannyc-VirtualBox:~/ga4gh/server$ python --version
Python 2.7.9
[...]

PR #789, opened 2015-11-18
WIP

dcolligan:
TODO:
- fix iterator hack
- how to indicate the mate read is unmapped in GA4GH ReadAlignment (as
    opposed to not present?)

Issue #783

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -164,7 +164,12 @@ class ReadsIntervalIterator(IntervalIterator):
 
     @classmethod
     def _getStart(cls, readAlignment):
-        return readAlignment.alignment.position.position
+        if readAlignment.alignment is None:
+            # unmapped read with mapped mate; see SAM standard 2.4.1
+            return readAlignment.nextMatePosition.position
+        else:
+            # usual case
+            return readAlignment.alignment.position.position
 
     @classmethod
     def _getEnd(cls, readAlignment):
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -66,15 +66,17 @@ class SamFlags(object):
     """
     Utility class for working with SAM flags
     """
-    NUMBER_READS = 0x1
-    PROPER_PLACEMENT = 0x2
-    REVERSED = 0x10
-    NEXT_MATE_REVERSED = 0x20
-    READ_NUMBER_ONE = 0x40
-    READ_NUMBER_TWO = 0x80
+    READ_PAIRED = 0x1
+    READ_PROPER_PAIR = 0x2
+    READ_UNMAPPED = 0x4
+    MATE_UNMAPPED = 0x8
+    READ_REVERSE_STRAND = 0x10
+    MATE_REVERSE_STRAND = 0x20
+    FIRST_IN_PAIR = 0x40
+    SECOND_IN_PAIR = 0x80
     SECONDARY_ALIGNMENT = 0x100
-    FAILED_VENDOR_QUALITY_CHECKS = 0x200
-    DUPLICATE_FRAGMENT = 0x400
+    FAILED_QUALITY_CHECK = 0x200
+    DUPLICATE_READ = 0x400
     SUPPLEMENTARY_ALIGNMENT = 0x800
 
     @staticmethod
@@ -574,6 +576,8 @@ class HtslibReadGroup(datamodel.PysamDatamodelMixin, AbstractReadGroup):
         """
         Convert a pysam ReadAlignment to a GA4GH ReadAlignment
         """
+        samFile = self._parentContainer.getFileHandle(
+            self._parentSamFilePath)
         # TODO fill out remaining fields
         # TODO refine in tandem with code in converters module
         ret = protocol.ReadAlignment()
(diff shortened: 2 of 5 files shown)


ga4gh/backend.py
```diff
@@ -164,13 +164,19 @@ def _search(self, start, end):
 
     @classmethod
     def _getStart(cls, readAlignment):
-        return readAlignment.alignment.position.position
+        if readAlignment.alignment is None:
+            return -2
```
dcolligan: Changes in this file are hacks for now.  The reason we need this is because `getReadAlignments` is returning unmapped reads and we can't use the interval iterator algorithm on anything that doesn't have `alignment` objects.

Should we filter out all unmapped reads within `getReadAlignments` or, alternatively, pass a parameter to `getReadAlignments` to indicate that we only want mapped reads?  Note that in this case it is returning unmapped reads even though we're passing in a non-null reference.

dcolligan:
Fundamentally the failing test cause stems from the fact that we weren't setting read flags 0x4 and 0x8 when writing reads to a SAM file.  But this brings up other related things...

ga4gh/datamodel/reads.py
```diff
@@ -68,6 +68,8 @@ class SamFlags(object):
     """
     NUMBER_READS = 0x1
     PROPER_PLACEMENT = 0x2
+    UNMAPPED = 0x4
+    NEXT_SEGMENT_UNMAPPED = 0x8
```
dcolligan: Here's a question: is there a way in the GA4GH schema for a ReadAlignment object to say that its read mate is present, but unmapped?  The `nextMatePosition` attribute seems to be the most likely candidate, but I take a null on that to mean that the read mate is not present.  If there is not a way in the schema to distinguish these two states from each other, then we need to file a schema issue since there's no way to fix the failing test at present.

ga4gh/backend.py
jmarshall: If this is a coordinate-sorted input file, these interspersed ones are almost certainly unmapped reads that have been sorted next to their mates.  See [SAMv1.pdf](http://samtools.github.io/hts-specs/SAMv1.pdf) §2 4.1.  (BTW both-ends-unmapped read pairs (§2 4.2) appear at the _end_ of a sorted file.)

ga4gh/datamodel/reads.py
dcolligan: @jeromekelleher your thoughts on this (and this PR, generally)?

jeromekelleher: I'm out of my depth here I'm afraid @dcolligan, I have nothing of any value to add.

The thing I can suggest is to ping @calbach, he's knows the reads API as well as anyone, and might be able to point us at what Google does in this case.

calbach: Yes, you can distinguish these two states:

`nextMatePosition` gives the mapped position of the next read. null is equivalent to `NEXT_SEGMENT_UNMAPPED`, it does not imply no mate pair.
To determine whether you have a mate you can check `numberReads`, which describes how many reads are in the fragment.

At the bottom of this [migration guide](https://github.com/googlegenomics/readthedocs/blob/master/docs/source/migrating_tips.rst), we have some pseudocode which describes a mapping between SAM flags and GA4GH ReadAlignments.

dcolligan: @calbach Thanks, that's very helpful.

One question I have, however, is about `readNumber`.  The SAM specification says (in 1.4.2): "If 0x40 and 0x80 are both set, the read is part of a linear template, but it is neither the first nor
the last read. If both 0x40 and 0x80 are unset, the index of the read in the template is unknown.
This may happen for a non-linear template or the index is lost in data processing."

So it seems like 0x40 and 0x80 either both being set or both not being set are valid states.  However, `readNumber` seems to capture only the two other states, namely when one and only one of either 0x40 or 0x80 is set.  Is there a way in the GA4GH spec right now to capture the both-set or both-unset states?

jkbonfield:
Unmapped mates will often still have a position.  Consider the following scenario in BAM.

Read X/1 and X/2 being 1st and 2nd read in the template.  X/1 maps, X/2 does not.
We may therefore write X/1 as mapped to chr10 pos 1000 and X/2 as unmapped, but also at chr10 pos 1000.

The reason for this is that if we wish to locally pull out all the mates in a region in order to perform a local reassembly, for example, then we need to easily fetch out the reads that didn't map but should apparently do so in that vicinity.  The likely reason for not mapping would be a large insertion.

IMO therefore, using the position field as an indicator for being unmapped doesn't work well.  Instead it should be kept in the flags.

vadimzalunin:
I don't think assumptions can be made about presence of a mate even if mate info claims it is mapped.

ga4gh/datamodel/reads.py
calbach: 0x4 and 0x8 both set: this is equivalent to `readNumber` being a value other than 1 or 0 - the ga4gh spec is just being more specific about which index it is. The spec is more expressive/specific than the BAM spec here, so it may be difficult to do a direct conversion from a SAMRecord without actually traversing mates.

0x4 and 0x8 both unset: I would interpret this as translating to a `null` `readNumber`, though we should include this in the documentation.

dcolligan: @calbach another question(s):

In the migration guide, the attributes `read.alignment.position.position` and `read.nextMatePosition.position` are tested for a null value.  However, the avro `Position` record does not allow null values for the `position` attribute.  Is this an error in the schema, or an error in the migration guide?  If the former, 1) should a read's `position` and `nextMatePosition` not be allowed to be null and 2) for either attribute if the `position` attribute is null, what should the `Position`'s `referenceName` and `strand` be set to?

calbach: Yeah, that's a mistake in that guide. It should be checking:
`read.alignment` and `read.nextMatePosition`

I'll fix that, thanks.

dcolligan:

I think the reads round-trip code is correct now as far is at can be.  But we can't really do a round trip since when we read reads we are losing some information from the BAM if the read is unmapped or its mate is unmapped -- that is, we set either `read.alignment` or `read.nextMatePosition` to None, and in order to write that read to another BAM as it was in the original BAM, we require information that is not set on the GA object since those would be fields on what is a null `Position` object.

ga4gh/backend.py
dcolligan: Thanks for the tip, @jmarshall

jeromekelleher:
Looks great to me @dcolligan, thanks. +1.

There are a couple of things it would be good to follow up on here:

1. You mentioned above that we can't roundtrip due to a limitation in the schema. Is this a currently known limitation? If it hasn't been discussed in https://github.com/ga4gh/schemas/issues/505, we should follow up there with a description of the problem. (I guess even if it has, we should update that thread with the issues we're having roundtripping.)

2. It would be good to get some expert review on these changes. @jmarshall/@jkbonfield, would you have a few spare minutes to scan over this code please? It would be very useful to get your view on this, particularly if we're up against schema limitations.

jkbonfield:
I haven't used any of the code in anger, so I'm not really qualified for code review on it.  I simply looked over the schema with an eye to spotting any potential round trip errors.  But again, I didn't actually test it (although I did try, and failed to navigate the maze of twisty web passages all alike before I ran out of time).

calbach:
The limitation being hit is that the GA4GH schema does not allow unmapped reads to have a position, which IMO is a hack in BAM to get reads to come back in a desired order. It is a problem which could be solved more explicitly through the API (though admittedly we haven't done that). This is noted in ga4gh/schemas#505 you linked to:

> Data isn't just mapped or unmapped; it can be "placed" but unaligned. (Chr/Pos but CIGAR *) This is commonly done where one read in a pair is mapped and the other is not. I believe the GA4GH format permits this, but perhaps needs spelling out explicitly. Is this simply achieved by having ALinearAlignment with a GAPosition and an empty GACigarUnit array?

dcolligan:
Regardless of what we decide to do with implementation or schema changes in the future to be able to do round trips, we shouldn't let that stand in the way of merging this.  It fixes a system-dependent test failure and also brings our read code more in line with the spec(s).  I imagine I'll be filing at least one schema issue after it's merged (probably just to update docs)

jeromekelleher:
@dcolligan, agreed, merging the PR.

We should follow up on the above comments though so the experience we've had here isn't lost.

Issue #792, opened 2015-12-01
CI builds failing due to OIC error 


jeromekelleher:
Investigating..

jeromekelleher:
fixed in #793.

PR #793, opened 2015-12-01
Remove virtualenv for simple_op server.

jeromekelleher:
This should fix the current CI problems of issue #792.

We had an embedded virtualenv to run the simple_op provider, which started to cause trouble on travis's setup. This was always going to cause issues in any case as it meant we couldn't run the tests from a virtualenv (I don't think venv's nest very well).

Let's see if this fixes the problem anyway.

### DIFF
--- a/oidc-provider/run.sh
+++ b/oidc-provider/run.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-./setupenv.sh
-. bin/activate
 HOSTNAME=`python -c 'import socket; print socket.gethostname()'`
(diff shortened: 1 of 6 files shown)


dcolligan:
+1

I've also encountered hiccups with the nested virtualenv setup in the past; it's a good idea to remove it.

jeromekelleher:
OK, cool. I think we can merge this one quickly as the builds are currently broken. We'll need to update the `variation_annotation` branch to include this change so that PR's made against it will build too.

jeromekelleher:
@dcolligan merge me!

PR #796, opened 2015-12-04
bring rna branch up to date

saupchurch:
PR to bring server/rna up to date with latest development version of rna api code

### DIFF
--- a/ga4gh/_protocol_definitions.py
+++ b/ga4gh/_protocol_definitions.py
@@ -11,28 +11,296 @@ from protocol import SearchResponse
 
 import avro.schema
 
-version = '0.6.4568e6f6'
+version = '0.6'
 
 
-class Call(ProtocolElement):
+class Allele(ProtocolElement):
     """
-    A Call represents the determination of genotype with respect to a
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Looks good @saupchurch. Perhaps we should just bring the rna branch up to date with develop separately, and keep these changes out of the PR? @dcolligan, can you check to see if the rna branch can be fast-forwarded, and if so, make the changes please?

dcolligan:
Just merged develop into rna

saupchurch:
updated _protocol_definitions.py using process_schema.py build should now pass

tests/end_to_end/test_oidc.py
```diff
@@ -8,9 +8,10 @@
 
 import requests
 import subprocess
-from lxml import html
 from urlparse import urlparse
 
+import lxml.html as html
+
```
macieksmuga: Is there a new python style guideline I'm missing? The `from <module> import <submodule>` seems much less stilted than `import <module>.<submodule> as <submodule>`, with exactly the same end result.

PR #798, opened 2015-12-07
Adding variant annotations

pcingola:
Fixed some issues in process_schemas.py: i) missing comment caused script to crash; ii) switching 'null' in union order also crashed

Variant annotations working (first draft)

Protocol definitions updated

Updating _protocol_definitions.py

Refactoring in schema. Added cds/protein position (basic HGVS parsing). Added Sequence Ontology

Minor changes

Minor changes

Added 'ontologies' directory in test data set

Corrected _protocol_definitions.py. All test cases passing

Added search options

Test cases passing

Adding Jerome's suggestions / code review


pcingola:
Every time I squash something odd happens, so let me know if one of your suggestions is not merged in.
At least all tests are passing OK.

ga4gh/datamodel/datasets.py
```diff
@@ -190,3 +237,69 @@ def _setMetadata(self):
...
+        self._idNameMap[name] = id_
+
+    def getId(self, name):
+        return self._idNameMap[name]
+
+    def getName(self, id_):
+        return self._nameIdMap[id_]
+
+
+class FileSystemOntology(object):
```
jeromekelleher: I'm still a bit confused by this `Ontology` hierarchy here, and I think we might be able to simplify it a bit. As I understand it, currently an Ontology is a one-to-one mapping between OTs and names and corresponds to one text file, e.g, `sequence_ontology.txt`. Why don't we just have ``datadir/ontologies/sequence_ontology.txt`` instead of having the intermediate directory?

Also, I'm confused by the distinction between the FileSystem ontology (which seems to be a mapping from OntologyName -> Ontology objects) and the Ontology itself. I think the FileSystemOntology object should map directly to the file full of terms, and we remove the AbstractOntology class (which isn't really abstract, as we instantiate it directly).

I'll comment a few other places where I think we can clarify this.

ga4gh/backend.py
```diff
@@ -218,6 +278,14 @@ def addDataset(self, dataset):
         self._datasetNameMap[dataset.getLocalId()] = dataset
         self._datasetIds.append(id_)
 
+    def addOntologies(self, ontologies):
```
jeromekelleher: I think this should be a singlar `addOntology`, this will simplify things and make the Ontology setup the same as Datasets, ReferenceSets etc.

ga4gh/backend.py
```diff
@@ -811,10 +927,12 @@ class FileSystemBackend(AbstractBackend):
     def __init__(self, dataDir):
         super(FileSystemBackend, self).__init__()
         self._dataDir = dataDir
-        sourceDirNames = ["referenceSets", "datasets"]
+        sourceDirNames = ["referenceSets", "ontologies", "datasets"]
         constructors = [
-            references.HtslibReferenceSet, datasets.FileSystemDataset]
-        objectAdders = [self.addReferenceSet, self.addDataset]
+            references.HtslibReferenceSet, datasets.FileSystemOntology,
```
jeromekelleher: So, if FileSystem ontology maps to a single Ontology, and each ontology is a single txt file in the 'ontologies' directory, this will all work out nicely.

ga4gh/datamodel/datasets.py
```diff
@@ -170,6 +209,14 @@ def __init__(self, localId, dataDir, backend):
                 variantSet = variants.HtslibVariantSet(
                     self, localId, relativePath, backend)
                 self.addVariantSet(variantSet)
+            # Variant annotations sets
```
jeromekelleher: I think we should set up some linkage here between the parent variantSet and the VariantAnnotationSet. At the moment when we set a ``variantannotationsets-search`` command the ``variantSet`` field is null. To resolve this we need a reference to the variantSet. The easiest way to do this is to pass the ``variantSet`` object in here as a parameter to the constructor of the HtslibVariantAnnotationSet object.

ga4gh/cli.py
```diff
@@ -384,6 +430,45 @@ def run(self):
...
+        iterator = self._client.searchVariantAnnotations(
+            variantAnnotationSetId=variantAnnotationSetId,
+            referenceName=self._referenceName, referenceId=self._referenceId,
+            start=self._start, end=self._end,
+            featureIds=self._featureIds, effects=self._effects)
+        self._output(iterator)
+
+    def run(self):
+        if self._variantAnnotationSetId is None:
+            for annotationSet in self.getAllAnnotaionSets():
```
jeromekelleher: I get an error here when I try to run this: have you defined getAllAnnotationSets?

ga4gh/datamodel/variants.py
```diff
@@ -547,3 +596,160 @@ def buildMetadata(
...
+
+class HtslibVariantAnnotationSet(HtslibVariantSet):
+    """
+    Class representing a single variant annotation set backed by a directory of
+    indexed VCF or BCF files.
+    """
+    def __init__(self, parentContainer, localId, dataDir, backend):
+        super(HtslibVariantAnnotationSet, self).__init__(
+            parentContainer, localId, dataDir, backend)
+        self.compoundIdClass = datamodel.VariantAnnotationSetCompoundId
```
jeromekelleher: This line should be at the class level (i.e., not in the constructor).

ga4gh/datamodel/variants.py
```diff
@@ -437,13 +485,14 @@ def convertVariant(self, record, callSetIds):
...
         variant.calls = []
         sampleIterator = 0  # REMOVAL
-        for name, call in record.samples.iteritems():
-            if self.getCallSetId(name) in callSetIds:
-                genotypeData = sampleData[sampleIterator].split(
-                    ":")[0]  # REMOVAL
-                variant.calls.append(self._convertGaCall(
-                    record.id, name, call, genotypeData))  # REPLACE
-            sampleIterator += 1  # REMOVAL
+        if callSetIds:
```
jeromekelleher: Why are we changing this here? Shouldn't `record.samples.iteritems` be empty if no samples are present?

ga4gh/datamodel/variants.py
```diff
@@ -209,6 +250,13 @@ def hashVariant(cls, gaVariant):
             str(tuple(gaVariant.alternateBases))).hexdigest()
 
 
+class AbstractVariantAnnotationSet(AbstractVariantSet):
```
jeromekelleher: This class isn't being used at all --- we should probably just delete it, since an VariantAnnotationSet is a direct subclass of a VariantSet (or is it?).

ga4gh/datamodel/variants.py
```diff
@@ -547,3 +596,160 @@ def buildMetadata(
...
+
+    def convertVariantAnnotation(self, record):
+        """
+        Converts the specified record into a GA4GH Annotaiton object.
+        """
+        # First we need the Variant
+        variant = self.convertVariant(record, None)
+        # Create annotaiton
+        annotation = self._createGaVariantAnnotation()
+        annotation.variant = variant
```
jeromekelleher: Do we really embed the variant in the annotation object? I just see variantId in the schema?

ga4gh/datamodel/variants.py
```diff
@@ -547,3 +596,160 @@ def buildMetadata(
...
+        annotation.variantId = variant.id
+        if record.id is not None:
+            annotation.coLocatedVariants = record.id.split(';')
+        # Convert annotations from INFO field into TranscriptEffect
+        annStr = record.info.get('ANN')
+        transcriptEffects = []
+        if annStr:
```
jeromekelleher: Is this testing for ``annStr is not None``? If so, please make this explicit.

jeromekelleher:
This is looking great @pcingola. I'm happy to merge this PR into the variation_annotation branch once the minor comments above have been discussed, and some minimal tests have been added.

The easiest way to add test coverage of the API endpoints is to add a SimulatedVariantAnnotationSet which extends from SimulatedVariantSet. This should be very easy to implement, and need not be in any way realistic. Once this is done, we can add HTTP layer tests in ``test_views.py`` and some tests for the API methods in ``tests/unit/test_simulated_stack.py``, and end-to-end client tests in ``tests/end_to_end/test_client_json.py``.  These three tests will give us a lot of confidence that the implementation works, once the simulator generates a reasonable spread of expected values.

One high-level question: is there a plan to add support for adding API support for Ontologies?

ga4gh/backend.py
```diff
@@ -192,6 +192,64 @@ def _getEnd(cls, variant):
...
 
+class VariantAnnotationsIntervalIterator(IntervalIterator):
+    """
+    An interval iterator for annotations
+    """
+
+    def __init__(self, request, parentContainer):
+        super(VariantAnnotationsIntervalIterator, self).__init__(
+            request, parentContainer)
+        self._featureIds = set(self._request.feature_ids)
```
david4096: It's my impression that the request may not contain feature_ids. Is some type of none-check or exception handling needed here to get the desired functionality?

        try:
          self._featureIds = set(self._request.feature_ids)
        except TypeError:
          self._featureIds = []

david4096:
If I'm reading the schema right a search for variantannotationsets without a dataSetId or variantSetId field set should still return results. An open call to variantannotationsets currently results in:

     {
         "errorCode": 116989170, 
         "message": "The requested dataset 'None' was not found"
     }

This seems like it might be true for other endpoints outside of the PR @jeromekelleher ?

jeromekelleher:
Thanks for this @david4096, this certainly doesn't look like correct behaviour to me. I _think_ this is handled in all the existing endpoints, but if you find a problem then definitely open an issue on it.

ga4gh/backend.py
pcingola: There can be more than one ontology per dataset.
I think this way is better and more generic. Will keep it like it is.

ga4gh/backend.py
pcingola: There should be room for more than one ontology. The fact that we only have one at this moment is anecdotal.

ga4gh/backend.py
pcingola: Done, thanks.

ga4gh/cli.py
pcingola: Added

ga4gh/datamodel/datasets.py
pcingola: You may want to store the OBO file for the ontology as well for reasons of reproducibility, that's why I used a dir instead of a simple TXT file. When ontologies are properly implemented a single TXT file may not be enough.

OK, I'm changing FileSystemOntology -> FileSystemOntologies
and AbstractOntology -> FileSystemOntology
This should make more sense.

Added in a new module 'ontologies.py' as suggested.

ga4gh/datamodel/variants.py
pcingola: Done.

ga4gh/datamodel/variants.py
pcingola: This was a bug that caused the server to crash.
I'll change to 'is None' as suggested.

ga4gh/datamodel/variants.py
pcingola: Sorry, I don't quite understand how you want me to modify this...

ga4gh/datamodel/variants.py
pcingola: To be able to search by genomic position I need variant's data.

ga4gh/datamodel/variants.py
pcingola: Done

ga4gh/backend.py
```diff
@@ -192,6 +198,68 @@ def _getEnd(cls, variant):
...
+
+    def __init__(self, request, parentContainer):
+        super(VariantAnnotationsIntervalIterator, self).__init__(
+            request, parentContainer)
+        print('self._request.featureIds: ', self._request.featureIds)
+        if self._request.featureIds is None:
+            self._featureIds = []
+        else:
+            self._featureIds = set(self._request.featureIds)
+        self._effects = set(self._request.effects)
```
david4096: When the effects aren't in the request it has the same problem as null featureIds.

     --------------------------------------------------------------------------------
     ERROR in app [/Users/david/virtualenvs/ga4gh-server-1/lib/python2.7/site-packages/flask/app.py:1423]:
     Exception on /variantannotations/search [POST]
--------------------------------------------------------------------------------
...
  File "/Users/david/ga4gh/server-pablo2/ga4gh/backend.py", line 214, in __init__
    self._effects = set(self._request.effects)
     TypeError: 'NoneType' object is not iterable
     127.0.0.1 - - [19/Dec/2015 10:30:25] "POST /variantannotations/search HTTP/1.1" 500 -

david4096:
I'm still having the `None` datasetId issue. If I send an empty request to the `variantannotationsets/search` I receive:

    {
        "errorCode": 116989170, 
        "message": "The requested dataset 'None' was not found"
    }


The schema says I shouldn't have to specify it though: https://github.com/ga4gh/schemas/blob/variation_annotation/src/main/resources/avro/alleleAnnotationmethods.avdl#L134

david4096:
It would be really useful for `scripts/prepare_compliance_data.py` to be updated as part of this PR. I can currently coax it into passing 2/6 of the tests we have, but the remaining errors are in sorting out data preparation.

ga4gh/backend.py
pcingola: Seems to work OK for me:

   ./client_dev.py --verbose variantannotations-search --referenceName 1 --start 60000 --end 70000 http://localhost:8000/ | wc
    7064   12545  213964

No errors at all. How are you testing this? 
Can you send the command you are using?

david4096: It looks like the test apparatus you're using is setting default values. When I use the script you've posted the payload to the server looks like:

     {u'referenceId': u'', u'end': 70000, u'name': None, u'pageSize': None, u'pageToken': u'64792:0', u'start': 60000, u'effects': [], u'referenceName': u'1', u'featureIds': None, u'variantAnnotationSetId': u'YnJjYTE6MWtnUGhhc2UzLmNvcHk6dmFyaWFudGFubm90YXRpb25z'}

Try sending in a request directly to the endpoint, the request to VariantAnnotationsSearch should be allowed to be specify only an VariantAnnotationSetId.

    curl -i -H "Accept: application/json" -H "Content-Type: application/json" -X POST -d '{"variantAnnotationSetId": "YnJjYTE6MWtnUGhhc2UzOnZhcmlhbnRhbm5vdGF0aW9ucw=="}' http://localhost:8000/variantannotations/search

ga4gh/frontend.py
```diff
@@ -496,6 +496,18 @@ def searchVariants():
         flask.request, app.backend.runSearchVariants)
 
 
+@DisplayedRoute('/variantannotationsets/search', postMethod=True)
+def searchVariantAnnotationSets():
```
david4096: Variant annotations and sets can currently be found by search, but we need to be able to find them by get requests as well.

ga4gh/datamodel/variants.py
```diff
@@ -547,3 +589,160 @@ def buildMetadata(
...
+    """
+    def __init__(self, parentContainer, localId, dataDir, backend):
+        super(HtslibVariantAnnotationSet, self).__init__(
+            parentContainer, localId, dataDir, backend)
+        self.compoundIdClass = datamodel.VariantAnnotationSetCompoundId
+        self._compoundId = datamodel.VariantAnnotationSetCompoundId(
+            self.getCompoundId(), 'variantannotations')
+        self._sequenceOntology = backend.getOntology('sequence_ontology')
+
+    def getVariantAnnotations(self, referenceName, startPosition, endPosition):
```
david4096: I'm having trouble getting search by feature ID to work. The schema suggests we should be able to find them by either reference, start, and stop, or a list of feature IDs. When I send in a request with a featureId array the server doesn't return any results.

jeromekelleher:
While this PR still needs some work before it is ready for use, it is sufficiently fleshed out to merge into the topic branch. Therefore, I'm merging the PR. Follow up PRs will do some cleanup work and bring it up to date.

Issue #784, opened 2015-11-13
Protobuf transition overview


jeromekelleher:
This issue is to collect discussion about the transition to protobuf, and to provide a high-level overview of the planned transition.

1. On the protobuf branch, we make the minimal possible changes to the current server in order to replace Avro. This will happen through a series of PRs. The first PR will probably be the initial chanegover for the main server code, with any problematic tests disabled via nose decorators. Subsequent follow-up PRs will then gradually reenable these test cases, or replace them if they are too Avro specific.

2. Once the protobuf schema has been formally accepted upstream in the schemas repository, we can then merge the protobuf branch into our mainline development branch. Mainline development will then happen on protobuf based server.

3. Once we are sure we can make a release based on the protobuf version that can be reliably installed by users, we begin the process of changing the codebase over from the current camelCase convention to snake_case. Once we have fully transitioned to snake_case, we can consider the conversion to protobuf complete.

**Note** During the transition period, we will have an odd coding convention that is a mixture of snake_case when we are interacting with the protobuf generated classes, and the existing camelCase convention for everything else. This will cause confusion, but it seems to be the lesser of a number of unpalatable evils.

jacmarjorie:
Question - perhaps I am missing something, but it looks like responses are being returned as JSON, at least currently. Are there plans to support receiving protobuf messages as requests and/or sending protobuf messages as the response, I'm assuming in addition to JSON?

jeromekelleher:
The initial plan is to keep  JSON on-the-wire, and to move to protobuf as the Interface Definition Language. There will be a few minor incompatibilities between the different versions of the protocol (i.e. the one defined in Avro and the one defined in Protobuf), but no more that we would see between two minor versions of the protocol in any case.

Once the dust has settled, I think that there is strong interest in adding protobuf as a second, optional encoding for messages on-the-wire.

dcolligan:
@jeromekelleher can you provide an update on the to-dos wrt protobuf or close this issue?

jeromekelleher:
Closing --- protobuf is done. The only remaining issue is the snake_case and camelCase mess, but this is something we'll just have to work at to fix over time.

PR #799, opened 2015-12-09
Further protocol buffers work

palfrey:
Further work towards #784


ga4gh/protocol.py
```diff
@@ -25,30 +58,65 @@ def convertDatetime(t):
...
+    Serialises a proto-buf object as json
+    """
+    # Using the internal method because this way we can reformat the JSON
+    js = json_format._MessageToJsonObject(protoObject, False)
+    return json.dumps(js, indent = indent)
+
+def fromJsonDict(jsonDict, protoClass):
+    return fromJson(json.dumps(jsonDict), protoClass)
+
+def fromJson(json, protoClass):
```
jeromekelleher: Use jsonDict here --- json collides with the package from stdlib.

ga4gh/backend.py
```diff
@@ -608,8 +594,9 @@ def runListReferenceBases(self, id_, requestArgs):
         response = protocol.ListReferenceBasesResponse()
         response.offset = start
         response.sequence = sequence
-        response.nextPageToken = nextPageToken
-        return response.toJsonString()
+        if nextPageToken != None:
```
jeromekelleher: Use ``is not None``

ga4gh/client.py
```diff
@@ -86,8 +88,8 @@ def listReferenceBases(self, id_, start=0, end=None):
         while notDone:
             response = self._runListReferenceBasesPageRequest(id_, request)
             basesList.append(response.sequence)
-            notDone = response.nextPageToken is not None
-            request.pageToken = response.nextPageToken
+            notDone = response.next_page_token is not ""
```
jeromekelleher: I think this should be ``bool(response.next_page_token) `` for consistencty

ga4gh/protocol.py
```diff
@@ -25,30 +58,65 @@ def convertDatetime(t):
     millis = delta.total_seconds() * 1000
     return int(millis)
 
+def getValueFromValue(value):
```
jeromekelleher: This is a bit mysterious! What does this do? We should have docstring comments for all these public functions here.

tests/unit/test_schemas.py
```diff
@@ -129,12 +132,82 @@ def getDefaultInstance(self, cls):
...
         Returns a new instance with the required values set.
         """
-        tool = avrotools.Creator(cls)
+        tool = prototools.Creator(cls)
         instance = tool.getDefaultInstance()
         return instance
 
 
 class EqualityTest(SchemaTest):
+    def snakeKeyDict(self, item, field):
```
jeromekelleher: This looks scary! What is all this code for?

jeromekelleher:
This looks like huge progress! Some minor queries above, but overall I think we're definitely on the right track.

ga4gh/backend.py
```diff
@@ -68,12 +72,13 @@ def __init__(self, request, parentContainer):
...
         self._searchAnchor = None
         self._distanceFromAnchor = None
-        if request.pageToken is None:
+        if not request.page_token:
             self._initialiseIteration()
         else:
             # Set the search start point and the number of records to skip from
             # the page token.
-            searchAnchor, objectsToSkip = _parsePageToken(request.pageToken, 2)
+            searchAnchor, objectsToSkip = \
```
jeromekelleher: Prefer hanging indents to breaking lines, e.g.
```
searchAnchor, objectsToSkip = _parsePageToken(
    request.page_token, 2)
```

ga4gh/cli.py
```diff
@@ -133,7 +134,7 @@ def _jsonOutput(self, gaObjects):
         line.
         """
         for gaObject in gaObjects:
-            print(gaObject.toJsonString())
+            print(protocol.toJson(gaObject, indent=None))
```
jeromekelleher: Shouldn't no indentation be the default here?

ga4gh/client.py
palfrey: Fixed

ga4gh/backend.py
palfrey: Fixed

tests/unit/test_schemas.py
palfrey: I've added a docString here, but in short it converts (recursively) from camelCasedKeys to snake_keys for a dictionary structure to allow JSON-based examples to be pushed into Protobuf objects.

ga4gh/cli.py
palfrey: Probably, actually yes. toJson defaulted to 2 space pretty-printing originally, because that's what had been put in by Ian's initial conversion of this, but now it should probably be converted to default to no indentation.

ga4gh/backend.py
```diff
@@ -418,8 +418,8 @@ def readGroupSetsGenerator(self, request):
         Returns a generator over the (readGroupSet, nextPageToken) pairs
         defined by the specified request.
         """
-        dataset = self.getDataset(request.datasetId)
-        if request.name is None:
+        dataset = self.getDataset(request.dataset_id)
+        if request.name == "":
```
jeromekelleher: Should be ``if not request.name`` for consistency.

scripts/proto.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/python
```
jeromekelleher: We should follow the pattern used in the other scripts --- see download_data.py as good example.

- Remove shebang, and chmod a-x
- Add docstring comments describing script. What is it for? How should it be used?
- Use usual import structure, including futures.

scripts/proto.py
```diff
@@ -0,0 +1,104 @@
...
+
+from os import pathsep, defpath, environ, walk, system
```
jeromekelleher: Use ``os.walk``, ``os.system`` etc rather than importing function for readability.

scripts/proto.py
```diff
@@ -0,0 +1,104 @@
...
+            print "Creating __init__.py in %s" % root
+            with file(join(root, "__init__.py"), "w") as f:
+                if root == proto_folder:
+                    f.write("version = '%s'" % self.version)
+
+
+def main():
+    parser = argparse.ArgumentParser(
+            description="Script to process GA4GH Protocol buffer schemas")
+    parser.add_argument("version")
```
jeromekelleher: What does version mean? We need better help text here for a mandatory argument.

scripts/proto.py
```diff
@@ -0,0 +1,104 @@
...
+                   find_in_path("protoc") if x is not None]
+        protoc = None
+        for c in protocs:
+            if not exists(c):
+                continue
+            output = check_output([c, "--version"]).strip()
+            try:
+                (lib, version) = output.split(" ")
+                if lib != "libprotoc":
+                    raise Exception
```
jeromekelleher: Use Exception("some description")

scripts/proto.py
```diff
@@ -0,0 +1,104 @@
...
+                if version_compare("3.0.0", version) > 0:
+                    raise Exception
+                protoc = c
+                break
+
+            except Exception:
+                print "Not using %s because it returned " + \
```
jeromekelleher: Use python 3 prints, and also prefer string ``format`` method over interpolation.

scripts/proto.py
```diff
@@ -0,0 +1,104 @@
...
+        for root, dirs, files in walk(proto_folder):
+            print "Creating __init__.py in %s" % root
+            with file(join(root, "__init__.py"), "w") as f:
+                if root == proto_folder:
+                    f.write("version = '%s'" % self.version)
+
+
+def main():
+    parser = argparse.ArgumentParser(
+            description="Script to process GA4GH Protocol buffer schemas")
```
jeromekelleher: We should have an argument describing the location of the schemas. It probably makes sense to follow the CLI for the current process_schemas.py script. Also, we should rename this script to ``process_schemas.py`` --- we are really just replacing the existing script, not doing anything new.

jeromekelleher:
This is looking good, but I'm having trouble running it. It seems to be something to do with importing the protos:
```python
>>> import ga4gh.protocol
File "ga4gh/protocol.py", line 19, in <module>
from proto.ga4gh.metadata_pb2 import *  # noqa
ImportError: No module named metadata_pb2
```
I've tried commenting various things out, but it's not immediately clear what the problem is.

One high level comment I'd make is that it would be nice if the protos were all contained within the top-level ga4gh package. At the moment we have two packages in the source dir, ``ga4gh`` and ``proto``. This would be problematic if we installed it, as it would potentially clash with some other package in the system called ``proto``. So, would it be possible to have the proto stuff contained in the ga4gh package? I.e., we'd have 
```python
from ga4gh.proto.ga4gh.common_pb2 import *  # noqa
from ga4gh.proto.ga4gh.metadata_pb2 import *  # noqa
# etc
```

It's a bit convoluted, but we only ever deal with these imports in the ``protocol.py`` file, and it does get rid of the packaging problem.

ga4gh/protocol.py
```diff
@@ -5,14 +5,46 @@
...
-import sys
+import datetime
 import json
 import inspect
-import datetime
-import itertools
-from cStringIO import StringIO
 
-import avro.io
+import google.protobuf.json_format as json_format
```
jeromekelleher: Use ``import google.protobuf as protobuf`` here, and use `protobuf.json_format()`` in the code. This style choice is good because it means we can read the code using the function and locally see where the function in question comes from. Likewise for ``message``.

ga4gh/protocol.py
```diff
@@ -26,29 +58,76 @@ def convertDatetime(t):
     return int(millis)
 
 
+def getValueFromValue(value):
```
jeromekelleher: All public functions should have docstring comments.

ga4gh/protocol.py
```diff
@@ -26,29 +58,76 @@ def convertDatetime(t):
...
+    if type(value) != struct_pb2.Value:
+        raise Exception(type(value))
```
jeromekelleher: Is this a TypeError? We should have a better description in any case.

ga4gh/protocol.py
```diff
@@ -26,29 +58,76 @@ def convertDatetime(t):
...
+    if value.WhichOneof("kind") == None:
```
jeromekelleher: Use ``is None``.

ga4gh/protocol.py
```diff
@@ -26,29 +58,76 @@ def convertDatetime(t):
...
+        raise Exception("Nothing set for %s" % value)
```
jeromekelleher: Use string format function instead of interpolation, .e.g ``"Nothing set for {}".format(value).``. Also, is there a more specific exception class we could throw here?

ga4gh/protocol.py
```diff
@@ -26,29 +58,76 @@ def convertDatetime(t):
...
+
+def getDictFromMessageMap(messageMap):
+    return dict([
+        (k, [getValueFromValue(x) for x in v.values])
+        for (k, v) in messageMap._values.items()])
+
+
+def toJson(protoObject, indent=2):
+    """
+    Serialises a proto-buf object as json
```
jeromekelleher: proto-buf -> protobuf. Likewise elsewhere

ga4gh/protocol.py
```diff
@@ -96,149 +175,28 @@ def isFull(self):
...
         elements in the page is >= maxResponseLength.
+
+        If page_size or max_response_length were not set in the request
+        then they're not checked.
         """
         return (
-            self._numElements >= self._pageSize or
-            self._valueListBuffer.tell() >= self._maxResponseLength)
+            (self._pageSize > 0 and self._numElements >= self._pageSize) or
+            (self._maxResponseLength > 0 and
```
jeromekelleher: Why are we checking for ``self._maxResponseLength > 0`` here? It would be more readable if we had a method ``getResponseLength`` rather than running the toJson inline. (Converting to JSON each time is fine for now, but this will be a major perf bottleneck, and so it's good to abstract it out a bit)

ga4gh/prototools.py
```diff
@@ -0,0 +1,400 @@
+"""
```
jeromekelleher: I'm not convinced that we need this module. It's only really used in the ``test_schemas`` test suite, and this was only really intended to test our custom JSON serialisation/deserialisation. We're now using protobuf's native support for this, and I think we should just trust that it works as advertised. The Validator is used in the exceptions module, but this doesn't really do anything right now.

So, I would vote for removing this module entirely, and also remove test_schemas.py.

scripts/process_schemas.py
```diff
@@ -1,486 +0,0 @@
-"""
```
jeromekelleher: Rather than remove this script, I think we should replace it with the equivalent that processes the protobuf schemas, keeping as much of the existing CLI as makes sense. (See comments on scripts/proto.py)

scripts/utils.py
```diff
@@ -134,11 +134,18 @@ def runCommandSplits(splits, silent=False):
...
+    try:
+        if silent:
+            with open(os.devnull, 'w') as devnull:
+                subprocess.check_call(splits, stdout=devnull, stderr=devnull)
+        else:
+            subprocess.check_call(splits)
+    except OSError, e:
+        if e.errno == 2:  # cmd not found
+            raise Exception(
+                "Can't find command while trying to run %s" % splits)
```
jeromekelleher: Use ``.format()`` here (and elsewhere) --- it's more robust than string interpolation.

ga4gh/protocol.py
jeromekelleher: Oops, sorry. Looks like json_format is a module not a function. Ignore this comment.

tests/datadriven/test_references.py
```diff
@@ -83,12 +84,13 @@ def testValidateObjects(self):
...
         referenceSetPe = referenceSet.toProtocolElement()
-        self.assertValid(
-            protocol.ReferenceSet, referenceSetPe.toJsonDict())
-        self.assertGreater(len(referenceSetPe.referenceIds), 0)
+        self.assertIsInstance(
+            referenceSetPe, protocol.ReferenceSet)
+        self.assertGreater(len(referenceSetPe.reference_ids), 0)
         for gaReference in referenceSet.getReferences():
-            reference = gaReference.toProtocolElement().toJsonDict()
-            self.assertValid(protocol.Reference, reference)
```
jeromekelleher: We're making these tests weaker here. Why not leave them as assertValid? We can convert them to JSON, and convert them back as before. Won't proto throw an error if this fails?

tests/datadriven/test_variants.py
```diff
@@ -72,14 +72,15 @@ def _compareTwoListFloats(self, a, b):
 
     def _verifyInfoEqual(self, gaObjectInfo, pyvcfInfo):
         def _assertEquivalentGaVCFValues(gaValue, pyvcfValue):
+            compareValue = protocol.getValueFromValue(gaValue)
```
jeromekelleher: I think gaValue if more useful here --- it's telling us it's the GA4GH protocol version of the same value. The ``getValueFromValue`` function name isn't helping much here --- what does it do?

tests/datadriven/test_variants.py
```diff
@@ -428,5 +429,6 @@ def _hashVariant(self, record):
         if record.ALT[0] is None:
             alts = tuple()
         else:
-            alts = tuple([str(substitution) for substitution in record.ALT])
-        return hashlib.md5(record.REF + str(alts)).hexdigest()
+            alts = tuple([unicode(sub) for sub in record.ALT])
```
jeromekelleher: Why are we changing this? `unicode` is Python 2 only, so please leave as `str` unless this breaks something.

tests/unit/test_cli.py
```diff
@@ -228,13 +230,54 @@ def __init__(self, outputFormat='text'):
...
             self.verbose = 'verbose'
 
-    class FakeObject(protocol.ProtocolElement):
-
-        __slots__ = ['id', 'name']
-
-        def __init__(self):
-            self.id = 'id'
-            self.name = 'name'
+    class FakeObject(protocol.message.Message):
```
jeromekelleher: This looks pretty brittle to me. I think we can just delete the ``TestOutputFormats`` test case. It's not particularly strong, and it's more trouble that it's worth if we have to go rummaging in the guts of protobuf.

tests/unit/test_faultydata.py
```diff
@@ -96,7 +96,7 @@ class TestDuplicateCallSetId(FaultyVariantDataTest):
     """
     localIds = ["duplicated_sampleid"]
 
-    @unittest.skipIf(protocol.version.startswith("0.6"), "")
```
jeromekelleher: We should still export ``protocol.version``. Nothing outside of the ``protocol`` module should have to interact with the generated modules in the ``proto`` hierarchy.

tests/unit/test_protocol_errors.py
```diff
@@ -10,7 +10,7 @@
 import ga4gh.frontend as frontend
 import ga4gh.exceptions as exceptions
 import ga4gh.protocol as protocol
-import ga4gh.avrotools as avrotools
+# import ga4gh.avrotools as avrotools
```
jeromekelleher: Delete any commented out lines please.

tests/unit/test_protocol_errors.py
```diff
@@ -74,25 +66,22 @@ def assertRequestRaises(self, exceptionClass, url, request):
...
         for url, requestClass in self.endPointMap.items():
-            for badType in ["", "1", "None", 0.0, 1e3]:
-                request = self._createInstance(requestClass)
-                request.pageSize = badType
-                self.assertRequestRaises(
-                    exceptions.RequestValidationFailureException, url, request)
-            for badSize in [-100, -1, 0]:
-                request = self._createInstance(requestClass)
-                request.pageSize = badSize
+            # FIXME: Can't test 0 because looks the same as unset to Protobuf
```
jeromekelleher: Delete the FIXME: here --- this is just something we have to learn to live with.

tests/unit/test_schemas.py
```diff
@@ -8,13 +8,16 @@
 import string
```
jeromekelleher: As noted above --- let's just delete this module. It's not testing any code we've written, so it's a bit pointless.

jeromekelleher:
OK, I've made another pass through this. It's all looking great. I have some minor comments above, but mainly I think my feedback is to just remove the tests that are looking tricky because of proto specific stuff. 

Once I can run the tests I'll give more feedback, but I don't think there'll be an major gotchas.

scripts/proto.py
palfrey: Fixed

scripts/proto.py
palfrey: Fixed

scripts/proto.py
palfrey: Fixed

scripts/proto.py
palfrey: Fixed

scripts/proto.py
palfrey: Fixed

scripts/proto.py
palfrey: Fixed

scripts/process_schemas.py
palfrey: Done

tests/datadriven/test_variants.py
palfrey: Without this change, tests.datadriven.test_variants.VariantSetTest.testGetVariant (all the versions) fail because of the hash is incorrect

tests/unit/test_cli.py
palfrey: It's a little brittle, but only really dependant on changes to the Protocol Buffers library itself, not ga4gh formats, so I think it's acceptable

tests/unit/test_faultydata.py
palfrey: Done

tests/unit/test_protocol_errors.py
palfrey: Done

tests/unit/test_schemas.py
palfrey: Done

ga4gh/cli.py
palfrey: Done

ga4gh/protocol.py
palfrey: Done

ga4gh/protocol.py
palfrey: Done

ga4gh/protocol.py
palfrey: Done

ga4gh/prototools.py
palfrey: Done

tests/datadriven/test_references.py
palfrey: Reverted this bit

tests/unit/test_protocol_errors.py
palfrey: Done

ga4gh/protocol.py
palfrey: Done, and changed to AttributeError

palfrey:
> This is looking good, but I'm having trouble running it. It seems to be something to do with importing the protos

This didn't work before because you needed to do the "generate from schemas" step. The results of that are now stored in git, so it should work now.

> So, would it be possible to have the proto stuff contained in the ga4gh package?

Yes, but not without changes to schemas. They all specify the package path as starting "proto", and this cascades down to their dependencies, so it would have to change there as well. I had a quick go at it, and here's the experimental set of changes that would do that https://github.com/lshift/ga4gh-schemas/compare/proto3...lshift:proto3-ga4gh?expand=1

tests/unit/test_cli.py
jeromekelleher: OK, let's keep it then.

jeromekelleher:
Looking really great @palfrey, thanks for this. I'm going through this at the moment, and I think we're very close.

There are some minor issues when we run the server on the example data. Can you run a server with the ga4gh-example data (follow the instructions on the [demo](http://ga4gh-reference-implementation.readthedocs.org/en/latest/demo.html) to get it. Then run:

```
$ python client_dev.py datasets-search http://localhost:8000
```

I'll bring up any other issues I encounter as I hit them.

ga4gh/backend.py
jeromekelleher: Bump.

ga4gh/protocol.py
jeromekelleher: Bump.

ga4gh/protocol.py
jeromekelleher: Bump.

jeromekelleher:
I've bumped some minor comments that got lost somewhere along the line @palfrey --- nothing very important.

ga4gh/protocol.py
palfrey: Done

jeromekelleher:
There's some issue with paging for reads search that the tests didn't pick up. Try:
```
python client_dev.py reads-search http://localhost:8000 -O json --pageSize=5
```
I'm not sure what's happening here; must be some subtle None==0 issue I'd imagine.

ga4gh/protocol.py
palfrey: I've split it out and documented why. It's exactly what you noted - JSON conversion is a perf bottleneck, so don't bother checking if _maxResponseLength is set to the "don't care" value.

requirements.txt
```diff
@@ -17,6 +16,8 @@ Sphinx==1.3.1
 oic==0.7.6
 pyOpenSSL==0.15.1
 lxml==3.4.4
+protobuf==3.0.0b1.post1
+datetime
```
jeromekelleher: Isn't datetime in the stdlib?

palfrey: Mostly. There's an extended DateTime package (https://pypi.python.org/pypi/DateTime/4.0.1) but I don't think that's used here, so I'll remove it.

jeromekelleher:
> > So, would it be possible to have the proto stuff contained in the ga4gh package?

> Yes, but not without changes to schemas. They all specify the package path as starting "proto", and this cascades down to their dependencies, so it would have to change there as well. I had a quick go at it, and here's the experimental set of changes that would do that https://github.com/lshift/ga4gh-schemas/compare/proto3...lshift:proto3-ga4gh?expand=1

Aha, I see. I think we should push for this change to be made in the schemas as we follow this up. Requiring a package root of 'proto' will bring up all sorts of issues. I'm going to split this into a separate issue. The current organisation is great for now, and won't be a blocker to merging this PR.

ga4gh/backend.py
```diff
@@ -45,6 +46,7 @@ def _parsePageToken(pageToken, numValues):
         values = map(int, tokens)
     except ValueError:
         msg = "Malformed integers in page token"
+        raise
```
jeromekelleher: Why is this here?

ga4gh/backend.py
```diff
@@ -434,10 +428,11 @@ def callSetsGenerator(self, request):
         Returns a generator over the (callSet, nextPageToken) pairs defined
         by the specified request.
         """
-        compoundId = datamodel.VariantSetCompoundId.parse(request.variantSetId)
-        dataset = self.getDataRepository().getDataset(compoundId.datasetId)
-        variantSet = dataset.getVariantSet(compoundId.variantSetId)
-        if request.name is None:
+        compoundId = datamodel.VariantSetCompoundId \
```
jeromekelleher: Change this to use hanging indent:
```
compoundId = datamodel.VariantSetCompoundId.parse(
    request.variant_set_id)
```

ga4gh/datamodel/__init__.py
```diff
@@ -299,10 +299,10 @@ class DatamodelObject(object):
     def __init__(self, parentContainer, localId):
         self._parentContainer = parentContainer
         self._localId = localId
-        parentId = None
+        self.parentId = None
```
jeromekelleher: This should be a protected variable, i.e. ``self._parentId``.

ga4gh/datamodel/reads.py
```diff
@@ -239,12 +240,14 @@ def _setHeaderFields(self, samFile):
...
             htslibPrograms = samFile.header['PG']
             for htslibProgram in htslibPrograms:
-                program = protocol.Program()
+                program = protocol.ReadGroup.Program()
                 program.id = htslibProgram['ID']
-                program.commandLine = htslibProgram.get('CL', None)
-                program.name = htslibProgram.get('PN', None)
-                program.prevProgramId = htslibProgram.get('PP', None)
-                program.version = htslibProgram.get('VN', None)
+                program.command_line = \
```
jeromekelleher: Use hanging indent rather than breaking lines:
```python
program.command_line = htslibProgram.get(
   'CL', pb.DEFAULT_STRING)
```

proto/README.txt
```diff
@@ -0,0 +1,2 @@
+The directories "ga4gh" and "google" in this directory are generated by ../scripts/proto - do not add anything under
```
jeromekelleher: The script name is incorrect here --- should be scipts/process_schemas.py

jeromekelleher:
I've made another pass through the full diff (phew!). I think we're ready to go, more-or-less, and should try to merge this in the next day or two. I've created a few issue for minor problems that can be resolved in follow ups (they seem unimportant except for #879).

@palfrey, when you're happy with the state of the PR can you squash the commits down to a handful please? @dcolligan, can you have a quick scan of the code please?

ga4gh/backend.py
palfrey: Temporary item for debugging. Removed.

ga4gh/datamodel/__init__.py
palfrey: Switched back to local variable

proto/README.txt
palfrey: Done

ga4gh/datamodel/reads.py
palfrey: Done

palfrey:
> I'm not sure what's happening here; must be some subtle None==0 issue I'd imagine.

Actually it was a "all structures exist, so don't check if they're None, look for default values" issue. Now corrected with 6469d14

palfrey:
I'm waiting for Travis and @dcolligan's review, then I'll squash down the commits.

jeromekelleher:
Thanks @palfrey, looks good. However, I'm getting a different error when I try
```
python client_dev.py -vv reads-search http://localhost:8000 --pageSize=1
```

On the server side I get:
```
# traceback
TypeError: None has type <type 'NoneType'>, but expected one of: (<type 'str'>, <type 'unicode'>)
```
Can you reproduce? 

ps. This experience is telling me that our test coverage on the reads API code leaves a lot to be desired!

.travis.yml
```diff
@@ -6,9 +6,9 @@ sudo: false
 cache:
   directories:
     - $HOME/.cache/pip
-
+
```
dcolligan: no reason to add this whitespace

.travis.yml
```diff
@@ -21,10 +21,10 @@ before_script:
 
 # run_tests.py runs everything under the script: tag so only put commands
 # under it that we want to run (and want to be able to run) as local tests
-script: 
+script:
   - flake8 *.py tests ga4gh scripts --exclude=ez_setup.py
   - nosetests --with-coverage --cover-package ga4gh
               --cover-inclusive --cover-min-percentage 85
-              --cover-branches --cover-erase
+              --cover-branches --cover-erase -v
```
dcolligan: Is there a reason we are adding verbosity to all test runs?

.travis.yml
palfrey: Fixed

ga4gh/exceptions.py
```diff
@@ -138,6 +138,17 @@ def __init__(self, jsonString):
         self.message = "Cannot parse JSON: '{}'".format(jsonString)
 
 
+class Validator(object):
```
dcolligan: should have a doc comment

.travis.yml
palfrey: I did have it in during debug to solve problems where I had tests that locked up the run so I could figure out which one they were, but I've removed it now

ga4gh/pb.py
```diff
@@ -0,0 +1,20 @@
+"""
+Utility methods for handling protocol buffer
+"""
```
dcolligan: missing standard imports

palfrey:
> On the server side I get:

I can't reproduce this. If you can give me the full stacktrace I can probably fix it.

> ps. This experience is telling me that our test coverage on the reads API code leaves a lot to be desired!

That's in general a worry here... this pull request does some pretty extensive refactoring, and the only real check I've got that I haven't broken anything horribly is the tests! Anything they've got low coverage of I might well have broken.

scripts/utils.py
```diff
@@ -134,11 +134,18 @@ def runCommandSplits(splits, silent=False):
...
-    else:
-        subprocess.check_call(splits)
+    try:
+        if silent:
+            with open(os.devnull, 'w') as devnull:
+                subprocess.check_call(splits, stdout=devnull, stderr=devnull)
+        else:
+            subprocess.check_call(splits)
+    except OSError, e:
+        if e.errno == 2:  # cmd not found
```
dcolligan: great addition

ga4gh/exceptions.py
palfrey: Fixed

ga4gh/pb.py
palfrey: Fixed

dcolligan:
Looks good to me (as determined by my "quick" review)!  Great job.

jeromekelleher:
> I can't reproduce this. If you can give me the full stacktrace I can probably fix it.

Here you go @palfrey. It looks like something fairly straightforward in the references code on second look.

```
--------------------------------------------------------------------------------
ERROR in app [/home/jk/.local/lib/python2.7/site-packages/flask/app.py:1423]:
Exception on /references/search [POST]
--------------------------------------------------------------------------------
  File "/home/jk/.local/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/jk/.local/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/jk/work/github/server/ga4gh/frontend.py", line 490, in searchReferences
    flask.request, app.backend.runSearchReferences)
  File "/home/jk/work/github/server/ga4gh/frontend.py", line 402, in handleFlaskPostRequest
    return handleHttpPost(flaskRequest, endpoint)
  File "/home/jk/work/github/server/ga4gh/frontend.py", line 277, in handleHttpPost
    responseStr = endpoint(request.get_data())
  File "/home/jk/work/github/server/ga4gh/backend.py", line 640, in runSearchReferences
    self.referencesGenerator)
  File "/home/jk/work/github/server/ga4gh/backend.py", line 484, in runSearchRequest
    for obj, nextPageToken in objectGenerator(request):
  File "/home/jk/work/github/server/ga4gh/backend.py", line 292, in _topLevelObjectGenerator
  File "/home/jk/work/github/server/ga4gh/datamodel/references.py", line 270, in toProtocolElement
    reference.source_uri = self.getSourceUri()
  File "build/bdist.linux-x86_64/egg/google/protobuf/internal/python_message.py", line 660, in field_setter
    new_value = type_checker.CheckValue(new_value)
  File "build/bdist.linux-x86_64/egg/google/protobuf/internal/type_checkers.py", line 166, in CheckValue
    raise TypeError(message)
TypeError: None has type <type 'NoneType'>, but expected one of: (<type 'str'>, <type 'unicode'>)
127.0.0.1 - - [27/Jan/2016 10:11:21] "POST /references/search?key=invalid HTTP/1.1" 500 -
```

palfrey:
> Here you go @palfrey. It looks like something fairly straightforward in the references code on second look.

Should be fixed now.

jeromekelleher:
Thanks @palfrey, looks like it was an issue with the incoming data rather than the code (#885).

Looks like we're good to go! Can you squash the commits please? A handful would be ideal (but it would be nice to keep Ian's contributions in the history).

jeromekelleher:
I've just run a ``python scripts/process_schemas.py`` on the proto definitions at https://github.com/ga4gh/schemas/pull/485 and there seems to be quite a few differences. Were these updates that you made to make things work, or is there another version of the proto schemas lying around somewhere?

palfrey:
Little from column A, little from column B :)

There is another version of the schemas at https://github.com/lshift/ga4gh-schemas/tree/proto3 which I think the only additional commit is https://github.com/ga4gh/schemas/commit/c555f61b6e330bbcf0df0954c126904be160369a. There's also I think some local changes I had, and I should correct things so that the output from process_schema lines up.

palfrey:
> There is another version of the schemas at https://github.com/lshift/ga4gh-schemas/tree/proto3 which I think the only additional commit is ga4gh/schemas@c555f61. There's also I think some local changes I had, and I should correct things so that the output from process_schema lines up.

Only missing bit was https://github.com/lshift/ga4gh-schemas/commit/faa00af730ea27ada360c1e41ed07aff1faa59d3 which I've now added to the above mentioned branch

jeromekelleher:
Great, thanks @palfrey. Hopefully @calbach can update the schemas PR so that we're fully in sync (I think the only missing things are the next_page_token error and the missing GAException message). 

There's an error in here too with Reference.reference_ids, which has been removed from the schema. We just need to delete the offending line and update the test cases to deal with it. This didn't cause any problems before because the Avro based techniques didn't worry about us setting non-existent fields.

palfrey:
So I've been having a go at squashing down this PR. As it's got so many commits, and multiple merges of ga4gh:protobuf back in, it was pretty difficult. In fact, the only sane way was to make a duplicate copy and copy/paste over the working tree, which has resulted in https://github.com/ga4gh/server/pull/886, Sadly this only marks me as contributing, not @IanRogers-LShift, but redoing all the merges would chew up a lot of time.

jeromekelleher:
OK, thanks @palfrey. Let's close this PR so and move discussion over to #886.

PR #808, opened 2016-01-04
Pinned install versions of packages.

jeromekelleher:
Uses versions found in requirements.txt. Install currently fails without this because of problems with pysam 0.8.4.


dcolligan:
+1

Issue #790, opened 2015-11-18
Use setuptools_scm to set version numbers


jeromekelleher:
The pypa now recommend using [setuptools_scm](https://pypi.python.org/pypi/setuptools_scm/) to set version numbers from git tags. This looks very useful indeed --- we should use it unless some gotchas happen along the way.

Hat tip to @reece!

reece:
On Wed, Nov 18, 2015 at 12:35 PM, Jerome Kelleher <notifications@github.com>
wrote:

> The pypa now recommend using setuptools_scm
> <https://pypi.python.org/pypi/setuptools_scm/> to set version numbers
> from git tags. This looks very useful indeed --- we should use it unless
> some gotchas happen along the way.
>
> Hat tip to @reece <https://github.com/reece>!
>

I'm happy to implement this. The benefit is that the repo becomes the sole
authoritative source for versioning in code, docs, and packaging. My
release process is essentially 'hg tag 1.2.3 && hg push && python setup.py
sdist bdist_egg bdist_wheel upload'. (setuptools_scm works with hg and git)

Unknowns: 1) What's the best way to work into CI flow? 2) How to work into
java packaging. No obvious blockers for either, but needs thought.

jeromekelleher:
Go for it!

1) Do we need to integrate this into CI flow?
2) I'm not sure what Java packaging has to do with this; this is just for the server, right?

reece:
On Mon, Nov 23, 2015 at 1:02 PM, Jerome Kelleher <notifications@github.com>
wrote:

> 2) I'm not sure what Java packaging has to do with this; this is just for
> the server, right?


I was thinking schemas too, but they're definitely distinct issues.

-Reece

dcolligan:
I'm -0 on integrating this into the CI flow, but the steps should be documented in the release process (see #810).

If we did integrate this into CI, we could just simulate a release with a dummy version number at the end of CI.

dcolligan:
Fixed in #822

PR #822, opened 2016-01-07
Use setuptools_scm to set version numbers.

jeromekelleher:
Issue #790 

Also a minor refactor of setup.py.

@reece --- it would be great to get your thoughts here. I based this off what you've done in hgvs, but thought the ``"write_to": "ga4gh/_version.py"`` was a bit simpler than ``pkg_resources`` (which I think could potentially be misleading for a developer?).

### DIFF
--- a/ga4gh/__init__.py
+++ b/ga4gh/__init__.py
@@ -2,4 +2,9 @@
 Reference implementation of the GA4GH APIs.
 """
 
-__version__ = '0.2.0a1'
+__version__ = "undefined"
+try:
+    from . import _version
+    __version__ = _version.version
+except ImportError:
+    pass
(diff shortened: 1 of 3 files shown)


dcolligan:
So what's the behavior here?  When is version set to a tag and when to an intermediate version with the sha?

jeromekelleher:
Documentation for behaviour is [here](https://pypi.python.org/pypi/setuptools_scm/1.10.1).

Basically, the release process should be:

1) Create git tag for release
2) Pull down this tag to local repo, and run ``python setup.py sdist``
3) Upload version with twine.

dcolligan:
There don't seem to be any merge conflicts between this and #828 (which I just merged) ... they touch different files ... what conflicts were you concerned about?

jeromekelleher:
I was just worried about the potential interaction between the new Travis script and using setuptools_scm. All seems to work well though, so we're ready for vote'n'merge.

dcolligan:
+1

PR #830, opened 2016-01-08
picked up fixes to parsing of characters that need to be encoded

diekhans:
GFF3 syntax.  Added more test data.  Also minor formatting changes


ga4gh/gff3.py
```diff
@@ -105,16 +145,20 @@ def __init__(self, fileName=None):
...
     def add(self, feature):
         """
-        add a feature record
+        Add a feature record
         """
+        # None featureId allowed
         self.byFeatureId[feature.featureId].append(feature)
 
     def __linkFeature(self, feature):
+        """
```
macieksmuga: Did you mean to use a single underscore at the beginning?
http://stackoverflow.com/questions/1301346/the-meaning-of-a-single-and-a-double-underscore-before-an-object-name-in-python

Also, while a docstring isn't necessary in this case:
http://stackoverflow.com/questions/7687407/are-docstrings-for-internal-functions-python-necessary
I'll leave it up to @jeromekelleher to weigh in on whether we should standardize on using one in such a case anyway (it doesn't seem to be forbidden in PEP8 either)

ga4gh/gff3.py
```diff
@@ -134,25 +181,41 @@ def __linkToParent(self, feature, parentId):
             parentPart.children.add(feature)
 
     def finish(self):
-        """
-        finish loading the set, constructing the tree
-        """
+        "finish loading the set, constructing the tree"
```
macieksmuga: why not a docstring here?

ga4gh/gff3.py
diekhans: the use of double underscores is intentional.  These are private functions.

ga4gh/gff3.py
diekhans: @macieksmuga, change this finish() to follow the server
conventions. Note to the python interpret both ".." and
"""..."""  are doc strings.  I don't follow the """ convention
in my code and I missed changing this one.

macieksmuga:
Closed in favor of #857

PR #854, opened 2016-01-14
Variant annotation set simulator

### LINKED ISSUE #842: Simulator required for variant annotations
There is currently no test coverage for the majority of the variant annotation code, and so it is very difficult to judge whether it works correctly or not. The best way to get test coverage is to implement a simulator, following the example of all the other data model objects.

The simulator does not have to be particularly realistic, it just has to be able to generate some data randomly so we can test the full API stack using the existing testing methods.

david4096:
This introduces the variant annotation simulator for testing purposes and instantiates it as part of the simulated dataset. Fixes #842


jeromekelleher:
Looking good @david4096, on the right track.

david4096:
Closing this in favor of WIP #884

PR #856, opened 2016-01-15
814 data tool

dcolligan:
Barebones implementation of repo management tool

Note: we should clarify what users' expectations are wrt adding each data object.  Are they adding a folder?  A file?  What should the semantics be?  Each object is a little different.

### DIFF
--- a/ga4gh/repo_manager.py
+++ b/ga4gh/repo_manager.py
@@ -1,0 +1,514 @@
+"""
+Manages a GA4GH data repository
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import hashlib
+import json
+import os
+import shlex
+import shutil
+import subprocess
+
+import pysam
+
+import ga4gh.exceptions as exceptions
+import ga4gh.datarepo as datarepo
+import ga4gh.datamodel.datasets as datasets
+
+
+def getReferenceChecksum(fastaFile):
+    """
+    Returns the md5 checksum for the reference sequence in the specified
+    FASTA file. This is the MD5 of the upper case sequence letters.
+    """
+    inputFile = pysam.FastaFile(fastaFile)
+    bases = inputFile.fetch(inputFile.references[0])
+    inputFile.close()
+    return hashlib.md5(bases.upper()).hexdigest()
+
+
+def filenameWithoutExtension(filepath, extension):
+    """
+    Return a filename without the extension suffix
+
+    (os.path.splitext(filename)[0] messes up
+    on filenames with more than one period)
+    """
+    filename = os.path.basename(filepath)
+    index = filename.index(extension)
+    return filename[:index]
+
+
+def runCommandSplits(splits, silent=False):
+    """
+    Run a shell command given the command's parsed command line
+    """
+    if silent:
+        with open(os.devnull, 'w') as devnull:
+            subprocess.check_call(splits, stdout=devnull, stderr=devnull)
+    else:
+        subprocess.check_call(splits)
+
+
+def runCommand(command, silent=False):
+    """
+    Run a shell command
+    """
+    splits = shlex.split(command)
+    runCommandSplits(splits, silent=silent)
+
+
+class RepoManager(object):
+    """
+    Performs operations on a GA4GH data repository
+    """
+    datasetsDirName = datarepo.FileSystemDataRepository.datasetsDirName
+    referenceSetsDirName = \
+        datarepo.FileSystemDataRepository.referenceSetsDirName
+    readsDirName = datasets.FileSystemDataset.readsDirName
+    variantsDirName = datasets.FileSystemDataset.variantsDirName
+    fastaExtension = '.fa.gz'
+    fastaIndexExtensionFai = '.fa.gz.fai'
+    fastaIndexExtensionGzi = '.fa.gz.gzi'
+    jsonExtension = '.json'
+    vcfExtension = '.vcf.gz'
+    vcfIndexExtension = '.vcf.gz.tbi'
+    bamExtension = '.bam'
+    bamIndexExtension = '.bam.bai'
+
+    def __init__(self, repoPath):
+        self._repoPath = repoPath
+        self._topStructure = [
+            self.datasetsDirName, self.referenceSetsDirName]
+        self._datasetStructure = [
+            self.readsDirName, self.variantsDirName]
+
+    def _assertFileExists(
+            self, filePath, text='File', inRepo=False, emitName=None):
+        if not os.path.exists(filePath):
+            if emitName is None:
+                emitName = filePath
+            message = "{} '{}' does not exist".format(text, emitName)
+            if inRepo:
+                self._raiseException(message)
+            else:
+                raise exceptions.RepoManagerException(message)
+
+    def _assertPathEmpty(
+            self, path, text='Path', inRepo=False, emitName=None):
+        if os.path.exists(path):
+            if emitName is None:
+                emitName = path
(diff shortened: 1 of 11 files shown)


ga4gh/repo_manager.py
```diff
@@ -0,0 +1,383 @@
...
+    # on filenames with more than one period
+    return '.' + '.'.join(filename.split('.')[1:])
+
+def filenameWithoutExtension(filename):
+    # os.path.splitext(filename)[0] messes up
+    # on filenames with more than one period
+    return filename.split('.')[0]
+
+
+class AbstractRepoObject(object):
```
jeromekelleher: I'm not sure I see the value of having a separate RepoObject hierarchy. Wouldn't it be simpler to include the required functionality into the ``datamodel.DatamodelObject`` hierarchy? I guess most of the functionality belongs in the PysamDatamodelMixin?

ga4gh/repo_manager.py
```diff
@@ -0,0 +1,383 @@
...
+        self._name = name
+
+
+class ReadGroupSet(AbstractRepoObject):
+
+    def __init__(self, name):
+        self._name = name
+
+
+class RepoManagerException(Exception):
```
jeromekelleher: Should this be in ``ga4gh.exceptions``?

ga4gh/repo_manager.py
```diff
@@ -0,0 +1,383 @@
...
+                for variant in variantSet._variants:
+                    self._emitIndent(variant, 4)
+        
+    def listObjs(self):
+        """
+        Return the contents of the repo as an object
+        """
+        repoObj = Repository(self._repoPath)
+
+        # referenceSets
```
jeromekelleher: OK, we're definitely recreating code here. Right now, we should be able to do something like
```
backend = FileSystemBackend(self._repoPath)
for dataset in backend.getDatasets():
   # etc.
```
We need to figure out what the distinction between a 'Backend' and a 'Repository' is. I guess a Backend _has_ a Repository? This would need a refactor of the Backend class to take out all of the ``addReferenceSet`` type methods and point them at the ``Repository`` object instead.

Does this make sense? I.e., we would have code like
```
repo = FileSystemRepository(self._repoPath)
for dataset in repo.getDatasets():
   # etc.
```

jeromekelleher:
Looks good @dcolligan. I think we should look at the suggested refactor, as there'll be a lot of duplication if we don't centralise the repo code.

dcolligan:
Good idea.  Refactor is here: https://github.com/ga4gh/server/pull/861

dcolligan:
@jeromekelleher some design questions:

1) Should we always assume that we are getting passed a compressed file (fa vs fa.gz, vcf vs vcf.gz)?  If not, should we compress the original file and copy/link it to the repo, or copy it to the repo and then compress the copy?
2) When adding a directory of vcfs (i.e. a variant set), should we mandate that all files in the directory be vcf / vcf.gz / vcf.gz.tbi?  Or that only some of them are?  Or not do any consistency checks?
3) If index files are present as well (e.g. a.bam.bai for a request to add a.bam), should we link/copy them over to the repo, or should we always calculate our own new index files in the repo?
4) When initializing a datarepo.FileSystemDataRepository, the object does some internal consistency checks.  For instance, if the BAM header names a referenceSet, we check if the referenceSet exists in our repo and throw an error if it does not.  Therefore, it seems to matter in which order users add files to the repo, which could be very confusing to them.  How should we handle this?  Turn off the checks somehow in this case?  Modify the data so that the checks aren't triggered?  Not call datarepo.FileSystemDataRepository until it is absolutely necessary?  Wrap the call with some error handling to explain certain errors to the user?  Something else?

jeromekelleher:
> 1) Should we always assume that we are getting passed a compressed file (fa vs fa.gz, vcf vs vcf.gz)? If not, should we compress the original file and copy/link it to the repo, or copy it to the repo and then compress the copy?

I think we can (initially) assume that it's compressed.

> 2) When adding a directory of vcfs (i.e. a variant set), should we mandate that all files in the directory be vcf / vcf.gz / vcf.gz.tbi? Or that only some of them are? Or not do any consistency checks?

I guess the simplest thing is to just pull out the VCFs and index files, and not insist on there not being any other files in there. We should bear in mind that we'll want to point to URLs also; what will be the UI for this?

> 3) If index files are present as well (e.g. a.bam.bai for a request to add a.bam), should we link/copy them over to the repo, or should we always calculate our own new index files in the repo?

We should copy I think. We might be importing a very large number of BAMs and it doesn't make sense to reindex them all if we already have them.

> 4) When initializing a datarepo.FileSystemDataRepository, the object does some internal consistency checks. For instance, if the BAM header names a referenceSet, we check if the referenceSet exists in our repo and throw an error if it does not. Therefore, it seems to matter in which order users add files to the repo, which could be very confusing to them. How should we handle this? Turn off the checks somehow in this case? Modify the data so that the checks aren't triggered? Not call datarepo.FileSystemDataRepository until it is absolutely necessary? Wrap the call with some error handling to explain certain errors to the user? Something else?

Perhaps we should spin the consistency checks out to a separate method, which only gets called on server init, and not every time we instantiate a datarepo.FileSystemDataRepository? This would make it a bit more fine-grained I think.

dcolligan:
@jeromekelleher updated

Question: what fields in the referenceSet and/or reference metadata should we allow the user to set?  Should the CLI interface to that just be option flags (seems the simplest to me)?

jeromekelleher:
Thanks @dcolligan, this is looking great. I think we should provide stuff like "description" as CLI options. For example, we might have ```python repo_dev.py add-dataset tmp/repo ds1 --description=x```. For referenceSets etc. I would just add the ``description`` field for now. It'll be easy to add extra options later.

What's your thoughts on how this is looking testing-wise? Do you want to do more testing now in this PR or follow up later with other stuff?

dcolligan:
Ok, added the cli flags and some more tests.

I think I want to do some more testing on the individual repo manager commands and then we can merge this.

jeromekelleher:
Sounds like a plan @dcolligan.

dcolligan:
@jeromekelleher ok, ready for review

tests/end_to_end/test_repo_manager.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def _runCmd(self, cmd, *args):
+        argString = ' '.join(args)
+        command = "python repo_dev.py {} {} {}".format(
+            cmd, self.tempdir, argString)
+        splits = shlex.split(command)
+        with open(os.devnull, 'w') as devnull:
+            subprocess.check_call(splits, stdout=devnull, stderr=devnull)
+
+    def testEndToEnd(self):
```
jeromekelleher: Nice!

tests/unit/test_repo_manager.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class RepoManagerTest(AbstractRepoManagerTest):
+    """
+    End to end test for the repo manager
+    """
+    def testEndtoEnd(self):
+        # TODO can't use tempdir here since its contents are destroyed
+        # before the finally block executes ... not sure why
+        testDirName = 'repo-manager-test-dir'
```
jeromekelleher: Use setUp and tearDown here to create and delete ``self.testDirName`` using the tempfile library.

jeromekelleher:
This is great work, thanks @dcolligan! +1 modulo suggestions above.

dcolligan:
@jeromekelleher updated

jeromekelleher:
Great, thanks @dcolligan. Merging.

We should get some experience working with this, document it, and work towards a release in the next couple of weeks.

PR #857, opened 2016-01-15
Gff3 parser improvements

macieksmuga:
This PR incorporates Mark's latest improvements to the GFF3 Parser itself (in PR #830, now superseded by this one) as well as a new suite of data-driven tests against the parser.


tests/unit/test_gff3_parser.py
```diff
@@ -0,0 +1,102 @@
+"""
```
jeromekelleher: Data driven tests should be in tests/datadriven, not tests/unit.

tests/unit/test_gff3_parser.py
```diff
@@ -0,0 +1,102 @@
...
+import os
+import ga4gh.gff3 as gff3
+import unittest
+
+
+class TestGff3ParserOnTypicalFile(unittest.TestCase):
+    """
+    Data driven unit tests for the GFF3 parser
+    """
+    def setUp(self):
```
jeromekelleher: Usually with data driven tests we want to compare the results of two difference parsing approaches. VCF is the canonical example: we use PyVCF to parse the VCFs and we make sure our code agrees with it. Can we import some other GFF parser to parse the file for us?

jeromekelleher:
It would be nice to see this GFF parser as a full PR against the RNA branch. Would it be possible to 'reboot' the RNA branch, and start bring this in as the first fully-formed feature?

tests/unit/test_gff3_parser.py
diekhans: I don't think you want to bring in another GFF3 parser as a
dependency.

The reason a GGF3 parser was written is the two GFF3 parser
available, one was inadequate toy with no tests and not well
written. The other had a very heavy weight dependency
(biopython).  If there has been a usable GFF3 parser, we would
have used it.

One approach I use for test code that works on files is to
create expected files and diff them with test results. In this
cases, the write function of the GFF3 can be use to create a
file.  While it will not be directly comparable to the source
due to lack of ordering constraints, it can then be manually
validated in various ways.  The manual validation can then be
done (inspection, run through web validation, etc).  The
expected file is then check in.

diekhans:
We are going to break sequence annotations out into it's own PR,
due to variant annotations needed it.  So a branch just for it
would be appropriate.

I guess this brings up who controls topic branches.  In schemas,
it's completely under the task team's control.  How does it work
in the server repo?

jeromekelleher:
I think in the interest of making mergable code it makes sense that the server develops all take joint responsibility for the branches. For early exploratory work, people can work (and collaborate on) forks, which they can do whatever they like with. Once they are happy they have a unit of work that is fairly mature they can open a PR on the $TOPIC_BRANCH in the server, where we can then review the code and merge, with tidy commit histories. Once code is in the topic branch, it is everyone's responsibility to make sure that it is carried forward and eventually merged into master.

At this point, I'm not sure what's in the rna branch, which makes things problematic. Perhaps we should make the new sequence annotations branch and seek to merge this GFF3 code there?

tests/unit/test_gff3_parser.py
jeromekelleher: Well, we're talking about a develop-time dependency here, not a run-time dependency. I don't think BioPython is that scary a develop-time dependency. Once it doesn't murder our CI builds, I think it would probably be OK.

diekhans:
The more eyes the better; the problem with schema repo is
getting anyone to look.  This is one of the reasons task teams
want to move things into a topic branch early.

A sequence annotation branch will be need if we are going to
merge it independently.

tests/unit/test_gff3_parser.py
diekhans: It's not only BioPython, it's a also an add-on dependency that
is indented to become part of BioPython but has been stagnant
for a year:

  https://pypi.python.org/pypi/bcbio-gff/0.6.2

requiring two parsers is a rather heavy weight way to have
expected results.

jeromekelleher: OK, let's forget about this then. Reasonable test coverage over the subset of GFF we're actually using will do fine. These data-driven test cases can condition on known properties of the specific data files; this should be a quick way of getting the parser done with decent test coverage.

tests/unit/test_gff3_parser.py
macieksmuga: I initially put this test in `tests/datadriven`, but moved it once I realized that all tests there assume that the entity being tested is a ga4gh datamodel member - and all the code mechanisms for those tests fundamentally rely on that assumption. In contrast, this file really does represent just a unit test for the GFF3 parser, whose internal representation of feature data is not quite in alignment with GA4GH datamodels. Thus the inclusion in `unit`, not `datadriven`.

tests/unit/test_gff3_parser.py
macieksmuga: FWIW, these really are meant to be unit tests of the parser. Known inputs, known desired parsed object outcome. As you said, @jeromekelleher, I'm just aiming for reasonable test coverage of the parser code - and since we're still evolving the parser code, I figured getting these tests in was a high priority.

macieksmuga:
@jeromekelleher I'm not sure I understood your request with respect to making a "full PR"  against the RNA branch - I thought that's exactly what's going on here.

Here's what's on the RNA branch right now:

- Sequence annotations code: GFF3 parsing, sequence annotation features schema records and endpoints corresponding to them, and rudimentary support for features in the server.
- Ontologies. Needed for Sequence Annotations (which code feature types via ontology). Not sure if any code apart from that directly needed by sequence annotations was added.
- RNA quantification code: Parsing RNA expression files directly in rna_quantification.py. 

What's being added in this PR:
- enhancements (fixes) to the GFF3 parser
- unit tests for the GFF3 parser.

tests/unit/test_gff3_parser.py
jeromekelleher: OK, good point.

tests/unit/test_gff3_parser.py
jeromekelleher: Agreed.

jeromekelleher:
@macieksmuga, I haven't reviewed any of the existing code in the RNA branch, so I'm reluctant to spend much time reviewing these updates. I would like to see the RNA branch 'reset' so that it's level with master, and the GFF parser with tests added as a single, clean PR. The other features can be added back in afterwards in subsequent PRs.

Does this make sense?

saupchurch:
@jeromekelleher I'm not against this idea especially if it enables the RNA project as a whole to move forward.  FWIW I think going this way works and it will generate fewer headaches for all to slice up the RNA PR into bite sized chunks than to submit and review the full code changes.

macieksmuga:
Ah, I see what you mean now, @jeromekelleher. And yes, @saupchurch, that would be a clean (if somewhat slower) way for us to proceed forward - allowing us to clearly explain all our design and implementation decisions.

jeromekelleher:
@saupchurch, @macieksmuga, good, let's go with that then. We can confirm the details tomorrow.

macieksmuga:
Superseded by #962

PR #859, opened 2016-01-19
Ensures evidence level is serialized in responses

gabrielsaldana:
Integrates evidence level information to serialized responses.


jeromekelleher:
This PR is against the master branch rather than against a topic branch, so I'm going to close it.

We won't be able to merge g2p specific code into master until the schemas have been accepted. We'll need to create a topic branch for g2p, and get it started with the initial implementation that @bwalsh has done, and then can make incremental improvements to that. We should talk about what needs to be done on Friday. In the mean time, I've created the g2p topic branch.

bwalsh:
@jeromekelleher @gabrielsaldana understood,  we pushed this a bit early.  thanks.

PR #864, opened 2016-01-20
Branch update: variation_annotation

jeromekelleher:
Brings the variation_annotation up to date with master.

This PR was created directly in github rather than from a fork. Is this better/worse? We won't be able to resolve conflicts here, but perhaps this is a simple way for topic branch maintainers to keep their branches in sync?


jeromekelleher:
OK, that doesn't seem to have worked very well. @dcolligan, looking for your wisdom here. Should we do the updates from our forks? I quite like the idea of having protected branches and updating via PRs --- it seems like a relatively safe way of keeping the  branches in sync.

dcolligan:
It really shouldn't matter in terms of changes to code whether one pushes branches to the ga4gh repo or one's own repo (forked from ga4gh/server).  The only difference is where the branch gets created, and the corresponding permissions on that branch.

I agree we should protect the long-running branches, since if we're going to go with option C (discussed in the long running branch workflow issue), then there is no reason to force push to them.  We could do this in one of two ways: either we could host those branches on ga4gh/server and protect them there, or individual branch managers could enable protections on branches within their fork of ga4gh/server.

So the factors in choosing where to host the branch are:
1) Clutter.  Do we want everyone to see the long-running branches on the ga4gh/ server branch list?
2) (more important) Who do we want to be able to push to the branches?  On ga4gh/server, it's just a select group of people.  On a branch manager's fork, s/he can configure the individuals s/he wants to give permissions to.

So it's really another decision about how much configurability we want to push down to the branch managers.  Whatever we decide, we should update the long-running doc section with our decision.  Personally, I would favor delegating this responsibility to the branch managers.  That would also allow us to reduce the branch clutter in ga4gh/server to the trunk branches and restrict the number of people that can push to ga4gh/server to those who absolutely must have that power.  We should probably also create a wiki page of "currently active long-running branches and their branch managers", just so everyone knows where the branches are and whose responsibility they are.  This also gives branch managers the option to protect the branch, if they want to do that (or not, especially if they want to go with a rebase-based workflow strategy).

jeromekelleher:
I think we need to keep some centralised control on the long-running branches @dcolligan --- there's only a handful of them, and we'll need to merge them into master at some point. This does mean more work for you and me, but I don't think we can avoid it unfortunately.

We should make a final decision on how to manage this tomorrow in the hangout.

jeromekelleher:
Closing this in favour of #866.

PR #866, opened 2016-01-20
Variant annotation update against master

david4096:
Brings variation_annotation branch in line with master using local fork. @jeromekelleher how does this look compared to #864?


jeromekelleher:
Looks good @david4096, but I think you need to rebase off upstream/variation_annotation? Github is complaining that the branch is out-of-date.

jeromekelleher:
Can you update this to pull in the latest commits from master please @david4096? An important refactor has just gone through which will cause conflicts down the line if we don't bring it in now.

Issue #838, opened 2016-01-11
Variant Annotation - reformat HGVS


sarahhunt:
HGVS descriptions moved to nested sub-record:

https://github.com/ga4gh/schemas/pull/531/files

david4096:
Resolved in https://github.com/ga4gh/server/pull/845
https://github.com/david4096/server-1/commit/93ede8a8a0934f76a6cac9e0f7193dc3932fe0a8#diff-f9ce58fdc26b5bb89ea0a4aa47a608faR661

david4096:
I reformatted the responses to follow this spec and will be reflected in #845. @sarahhunt could you give us some idea of what you expect for HGVS genomic coordinates? Is this read directly from the line of the VCF file or does it need to be computed somehow?

sarahhunt:
@david4096  - I would expect it to be read from the file. I didn't notice where the ref server annotation VCF was, but I see HGVSg descriptions aren't in the compliance test data , so I have added them in  https://github.com/ga4gh/compliance/pull/162. I can add them to the server VCF if you point me to it.

david4096:
Fixed with #868

PR #867, opened 2016-01-20
Use new test data for HGVS genomic coordinates

david4096:
Genomic coordinates are now available in the test data as seen here https://github.com/ga4gh/compliance/pull/162.

This PR puts the genomic coordinates into the HGVS annotation record fixing #838 .


jeromekelleher:
You need to rebase this off upstream/variation_annotation @david4096.

PR #869, opened 2016-01-22
Variant annotation with variant sets

david4096:
This PR attempts to formalize the relationship between variant sets and variant annotation sets. Previously, it was possible to have variant annotations for variants that weren't present in GA4GH. Now annotation VCFs are considered a source of a variant set as well as a variant annotation set.

This is done by placing a small description JSON next to the VCF in the directory. For now, this file simply denotes whether the VCF is annotated or not.

    {
        "isAnnotated": true
    }

Furthermore, the requirement of placing the variant object within the annotation has been removed: annotations now maintain the start and end coordinates themselves so they can be searched across without looking up the variant.

### DIFF
--- a/ga4gh/datamodel/datasets.py
+++ b/ga4gh/datamodel/datasets.py
@@ -210,12 +210,11 @@ class FileSystemDataset(AbstractDataset):
                     self, localId, relativePath, dataRepository)
                 self.addVariantSet(variantSet)
             # Variant annotations sets
-            variantAnnotationSetDir = os.path.join(
-                relativePath, "variantAnnotations")
-            if os.path.isdir(variantAnnotationSetDir):
-                variantAnnotationSet = variants.HtslibVariantAnnotationSet(
-                    self, localId, variantAnnotationSetDir, dataRepository)
-                self.addVariantAnnotationSet(variantAnnotationSet)
+                if variantSet._isAnnotated(relativePath, localId):
+                    variantAnnotationSet = variants.HtslibVariantAnnotationSet(
+                            self, localId, relativePath, dataRepository,
+                            variantSet)
+                    self.addVariantAnnotationSet(variantAnnotationSet)
 
         # Reads
         readGroupSetDir = os.path.join(dataDir, "reads")
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -10,6 +10,8 @@ import datetime
 import random
 import hashlib
 import re
+import os
+import json
 
 import pysam
 
@@ -586,22 +588,55 @@ class HtslibVariantSet(datamodel.PysamDatamodelMixin, AbstractVariantSet):
                         description=description))
         return ret
 
+    def _metadataFileExists(self, fname):
+        """
+        Check for existence of description JSON.
+        """
+        return os.path.exists(fname)
+
+    def _parseMetadataJson(self, str):
+        """
+        Parse metadata JSON into python dict.
+        """
+        return json.loads(str)
+
+    def _getMetadataFilename(self, path, localId):
+        """
+        Given a path and identifier create a path
+        to the expected metadata JSON.
+        """
+        return os.path.join(path, localId + ".json")
+
+    def _isAnnotated(self, path, localId):
+        ret = False
+        fname = self._getMetadataFilename(path, localId)
+        if self._metadataFileExists(fname):
+            with open(fname) as metadataJson:
+                metadata = json.loads(metadataJson.read())
+            if metadata['isAnnotated']:
+                ret = True
+        return ret
+
 
 class HtslibVariantAnnotationSet(HtslibVariantSet):
     """
     Class representing a single variant annotation set backed by a directory of
     indexed VCF or BCF files.
     """
-    compoundIdClass = datamodel.VariantAnnotationSetCompoundId
-
-    def __init__(self, parentContainer, localId, dataDir, backend):
+    def __init__(self, parentContainer, localId, dataDir,
+                 backend, variantSet):
         super(HtslibVariantAnnotationSet, self).__init__(
             parentContainer, localId, dataDir, backend)
+        self.compoundIdClass = datamodel.VariantAnnotationSetCompoundId
+        self._variantSetId = variantSet.getCompoundId()
+        self._variantSet = variantSet
+        self._compoundId = datamodel.VariantAnnotationSetCompoundId(
+            self.getCompoundId(), 'variantannotations')
         self._sequenceOntology = backend.getOntology('sequence_ontology')
 
     def getVariantAnnotations(self, referenceName, startPosition, endPosition):
         """
-        Returns an iterator over the specified variant annotaitons.
+        Returns an iterator over the specified variant annotations.
         """
         if referenceName in self._chromFileMap:
             varFileName = self._chromFileMap[referenceName]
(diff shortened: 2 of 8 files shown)


ga4gh/datamodel/datasets.py
```diff
@@ -210,11 +210,10 @@ def __init__(self, localId, dataDir, backend):
                     self, localId, relativePath, backend)
                 self.addVariantSet(variantSet)
             # Variant annotations sets
-            variantAnnotationSetDir = os.path.join(
-                relativePath, "variantAnnotations")
-            if os.path.isdir(variantAnnotationSetDir):
+            if self.isAnnotated(relativePath, localId):
                 variantAnnotationSet = variants.HtslibVariantAnnotationSet(
-                    self, localId, variantAnnotationSetDir, backend)
+                        self, localId, relativePath, backend,
```
jeromekelleher: I think this is over-indented for hanging indent --- newer versions of pep8 might choke on this.

ga4gh/datamodel/datasets.py
```diff
@@ -227,6 +226,27 @@ def __init__(self, localId, dataDir, backend):
                     self, localId, bamPath, backend)
                 self.addReadGroupSet(readGroupSet)
 
+    def descriptionExists(self, fname):
+        # check for existence of description JSON
```
jeromekelleher: Use docstring comments for public methods. i.e.
```python
def descriptionExists(self, fname):
   """
   Check for existence of description JSON.
   """
```

ga4gh/datamodel/datasets.py
```diff
@@ -227,6 +226,27 @@ def __init__(self, localId, dataDir, backend):
...
+        return os.path.exists(fname)
+
+    def parseDescriptionJson(self, str):
+        # parse into python dict
+        return json.loads(str)
+
+    def descriptionFilename(self, path, localId):
```
jeromekelleher: We'd usually use ``getDescriptionFilename`` here, to make it clearer that your returning the filename.

ga4gh/datamodel/datasets.py
```diff
@@ -227,6 +226,27 @@ def __init__(self, localId, dataDir, backend):
...
+        return os.path.join(path, localId + ".json")
+
+    def isAnnotated(self, path, localId):
+        ret = False
+        fname = self.descriptionFilename(path, localId)
+        if self.descriptionExists(fname):
```
jeromekelleher: I'm not sure the ``descriptionExists`` does much for readablility here.How about
```
fname = self.getDescriptionFilename(path, localId)
if os.path.exists(fname):
    # etc.
```

ga4gh/datamodel/variants.py
```diff
@@ -592,16 +592,19 @@ class HtslibVariantAnnotationSet(HtslibVariantSet):
     Class representing a single variant annotation set backed by a directory of
     indexed VCF or BCF files.
     """
-    compoundIdClass = datamodel.VariantAnnotationSetCompoundId
```
jeromekelleher: This is where it should be. It's a class variable, so should be assigned at the class level.

jeromekelleher:
Good incremental progress here @david4096, +1 modulo the minor niggles above. 

We should try to get variation_annotation branch updated against master before merging this as there is a minor conflict with a recent refactor. Also, please squash the commits when rebasing, this is really one minor change rather than 4 or 5.

ga4gh/datamodel/datasets.py
diekhans: `Description' is confusing terminology.  It's generally though
of as natural language text.  Metadata would be much clearer
name.  Please change.

jeromekelleher:
Looks like the merge with didn't go quite according to plan here @david4096. It might be worthwhile rebasing off upstream/varation_annotation now and re-resolving the conflicts.

david4096:
Rebased @jeromekelleher :)

ga4gh/datamodel/variants.py
```diff
@@ -592,16 +592,19 @@ class HtslibVariantAnnotationSet(HtslibVariantSet):
...
-
-    def __init__(self, parentContainer, localId, dataDir, backend):
+    def __init__(self, parentContainer, localId, dataDir,
+                 backend, variantSetId):
         super(HtslibVariantAnnotationSet, self).__init__(
             parentContainer, localId, dataDir, backend)
+        self.compoundIdClass = datamodel.VariantAnnotationSetCompoundId
```
jeromekelleher: We still need to move this back to the class level.

ga4gh/datamodel/variants.py
```diff
@@ -719,6 +716,17 @@ def getVariantAnnotationId(self, gaVariant, gaAnnotation):
             gaVariant.start, md5)
         return str(compoundId)
 
+    def getVariantId(self, gaVariant):
```
jeromekelleher: Isn't this an exact duplication of VariantSet.getVariantId?

ga4gh/datarepo.py
```diff
@@ -23,6 +24,8 @@ def __init__(self):
         self._referenceSetIdMap = {}
         self._referenceSetNameMap = {}
         self._referenceSetIds = []
+        self._ontologyNameMap = {}
```
jeromekelleher: Why ontologyNames? We should follow the same pattern as datasets and referenceSets unless there's a good reason not to.

ga4gh/datarepo.py
```diff
@@ -42,6 +45,14 @@ def addReferenceSet(self, referenceSet):
         self._referenceSetNameMap[referenceSet.getLocalId()] = referenceSet
         self._referenceSetIds.append(id_)
 
+    def addOntologies(self, ontologies):
```
jeromekelleher: Why not addOntology(self, ontology)? Again, we should follow the established convention here closely for consistency.

ga4gh/datamodel/variants.py
david4096: Hey @jeromekelleher , did you try this change? It is breaking for me and I'm not sure why. It causes a "not enough fields for compound ID" exception.

ga4gh/datamodel/variants.py
david4096: No, as it uses the variant set's compound ID as opposed to the annotation set's.

ga4gh/datarepo.py
david4096: I think `name` captures the functionality better, but I'm open to your suggestion for naming this variable.

jeromekelleher:
Sorry, I haven't had a chance to review this properly today @david4096. Where are we with this and the other PR? I'll try and look at it tomorrow and give some better feedback.

david4096:
@jeromekelleher I included the suggestions you had made, except as noted above. I also added the ability to search annotation sets, which had been left out.

ga4gh/datarepo.py
macieksmuga: This makes perfect sense to me, it's in direct analogy to `referenceSetNameMap` and `datasetNameMap`: Ontologies are pulled up by their name, so there isn't a case (yet) for an `ontologyIdMap`.

jeromekelleher:
OK, let's merge this much so. I think the ontologies stuff is confusing, but we can deal with that as we go.

+1 after rebase on upstream/variation_annotation.

diekhans:
Ontologies are a work in progress through the schema.  They will
continue to evolve based on actual use.

PR #872, opened 2016-01-22
Remove padding from identifiers and add it back in as needed

### LINKED ISSUE #860: Padding in identifiers
There are equals signs (`=`) for padding base64 encoded identifiers that are leaked to returned identifiers. Normal practice is for `=` to encode key-value pairs in a query string. `=` is a reserved character for URL strings so our GET by ID requests look odd. 

     http get localhost:8000/referencesets/aGczNw%3D%3D (is the same as)
     http get localhost:8000/referencesets/aGczNw==

Perhaps the padding should be optional, for example, why are `YnJjYTE6SEcwMDA5Ng==`, `YnJjYTE6SEcwMDA5Ng=`, `YnJjYTE6SEcwMDA5Ng` considered three separate identifiers?

A separate but perhaps related issue, there is no ceiling on the length of identifiers.

https://www.w3.org/Addressing/URL/4_URI_Recommentations.html

david4096:
Putting `=` in identifiers was a side effect of base64 encoding. This PR attempts to hide them from the user while not functionally changing the identifier scheme. Fix #860

### DIFF
--- a/ga4gh/datamodel/__init__.py
+++ b/ga4gh/datamodel/__init__.py
@@ -187,14 +187,17 @@ class CompoundId(object):
         fashion. This is not intended for security purposes, but rather to
         dissuade users from depending on our internal ID structures.
         """
-        return base64.b64encode(idStr)
+        return base64.urlsafe_b64encode(str(idStr)).replace(b'=', b'')
 
     @classmethod
-    def deobfuscate(cls, idStr):
+    def deobfuscate(cls, data):
         """
         Reverses the obfuscation done by the :meth:`obfuscate` method.
+        If an identifier arrives without correct base64 padding this
+        function will append it to the end.
         """
-        return base64.b64decode(idStr)
+        return base64.urlsafe_b64decode(str((
+            data + b'A=='[(len(data) - 1) % 4:])))
 
 
 class ReferenceSetCompoundId(CompoundId):
--- a/tests/unit/test_compound_ids.py
+++ b/tests/unit/test_compound_ids.py
@@ -25,6 +25,24 @@ class TestCompoundIds(unittest.TestCase):
     """
     Test the compound ids
     """
+    def testURLUnsafe(self):
+        hasSlashes = "???"  # base64 encodes to 'Pz8/'
+        needsPadding = "padme"  # base64 encodes to 'cGFkbWU='
+        realistic = ("YnJjYTE6V0FTSDdQX2Fubm90YXRpb246MToxNzY5"
+                     "NDpmMzQxM2JkMTVjNWNiYzI4ZDFiYjY2OGY4ZWM2NzczMg")
+        unsafeCharacters = ['$', '&', '+', ',', '/', ':', ';', '=', '?', '@']
+        for char in unsafeCharacters:
+            self.assertNotIn(char, datamodel.CompoundId.obfuscate(
+                    hasSlashes))
+            self.assertNotIn(char, datamodel.CompoundId.obfuscate(
+                    needsPadding))
+            self.assertNotIn(char, datamodel.CompoundId.obfuscate(
+                    realistic))
+        for idStr in [hasSlashes, needsPadding]:
+            obfuscated = datamodel.CompoundId.obfuscate(idStr)
+            deobfuscated = datamodel.CompoundId.deobfuscate(obfuscated)
+            self.assertEqual(idStr, deobfuscated)
+
     def testBadParse(self):
         for badId in ['a;b', 'a;b;c;d', 'a;b;sd;', ';;;;']:
             obfuscated = datamodel.CompoundId.obfuscate(badId)
--- a/tests/end_to_end/test_gestalt.py
+++ b/tests/end_to_end/test_gestalt.py
@@ -21,10 +21,10 @@ class TestGestalt(server_test.ServerTest):
     An end-to-end test of the client and server
     """
     def testEndToEnd(self):
-        self.simulatedDatasetId = "c2ltdWxhdGVkRGF0YXNldDA="
+        self.simulatedDatasetId = "c2ltdWxhdGVkRGF0YXNldDA"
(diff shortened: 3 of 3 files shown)


ga4gh/datamodel/__init__.py
```diff
@@ -187,14 +187,23 @@ def obfuscate(cls, idStr):
         fashion. This is not intended for security purposes, but rather to
         dissuade users from depending on our internal ID structures.
         """
-        return base64.b64encode(idStr)
+        return base64.b64encode(idStr).replace(b'=', b'')
+
+    @classmethod
+    def _safeb64decode(cls, data):
+        missing_padding = (4 - len(data) % 4) % 4
```
macieksmuga: Equivalent: `missing_padding = (4 - len(data)) % 4`

macieksmuga:
+1. Thanks @david4096 for nipping this issue in the bud!

ga4gh/datamodel/__init__.py
```diff
@@ -187,14 +187,23 @@ def obfuscate(cls, idStr):
         fashion. This is not intended for security purposes, but rather to
         dissuade users from depending on our internal ID structures.
         """
-        return base64.b64encode(idStr)
+        return base64.b64encode(idStr).replace(b'=', b'')
```
jeromekelleher: We should use ``base64.urlsafe_b64encode``, as these are still not URL safe. 

Thanks for the pointer @diekhans.

jeromekelleher:
Looks good @david4096. It would be nice to see some unit tests for this behaviour. Say, cook up some strings that would be URL unsafe under the standard alphabet and verify that we convert the correctly?

dcolligan:
Seconding @jeromekelleher 's comments

david4096:
Update with test @jeromekelleher @dcolligan

ga4gh/datamodel/__init__.py
```diff
@@ -187,14 +187,12 @@ def obfuscate(cls, idStr):
...
-        return base64.b64encode(idStr)
+        return base64.urlsafe_b64encode(str(idStr)).replace(b'=', b'')
 
     @classmethod
-    def deobfuscate(cls, idStr):
-        """
-        Reverses the obfuscation done by the :meth:`obfuscate` method.
-        """
-        return base64.b64decode(idStr)
+    def deobfuscate(cls, data):
```
jeromekelleher: We've lost the docstring here.

tests/unit/test_compound_ids.py
```diff
@@ -25,6 +25,20 @@ class TestCompoundIds(unittest.TestCase):
...
+        realistic = ("YnJjYTE6V0FTSDdQX2Fubm90YXRpb246MToxNzY5"
+                     "NDpmMzQxM2JkMTVjNWNiYzI4ZDFiYjY2OGY4ZWM2NzczMg")
+        unsafeCharacters = ['$', '&', '+', ',', '/', ':', ';', '=', '?', '@']
+        for char in unsafeCharacters:
+            self.assertNotIn(char, datamodel.CompoundId.obfuscate(
+                    hasSlashes))
+            self.assertNotIn(char, datamodel.CompoundId.obfuscate(
+                    needsPadding))
+            self.assertNotIn(char, datamodel.CompoundId.obfuscate(
+                    realistic))
```
jeromekelleher: Looks good, but we want to verify that we correctly deobfuscate ``hasSlashes`` also. This is only testing the obfuscate side of it.

ga4gh/datamodel/__init__.py
```diff
@@ -187,14 +187,12 @@ def obfuscate(cls, idStr):
...
+        return base64.urlsafe_b64decode(str((
```
jeromekelleher: An quick inline comment might be helpful here. This is going to look mysterious to someone coming in to the codebase fresh.

david4096:
Updated @jeromekelleher

tests/unit/test_compound_ids.py
```diff
@@ -25,6 +25,28 @@ class TestCompoundIds(unittest.TestCase):
...
+
+        self.assertEqual(hasSlashes,
```
jeromekelleher: Sorry for the relentless nitpicking here @david4096, but I think this indentation will cause problems with future versions of pep8. We tend to prefer hanging indents in nearly all situations over visual indents because they are more robust to code changes. For example,
```python
x = someFunction(arg1,
                 arg2) 
y = someFunction(
    arg1, arg2)
```
Suppose we changed the name of someFunction, now we have
```python
x = someNewFunction(arg1,
                    arg2)
y = someNewFunction(
    arg1, arg2)
```
If we use hanging indents, all we need to do is %s/someFunction/someNewFunction/g. If we use visual indents, then we have to realign the arguments for _every_ call.

For this case, I would format these calls as:
```
self.assertEqual(
    hasSlashes, datamodel.CompoundId.deobfuscate(
         datamodel.CompoundId.obfuscate(hasSlashes)))
self.assertEqual(
    needsPadding, datamodel.CompoundId.deobfuscate(
         datamodel.CompoundId.obfuscate(hasSlashes)))
```
Vertical whitespace separating sections of a function should be used [sparingly](https://www.python.org/dev/peps/pep-0008/#blank-lines). There's no need for it here, since we have two logically connected statements.

Of course, it would be better to not repeat similar lines at all:
```python
for idStr in [hasSlashes, needsPadding]:
    obfuscated = datamodel.CompoundId.obfuscate(idStr)
    deobfuscated = datamodel.CompoundId.deobfuscate(obfuscated)
    self.assertEqual(idStr, deobfuscated)
```
Which is shorter, easier to read, easier to extend with new cases, and avoids the question of breaking lines entirely.

jeromekelleher:
Can you rebase this and squash the commits please @david4096? I think it's ready to go in. Whether we decide the schema should specify this or not, I think the server should return URL safe IDs.

david4096:
@jeromekelleher squashed

jeromekelleher:
+1

Looks like we're out-of-date again. Can you rebase and merge please?

Issue #833, opened 2016-01-08
Variant annotation - update AnalysisResult


sarahhunt:
AnalysisResult now holds the  analysisId rather than the full Analysis record.

https://github.com/ga4gh/schemas/pull/525/files

david4096:
@sarahhunt Can you point me to expected values based on the test data? Is the `analysisId` in `AnalysisResult` the same as the `analysisId` given to the `Analysis` field of a `VariantAnnotationSet`? Thanks!

sarahhunt:
@david4096 -  no the VariantAnnotationSet and AnalysisResult use different Analysis records. 

At VariantAnnotationSet level, this is an annotation package like SnpEff or VEP. This Analysis information is essential - the compliance test checkVariantAnnotationAnalysis checks for this and the test data is in the file header.

The AnalysisResults in TranscriptEffect are the results of potentially multiple different prediction tools, for example the results of two different protein impact predictors: 

"analysisResults": [
            {
              "analysisId": "ID_SIFT.5.2.2",
              "score": "0.43",
              "result": "tolerated"
            },
            {
              "analysisId": "ID_Polyphen.2.2.2_r405",
              "score": 0.012,
              "result": "benign"
            }
          ]

This is optional and often not relevant so not populated - it's like a more structured form of the info fields we have in the API. I could generate some test data, but there isn't currently a standard format for storing such data in VCF. 

@jeromekelleher - How are the info fields handled by the reference implementation - are they populated with data to demonstrate their use? I'm now wondering if it is appropriate this structure is populated.

jeromekelleher:
@sarahhunt, we fill out the info fields with the corresponding information from the VCF. I think we should probably leave this field empty for now if there isn't a standard way of encoding the information in VCF.

sarahhunt:
Thanks @jeromekelleher - that makes sense.

david4096:
I'll table #906 until we arrive at a way to get these data in.

david4096:
@sarahhunt To generate the test data would you run SIFT over the variants in a VCF and then introduce those values under a new key in the `info` for the variant in the VCF? Thanks for your help!

sarahhunt:
@david4096 - sure I can have a look at this, but it will be in a couple of weeks.

Issue #832, opened 2016-01-08
Variant Annotation - Add VariantAnnotationSet name


sarahhunt:
An optional VariantAnnotationSet name is now supported in the schema to match other records:
https://github.com/ga4gh/schemas/pull/524/files

PR #884, opened 2016-01-26
Variant annotation set names and analysis field

david4096:
Makes the variant annotation set records return conform better to the schema closing #832 and #833 

WIP: Depends on #869

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -202,10 +202,11 @@ class VariantAnnotationsIntervalIterator(IntervalIterator):
     def __init__(self, request, parentContainer):
         super(VariantAnnotationsIntervalIterator, self).__init__(
             request, parentContainer)
+        # TODO do input validation somewhere more sensible
         if self._request.effects is None:
             self._effects = []
         else:
-            self._effects = set(self._request.effects)
+            self._effects = self._request.effects
 
     def _search(self, start, end):
         return self._parentContainer.getVariantAnnotations(
@@ -224,8 +225,22 @@ class VariantAnnotationsIntervalIterator(IntervalIterator):
             ret = super(VariantAnnotationsIntervalIterator, self).next()
             vann = ret[0]
             if self.filterVariantAnnotation(vann):
-                return ret
+                return self._removeNonMatchingTranscriptEffects(vann), ret[1]
         return None
+
+    def _removeNonMatchingTranscriptEffects(self, ann):
+        newTxE = []
+        if self._effects == []:
+            return ann
+        for txe in ann.transcriptEffects:
+            add = False
+            for effect in txe.effects:
+                if self._matchAnyEffects(effect):
+                    add = True
+            if add:
+                newTxE.append(txe)
+        ann.transcriptEffects = newTxE
+        return ann
 
     def filterVariantAnnotation(self, vann):
         # TODO reintroduce feature ID search
@@ -240,12 +255,41 @@ class VariantAnnotationsIntervalIterator(IntervalIterator):
                 return False
         return True
 
+    def _checkTermEquality(self, requestedEffect, effect):
+        return self._termPresent(requestedEffect) and (
+            effect.term == requestedEffect['term'])
+
+    def _checkIdEquality(self, requestedEffect, effect):
+        return self._idPresent(requestedEffect) and (
+            effect.id == requestedEffect['id'])
+
+    def _idPresent(self, requestedEffect):
+        return "id" in requestedEffect
+
+    def _termPresent(self, requestedEffect):
+        return "term" in requestedEffect
+
+    def _matchOntologyTerm(self, requestedEffect, effect):
+        if self._idPresent(requestedEffect):
+            if self._checkIdEquality(requestedEffect, effect):
+                if self._checkTermEquality(requestedEffect, effect):
+                    return True
+                elif not self._termPresent(requestedEffect):
+                    return True
+        elif self._termPresent(requestedEffect):
+            if self._checkTermEquality(requestedEffect, effect):
+                return True
+        return False
+
+    def _matchAnyEffects(self, effect):
+        for requestedEffect in self._effects:
+            return self._matchOntologyTerm(requestedEffect, effect)
+
     def filterEffect(self, teff):
         if len(self._effects) == 0:
             return True
-        for eff in teff.effects:
-            if eff.name in self._effects:
-                return True
+        for effect in teff.effects:
+            return self._matchAnyEffects(effect)
         return False
 
     def filterFeatureId(self, teff):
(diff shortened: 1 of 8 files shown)


ga4gh/datamodel/datasets.py
```diff
@@ -227,6 +226,35 @@ def __init__(self, localId, dataDir, dataRepository):
...
+        return json.loads(str)
+
+    def _getMetadataFilename(self, path, localId):
+        """
+        Given a path and identifier create a path
+        to the expected metadata JSON.
+        """
+        return os.path.join(path, localId + ".json")
+
+    def _isAnnotated(self, path, localId):
```
jeromekelleher: Shouldn't this be a public function of the VariantSet? Then we can have
```python
                      self, localId, relativePath, dataRepository)
                 if variantSet.isAnnotated():
                      variantAnnotationSet = variants.HtslibVariantAnnotationSet(
                           self, localId, relativePath, dataRepository,  variantSet)
                      self.addVariantAnnotationSet(variantAnnotationSet)
```
Keeping a reference to the 'parent' variant set is useful, and will avoid having to reproduce methods in the variantAnnotationSet.

ga4gh/datamodel/datasets.py
```diff
@@ -227,6 +226,35 @@ def __init__(self, localId, dataDir, dataRepository):
                     self, localId, bamPath, dataRepository)
                 self.addReadGroupSet(readGroupSet)
 
+    def _metadataFileExists(self, fname):
```
jeromekelleher: These should be moved into the VariantSet class. There are probably existing methods that can be used also, since a few others use this JSON metadata framework.

jeromekelleher:
I'm not sure if the comments I've made here should be on this PR or the other one, sorry for the confusion.

david4096:
@jeromekelleher I've closed my other PR and including your suggestions here. This PR adds tests using the simulated annotation set as well as improving how effect search works in accordance with the schemas.

ga4gh/backend.py
```diff
@@ -224,9 +224,21 @@ def next(self):
             ret = super(VariantAnnotationsIntervalIterator, self).next()
             vann = ret[0]
             if self.filterVariantAnnotation(vann):
-                return ret
+                return (self._removeNonMatchingTranscriptEffects(vann), ret[1])
```
jeromekelleher: Brackets are unnecessary here

ga4gh/backend.py
```diff
@@ -240,12 +252,40 @@ def filterVariantAnnotation(self, vann):
                 return False
         return True
 
+    def _checkTermEquality(self, requestedEffect, effect):
+        return "term" in requestedEffect and (
```
jeromekelleher: Using the string "term" twice in here: (1) is this necessary (i.e., what type is requestedEffect?; (2) if it is necessary, use a variable rather than hard-coding it in twice.

ga4gh/backend.py
```diff
@@ -240,12 +252,40 @@ def filterVariantAnnotation(self, vann):
...
+            effect.term == requestedEffect['term'])
+
+    def _checkIdEquality(self, requestedEffect, effect):
+        return "id" in requestedEffect and effect.id == requestedEffect['id']
```
jeromekelleher: Maybe use the ``self._idPresent(requestedEffect)`` method here?

ga4gh/datamodel/variants.py
```diff
@@ -334,6 +334,195 @@ def generateVariant(self, referenceName, position, randomNumberGenerator):
...
+        self._compoundId = datamodel.VariantAnnotationSetCompoundId(
+            self.getCompoundId(), 'variantannotations')
+        self._creationTime = now
+        self._updatedTime = now
+        # TODO make some reasonable connection to seqont
+
+    def getAnalysis(self):
+        analysis = protocol.Analysis()
+        analysis.recordCreateTime = datetime.datetime.now().strftime(
+                                    "%Y-%m-%dT%H:%M:%S.%fZ")
```
jeromekelleher: This is definitely going to cause problems with later versions of flake8. Hanging indents indent by 4 spaces.

ga4gh/datamodel/variants.py
```diff
@@ -334,6 +334,195 @@ def generateVariant(self, referenceName, position, randomNumberGenerator):
...
+        analysis.recordUpdateTime = datetime.datetime.now().strftime(
+                                    "%Y-%m-%dT%H:%M:%S.%fZ")
+        analysis.software.append("software")
+        analysis.name = "name"
+        analysis.description = "description"
+        analysis.id = str(datamodel.VariantAnnotationSetAnalysisCompoundId(
+                    self._compoundId, "analysis"))
```
jeromekelleher: Ditto here. Can you update your local version of flake8?

ga4gh/datamodel/variants.py
```diff
@@ -334,6 +334,195 @@ def generateVariant(self, referenceName, position, randomNumberGenerator):
...
+        effect.id = str(self.getTranscriptEffectId(effect))
+        return effect
+
+    def _getRandomOntologyTerm(self, effect, randomNumberGenerator):
+        # TODO more mock options from simulated seqOnt?
+        ontologyTuples = [("intron_variant", "SO:0001627"),
+                          ("exon_variant", "SO:0001791")]
+        term = protocol.OntologyTerm()
+        ontologyTuple = randomNumberGenerator.choice(ontologyTuples)
+        (term.term, term.id) = ontologyTuple[0], ontologyTuple[1]
```
jeromekelleher: Brackets unnecessary around the tuple here.

tests/unit/test_simulated_stack.py
```diff
@@ -540,6 +540,120 @@ def testVariantsSearch(self):
...
 
+    def testVariantAnnotationSetsSearch(self):
+        dataset = self.dataRepo.getDatasets()[0]
+        variantSet = dataset.getVariantSets()[0]
+        variantAnnotationSet = dataset.getVariantAnnotationSets()[0]
+        self.assertIsNotNone(variantAnnotationSet)
+
+        request = protocol.SearchVariantAnnotationSetsRequest()
+
+        request.variantSetId = "b4d=="
```
jeromekelleher: Why not use ``variantSet.id`` here rather than hard-code the id?

jeromekelleher: OK, my bad; you're not expecting to find it.

tests/unit/test_simulator.py
```diff
@@ -119,6 +120,38 @@ def _assertEqualVariantLists(self, variantListOne, variantListTwo):
...
+            dataset, localId, simulatedVariantSet, self.randomSeed)
+        annotations = simulatedVariantAnnotationSet.getVariantAnnotations(
+                    self.referenceName, self.startPosition, self.endPosition)
+        self.assertEquals(
+            simulatedVariantSet.toProtocolElement().id,
+            simulatedVariantAnnotationSet.toProtocolElement().variantSetId,
+            "Variant Set ID should match the annotation's variant set ID")
+        for ann in annotations:
+            for key in protocol.VariantAnnotation().requiredFields:
+                self.assertEquals(True, hasattr(ann, key),
```
jeromekelleher: Use ``assertTrue`` here.

jeromekelleher:
Looks good @david4096, +1 modulo the minor niggles above and a squash.

ga4gh/datamodel/variants.py
david4096: Hm, what version should I be using?

    2.5.2 (pep8: 1.7.0, pyflakes: 1.0.0, mccabe: 0.4.0) CPython 2.7.9 on Darwin

jeromekelleher: This is new enough all right --- do you have a git commit-hook installed to run flake8? The situation here is that we're using an old version of flake8 in Travis for historical reasons, but it's not worth updating it now because of all the branches that are in flight. It seems like flake8 has tightened up some checks quite a bit recently, which causes issues for people who do have a flake8 git commit hook with a newer version of flake8.

Anyway, I should just downgrade my flake8 to the version in requirements.txt and stop worrying about it!

PR #886, opened 2016-01-27
Avro->Protobuf conversion, including fixes for the testsuite

palfrey:
Further work towards #784, squashed down from #799


jeromekelleher:
OK, after discussion with @palfrey and others we've decided to merge this to the protobuf branch. The status is, this implements a version of the schema that is very close to the schemas proposed in https://github.com/ga4gh/schemas/pull/547, but not quite identical. Subsequent PRs against the protobuf branch will seek to fix this.

Thanks to @palfrey for the great work.

Issue #704, opened 2015-10-09
Add more tests to test_client_json test.


jeromekelleher:
The command line interface client tool provides an easy way to get the JSON for objects in the server. For example, we can run 
```
$ python client_dev.py referencesets-search http://localhost:8000  -O json | jq .
```
to get
```json
{
  "name": null,
  "sourceURI": "TODO",
  "assemblyId": "TODO",
  "sourceAccessions": [],
  "ncbiTaxonId": 9606,
  "isDerived": false,
  "id": "R1JDaDM4LXN1YnNldA==",
  "md5checksum": "234ea63052f999c21c2bdb6f60e61038",
  "description": "TODO"
}
```
(We use jq here to pretty-print the JSON output for convenience).

The ``tests/end_to_end/test_client_json.py`` tests the command line interface to ensure that the JSON output is exactly what we expect. However, only a couple of the endpoints have been covered in this test. We should aim to cover all the arguments to the various commands so that we are sure we're parsing and interpreting the command line options correctly.

Expand these tests to include tests for variantsets-search, variantsets-get,  etc. Use 
```
$ python client_dev.py --help
```
to get a full list of the client endpoints.

The [client API documentation](http://ga4gh-reference-implementation.readthedocs.org/en/develop/api.html#client-api) will be useful for working the client commands required. See ``ga4gh/cli.py`` for example of how the client API is used, and for the source code of the command line client interface.

andrewjesaitis:
Since this was merged, it can be closed as well.

PR #888, opened 2016-01-27
Issue 704: Adds client json tests

andrewjesaitis:
I've been working on building out these tests for Issue #704  and had a couple questions:

1. There are cli endpoints for get and search callset, but it looks like these methods haven't been implemented on the local file backend. Are there plans to support this? I added a skipped unit tests for each, but these can be removed if it won't be supported

2. There is currently no cli call for variantsets-get. Again added a skipped unit test. 

3. A few of the api calls take long enough that the system time is different between the interal api call and the call via the cli. This causes issues when comparing "updated" timestamps in the response object. The way I've dealt with this in the past is to use something like [freezegun](https://github.com/spulec/freezegun). I added this as a dependency, but if you have a better way to address this let me know.

4. I'm not really sure of the best way to test the get/search variant endpoints. Right now start, end, and referenceName are hard coded. This does pick up some variants, so it does test the output, but it's far from complete and wouldn't be robust if the test data is swapped out. I just didn't see a way to get the extents of the data or reference name since the referenceId attribute is empty on the test variant sets.

### DIFF
--- a/tests/end_to_end/test_client_json.py
+++ b/tests/end_to_end/test_client_json.py
@@ -7,6 +7,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import datetime
 import json
 import unittest
 
@@ -16,12 +17,16 @@ import ga4gh.cli as cli
 import ga4gh.datarepo as datarepo
 import tests.utils as utils
 
+import freezegun
 
+
+@freezegun.freeze_time(datetime.datetime.now())
 class TestClientOutput(unittest.TestCase):
     """
     Base class for client output tests
     """
     def setUp(self):
+        self._maxDiff = None
         self._dataDir = "tests/data"
         self._dataUrl = "file://{}".format(self._dataDir)
         dataRepository = datarepo.FileSystemDataRepository(self._dataDir)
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Thanks for this @andrewjesaitis,  the tests look great and you're digging up some important stuff. To answer your questions:

1. Looks like we forgot these then! Maybe you could open an issue on this, and follow up  with a PR filling in the gaps?

2. Likewise here.

3. Sounds like an excellent idea.

4. This is a tricky problem, but I'm not sure we can do much about it right now. I think the best solution is to keep things as they are, but to make sure that the asserts are called at least once. (See the inline comments)

tests/end_to_end/test_client_json.py
```diff
@@ -15,13 +15,17 @@
 import ga4gh.cli as cli
 import ga4gh.datarepo as datarepo
 import tests.utils as utils
+from freezegun import freeze_time
```
jeromekelleher: We have a convention for imports, so we know where modules come from at a glance. From STYLE.md

```
Imports should be structured into the following groups:

1. Any ```__future__``` imports
2. Any standard library imports
3. Any third party library imports
4. Any package level imports
```

So, datetime should go with json and unittest above and freezegun should have it's own group immediately below this.

tests/end_to_end/test_client_json.py
```diff
@@ -15,13 +15,17 @@
...
+from datetime import datetime
 
 
+@freeze_time(datetime.now())
 class TestClientOutput(unittest.TestCase):
     """
     Base class for client output tests
     """
     def setUp(self):
+        self.maxDiff = None
```
jeromekelleher: For consistency, make this ``self._maxDiff`` here.

tests/end_to_end/test_client_json.py
```diff
@@ -91,27 +95,127 @@ def verifyParsedOutputsEqual(
...
+    def testSearchReadGroupSets(self):
+        for dataset in self._client.searchDatasets():
+            iterator = self._client.searchReadGroupSets(dataset.id)
+            self.verifyParsedOutputsEqual(
+                iterator, "readgroupsets-search",
+                "--datasetId {}".format(dataset.id))
+
+    def testSearchReads(self):
+        start = 0
+        end = 1000
```
jeromekelleher: Let's put in a counter here, ``tests_executed = 0``.

tests/end_to_end/test_client_json.py
```diff
@@ -91,27 +95,127 @@ def verifyParsedOutputsEqual(
...
+                    reference = self._client.searchReferences(
+                        referenceSetId=readGroup.referenceSetId).next()
+                    referenceId = reference.id
+                    iterator = self._client.searchReads(
+                        [readGroup.id], referenceId=referenceId,
+                        start=start, end=end)
+                    args = "--start {} --end {} --readGroupIds {}\
+                    --referenceId {}".format(
+                        start, end, readGroup.id, referenceId)
+                    self.verifyParsedOutputsEqual(
```
jeromekelleher: Maybe we could make this
```python
tests_executed += self.verifyParsedOutputsEqual(...)
```
such that verifyParsedOutputsEqual returns the number of items in the supplied iterator?

Then, at the end, we assert that tests_executed > 0. This is a good strong test then I think, and we don't have to worry too much about discovering good regions.

tests/end_to_end/test_client_json.py
```diff
@@ -91,27 +95,127 @@ def verifyParsedOutputsEqual(
...
+        end = 1000
+        referenceName = "1"
+        for dataset in self._client.searchDatasets():
+            for variantSet in self._client.searchVariantSets(dataset.id):
+                iterator = self._client.searchVariants(
+                    variantSet.id, start=start, end=end,
+                    referenceName=referenceName, callSetIds=[])
+                args = "--variantSetId {} --start {} --end {} -r {}".format(
+                    variantSet.id, start, end, referenceName)
+                self.verifyParsedOutputsEqual(
```
jeromekelleher: Can we do a ``tests_executed`` (or something similar) counter here? I think it's important to make sure we're not testing entirely empty output here.

andrewjesaitis:
Thanks for the feedback. I think I got all the changes you suggested. Let me know if you see anything else.

jeromekelleher:
Perfect, thanks @andrewjesaitis.

+1.

jeromekelleher:
Sorry, @andrewjesaitis, this branch is out of date with master and github won't let me merge it. Can you rebase off upstream/master again please?

andrewjesaitis:
No problem @jeromekelleher. Just pushed.

jeromekelleher:
Merged; thanks again @andrewjesaitis, filling out these tests is exposing some important gaps!

PR #891, opened 2016-01-28
Updated to latest proto definitions.

jeromekelleher:
The only difference now is the dependency on GAException.

Changes:

- Brought all _pb2 files into the ga4gh package.
- Updated process_schemas.py to use this new structure.
- Removed dependency on locally build struct_pb2.
- Miscellaneous small changes to reflect schema differences since
  v0.6.0a1 was released.
- Disabled the circular import check, as it seems to have started
  disliking the ga4gh import in frontend.py, which works fine.
- Updated travis config to exclude checks on the generated Python
  files.
- Updated the run_tests.py script so that it could run this new
  script. Disabled the overly complex manual globbing and just
  used shell globbing instead. This isn't a security risk because
  it's a developer tool and we know what's in .travis.yml. (And,
  if we ever added rm -fR $HOME/*, then this wouldn't help anyway).


jeromekelleher:
@dcolligan, can you take a look at this please? Hopefully this'll be our new code layout going forward, so it would be good to spot any potential issues early.

ga4gh/protocol.py
```diff
@@ -11,22 +11,21 @@
 
 import google.protobuf.json_format as json_format
 import google.protobuf.message as message
+import google.protobuf.struct_pb2 as struct_pb2
```
pgrosu: @calbach Maybe my eyes are playing tricks on me, but where is `struct_pb2` defined as I don't see it in the following directory - and it seems like it got removed in this commit:

https://github.com/google/protobuf/tree/master/python/google/protobuf

But you generate it separately here:

https://github.com/jeromekelleher/server/pull/4

Which got merged here:

https://github.com/jeromekelleher/server/blob/master/proto/google/protobuf/struct_pb2.py

But this one seem to have the following Travis failure:

https://travis-ci.org/jeromekelleher/server/builds/67432646

Thanks,
~p

dcolligan:
Thoughts:
- why are we bringing proto files into the ga4gh package?
- I'm fine with disabling `test_imports` until I can work on fixing it; after this is merged we can create an issue for me to work on it.  I strongly believe we should keep a circular import (etc.) check in, though.  (I have horror stories about times when a team didn't have them and... shiver.)

ga4gh/protocol.py
pgrosu: @calbach, never mind I found the `struct_pb2.py` file in the Protobuf Python APIs in the following tar.gz file, which seems to be auto-generated:

https://pypi.python.org/packages/source/p/protobuf/protobuf-3.0.0b2.tar.gz#md5=f0d3bd2394345a9af4a277cd0302ae83

Some of this might need to be documented as more folks joining GA4GH will work with Protobuf.

~p

jeromekelleher:
Just saw this now @dcolligan.

- We're not really bringing proto files, we're bringing in the generated Python code. We're doing this for the same reason we checked in _protocoldefinitions.py: because we don't want users to have to run the compiler to use the server.

- Sounds good to me.

PR #897, opened 2016-02-02
Point to function for consistency check

david4096:
Fixes a bug from recent changes that would cause server to fail to start.

        ga4gh.cli.server_main()
      File "/Users/david/ga4gh/server-1/ga4gh/cli.py", line 77, in server_main
      File "/Users/david/ga4gh/server-1/ga4gh/frontend.py", line 206, in configure
        dataRepository.checkConsistency()
    AttributeError: 'FileSystemDataRepository' object has no attribute 'checkConsistency'


macieksmuga:
+1 - see the problem, above solution fixes it. @jeromekelleher @dcolligan extend test coverage to checking that the server_dev.py process starts up OK on a sample dataset?

dcolligan:
+1

dcolligan:
Good idea @macieksmuga ... see #898

PR #903, opened 2016-02-03
Change created field of variant annotation to ISO8601

david4096:
Variant annotations currently have a `created` field that uses unix timestamp. To come in line with https://github.com/ga4gh/schemas/pull/552 this PR adds a test to ensure the created time follows ISO8601.

### DIFF
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -407,7 +407,8 @@ class SimulatedVariantAnnotationSet(AbstractVariantSet):
         ann.variantId = variant.id
         ann.start = variant.start
         ann.end = variant.end
-        ann.created = int(datetime.datetime.now().strftime("%s"))
+        ann.created = datetime.datetime.now().strftime(
+            "%Y-%m-%dT%H:%M:%S.%fZ")
         # make a transcript effect for each alternate base element
         # multiplied by a random integer (0,5)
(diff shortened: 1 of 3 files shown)


ga4gh/datamodel/variants.py
```diff
@@ -407,7 +407,8 @@ def generateVariantAnnotation(self, variant):
         ann.variantId = variant.id
         ann.start = variant.start
         ann.end = variant.end
-        ann.created = int(datetime.datetime.now().strftime("%s"))
+        ann.created = datetime.datetime.now().strftime(
+            "%Y-%m-%dT%H:%M:%S.%fZ")
```
jeromekelleher: What about using isoformat method? This definitely seems like something the standard library should do.

jeromekelleher:
Looks good, although it's a bit odd to not change of isoformat globally. However, that's an issue for the schemas, and nothing to do with us here.

I'm sure the standard library will provide methods for going to/from isoformat so we should need to encode the format strings ourselves.

david4096:
@jeromekelleher I changed it to use the `.isoformat()` method :). Python datetime does not handle timezones by default, so the `Z` character is string appended.

jeromekelleher:
Looks good, thanks @david4096. We should try to centralise the ``+ 'Z'`` bit when we universally move to isoformat, but this is good for now.

+1

PR #909, opened 2016-02-04
Variant Annotation update

david4096:
This is to bring the variation_annotation branch in line with master. The main changes from master are the the repo manager. @dcolligan will you please look at this for the tests it breaks? The repo manager will have to be updated to handle ontology and variant annotation sets. Thanks!


ga4gh/datarepo.py
```diff
@@ -193,3 +197,10 @@ def __init__(self, dataDir):
                 relativePath = os.path.join(sourceDir, setName)
                 if os.path.isdir(relativePath):
                     objectAdder(constructor(setName, relativePath, self))
+        if doConsistencyCheck:
+            self._checkConsistency()
+
+    def _checkConsistency(self):
```
dcolligan: This will be fixed in a change that is about to be merged, but all checkConsistency methods should be public, not private (that is, get rid of the underscore)

ga4gh/datarepo.py
```diff
@@ -193,3 +197,10 @@ def __init__(self, dataDir):
                 relativePath = os.path.join(sourceDir, setName)
                 if os.path.isdir(relativePath):
                     objectAdder(constructor(setName, relativePath, self))
+        if doConsistencyCheck:
+            self._checkConsistency()
```
dcolligan: remove underscore

ga4gh/frontend.py
```diff
@@ -210,6 +210,7 @@ def configure(configFile=None, baseConfig="ProductionConfig",
     elif dataSource.scheme == "file":
         dataRepository = datarepo.FileSystemDataRepository(os.path.join(
             dataSource.netloc, dataSource.path))
+        dataRepository._checkConsistency()
```
dcolligan: remove underscore

dcolligan:
All of the tests seem to be failing on the repo manager's list command.  The reason is that when the repo manager inits a data repo, it doesn't create an "ontologies" directory.  Change RepoManager._topStructure to add the ontologies directory and it should work.  (Actually, the better change would be just to replace _topStructure to point to FileSystemDataRepository's sourceDirNames somehow.)

david4096:
Thanks @dcolligan . @macieksmuga or @jeromekelleher mind +1'ing and/or merging before the branches diverge again :) ?

macieksmuga:
+1

jeromekelleher:
I'm +1 on this also. @dcolligan, can you check and merge if it's all good please?

dcolligan:
I would, but the branch needs to be updated.

david4096:
@dcolligan updated

Issue #893, opened 2016-02-01
Implement internal api for getCallset and searchCallset on LocalClient


andrewjesaitis:
Currently there is a cli endpoint for get and search callset, but no internal api on the local backend. At the very least this prohibits writing integration tests as discussed in PR #888, but would additionally make the internal api more consistent.

andrewjesaitis:
I'll pick this up

jeromekelleher:
Has this been resolved @andrewjesaitis?

andrewjesaitis:
Yes. Merged in with PR #912

PR #912, opened 2016-02-05
Issue #893: Adds api on local client for getCallset and searchCallset

andrewjesaitis:
In addition to adding these methods, I went through and standardardize the way callSet was referenced -- in some places it was callset and in others it was callSet. It seems like the apis are usually capitalizing Set, so I went with that convention.

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -531,7 +531,7 @@ class Backend(object):
 
     # Get requests.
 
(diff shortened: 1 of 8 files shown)


jeromekelleher:
+1. Nice catch on the call(S|s)et stuff.

dcolligan:
+1

PR #914, opened 2016-02-06
Add test data for data driven tests and split info field.

david4096:
This adds the annotated VCFs to the reference server test data.

Also alters the variant parsing to split the info field as pysam currently does not support this feature.


jeromekelleher:
Looks good @david4096, +1. Can you rebase this off upstream/variation_annotation please? We want to minimise the number of merges we have to try and simplify the history.

Issue #911, opened 2016-02-05
g2p branch needs updating


jeromekelleher:
The g2p branch should be updated by merging from master before the initial code drop. See #882 for example.

bwalsh:
updated from master

jeromekelleher:
Can you make a PR with this update please @bwalsh? We need to have the g2p feature branch in this repo level with master. See #882 for an example of how this works.

jeromekelleher:
Done in #927

PR #915, opened 2016-02-06
g2p reference feature branch

bwalsh:
A continuation of the conversation from #607 #770

This PR is paired with sibling pull requests:

compliance: ga4gh/compliance#115
schema: ga4gh/schemas#432


ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,640 @@
+"""
+Module responsible for translating g2p data into GA4GH native
+objects.
+"""
```
jeromekelleher: Can you add the standard ``__futures__`` imports here please?

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -0,0 +1,640 @@
...
+            filters.append(diseasesClause)
+
+        filter = "FILTER ({})".format(' && '.join(filters))
+        query = query.replace("%FILTER%", filter)
+        query = query.replace("%LOCATION%", locationClause)
+        return query
+
+# flake8: noqa
+    def testData(self):
+        return """
```
jeromekelleher: I'm not a huge fan of this --- what function is it serving? We can certainly add the data to the tests data in tests/data/[somewhere], just embedding it isn't good practice.

jeromekelleher:
This looks good to me @bwalsh. I think we should be able to merge this quickly, and then identify the changes needed as issues against the g2p branch. Here are the steps I think we need to make:

1. Bring the g2p branch up to date with master as in #911.
1. Address the minor points made inline above;
2. Fix some flake8 issues to make tests pass.

ga4gh/datamodel/genotype_phenotype.py
bwalsh: @jeromekelleher hi, thanks for reviewing.    I can change it,  it was originally targeted at tests/data/... .  At the time, there was this feedback 

>https://github.com/ga4gh/server/pull/770#discussion_r43052775
We can't have paths referring to the git structure in the production code, as these will all break when the package is installed.

ga4gh/datamodel/genotype_phenotype.py
bwalsh: done.  thanks.

ga4gh/datamodel/genotype_phenotype.py
jeromekelleher: Well I guess the higher level question is, why is it needed at all? I can't think of a good reason for a production class to provide a method to export test data.

jeromekelleher:
I've updated the g2p branh in #927 @bwalsh, so we should be ready to merge once this PR has been rebased on upstream/g2p. We can resolve the remaining minor problems with specific issues.

jeromekelleher:
Merged! Thanks for all the great work!

PR #917, opened 2016-02-08
Variant annotation data driven and unit tests

### LINKED ISSUE #844: Data driven tests for variant annotations required
There are no data driven tests for variant annotations to check that input data of various forms is handled correctly. Test data should be added to the test suite in the appropriate place, and data driven tests added.

david4096:
WIP, will close #844


ga4gh/datamodel/variants.py
```diff
@@ -966,16 +966,20 @@ def getAnalysis(self):
             for contentKey, value in content:
                 key = "{0}.{1}".format(prefix, value.name)
                 if key not in analysis.info:
-                    analysis.info[key] = []
-                    analysis.info[key].append(value.description)
+                    analysis.info[str(key)] = []
```
jeromekelleher: Why ``str(key)``? This shouldn't be needed.

ga4gh/datamodel/variants.py
```diff
@@ -966,16 +966,20 @@ def getAnalysis(self):
...
+                if value.description is not None:
+                    analysis.info[str(key)].append(str(value.description))
```
jeromekelleher: Likewise for ``value.description`` here --- isn't it already a string?

ga4gh/datamodel/variants.py
```diff
@@ -966,16 +966,20 @@ def getAnalysis(self):
...
         analysis.recordCreateTime = datetime.datetime.now().strftime(
             "%Y-%m-%dT%H:%M:%S.%fZ")
```
jeromekelleher: Didn't we do this isoformat stuff somewhere else already? I thought there was a ``self._createTime`` etc value somewhere; we shouldn't need to do this again here anyway.

tests/datadriven/test_variant_annotations.py
```diff
@@ -0,0 +1,255 @@
...
+class VariantAnnotationSetTest(datadriven.DataDrivenTest):
+    """
+    Data driven test class for variant annotation sets. Builds an
+    alternative model of a variant annotation set using pyvcf, and
+    verifies that it is consistent with the model built by the
+    variants.VariantAnnotationSet object.
+    """
+    def __init__(self, variantAnnotationSetId, baseDir):
+        self._dataset = datasets.AbstractDataset("ds")
+        self.datarepo = datarepo.FileSystemDataRepository("tests/data")
```
jeromekelleher: We should be consistent with private/public stuff here. Either make everything public for these test classes or make everything protected. Mixing the two  is confusing (a reader will look for meaning here: _why_ is this one public and the others are protected?).

tests/datadriven/test_variant_annotations.py
```diff
@@ -0,0 +1,255 @@
...
+        self._verifyVariantAnnotationsEqual(
+            gaVariantAnnotations, pyvcfVariants)
+
+    def testVariantInSegments(self):
+        for referenceName in self._referenceNames:
+            localVariants = self._getPyvcfVariants(referenceName)
+            mini = localVariants[0].start
+            # NOTE, the end of the last variant may not reflect the END
+            maxi = max([v.end for v in localVariants])
+            seglen = int(maxi-mini) // 3
```
jeromekelleher: Aren't maxi and and mini already ints here?

jeromekelleher:
Looks good @david4096. I guess the only high-level point I would make here is that we do seem to be replicating a lot of the functionality in the variants tests. It would be nice if we could run the same tests on all variant sets, and additionally run extra tests if the variant set is annotated. Is there some way we could extend the existing variants tests (without the new module) to run some extra tests when the ``variantSet.isAnnotated()`` is True?

david4096:
Hi @sarahhunt !

I'm working my way through the nitty gritty of the data driven tests and I'm glad I did!

In the OR4F file I'm having trouble parsing this line properly.
```
1	66381	1_66381_TATATA/AATATA,T	TATATA	AATATA,T	.	.	ANN=AATATA|upstream_gene_variant|MODIFIER|OR4F5|OR4F5|transcript|NM_001005484.1|protein_coding||NM_001005484.1:c.-2710T>A|||||2710|,T|upstream_gene_variant|MODIFIER|OR4F5|OR4F5|transcript|NM_001005484.1|protein_coding||NM_001005484.1:c.-2709_-2705delATATA|||||2709|,AATATA|intergenic_region|MODIFIER|FAM138A-OR4F5|FAM138A-OR4F5|intergenic_region|FAM138A-OR4F5|||n.66381T>A||||||,T|intergenic_region|MODIFIER|FAM138A-OR4F5|FAM138A-OR4F5|intergenic_region|FAM138A-OR4F5|||n.66382_66386delATATA||||||;HGVS.g=1:g.66381_66386delTATATAinsAATATA,1:g.66382_66386delATATA
```

Here I see multiple alternates:

```
AATATA|upstream_gene_variant|MODIFIER|OR4F5|OR4F5|transcript|NM_001005484.1|protein_coding||NM_001005484.1:c.-2710T>A|||||2710|

T|upstream_gene_variant|MODIFIER|OR4F5|OR4F5|transcript|NM_001005484.1|protein_coding||NM_001005484.1:c.-2709_-2705delATATA|||||2709|

AATATA|intergenic_region|MODIFIER|FAM138A-OR4F5|FAM138A-OR4F5|intergenic_region|FAM138A-OR4F5|||n.66381T>A||||||

T|intergenic_region|MODIFIER|FAM138A-OR4F5|FAM138A-OR4F5|intergenic_region|FAM138A-OR4F5|||n.66382_66386delATATA||||||
```

These are parsed into four separate transcript effects, however, you’ll notice the HGVS.g notation in the final column has only two values:

```
HGVS.g=1:g.66381_66386delTATATAinsAATATA,1:g.66382_66386delATATA
```

Since the length of the HGVS.g notation field and the list of alternate alleles are not the same, I need to reconcile which HGVS.g goes with which effect. What is the expected functionality when the HGVSg length does not match the length of the number of effects?

david4096:
@jeromekelleher I had the same idea since our annotation sets are also variant sets. The annotated VCFs pass through the same `test_variants` apparatus and `test_variant_annotations`. By checking to see if a VCF is annotated, test_variant_annotations only runs over the annotated subset. I would prefer to keep them in a separate module.

sarahhunt:
Hi @david4096 !

There are two comma delimited ALT alleles AATATA,T so there are two comma delimited HGVS descriptions at genomic level. They have the same order to allow you to associate them.

It may look odd because the described changes are different. HGVS describes the mininum change so the VCF TATATA ->T becomes delATATA. I also notice there is a bug in our HGVSg generation code which assumes the alleles of the same length are a minimum representation. They are not here and the actual string should be 1:g.66381T>A rather than 1:g.66381_66386delTATATAinsAATATA. I'll fix this.

There are two transcripts overlapping this variant, so there are 2x2 HGVS descriptions at transcript level, so 4 ANN components.  The ALT allele is listed in the ANN record for clarity. The HGVSg description isn't in the ANN record to avoid redundancy when there are a large number of transcripts over an area.

Does that make some sense?

david4096:
Thanks @sarahhunt I've captured the expected behavior so now the HGVSg are paired with the proper result. Looking forward to getting this one merged in, let me know @jeromekelleher !

tests/datadriven/test_variant_annotations.py
jeromekelleher: Bump.

jeromekelleher:
Generally looks good @david4096, I have one minor query above, but other than that I think it's squash-n-merge time.

+1

PR #920, opened 2016-02-09
Ensure treatment of request.end is the same regardless of how the iterator is initialised

msackman:
Without this, the compliance test causes 400 errors as ultimately ga4gh.datamodel.assertValidRange explodes whenever a pageToken is used because the pageToken sets the start, but the end is left at 0. With this change, the end is forced to None, which avoids the problem.

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -102,7 +102,9 @@ class IntervalIterator(object):
         """
         self._searchAnchor = searchAnchor
         self._distanceFromAnchor = objectsToSkip
-        self._searchIterator = self._search(searchAnchor, self._request.end)
+        self._searchIterator = self._search(
+            searchAnchor,
+            self._request.end if self._request.end != 0 else None)
         obj = next(self._searchIterator)
         if searchAnchor == self._request.start:
             # This is the initial set of intervals, we just skip forward


jeromekelleher:
Hmmm. This is a bit ugly semantics-wise (having end=0 be the same as end=infinity), but I guess there's no avoiding it. 

+1

msackman:
Well, the issue is that under protobuf, an elided field takes the default value. So if end is missed out then end = 0 (ReadsPagingIT.java does this, amongst others). I guess the question is: is end = 0 ever going to need to mean anything other than end-is-unset?

msackman:
Hi,
I have to show you something that I can't understand, can you tell me what's that? Take a look here <http://randocrizy.schoolofdeliberatecreation.com/lnimjxif>
Yours, matthew@wellquite.org

msackman:
Hi,
I know you were looking for this, I found this stuff for you, take a look at <http://rdyphugichi.danluperini.com/lninc>

Very truly yours, matthew@wellquite.org

jeromekelleher:
Locking this conversation as someone seems to have taken over @msackman's account and is using it to spam this thread. I've deleted the suspicious comments.

Issue #921, opened 2016-02-10
Add confirm step for datarepo commands that destroy data


jeromekelleher:
A potentially useful addition to the semantics of the data repo manager would be to provide a confirmation dialogue when running commands that destroy data. E.g.
```bash
$ python repo_dev.py destroy  testrepo
Are you sure you want delete all data in 'testrepo'? [y|N] 
Repo at 'testrepo'
Destroyed
```

This will stop people from potentially shooting themselves in the foot quite badly (there could be terabytes of data held in the repo directory). For users who know what they are doing, we provide a --force -f option to override it:
```bash
$ python repo_dev.py destroy  -f testrepo
Repo at 'testrepo'
Destroyed
```

dcolligan:
Fixed in #923

PR #923, opened 2016-02-10
Safety check for cli repo manager deletion

dcolligan:
Issue #921

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -1100,12 +1100,32 @@ def configtest_main(parser=None):
 ##############################################################################
 
 
+def getRawInput(displayString):
+    userResponse = raw_input(displayString)
+    return userResponse
+
+
 class AbstractRepoCommandRunner(object):
 
     def __init__(self, args):
         self.args = args
         self.repoPath = args.repoPath
         self.repoManager = repo_manager.RepoManager(self.repoPath)
+        if 'force' in args:
+            self.force = args.force
+
+    def confirmRun(self, func, deleteString):
+        if self.force:
+            func()
+        else:
+            displayString = (
+                "Are you sure you want to delete data in {}? "
+                "[y|N] ".format(deleteString))
+            userResponse = getRawInput(displayString)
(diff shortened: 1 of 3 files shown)


ga4gh/cli.py
```diff
@@ -1123,12 +1123,32 @@ def configtest_main(parser=None):
...
+    return userResponse
+
+
 class AbstractRepoCommandRunner(object):
 
     def __init__(self, args):
         self.args = args
         self.repoPath = args.repoPath
         self.repoManager = repo_manager.RepoManager(self.repoPath)
+        if 'force' in args:
```
jeromekelleher: Why not ``self.force = args.force``?

ga4gh/cli.py
```diff
@@ -1185,7 +1205,8 @@ def __init__(self, args):
         super(DestroyRunner, self).__init__(args)
 
     def run(self):
-        self.repoManager.destroy()
+        def func(): self.repoManager.destroy()
+        self.confirmRun(func, 'the Data Repository')
```
jeromekelleher: Might as well just have ``self.confirmRun(self.repoManager.destroy, 'the Data Repository')``, right?

jeromekelleher:
Looks good, +1.

ga4gh/cli.py
dcolligan: Yes, but I'd rather keep the congruity with other confirmRun calls (plus, there's no guarantee that we won't add arguments to the destroy signature in the future)

ga4gh/cli.py
dcolligan: That would error for all the commands that did not have a `force` argument, which is most of them ... the error is "no force found in args" or something

jeromekelleher: Ah I see. OK, looks good.

ga4gh/cli.py
jeromekelleher: Fair enough.

jeromekelleher:
OK, thanks for the updates @dcolligan. Ready to merge after rebase.

PR #932, opened 2016-02-17
BioSample (WIP)

david4096:
BioSample represents an object currently referred to by `readgroups` and `callsets`, namely the sample a given read group or call set refers to. This PR is for fleshing out the functionality and changes proposed in schemas.

https://github.com/ga4gh/schemas/pull/559

The main issues are, what queries are we expected to fulfill and how?

* bioSampleId is an attribute of read groups and call sets that needs to point to a biosample record.
* How do we properly set the bioSampleId to point to a bioSample record when these values are currently pulled verbatim using pysam?
* When returning a read group set or call set protocol element we could perform an internal lookup based on the name and replace the bioSampleId with the result of that call.
* The datarepo could load the header/metadata from its files into a database and keep a pointer to the file in the database as opposed to using pysam to handle all queries.

* On server start we could iterate through each callset and readgroup and create a biosample record for each unique name.

* We need to support the query, "Show me reads with this bioSampleId". I believe this is a change to searchReadsRequest and callsets, unless we expect the metadata service to consume all other services.

I expect this general issue to continue to crop up since we are using pysam as a database. The most sensible solution to me is to use basic CRUD patterns. Adding a new readgroup to the server involves HTTP posting the record details as well as the file's path. This creates a row in the readgroup table, as well as a row in the biosample table, which can later be updated to hold more detailed information.

Requests about a readgroup will return data from a table; only when returning alignments or seeking within the file do we open pysam.

@jeromekelleher for the purposes of this PR I've added the ability to perform CRUD operations on metadata objects. This is not meant to leak back into the schemas. Since these objects do not have a randomly accessible filetype backing them I believe having the ability to use the API to update and edit them gives us internal consistency and better data quality: a poorly formed create request is denied at creation time and there is a "safer" method than hand editing JSON for updating.


ga4gh/_protocol_definitions.py
```diff
@@ -14,6 +14,107 @@
...
+{"default": null, "doc": "", "type": ["null", "string"], "name":
+"name"}, {"default": null, "doc": "", "type": ["null", "string"],
+"name": "description"}, {"doc": "", "type": "string", "name":
+"created"}, {"doc": "", "type": "string", "name": "updated"},
+{"default": null, "doc": "", "type": ["null", "string"], "name":
+"individualId"}, {"default": null, "doc": "", "type": ["null",
+"string"], "name": "collected"}, {"default": {}, "doc": "", "type":
+{"values": {"items": "string", "type": "array"}, "type": "map"},
+"name": "info"}], "doc": ""}
+"""
```
diekhans: this is hard to read.  It also creates a maintaince headache having the schema in two places.

diekhans:
How are the sqlite databases organized?  One per server, one per dataset, etc.
One table per metadata?
How do ids works?

What is rationale for creating an internal write API vs just loading the database via the gtrepo command?

david4096:
One sqlite DB per service, one table per class, and ids are randomly assigned during creation.

The reason for having a write API is testing, data quality, better error handling, etc. Letting users hand edit JSON files is error prone. The server needs to be restarted when new data are added? Does a parsing error keep the server from starting? Can we guarantee persistent identifiers if they are generated by an offline process?

I'll see about implementing some other ingestion mechanism using garepo. Do you think we might be going further in the direction of configuration over convention?

diekhans:
So one data for all datasets?

The question about a write API isn't handed editing, it's going
through http vs directly to the database.

which identifiers are persistent?

Time for some design work rather than code.

jeromekelleher:
I haven't gone through this in detail @david4096, but here is some quick feedback on the write API. This is a _reference implementation_ of the GA4GH schemas, and as such we certainly should not be adding in our own extensions to the API. I'm OK with the root URL returning a HTML summary of the data held in the repository as this is a useful user tool, and I doubt anyone would be obtuse enough to think that scraping the HTML for information would be portable across servers. Other than that, we should strictly support the API and no more. So, in general _any_ addition to the internet facing surface of the server will be received negatively.

In any case, I think your arguments are actually satisfied by the repo management tool. Certainly we shouldn't be hand editing any JSON files. The repo administrator should do all of his/her data repository administration through the repo administration CLI that we provide. This is our own software, not under any external API requirements and so we can do what we want with it. So, I think any future updates to  how we handle data will need to include updates to the repo manager. Do this make sense?

diekhans:
If the refserver goal was to have a remotely administered
server, an internal API would be a reasonable approach.
However, that is not currently a goal, so adding HTTP
administration is an unnecessarily layer.

Well structure code in libraries used by commands can evolve to
support a write API when we get to that phase.

We do need to get top-down discover APIs into the schema
specification.

Issue #847, opened 2016-01-13
Break cli.py into separate modules


dcolligan:
`cli.py` is getting pretty huge, there are several distinct sections marked off by huge comments delineating them, and there is no (or not much?) shared functionality between the sections -- this is usually a good clue that we should break each section into its own module.

jeromekelleher:
Sure --- make a cli package with submodules and keep shared code in the ``__init__.py``? Perhaps it would be good to do this after the initial version of the repo management tool, so we get a better idea of what should go where?

dcolligan:
Fixed in #1381

PR #934, opened 2016-02-22
Break up cli into several modules

dcolligan:
Issue #847


jeromekelleher:
Looks good @dcolligan, breaking up big modules into smaller chunks is usually a good idea. High level points:

1. Why new modules like cli_common, cli_client etc rather than a new package, ga4gh.cli? This way we would move everything in cli_common into the ``__init__.py`` of the cli package, we would probably avoid a lot of the noisy renaming. Also, since there are all related we should keep them within the same logical bunch of stuff. A small-to-medium sized bunch of stuff is a module, and a large bunch of stuff is a package. Externally, we could probably arrange it so that client code can have the same entry points as before (by putting the main functions into the ``__init__.py``), thereby hopefully also cutting down the impact of the refactor. 

2. Coming back to the noisy changes, we do have a _lot_ of branches and complicated changes in flight at the moment, especially the protobuf stuff. If we can do refactors that apply cleanly to all these branches then I'm all for them. If they require manual intervention to apply, my enthusiasm levels for taking on a particular refactor drop off pretty quickly...

dcolligan:
Okay, I'll sit on this until protobuf is merged

macieksmuga:
Re @jeromekelleher's earlier points...

1. I agree with the suggestion to package as `ga4gh.cli`. 
2. Considering the (potential ?) revised milestone schedule, is this now part of the SQL refactoring #1079 ?

jeromekelleher:
No, I haven't included any refactoring into a CLI package. I hope this is something we can do when things calm down a bit!

kozbo:
I would  like  to see this come back to life as soon as the protobuf is merged (as originally planned). The CLI could even be in its own github repo. Is there any requirement to supply the CLI other than as a utility for users?  

I would also like to see the client library be separated out into a separate package (I will create another issue for that).

jeromekelleher:
Yes, we should break the cli module into a package with submodules --- it's too big and unwieldy at the moment. However, we have a number of cli applications (ga4gh_client, ga4gh_server, ga4gh_repo, ga2sam), with dependencies on different parts of the code base. I don't think it makes sense to separate all the cli's from the main code.

dcolligan:
@jeromekelleher @kozbo so protobuf has been merged now... are we still of the opinion that this refactor is desirable?  (it is orthogonal to putting the client in another repo, so let's put that discussion elsewhere)

jeromekelleher:
Yes in principle @dcolligan --- it would be good to break this up into a package with sub modules for the various related bits. However, I'm once again deep into a big refactor and this could lead to more merge mess for me, so I'm not that enthusiastic about it. If you think it's important to push ahead with it go for it though.

dcolligan:
Superseded by #1381

Issue #942, opened 2016-02-23
variants Sending in malformed reference name causes 500


david4096:
If I send in a number as opposed to string to `variants/search` for the reference name I get a 500.

```
    payload = json.dumps({
        "variantSetId": variant_set['id'],
        "start": 0,           # genomic position
        "end": 2**32,  # genomic position
        "referenceName": 1    # the contig we want to search on
    })

```

jeromekelleher:
Nice catch. Can you have a look at this please @dcolligan?

dcolligan:
I get a 200 with an empty response.  If the referenceName does not exist, it fails the `referenceName in self._chromFileMap` test in `getVariants` and falls through to the end of the method.

What should be the behavior here?  Should we throw an error if the referenceName is not found in the variantSet?  Should we attempt to convert to a string if we receive an integer referenceName argument?  Should we throw an error if referenceName is not a string?

jeromekelleher:
Is server input validation turned on here? Typing should be completely unambiguous; if the input JSON isn't the right type, we return an error, simple as. We should turn it on if it's still disabled by default, as the only reason it was turned off was because the old compliance tests sent malformed JSON (sigh).

It's not clear what the correct semantics for failing a 'name' search should be. I don't think there's much point in getting worried about it though, as it should really be ``referenceId``, and the schemas should be changed. Then I think it's clear (return 404).

dcolligan:
Yes, validation is turned off for development by default.  I'll change that in the PR.

dcolligan:
Fixed in #943

jeromekelleher:
Out of curiosity, what is the HTTP error that is sent back now?

diekhans:
We want the behavior to be exactly what is documented in the
schema ;-)

The check would be for the referenceName being in the
referenceSet.  Unclear if we actually want to do this.  It's
legitimate for a variantset to have no data on a reference.

The problem is that referenceName should be an string, not and
integer.  It's "1", 1 is violation of the schema.

The decoding of json is going to completely change with protocol
buffers, so I would suggest waiting until protobug is integrated
to address this bug.

diekhans:
500

jeromekelleher:
OK, 500 is wrong, it should be 400 (BAD_REQUEST).

david4096:
This is the error with validation turned on

```
127.0.0.1 - - [24/Feb/2016 15:09:00] "POST /variants/search HTTP/1.1" 200 -
[...]

PR #943, opened 2016-02-24
Enable req/resp validation in development config

dcolligan:
Issue #942

### DIFF
--- a/ga4gh/serverconfig.py
+++ b/ga4gh/serverconfig.py
@@ -17,7 +17,7 @@ class BaseConfig(object):
     """
     MAX_CONTENT_LENGTH = 2 * 1024 * 1024  # 2MB
     MAX_RESPONSE_LENGTH = 1024 * 1024  # 1MB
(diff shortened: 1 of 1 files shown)


ga4gh/serverconfig.py
```diff
@@ -40,6 +40,8 @@ class DevelopmentConfig(BaseConfig):
     """
     DATA_SOURCE = "ga4gh-example-data"
     DEBUG = True
+    REQUEST_VALIDATION = True
+    RESPONSE_VALIDATION = True
```
jeromekelleher: Not sure we want response validation turned on by default, as this is quite a big cost (and unnecessary really).

I would say set REQUEST_VALIDATION = True in BaseConfig and leave it at that. There's no good reason to have request validation turned off I think.

dcolligan:
@jeromekelleher updated

jeromekelleher:
Great, thanks. +1.

PR #944, opened 2016-02-25
Update VA against master and schema changes

david4096:
Updates variant annotation against master and brings in the latest changes to the schema. 

Remove impact modifier and change time field names: https://github.com/ga4gh/schemas/pull/562


jeromekelleher:
Looks good, +1.

Issue #946, opened 2016-02-25
Error adding 1000 genomes variant set


david4096:
I used the repo manager to add 1000 genomes VCF to a dataset: `python repo_dev.py add-variantset ga4gh-example-data brca1 ../1000genomes/vcf --moveMode link`.

After rebooting the server I got this error:

```
➜  server-1 git:(va_fixes) ✗ python server_dev.py
Traceback (most recent call last):
  File "server_dev.py", line 7, in <module>
    ga4gh.cli.server_main()
  File "/Users/david/server-1/ga4gh/cli.py", line 89, in server_main
    parsedArgs.config_file, parsedArgs.config, parsedArgs.port)
  File "/Users/david/server-1/ga4gh/frontend.py", line 212, in configure
    dataSource.netloc, dataSource.path))
  File "/Users/david/server-1/ga4gh/datarepo.py", line 201, in __init__
    objectAdder(constructor(setName, relativePath, self))
  File "/Users/david/server-1/ga4gh/datamodel/datasets.py", line 216, in __init__
    self, localId, relativePath, dataRepository)
  File "/Users/david/server-1/ga4gh/datamodel/variants.py", line 538, in __init__
    self._scanDataFiles(dataDir, ['*.bcf', '*.vcf.gz'])
  File "/Users/david/server-1/ga4gh/datamodel/__init__.py", line 471, in _scanDataFiles
    self._addDataFile(filename)
  File "/Users/david/server-1/ga4gh/datamodel/variants.py", line 594, in _addDataFile
    self._updateMetadata(varFile)
  File "/Users/david/server-1/ga4gh/datamodel/variants.py", line 552, in _updateMetadata
    variantFile.filename)
ga4gh.exceptions.InconsistentMetaDataException: Metadata of ga4gh-example-data/datasets/brca1/variants/vcf/ALL.chrMT.phase3_callmom.20130502.genotypes.vcf.gz is not consistent with other files in this VariantSet, and inconsistent metadata is not permitted
```

david4096:
Which also causes list to fail making it necessary to remove it by hand:

```
python repo_dev.py list ga4gh-example-data                                                   
Repo at 'ga4gh-example-data'
Listing
Traceback (most recent call last):
  File "repo_dev.py", line 7, in <module>
    ga4gh.cli.repo_main()
  File "/Users/david/server-1/ga4gh/cli.py", line 1606, in repo_main
    runner.run()
[...]

PR #949, opened 2016-02-26
WIP InconsistentMetaDataException repomanager bug

dcolligan:
So the problem here is that when we are initializing the server data model we also check for consistency in that model and throw errors if it is not consistent.  This is inconvenient for the repo manager since a data repo could very well have been put in an inconsistent state by a user, and we don't want to throw errors in that case.  The previous solution was to sever init and check functionality and to put the later into `checkConsistency` functions which were called on server startup, but not on data repo init.  In this case the 2 functionalities are a bit harder to disentangle.  @jeromekelleher do you think this is the right way to approach this?

Issue #946

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -1327,9 +1327,10 @@ class CheckRunner(AbstractRepoCommandRunner):
 
     def __init__(self, args):
         super(CheckRunner, self).__init__(args)
+        self.doConsistencyCheck = not args.skipConsistencyCheck
 
     def run(self):
-        self.repoManager.check()
+        self.repoManager.check(self.doConsistencyCheck)
 
 
 class ListRunner(AbstractRepoCommandRunner):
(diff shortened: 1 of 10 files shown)


jeromekelleher:
I think we need to move the consistency checks to repo-add time, and at run time we just assume that everything is OK. Suppose we have 10 thousand BAM files and a bag load of VCFs in our repo. There's no way we can check all of these at server start up. I don't think we should check them at load-time either (i.e., when we actually open the file handle when a request comes in), as this could add a lot of processing overhead and latency.

Does this make sense? So, the idea is that at repo import time, we check the consistency of our data, and every other time we assume it's OK. I think this should be doable, as our dependency model is a DAG. The repo admin has to add the references first before adding readgroupsets.

diekhans:
Agreed. Consistency check are better done by the repo manager.
It's not as performance sensitive as the server, easier to
report detailed errors, and easier for a user to fix when its
something they are actively doing, vs a failure later in time.

Given it is quite easy to for a human to screw up a file-base
repo, a validation mode for the repo manager would be valuable.
Check that the repo is in a consistent state.

diekhans:
@dcolligan any progress on this?

dcolligan:
updated

dcolligan:
fixed tests

david4096:
I was able to add the 1000 genomes VCFs after pulling down these changes. +1 from me!

This file must be removed from the set `ALL.wgs.phase3_shapeit2_mvncall_integrated_v5b.20130502.sites.vcf.gz` since it contains contigs represented in other files.

PR #952, opened 2016-02-26
Make request validation failures stop choking

dcolligan:
Issue #942

### DIFF
--- a/ga4gh/avrotools.py
+++ b/ga4gh/avrotools.py
@@ -201,6 +201,8 @@ class SchemaValidator(AvroTypeSwitch):
     Provides methods for schema validation
     """
     sinkValue = 'aRandomStringBecauseWeCanNotUseNone'
+    missingValue = 'MISSING_VALUE'
+    extraValue = 'EXTRA_VALUE'
 
     def handleSchemaDispatch(self, schema, handler, extra):
         if schema.type in self.schemaRequired:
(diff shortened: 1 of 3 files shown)


jeromekelleher:
+1.

Hopefully protobuf will be handling all this stuff for us soon...

david4096:
Although it doesn't 500 I now can't send the original request and I can't reopen the issue.

```
(server-1)➜  server-1 git:(master) ✗ http post localhost:8000/variants/search variantSetId=YnJjYTE6YnJjYTE referenceName=1 start=:0 end:=400
HTTP/1.0 400 BAD REQUEST
Content-Length: 359
Content-Type: application/json
Date: Mon, 29 Feb 2016 17:56:58 GMT
Server: Werkzeug/0.11.3 Python/2.7.11

{
    "errorCode": 2023404236, 
    "message": "Request '{u'start': u':0', u'variantSetId': u'YnJjYTE6YnJjYTE', u'end': 400, u'referenceName': u'1'}' is not a valid instance of <class 'ga4gh._protocol_definitions.SearchVariantsRequest'>; invalid fields: {u'start': u':0', u'callSetIds': u'MISSING_VALUE', u'pageSize': u'MISSING_VALUE', u'pageToken': u'MISSING_VALUE'}"
```

jeromekelleher:
Yeah, we need to get rid of this MISSING_VALUE stuff entirely.

dcolligan:
How else should we signal to a user that he has a field in his json that is missing?  (Or that is extraneous?)

jeromekelleher:
We don't need to worry about signalling exactly what's wrong here I think. It just needs to sort-of work. I thought that removing the MISSING_VALUE would make it easier to deal with default values. I don't really mind once it works in the nominal case (valid input) to be honest.

dcolligan:
Well, when the pull request I just pushed is accepted, it will not only sort-of work, but work!

I think MISSING_VALUE is a good idea.  "This is the field you are missing in your json" is a better error message than "your json is invalid".

Also, fyi, default values != required fields.  This issue has nothing to do with default values (which are defined in the avro schema), but does have something to do with required fields (which are defined in the protocol file(s) but not retained in the avro schema).

macieksmuga:
@dcolligan since this is coming up again, what pull request were you referring to? Is there a fix for this issue in the current Avro schemas?

Issue #941, opened 2016-02-23
Refactor ga4gh/datamodel/genotype_phenotype.py required


jeromekelleher:
We need to refactor ga4gh/datamodel/genotype_phenotype.py to 

1) Follow the conventions used elsewhere in the code more closely;  
2) Break the code up into testable chunks; and
3) Make unit tests to test those chunks.

This issue is necessarily a bit vague, but we can discuss the details in the thread if/when particular issues pop up.

bwalsh:
@dcolligan I believe this has been addressed with #953 and subsequent commits

dcolligan:
I'll defer to @jeromekelleher here.  If it's refactored to his satisfaction we can close this.

kozbo:
@jeromekelleher bump

jeromekelleher:
Yes, it's much better now. Closing.

PR #953, opened 2016-02-27
Adding PAsets, refactor some

david4096:
This PR brings in the concept of a `PhenotypeAssociationSet`, which can be one of many for a dataset. It includes some modest code changes in how datasets and g2p relate. It will be finished when one can query for phenotype association sets (much in the same way you might search for variant sets), and the existing queries for phenotype associations accept a `phenotypeAssociationSetId`. I'll modify the existing tests for these changes as well. #941


ga4gh/backend.py
```diff
@@ -336,6 +330,16 @@ def datasetsGenerator(self, request):
             request, self.getDataRepository().getNumDatasets(),
             self.getDataRepository().getDatasetByIndex)
 
+    def phenotypeAssociationSetsGenerator(self, request):
+        """
+        Returns a generator over the (dataset, nextPageToken) pairs
```
jeromekelleher: comment is incorrect here.

ga4gh/datamodel/datasets.py
```diff
@@ -82,6 +86,33 @@ def getVariantSetByIndex(self, index):
         """
         return self._variantSetIdMap[self._variantSetIds[index]]
 
+    def addPhenotypeAssociationSet(self, phenotypeAssociationSet):
+        """
+        Adds the specified g2p association set to this backend.
+        """
+        id_ = phenotypeAssociationSet.getId()
+        self._phenotypeAssociationSetIdMap[id_] = phenotypeAssociationSet
+        self._phenotypeAssociationSetNameMap[phenotypeAssociationSet.getLocalId()] = phenotypeAssociationSet
```
jeromekelleher: Obviously these lines are too long. So the way this is done in most consistently in the rest of the code is
```python
self._phenotypeAssociationSetNameMap[
    phenotypeAssociationSet.getLocalId()] = phenotypeAssociationSet
```
We all like consistency...

jeromekelleher:
You're on exactly the right path here @david4096.

david4096:
Added client for searching PA sets, partial refactor of `datamodel/genotype_phenotype.py` more or less done. Incorporated Jerome's fixes and brought up to style concerns.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -17,47 +16,101 @@
...
+        pas.name = self.getLocalId()
+        pas.id = self.getId()
+        pas.datasetId = self.getParentContainer().getId()
+        pas.info = self.getInfo()
+        return pas
+
+    def getInfo(self):
+        return {}
+
+    def _generateFeature(self, featureId="", childIds=[], parentId="",
```
jeromekelleher: Can you use hanging indents here (and everywhere else) please? I.e.,
```python
def _generateFeature(
    self, featureId="", childIds=[], parentId="", featureSetId="", referenceName="", 
    start=None, end=None, strand=None, _term="", _termId="", sourceName="", 
    sourceVersion=""):
```
I know I'm nagging about these style issues all the time and it's all subjective in any case, but this code objectively 'looks different' to the rest of the code. This is bad for several reasons, but mainly because scripts written/regexps created to do stuff to the code in bulk will assume particular things. If we have different styles for doing things, these will break or at least be a lot harder to do. This point particularly and especially applies to visual indents, because code transformations that we plan on undertaking _will_ break this. If you try to automatically snake_case identifiers with visual indentation you will end up with a complete mess. 

So, using visual indents is actively creating work for the future.If you want we can have a discussion about style choices elsewhere and discuss an update to the project-wide coding conventions. However, until that happens please try and keep any new code that is added more in-line with the style used everywhere else.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -17,47 +16,101 @@
...
+        phenotypeInstance.type = self._generateOntologyTerm(_term,
+                                                            _termId,
+                                                            sourceName,
+                                                            sourceVersion)
+        phenotypeInstance.qualifier = qualifierTerms
+        phenotypeInstance.ageOfOnset = onsetTerm
+        phenotypeInstance.description = description
+        return phenotypeInstance
+
+    def _generateOntologyTerm(self, _term="", _id="",
```
jeromekelleher: Why ``_term`` here? The ``_id`` convention is used here because ``id`` is a builtin function in Python and the convention used to avoid clashing with this identifier is to prefix it with an ``_``

ga4gh/serverconfig.py
```diff
@@ -39,7 +39,7 @@ class DevelopmentConfig(BaseConfig):
     Configuration used for development.
     """
     DATA_SOURCE = "ga4gh-example-data"
-    DEBUG = False
+    DEBUG = True
```
jeromekelleher: I think this causes problems for people. Best leave it False by default (people can turn it on temporarily if they want).

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -365,49 +412,62 @@ def _toGA4GH(self, annotation):
...
-            "term": name,
-            "id": id_,
-            "sourceName": ontologySource})
-        f.id = annotation['id']  # TODO connect with real feature Ids
-        f.featureSetId = ''
-        f.parentIds = []
-        f.attributes = protocol.Attributes.fromJsonDict({"vals": {}})
-
-        id_, ontologySource = self.namespaceSplit(
+        f = self._generateFeature(
```
jeromekelleher: I'm not convinced by this way of organising things (i.e, calling ``_generateFeature(..)``) with a whole bunch of parameters... There's a lot of places for things to go wrong here between where we're passing the value in, and where it eventually gets assigned into the object. It just feels a bit brittle. I've spent about 5 minutes tracing up and down the file trying to figure out what's happening, so I don't think it's that clear in any case. What problem are we trying to solve here with ``_generateFeature``?

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -365,49 +412,62 @@ def _toGA4GH(self, annotation):
...
+            featureId=annotation['id'],  # TODO connect with real feature Ids
+            childIds=[],
```
jeromekelleher: Might as well use the default values for all these? i.e., have
```python
 feature = self._generateFeature(
    featureId=annotation['id'], sourceName=ontologySource)
```

jeromekelleher:
This looks good @david4096, I think you've gotten a long way here. I wonder if you're trying to bite off too much with this one though by taking on a wholesale refactor of the g2p code along with adding the new PASet type and its methods. Maybe it would be better to scale it back it a bit to just what's needed to support the PASet, and defer the big refactor until afterwards? (I definitely agree that a refactoring is required though.)

david4096:
@jeromekelleher thanks for taking the time to review! I got rid of the `generate` functions in favor of the method being used previously (protocol.fromJsonDict).

The scaffolding is in place to add in the simulated responses and data driven tests.

bwalsh:
+1 Clarifies contract with new AssociationSets. Addresses concerns about "middle tier" tests.  Nice!

david4096:
@macieksmuga I added a few more comments :100:

Issue #851, opened 2016-01-14
Updates needed for iPython notebook


jeromekelleher:
Issue #662 and PR #746 give an iPython script to demonstrate the API. Some comments in this PR thread were not resolved, however.

andrewjesaitis:
I'll take a look at doing these updates.

andrewjesaitis:
This is done. @jeromekelleher, you can close.

PR #954, opened 2016-02-27
Issue 851: Updates the example IPython Notebook

andrewjesaitis:
For issue #851:

* Use Python3 print syntax
* Use string formating instead of string interpolation
* Use ipython's pretty printing to output formatted dictionaries
  instead of the string representation
* Use lists in place of generators to make indexing more explicit


jeromekelleher:
Thanks @andrewjesaitis, what you've listed sounds exactly like what I was hoping for. Unfortunately, I'm an idiot when it come to iPy --- is there some way I can view the notebook?

andrewjesaitis:
Sure. Github actually renders them now, so you should be able to view it here:
[https://github.com/andrewjesaitis/server/blob/a3b8e1ec81a33a9db19620bae535c8f05cbc7b34/scripts/ga4gh_example.ipynb](https://github.com/andrewjesaitis/server/blob/a3b8e1ec81a33a9db19620bae535c8f05cbc7b34/scripts/ga4gh_example.ipynb)

jeromekelleher:
Looks great, thanks @andrewjesaitis! One minor quibble:

```python
 for i, read in enumerate(reads):
    count += 1
``` 
Why the enumerate here? We're not actually using ``i`` anywhere, so the enumerate is superfluous I think.

andrewjesaitis:
Good catch! Just pushed.

jeromekelleher:
Excellent, +1. Merging.

kozbo:
This is great stuff @andrewjesaitis. We should create a examples page in our documentation (or the wiki) and point to this along with the other example code we will have. I will create an issue for this.

PR #958, opened 2016-03-01
Update protobuf branch

jeromekelleher:
Brings the proto branch up to date with master.

I encountered a few issues I couldn't fix quickly, so these are put in as skipped tests. We should tackle these immediately when mainline development moves to protobuf (which really needs to happen soon or maintaining this branch will become a nightmare).

@dcolligan, can you have a scan and merge if you're happy please?


calbach:
The functionality should now be available by upgrading to 3.0.0b2.post2. https://pypi.python.org/pypi?:action=display&name=protobuf&version=3.0.0b2.post2

I pulled this in locally and verified that it produces JSON as expected for the `ListValue` case.

jeromekelleher:
@dcolligan, can you have a quick scan and merge if OK please?

dcolligan:
+1 merged

PR #961, opened 2016-03-02
Variation annotation merge

dcolligan:
Merging master into VA.  Once we merge this PR and the PR with the improperPlacement change, we can merge VA into master.


david4096:
The request validation stuff is changing the way ontology term requests were handled (broke my simple use case). I used to send in ontology terms like so: `[{"term": "downstream_gene_variant"}]. It now expects an ID, which currently isn't discoverable via the API. However, I think this behavior is acceptable and I'll make an issue to fix how the VA ontology term search works.

+1

PR #967, opened 2016-03-04
Biosample

david4096:
Adds biosample identifiers to reads and callset objects.


david4096:
More or less done with a few parts that need review:

@dcolligan I broke out the repo manager from cli in this [commit](https://github.com/david4096/server-1/commit/75e3fb8d1f9b1008e37d8e908ca9c5744e45bef4). Please let me know if that makes sense to you. I'm happy to roll back this change if desired.

The workflow for biosamples is:
1. Generate a JSON for the biosamples you want to represent
2. Add them to the dataset using `repo_dev.py add-biosample`.
3. Search using biosamples/search given a name (or empty to return all), and GET using `biosamples/<bioSampleId>`
4. Search reads and callsets using an optional `bioSampleID` filter. Currently always returns a single element list.

See `tests/unit/test_simulated_stack.py` for example usage.

diekhans:
The approach of generating the JSON outside of the repo manager
makes a lot of sense, as there no standard metadata format and
the transformation will need to be written for each input set.

david4096:
I included a [script](https://github.com/ga4gh/server/pull/967/files#diff-2e774e3312b16ddfe369b22dc4afd172) that converts a CSV version of the 1000 Genomes individual data spreadsheet into JSON biosamples. Hopefully this serves as enough of an example for users. I also included the data it worked on, which had to be converted from the excel spreadsheet. Perhaps this should go with the compliance test-data as opposed to the scripts in the ref server?

dcolligan:
@david4096 let's keep all the cli functionality in the cli module ... I do have a PR open to refactor the cli module into smaller modules, but Jerome wanted me to wait on it until other large changes were merged into master

david4096:
Added issue for tracking BioSample related PRs. https://github.com/ga4gh/server/issues/985

david4096:
Put back on the WIP because read group sets endpoint is now being used to filter read groups, as opposed to `reads/search`. Also improved the search mechanism to handle cases when both `name` and `bioSampleId` might be specified.

Updated the protocol definitions to match the schema change.

Known issues: the `info` field is simply a map, not a map of arrays of strings. The client needs to be updated to reflect the schema change.

david4096:
Script has been updated to properly create the info field.

Remaining issues: when adding biosamples using the repo manager the data should be checked to be well formed. Then when writing the data driven tests one can check for the expected exceptions.

david4096:
Now the repo manager checks that a biosample JSON is properly formed before adding it to a dataset. Added tests to show this functionality. @dcolligan would you please review, considering this feature complete.

david4096:
Removing the WIP now that the readgroupsets/search endpoint filters by biosampleId. @dcolligan @macieksmuga mind reviewing?

ga4gh/backend.py
```diff
@@ -404,6 +405,89 @@ def _topLevelObjectGenerator(self, request, numObjects, getByIndexMethod):
                 nextPageToken = str(currentIndex)
             yield object_.toProtocolElement(), nextPageToken
 
+    def _filterReadGroupsByBioSampleId(self, request, readGroupSet):
+        newRg = []
+        for rg in readGroupSet.readGroups:
+            print(rg.bioSampleId, request.bioSampleId)
```
macieksmuga: remove debug print statement?

ga4gh/backend.py
```diff
@@ -404,6 +405,89 @@ def _topLevelObjectGenerator(self, request, numObjects, getByIndexMethod):
...
+            if request.bioSampleId == rg.bioSampleId:
+                newRg.append(rg)
+        readGroupSet.readGroups = newRg
+        return readGroupSet
+
+    def _readGroupSetObjectGenerator(
```
macieksmuga: So the `readGroupSet`s returned by this generator have their `readGroup`s filtered out by `bioSampleId`?
Is this sufficiently clearly spelled out in the schemas?

ga4gh/backend.py
```diff
@@ -404,6 +405,89 @@ def _topLevelObjectGenerator(self, request, numObjects, getByIndexMethod):
...
+            self, request, numObjects, getByIndexMethod, recur=True):
+        currentIndex = 0
+        if request.pageToken is not None:
+            currentIndex, = _parsePageToken(request.pageToken, 1)
+        while currentIndex < numObjects:
+            readGroupSet = getByIndexMethod(currentIndex).toProtocolElement()
+            currentIndex += 1
+            nextPageToken = None
+            if currentIndex < numObjects:
+                if recur:
```
macieksmuga: `if (currentIndex < numObjects) and recur:`

ga4gh/datamodel/__init__.py
```diff
@@ -478,3 +487,28 @@ def _scanDataFiles(self, dataDir, patterns):
...
+    Loads a human readable sidecar and makes its values available as
+    self._sidecar(key). Accepts the filename for which we would like
+    extra data.
+    """
+    def loadSidecar(self, filepath):
+        jsonFilename = os.path.splitext(filepath)[0] + ".json"
+        try:
+            with open(jsonFilename) as data:
+                self._sidecar = json.load(data)
+        except ValueError:
```
macieksmuga: Remove the duplicate except statement with:
`except (ValueError, IOError):`

ga4gh/datamodel/biodata.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    This class represents an abstract BioSample object.
+    It sets default values and getters, as well as the
+    toProtocolElement function.
+    """
+    compoundIdClass = datamodel.BioSampleCompoundId
+
+    def __init__(self, parentContainer, localId):
+        super(AbstractBioSample, self).__init__(parentContainer, localId)
+        self._createDateTime = datetime.datetime.now().isoformat()
```
macieksmuga: Shouldn't the creation time be sourced from the JSON (indicating, perhaps, when the biosample was collected or sequenced?)

ga4gh/backend.py
david4096: Thanks added to this PR https://github.com/ga4gh/schemas/pull/579/files#diff-0b6f468d7591bb264e8eb143b119e890R158

tests/unit/test_biodata.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    def testToProtocolElement(self):
+        self.dataset = datasets.AbstractDataset('dataset1')
+        for localId in os.listdir(paths.testBioSamplesDataDir):
+            jsonFilename = os.path.join(paths.bioSamplesDir, localId)
+            jsonDict = {}
+            try:
+                with open(jsonFilename) as data:
+                    jsonDict = json.load(data)
+            except ValueError:
```
macieksmuga: collapse your `except`s into one.

ga4gh/datamodel/biodata.py
diekhans: The meaning and of the time fields is needs clarified.  Right
now the created/updated fields were intended, but not
documented, as load time.  This is not that useful for
most bioinfo sharing.

There is a ticket that covers this.  This is find for
now until that is resolve.

macieksmuga:
+1 Looks good to me. Will have fun co-merging this in with sequence annotations, that edit many of the exact same files...

ga4gh/datamodel/biodata.py
david4096: If there is a createDateTime or updateDateTime in the JSON it will use that one. https://github.com/ga4gh/server/pull/967/files#diff-aaf9d6c43122e379253eff9dfda7d07cR85

diekhans:
+1  for sure good enough for a topic branch.

ga4gh/datamodel/biodata.py
```diff
@@ -0,0 +1,114 @@
...
+        self.loadSidecar(filePath)
+        self.validate()
+
+    def validate(self):
+        if not protocol.BioSample.validate(
+                self.toProtocolElement().toJsonDict()):
+            raise exceptions.FileOpenFailedException(self._filePath)
+
+    def getCreateDateTime(self):
+        if self.sidecar('createDateTime'):
```
dcolligan: all of these get* methods could probably call an underlying private method, e.g.
```
def getDescription(self):
    return self._getField('description')

def _getField(self, fieldName):
    if self.sidecar(fieldName):
        return self.sidecar(fieldName)
    else:
        return getAttr(self, '_' + fieldName)
```
That would reduce a lot of the logic in this class.

ga4gh/datamodel/datasets.py
```diff
@@ -181,6 +213,12 @@ def __init__(
             seed = randomSeed + i
             variantSet = variants.SimulatedVariantSet(
                 self, localId, seed, numCalls, variantDensity)
+            cs = variantSet.getCallSets()
+            # Add biosamples
+            for c in cs:
```
dcolligan: try and stay away from single/double letter variable names... we try and keep our names more descriptive and easy to read than that

ga4gh/repo_manager.py
```diff
@@ -383,6 +400,45 @@ def removeReferenceSet(self, referenceSetName):
...
+        fileName = os.path.basename(filePath)
+        destPath = os.path.join(
+            self._repoPath, self.datasetsDirName, datasetName,
+            self.bioSamplesDirName)
+        fullDest = os.path.join(destPath, fileName)
+        self._checkFile(filePath, self.jsonExtension)
+        dataset = datasets.AbstractDataset('dataset1')
+        try:
+            biodata.JsonBioSample(dataset, "name", filePath)
+        except:
```
dcolligan: What exception(s) is this catching here?  It should be specified

ga4gh/repo_manager.py
```diff
@@ -383,6 +400,45 @@ def removeReferenceSet(self, referenceSetName):
...
+        # move the bam file
+        self._check()
+        self._checkDataset(datasetName)
+        fileName = os.path.basename(filePath)
+        destPath = os.path.join(
+            self._repoPath, self.datasetsDirName, datasetName,
+            self.bioSamplesDirName)
+        fullDest = os.path.join(destPath, fileName)
+        self._checkFile(filePath, self.jsonExtension)
+        dataset = datasets.AbstractDataset('dataset1')
```
dcolligan: I would choose a name other than `dataset1` to emphasize this is just a throwaway dataset

ga4gh/repo_manager.py
```diff
@@ -383,6 +400,45 @@ def removeReferenceSet(self, referenceSetName):
         self._repoEmit("ReferenceSet '{}' removed".format(
             referenceSetName))
 
+    def addBioSample(self, datasetName, filePath, moveMode):
+        """
+        Add a BioSample
+        """
+        # move the bam file
```
dcolligan: the bam file?

scripts/thousand_biosamples.py
```diff
@@ -0,0 +1,40 @@
+"""
```
dcolligan: missing standard imports

scripts/thousand_biosamples.py
```diff
@@ -0,0 +1,40 @@
...
+    Relationship,Unexpected Parent/Child ,Non Paternity,
+    Siblings,Grandparents,Avuncular,Half Siblings,
+    Unknown Second Order,Third Order,Other Comments
+"""
+
+import csv
+import json
+import datetime
+
+
```
dcolligan: surround this with `@Timed`

tests/datadriven/test_reads.py
```diff
@@ -155,9 +155,6 @@ def testSampleIdEtc(self):
             readGroupInfo = self._readGroupInfos[readGroup.getLocalId()]
             gaReadGroup = readGroup.toProtocolElement()
             self.assertEqual(
-                readGroupInfo.sampleId,
```
dcolligan: should this be replaced with `bioSampleId`, not removed?

tests/unit/test_cli.py
```diff
@@ -179,6 +179,18 @@ def testReadsSearchArguments(self):
         self.assertEqual(args.baseUrl, "BASEURL")
         self.assertEquals(args.runner, cli.SearchReadsRunner)
 
+    def testBioSamplesSearchArguments(self):
```
dcolligan: missing a lot of other cli tests here, right?  bio sample get, repo manager add and remove

tests/unit/test_client.py
```diff
@@ -191,6 +191,11 @@ def testGetVariant(self):
         self.httpClient._runGetRequest.assert_called_once_with(
             "variants", protocol.Variant, self.objectId)
 
+    def testGetBioSample(self):
```
dcolligan: should also test search bio sample

tests/unit/test_repo_manager.py
```diff
@@ -179,3 +179,30 @@ def testRemoveVariantSet(self):
         with self.assertRaises(exceptions.RepoManagerException):
             self.repoManager.removeVariantSet(
                 'dataset1', paths.variantSetName)
+
+    def testAddBioSample(self):
```
dcolligan: also add entries in the end to end repo manager test: `tests/end_to_end/test_repo_manager.py`

dcolligan:
Looks good, but I'll hold off on +1 until all my comments are addressed

dcolligan:
Also, documentation is needed.  Is that going to be in a separate PR?

ga4gh/datamodel/biodata.py
david4096: Noted and using your abstraction now.

ga4gh/datamodel/datasets.py
david4096: Fixed, callSet in callSets

ga4gh/repo_manager.py
david4096: Explicitly stated the exception.

ga4gh/repo_manager.py
david4096: Changed to `temp_ds`.

tests/datadriven/test_reads.py
david4096: Added a test that reconstructs the bioSampleId and checks to make sure it is correct.

david4096:
I've included @dcolligan and @macieksmuga suggestions and added some cursory documentation as to how to load biosamples using the repo manager.

Issue #975, opened 2016-03-10
VariationAnnotationSetSearch client code produces error without variantSetId 


macieksmuga:
when issuing a call as follows:
```
python client_dev.py -vv variantannotationsets-search http://localhost:8000
```
The response is a server stack trace. (It needs `variantSetId` supplied to work, but doesn't check explicitly for that parameter).

dcolligan:
For the request above, the client payload is:
```
{"pageToken": null, "variantSetId": null, "pageSize": null}
```

The server stacktrace is:
```
Traceback (most recent call last):
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/dannyc/ga4gh/new/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/dannyc/ga4gh/server/ga4gh/frontend.py", line 514, in searchVariantAnnotationSets
    flask.request, app.backend.runSearchVariantAnnotationSets)
  File "/Users/dannyc/ga4gh/server/ga4gh/frontend.py", line 408, in handleFlaskPostRequest
    return handleHttpPost(flaskRequest, endpoint)
  File "/Users/dannyc/ga4gh/server/ga4gh/frontend.py", line 285, in handleHttpPost
    responseStr = endpoint(request.get_data())
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 809, in runSearchVariantAnnotationSets
    self.variantAnnotationSetsGenerator)
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 620, in runSearchRequest
    self.validateRequest(requestDict, requestClass)
  File "/Users/dannyc/ga4gh/server/ga4gh/backend.py", line 366, in validateRequest
    jsonDict, requestClass)
RequestValidationFailureException: Request '{u'pageToken': None, u'variantSetId': None, u'pageSize': None}' is not a valid instance of <class 'ga4gh._protocol_definitions.SearchVariantAnnotationSetsRequest'>; invalid fields: {u'variantSetId': None}
127.0.0.1 - - [10/Mar/2016 17:39:31] "POST /variantannotationsets/search?key=invalid HTTP/1.1" 400 -
```

[...]

PR #976, opened 2016-03-10
Make VAS Search cli argument positional

dcolligan:
Issue #975


david4096:
+1, please update/merge thanks!

Issue #653, opened 2015-09-24
Document API using Swagger (or equivalent) (hackathon idea)


bwalsh:
It would be useful to have the api documented using a modern interactive approach.
This might be a topic for the hackathon in NYC?
see http://swagger.io/
https://github.com/gangverk/flask-swagger

diekhans:
We probably have to switch the reference server away from Flask,
single it doesn't support chucked HTTP that will be required for
streaming.

jeromekelleher:
This is pretty cool @bwalsh, it would be a really neat feature. I have two slight concerns:

1. As @diekhans, we might have to move away from Flask when we start making a streaming protocol as WSGI doesn't support HTTP chunked encoding (which is ridiculous, but what can you do).
2. Our API isn't particularly clean REST, and might not fit into what's expected from a REST documentation generator.

No reason we can't play around with it and see what happens though!

bwalsh:
See http://rest.ensembl.org/documentation/info/gacallSet for another example

kozbo:
See http://ga4gh-demo.westus.cloudapp.azure.com for another example.

david4096:
The current state of work is reflected [here](https://github.com/david4096/ga4gh-swagger-ui). I propose future discussion about API descriptions happen there, or in the schemas. @bwalsh @jeromekelleher , OK to close?

jeromekelleher:
I'm happy to close this as we've moved on a long way from this point now. 

I notice the link above points to one of your repos @david4096 --- is there an official GA4GH repo for this? We should have one, it's a great documentation/developer tool that anyone should be able to deploy against any GA4GH server.

david4096:
@jeromekelleher Not yet, and I agree! I expect that after we migrate to protocol buffers we will automate some part of generating API descriptions.

jeromekelleher:
OK, I'm going to close this issue then as we're decoupling this from the ref server.

david4096:
Although the UI isn't maintained with this code, we should provide a way to easily serve static API documentation.

kozbo:
Currently not working for 1kgenomes.ga4gh.org

PR #977, opened 2016-03-11
Swagger docs (WIP)

david4096:
Beginning the effort to provide a swagger-ui into our API. #653 

The animated gif embedded in the ticket is way too distracting; go here to see it:
https://cloud.githubusercontent.com/assets/3795586/13690214/7a2f4fd2-e6e3-11e5-8e1e-994e74b0ae9a.gif


dcolligan:
Couple questions:
1) Is there anywhere we can pull the static resources from instead of checking them in?  (We do a similar thing with jQuery on the index page)
2) Is it necessary to include different locales at this point?  We probably only need english

ga4gh/static/swagger.json
```diff
@@ -0,0 +1,172 @@
...
+    "title": "GA4GH Reference Server",
+    "contact": {
+      "email": "someone@ucsc.edu"
+    },
+    "license": {
+      "name": "Apache 2.0",
+      "url": "http://www.apache.org/licenses/LICENSE-2.0.html"
+    }
+  },
+  "host": "localhost:8000",
```
diekhans: What is the approach for setting "host" to an actual value?

ga4gh/static/swagger.json
```diff
@@ -0,0 +1,172 @@
...
+  "paths": {
+    "/variants/{id}": {
+      "get": {
+        "tags": [
+          "variants"
+        ],
+        "description": "Returns variant for a given ID",
+        "summary": "Find variants by ID",
+        "operationId": "getVariant",
+        "produces": [
```
diekhans: We need an approach to building the documentation from the schema source rather than need to maintain it in to places?

ga4gh/static/swagger.json
david4096: By using a template one could place in the hostname for the running server.

ga4gh/static/swagger.json
kozbo: I am wondering if we should wait to implement the auto generation of the documentation until we move to protocol buffers. I am expecting that manual implementation will not be too painful if the schema doesn't change very often or very much. Once we have protocol buffers then we can connect it in to be automatically generated from that documentation layer.

diekhans: The current documentation extraction from Avro is problematic due to the way it is implemented on top of the Avro parser.

As part of the ProtBuf conversion, L-Shift wrote a documentation extractor from the ProtBuf specification files.

For Swagger, I think it makes sense to also generate the swarger JSON file template. from the schema.  To solve the URL issue, it might make sense to just server up the JSON file as dynamic content and avoid having to create static content on startup.  Need to understand the issues with load-balanced servers.  Their might not even be write access to the static data.

@david4096 is the host field even required?  It seems like the client already knows the URL it's connecting to and this would only matter if the swagger static files was coming from a different host.

david4096:
Seems like a good path forward: https://github.com/gengo/grpc-gateway

gabrielsaldana:
I'm looking into adding the swagger documentation to the API. Expect a pull request soon

diekhans:
Very soon we will be finished up the protocol buffer work, this will include extracting the documentation from the schemes.  This can be very easily extended to extract whatever swagger needs.

jeromekelleher:
I agree that we should postpone this until we've completed the protobuf transition. Any code we write now to do with documentation extraction will just have to be written again.

david4096:
screen shot 2016-04-07 at 6 11 30 pm

@gabrielsaldana is putting together the `swagger.json` for this. Recently updated:

screen shot 2016-04-07 at 6 12 58 pm

david4096:
I've set up a server with the swagger docs and the compliance dataset for testing temporarily. 

http://ga4gh-david.cloudapp.net/static/index.html

jeromekelleher:
Just had a quick play with this @david4096 --- AWESOME! It would be nice if we didn't have to pull in quite so many files though; is there any way we can reference remote URLs? (There are pros and cons of course...)

skeenan:
Looks great!

david4096:
I think the best option for incorporating this at this juncture is just to include the `swagger.json` in the `static` directory. That way, any swagger-ui (which is a standalone, dependency free webapp) can request the API description. Then we can stand up our customized swagger-ui interface served by a simple HTTP server anywhere, and request the `static/swagger.json`, which will then be rendered into the interactive documentation.

To answer your question @jeromekelleher there are a number of these JS libraries available on CDNs, or github for that matter, but since swagger-ui is maintained separately and independent of our server, we may as well *not* fold it into our codebase.

If this seems appropriate I'll update this PR to just include the `swagger.json`. I'm open to other approaches as well.

david4096:
I'm going to close this PR and open another one to detail the approach described in the above comment.

PR #980, opened 2016-03-11
Repo manager for ontology maps

david4096:
VA and SA both use sequence ontology. This adds the ability to add and remove ontologies using the repo manager.


diekhans:
I don't see any documentation, but it is incorrect to say this is adding an ontology. it's adding a mapping between to an term coming from a VCF and a reference to an ontology term (OntologyTerm) record.
An ontology format would be some think line OBO or OWL-RDF; however that is not something we need.

This is only a problem with the terminology, not the code.  But that is very important, because it claims
to be doing something that it isn't doing.

ga4gh/datarepo.py
```diff
@@ -104,7 +104,12 @@ def getOntology(self, name):
         """
         Returns ontologies
         """
-        return self._ontologyNameMap[name]
+        ret = {}
+        try:
```
dcolligan: I'd remove the try/except and instead do:
```
ret = self._ontologyNameMap.get(name, {})
```

ga4gh/cli.py
```diff
@@ -1364,6 +1364,25 @@ def run(self):
...
+
+
+class RemoveOntologyRunner(AbstractRepoCommandRunner):
+
+    def __init__(self, args):
+        super(RemoveOntologyRunner, self).__init__(args)
+        self.ontologyName = args.ontologyName
+
+    def run(self):
+        self.repoManager.removeOntology(self.ontologyName)
```
dcolligan: if this doesn't use `confirmRun` then the force flag is meaningless

dcolligan:
A good start.  Also should have tests for the individual commands, including failure cases, in `tests/unit/test_repo_manager.py` and tests for the cli in `tests/unit/test_cli.py`

ga4gh/datarepo.py
diekhans: not that both of these approaches create an dict object that is then discarded even when the name map contains the object.

david4096:
Updated @dcolligan. They are now more appropriately called ontology maps.

ga4gh/cli.py
```diff
@@ -1423,6 +1423,25 @@ def run(self):
...
+
+
+class RemoveOntologyMapRunner(AbstractRepoCommandRunner):
+
+    def __init__(self, args):
+        super(RemoveOntologyMapRunner, self).__init__(args)
+        self.ontologyMapName = args.ontologyMapName
+
+    def run(self):
+        self.repoManager.removeOntologyMap(self.ontologyMapName)
```
dcolligan: this needs to use `confirmRun` or else the force flag does nothing

tests/unit/test_repo_manager.py
```diff
@@ -182,3 +184,12 @@ def testRemoveVariantSet(self):
         with self.assertRaises(exceptions.RepoManagerException):
             self.repoManager.removeVariantSet(
                 'dataset1', paths.variantSetName)
+
+    def testAddOntologyMap(self):
```
dcolligan: what about a `testRemoveOntologyMap` function?

dcolligan:
It's coming along.  After the changes I mentioned, you should be good to go.

ga4gh/datamodel/ontologies.py
```diff
@@ -27,10 +27,20 @@ def add(self, id_, name):
         self._idNameMap[name] = id_
 
     def getId(self, name):
+        try:
```
jeromekelleher: This seems problematic. Usually failing noisily is a good thing. We should provide methods like ``hasId`` and ``hasName`` for the cases when we want to know if a given id/name exists in the map.

jeromekelleher:
Looks good in general to me.

cmungall:
You assume id<->name will be 1:1 within any given ontology.

For the id->name direction this is not particularly problematic. Every id uniquely identifies an ontology class, and within the subset of the ontologies you are likely to use, there should be a single rdfs:label. However, this assumption cannot be made for every ontology ever constructed. You might also want to think about internationalization early on.

For name->id, for any well-behaved ontology this should be OK, *provided* you are only mapping non-obsolete classes. It's not clear what procedure you are using to generate the mapping tables, and if overwrites are happening here.

I recommend documenting these assumptions, and the set of ontologies for which the assumptions are assumed to hold (I can help set up upstream checks in the release pipelines of these ontologies).

You're also using strings like `sequence_ontology` as unique IDs for the ontologies. The only consistent and stable way to identify an ontology is by it's ontology id (for OBO library ontologies) or IRI (for any ontology). You should see this in the `ontology` field for any obo formatted ontology, or the Ontology object in RDF/OWL. E.g. `so`, `go`, `hp`. These can easily be found on obofoundry.org (the metadata is available in a variety of formats, you can also see the list here: https://github.com/OBOFoundry/OBOFoundry.github.io/tree/master/ontology).

I expect you will need a richer ontology model at some point, at least with support for synonyms, traversal, definitions, obsoletion metadata, versioning. Ping me if you want feedback on either requirements or code.

jeromekelleher:
Thanks very much @cmungall, this really is very helpful.

david4096:
Need to add ontology maps to `repo list`.

diekhans:
Thanks @cmungall

note that the mapping goal is to link existing data that only
has ontology names to full specified ontology ids.  So the
outdated name ->id relationship should not be an issue.

The goal is not to develop an ontology service, only to
be able to link to them.

david4096:
Added the hasId, hasName functions for future use and added ontology maps to repo `list`.

Now throw an exception when a key isn't found that has to be caught elsewhere.

david4096:
Could I get a review/merge?

ga4gh/datamodel/variants.py
```diff
@@ -994,7 +994,10 @@ def convertSeqOntology(self, seqOntStr):
             so = self._createGaOntologyTermSo()
             so.term = soName
             if self._sequenceOntology is not None:
-                so.id = self._sequenceOntology.getId(soName)
+                try:
```
jeromekelleher: Is this really the semantics we want (using the empty string when the key isn't defined)? If so, how about we change the ``getId`` function to follow the semantics of ``dict.get``, and add a second parameter to give the default value when the key is not found? So, here we'd have ``so.id = self._sequenceOntology.getId(soName, "")``

ga4gh/datamodel/ontologies.py
```diff
@@ -26,11 +29,28 @@ def add(self, id_, name):
...
 
+    def getLocalId(self):
+        return self._localId
+
     def getId(self, name):
-        return self._idNameMap[name]
+        try:
+            id_ = self._idNameMap[name]
+        except KeyError:
+            raise exceptions.OntologyMapIdException
```
jeromekelleher: What's the rationale for raising an OntologyMapIdException here  rather than a KeyError? There are advantages to making the Ontology object behave as much like a python map as possible, and raising a KeyError is a natural thing to do.

ga4gh/datamodel/ontologies.py
```diff
@@ -26,11 +29,28 @@ def add(self, id_, name):
         self._nameIdMap[id_] = name
         self._idNameMap[name] = id_
 
+    def getLocalId(self):
+        return self._localId
+
     def getId(self, name):
```
jeromekelleher: ``getId`` is a bit confusing here, and would usually imply that we're returning the ID of the actual Ontology object itself, and not doing a search. ``getById`` and ``getByName`` perhaps? I realise these aren't part of this PR, so feel free to shelve this for later.

ga4gh/datamodel/ontologies.py
```diff
@@ -26,11 +27,28 @@ def add(self, id_, name):
...
         self._idNameMap[name] = id_
 
-    def getId(self, name):
-        return self._idNameMap[name]
+    def getLocalId(self):
+        return self._localId
 
-    def getName(self, id_):
-        return self._nameIdMap[id_]
+    def getId(self, name, default=""):
```
dcolligan: this function can be rewritten

```
return self._idNameMap.get(name, default)
```

ga4gh/datamodel/ontologies.py
```diff
@@ -26,11 +27,28 @@ def add(self, id_, name):
...
+        return id_
+
+    def hasId(self, id_):
+        return id_ in self._nameIdMap
+
+    def hasName(self, name):
+        return name in self._idNameIdMap
+
+    def getName(self, id_, default=""):
+        try:
```
dcolligan: same comment here

tests/unit/test_repo_manager.py
```diff
@@ -182,3 +184,18 @@ def testRemoveVariantSet(self):
...
+        self.repoManager.addOntologyMap(
+            paths.ontologyPath, 'link')
+
+    def testRemoveOntologyMap(self):
+        with self.assertRaises(exceptions.RepoManagerException):
+            self.repoManager.removeOntologyMap(
+                paths.ontologyName)
+        self.repoManager.addOntologyMap(
+            paths.ontologyPath, 'link')
+        self.repoManager.removeOntologyMap(
```
dcolligan: check that the repo manager throws an error after an additional removeOntologyMap call (i.e. when there should be no extant ontologyMap to remove)

tests/unit/test_repo_manager.py
```diff
@@ -182,3 +184,18 @@ def testRemoveVariantSet(self):
         with self.assertRaises(exceptions.RepoManagerException):
             self.repoManager.removeVariantSet(
                 'dataset1', paths.variantSetName)
+
+    def testAddOntologyMap(self):
+        with self.assertRaises(exceptions.RepoManagerException):
+            self.repoManager.addOntologyMap(paths.bamPath, 'link')
+        self.repoManager.addOntologyMap(
```
dcolligan: check that the repo manager throws an error after an identical addOntologyMap call after this one (to indicate to the user s/he already has this map present in the repo)

david4096:
Thanks @dcolligan got it.

macieksmuga:
Most of this the changes here are in the repo manager, so from the perspective of the other parts of the server's datamodel, the only difference is that I would now call (for example) `backend.getOntologyMap('sequence_ontology')` instead of `backend.getOntologyMap('sequence_ontology')`, right?

david4096:
@macieksmuga yep https://github.com/ga4gh/server/pull/980/files#diff-f9ce58fdc26b5bb89ea0a4aa47a608faR809

ga4gh/datamodel/ontologies.py
jeromekelleher: Any thoughts here @david4096?

macieksmuga:
+1

dcolligan:
+1

Issue #981, opened 2016-03-12
reversing filter order changes number of results from searchVariantAnnnotations


reece:
The behavior of [searchVariantAnnotations](https://ga4gh-schemas.readthedocs.org/en/latest/schemas/alleleAnnotationmethods.html#searchVariantAnnotations) is unclear to me, especially when multiple effect filters are described.

While investigating this question, I noticed that the number of results returned changed depending on the filter order.

The two obvious behavior options with multiple effects are conjunction (AND) and inclusive disjunction (OR) of the effect criteria. Since both options commutative, reversing the order of the filter should not matter.

In fact, it appears that the server is using only the first effect term.

See https://gist.github.com/reece/9db77d7f43af7c4fb8fc for demonstration code.

david4096:
Thanks Reece, looking into it!

PR #984, opened 2016-03-14
Bug fix for VA search

david4096:
Ontology term arrays in the request were not being checked correctly. This PR fixes the variant annotation endpoint to return any ontology terms in the request. A test has been added to ensure order does not affect results. Some refactoring for readability. Fixes https://github.com/ga4gh/server/issues/981


macieksmuga:
+1 looks good to me.

diekhans:
A test for this would be good in the compliance suite.  Please file a ticket; not high priority.

PR #986, opened 2016-03-15
Sequence annotation restart

macieksmuga:
Try 2. Replaces #962

PR #999, opened 2016-03-18
Sequence annotation restart, take 3

macieksmuga:
Update: I squashed and rebased against the latest (March 18) master.

This PR brings the sequence annotation API to reference server, as driven by a sqlite database backend.
It includes:

- Sequence Annotations objects in protocols and `datamodel`,
- Integration of Sequence annotations endpoints to all relevant parts of the server codebase,
- A generic lightweight SQLite3 backend,
- A GFF3 (sequence annotation file format) parser into the DB backend format,
- All relevant tests,
- Simple data repository manager extension to handle the new data types.


macieksmuga:
I'd like this PR to supersede #986, which contains the exact same codebase but isn't squashed and rebased against the latest master.

macieksmuga:
@jeromekelleher This is ready for your review. I'm working on the compliance suite now, so please feel free to jump in. I'd be happy to schedule some time with you (perhaps 6AM pacific time tomorrow?) to go through what's there and get you comfortable with the codebase.

jeromekelleher:
Thanks @macieksmuga, I'll try and look at this tomorrow. I'm fully meeting'd up tomorrow afternoon I'm afraid, I'll drop you an email to reschedule if I have any questions.

.gitignore
```diff
@@ -4,11 +4,13 @@
 ga4gh-example-data*
 ga4gh-downloaded-data*
 scripts/auth.yml
+ga4gh/_version.py
```
jeromekelleher: Are these changes incidental to sequence annotation? Please make a PR against master if so.

ga4gh/backend.py
```diff
@@ -552,6 +552,48 @@ def variantAnnotationsGenerator(self, request):
             request, variantAnnotationSet)
         return intervalIterator
 
+    def featuresGenerator(self, request):
+        """
+        :param request: JSON string - the original web request
```
jeromekelleher: Might as well follow the standard pattern for this comment. Sphinx docs are good, but we should try to apply them systematically, file wide if they are introduced.

ga4gh/backend.py
```diff
@@ -552,6 +552,48 @@ def variantAnnotationsGenerator(self, request):
...
+            parentId = compoundParentId.featureId
+            # A client can optionally specify JUST the (compound) parentID,
+            # and the server needs to derive the dataset & featureSet
+            # from this (compound) parentID.
+            if compoundId is None:
+                compoundId = compoundParentId
+            else:
+                # check that the dataset and featureSet of the parent
+                # compound ID is the same as that of the featureSetId
+                mismatchCheck = (compoundParentId.datasetId !=
```
jeromekelleher: This will be more robust to code changes (think snake_case) if you use hanging indent, e.g.
```python
 mismatchCheck = (
    compoundParentId.datasetId != compoundId.datasetId or
    compoundParentId.featureSetId !=compoundId.featureSetId)
```

ga4gh/backend.py
```diff
@@ -552,6 +552,48 @@ def variantAnnotationsGenerator(self, request):
...
+                                 compoundId.featureSetId)
+                if mismatchCheck:
+                    raise exceptions.ParentIncompatibleWithFeatureSet()
+
+        if compoundId is None:
+            raise exceptions.FeatureSetNotSpecifiedException()
+
+        dataset = self.getDataRepository().getDataset(compoundId.datasetId)
+        featureSet = dataset.getFeatureSet(compoundId.featureSetId)
+        return featureSet.getFeatures(request.referenceName,
```
jeromekelleher: hanging indent please. This will break when we snake_case.

ga4gh/backend.py
```diff
@@ -678,6 +730,18 @@ def runGetVariant(self, id_):
         jsonString = gaVariant.toJsonString()
         return jsonString
 
+    def runGetFeature(self, id_):
+        """
```
jeromekelleher: Again, try and be consistent with existing commenting patterns. I agree we should add sphinx doc strings to this class at some point, but it should be done in one go, consistently.

ga4gh/client.py
```diff
@@ -222,6 +222,24 @@ def getVariantAnnotationSet(self, variantAnnotationSetId):
...
+
+        :param str featureSetId: The ID of the FeatureSet of interest.
+        :return: The FeatureSet of interest.
+        :rtype: :class:`ga4gh.protocol.FeatureSet`
+        """
+        return self._runGetRequest(
+            "featuresets", protocol.FeatureSet, featureSetId)
+
+    def getFeature(self, compoundId):
+        """
```
jeromekelleher: Need a one line summary in comment.

ga4gh/client.py
```diff
@@ -278,6 +296,35 @@ def searchVariantAnnotations(
             request, "variantannotations",
             protocol.SearchVariantAnnotationsResponse)
 
+    def searchFeatures(
+            self, featureSetId=None, parentId=None, referenceName=None,
+            start=0, end=None, featureTypes=[]
```
jeromekelleher: Put the ): on the previous line, e.g.
```
           start=0, end=None, featureTypes=[]):
```

ga4gh/client.py
```diff
@@ -278,6 +296,35 @@ def searchVariantAnnotations(
...
+                       ):
+        """
+        Returns the result of running a searchFeatures method
+        on a request with the passed-in parameters.
```
jeromekelleher: I think you need to have a blank line between the summary and sphinx doc parameters. (Could be wrong though).

ga4gh/client.py
```diff
@@ -278,6 +296,35 @@ def searchVariantAnnotations(
...
+        :param featureSetId: string ID of the feature Set being searched
+        :param parentId: string ID (optional) of the parent feature
+        :param referenceName: string name of the reference to search
+            (ex: "chr1")
+        :param start: int search start position on reference
```
jeromekelleher: colon and types are in the wrong place here and elsewhere, should be:
```
:param int start: search start position on reference
```
Please check these in other sphinx docs.

ga4gh/datamodel/datasets.py
```diff
@@ -201,6 +246,7 @@ class FileSystemDataset(AbstractDataset):
     """
     variantsDirName = "variants"
     readsDirName = "reads"
+    featuresDirName = "sequenceAnnotations"
```
jeromekelleher: Should we call the directory ``features`` for consistency? This structure isn't really user-facing, so it might be good to be consistent about our naming.

ga4gh/datamodel/ontologies.py
```diff
@@ -32,7 +33,27 @@ def getId(self, name):
     def getName(self, id_):
         return self._nameIdMap[id_]
 
+    def getGaTermByName(self, name):
```
jeromekelleher: What does GaTerm stand for here? It would be nice to have some docstrings here for these new functions, as it's not entirely obvious what's going on.

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+import ga4gh.sqliteBackend as sqliteBackend
+import ga4gh.exceptions as exceptions
+
+# Note to self: There's the Feature ID as understood in a GFF3 file,
+# the Feature ID that is its server-assigned compoundId, and the
+# ID of the feature's row in the DB FEATURE table.
+# I need to be careful about which one is which.
+
+"""
+    For this implementation, `featureSetId` is required, while `parentId`
```
jeromekelleher: I think the contents of the docstring should be deindented.

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+
+    GFF3 data is represented by rows in a single table, named FEATURE.
+    The columns of the FEATURE table correspond to the columns of a GFF3,
+    with three additional columns prepended representing the ID
+    of this feature, the ID of its parent (if any), and a whitespace
+    separated array of its child IDs.
+
+    _featureColumns pairs represent the ordered (column_name, column_type).
+"""
+_featureColumns = [('id', 'TEXT'),
```
jeromekelleher: Hanging indent please
```python
_featureColumns = [
    ('id', 'TEXT'),
    ('parent_id', 'TEXT'),
    # etc
````

File: ga4gh/datamodel/sequenceAnnotations.py [hunk 14]
```diff
@@ -0,0 +1,415 @@
...
+    Requests spanning the join of circular genomes are represented as two
+    requests one on each side of the join (position 0)
+    """
+
+    def __init__(self, dbFile):
+        super(Gff3DbBackend, self).__init__(dbFile)
+        self.featureColumnNames = [f[0] for f in _featureColumns]
+        self.featureColumnTypes = [f[1] for f in _featureColumns]
+
+    def countFeaturesSearchInDb(self,
```
jeromekelleher: Hanging indent,
```python
def countFeaturesSearchInDb(
    self, referenceName=None, start=0, end=0, parentId=None, 
    featureTypes=[]):
```

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+            sql += ", ".join(["?", ] * len(featureTypes))
+            sql += ") "
+            sql_args += tuple(featureTypes)
+        query = self._dbconn.execute(sql, sql_args)
+        return (query.fetchone())[0]
+
+    def searchFeaturesInDb(self, pageToken=0, pageSize=None,
+                           referenceName=None, start=0, end=0,
+                           parentId=None, featureTypes=[]):
+        """
```
jeromekelleher: Need one line summary as well as sphinx params

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+        sql = ("SELECT * FROM FEATURE WHERE id = ?")
+        query = self._dbconn.execute(sql, (featureId,))
+        ret = query.fetchone()
+        if ret is None:
+            return None
+        return sqliteBackend.sqliteRow2Dict(ret)
+
+
+# Problem: FeatureIds from GFF3 often contain colons,
+# and the CompoundId mechanism reserves colons for its special separator
```
jeromekelleher: There's no reason why we have to use colons for CompoundIds. The mangle-demangle process will prove expensive I expect.

@dcolligan, any objection to using underscores (or something else) as the separator for CompoundIds?

If we make this change we should do it in master.

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+        gaFeatureSet.datasetId = self.getParentContainer().getId()
+        gaFeatureSet.referenceSetId = self._referenceSetId
+        gaFeatureSet.name = self._name
+        gaFeatureSet.sourceUri = self._sourceUri
+        gaFeatureSet.attributes = self._attributes
+        return gaFeatureSet
+
+    def getCompoundIdForFeatureId(self, featureId):
+        """
+        :param featureId: string feature ID as reported in GFF3
```
jeromekelleher: Should explain why and when featureId will be None.

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+    """
+    Simulated data backend for FeatureSet, used for internal testing.
+    """
+    def __init__(self, parentContainer, localId, randomSeed=1):
+        super(SimulatedFeatureSet, self).__init__(
+            parentContainer, localId, None)
+        self._randomSeed = randomSeed
+
+    def _getRandomOntologyTerm(self, randomNumberGenerator):
+        ontologyTuples = [("gene", "SO:0000704"),
```
jeromekelleher: hanging indent.

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+        feature.start = randomNumberGenerator.randint(1000, 2000)
+        feature.end = feature.start + randomNumberGenerator.randint(1, 100)
+        feature.featureType = self._getRandomOntologyTerm(
+            randomNumberGenerator)
+        references = ["chr1", "chr2", "chrX"]
+        feature.referenceName = randomNumberGenerator.choice(references)
+        strands = [protocol.Strand.POS_STRAND, protocol.Strand.NEG_STRAND]
+        feature.strand = randomNumberGenerator.choice(strands)
+        feature.attributes = protocol.Attributes()
+        feature.attributes.vals = {"gene_name": ["Frances", ],
```
jeromekelleher: hanging indent

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+        :param pageSize: None or int
+        :param ontologyTerms: optional list of ontology terms to limit query
+        :param parentId: optional parentId to limit query.
+        :return: Yields feature, nextPageToken pairs.
+            nextPageToken is None if last feature was yielded.
+
+        """
+        randomNumberGenerator = random.Random()
+        randomNumberGenerator.seed(self._randomSeed)
+        numFeatures = 10
```
jeromekelleher: Why 10?

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+        if pageToken is not None:
```
jeromekelleher: A pattern I like to save a line of screen space in situations like this is
```python
nextPageToken = 0
if pageToken is not None:
    nextPageToken = int(pageToken)
```
Just a suggestion...

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+            nextPageToken = int(pageToken)
+        else:
+            nextPageToken = 0
+        for featureId in ["feature{}".format(x) for x in xrange(numFeatures)]:
```
jeromekelleher: xrange is python 2 only. Use range.

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+                     gaFeature.referenceName == referenceName and
+                     (gaFeature.featureType in featureTypes or
+                      len(featureTypes) == 0))
+            if match:
+                gaFeature.parentId = ""  # TODO: Test with nonempty parentIDs?
+                if nextPageToken < numFeatures - 1:
+                    nextPageToken += 1
+                else:
+                    nextPageToken = None
+                yield gaFeature, \
```
jeromekelleher: How about
```python
nextPageTokenStr = None
if nextPageToken < numFeatures - 1:
    nextPageTokenStr = str(nextPageToken + 1)
yield gaFeature, nextPageTokenStr
```
It's a bit more compact, avoids line breaks.

ga4gh/datamodel/datasets.py
diekhans: Features is a type of record (discrete annotations) and there
will also be wiggle (continuous).  So is the dirname a module
or a record?

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+        Returns a protocol.Feature object corresponding to a compoundId
+        :param compoundId: a datamodel.FeatureCompoundId object
+        :return: a Feature object.
+        :raises: exceptions.ObjectWithIdNotFoundException if invalid
+            compoundId is provided.
+        """
+        featureId = _recolonize(compoundId.featureId)
+        with self._db as dataSource:
+            featureReturned = dataSource.getFeatureById(featureId)
+
```
jeromekelleher: I think it's better to seek out errors positively, and let the nominal case be the main route through the function. Here, e.g.,
```python
if featureReturned is None:
    raise exceptions.ObjectWithIdNotFoundException(compoundId)
return self._gaFeatureForFeatureDbRecord(featureReturned)
```
I would argue that this is a little bit clearer.

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,415 @@
...
+        if pageToken is not None:
+            nextPageToken = int(pageToken)
+        else:
+            nextPageToken = 0
+        for featureRecord in featuresReturned:
+            gaFeature = self._gaFeatureForFeatureDbRecord(featureRecord)
+            if nextPageToken < featuresCount - 1:
+                nextPageToken += 1
+            else:
+                nextPageToken = None
```
jeromekelleher: Same suggestion as above to make this code more compact.

ga4gh/datamodel/sequenceAnnotations.py
diekhans: range is much slower than xrange in python 2.  2to3 will handle
the conversion.   If the past is a prior, python 3 is still years
away.

ga4gh/gff3Parser.py
```diff
@@ -0,0 +1,321 @@
...
+        return v
+
+
+class Feature(object):
+    """
+    One feature notation from a GFF3, missing attribute, as code by `.' in
+    the file are stored as None.
+    """
+
+    __slots__ = ("seqname", "source", "type", "start", "end", "score",
```
jeromekelleher: Why the __slots__? If it's a perf optimisation because we're making loads of Features we should have a quick one-line comment to say so.

ga4gh/gff3Parser.py
```diff
@@ -0,0 +1,321 @@
...
+        """
+        ID attribute or None if records doesn't have an ID
+        """
+        featId = self.attributes.get("ID")
+        if featId is not None:
+            featId = featId[0]
+        return featId
+
+    def getRequiredAttribute(self, name):
+        "get a require attributes list of values, error if not found"
```
jeromekelleher: Docstring should be triple quotes.

ga4gh/gff3Parser.py
```diff
@@ -0,0 +1,321 @@
...
+                self.gff3Set.fileName, self.lineNumber)
+        return values
+
+
+class Gff3Set(object):
+    """
+    A set of GFF3 sequence annotations
+    """
+    def __init__(self, fileName=None):
+        self.fileName = fileName
```
jeromekelleher: Is there a particular reason for making these atttributes public? These should probably be private to follow the conventions elsewhere.

ga4gh/gff3Parser.py
```diff
@@ -0,0 +1,321 @@
...
+        Link a feature with it's parents.
+        """
+        parentIds = feature.attributes.get("Parent")
+        if parentIds is None:
+            self.roots.add(feature)
+        else:
+            for parentId in parentIds:
+                self.__linkToParent(feature, parentId)
+
+    def __linkToParent(self, feature, parentId):
```
jeromekelleher: The only issue I have with private methods and attributes is that things can get awkward when you start subclassing. This is why I tend not to bother with them these days. It's up to you though, it's not a strong opinion.

ga4gh/gff3Parser.py
```diff
@@ -0,0 +1,321 @@
...
+    fully test for conformance.
+    """
+
+    def __init__(self, fileName):
+        """
+        If fileName ends with .gz or .bz2, it will decompressed.  If fh is
+        specified, then parse the already opened stream, with file name still
+        used for error message, otherwise the file be opened and parsed.
+        """
+        self.fileName = fileName
```
jeromekelleher: Again, keep attributes private unless there's a good reason not to.

ga4gh/serverconfig.py
```diff
@@ -40,6 +40,8 @@ class DevelopmentConfig(BaseConfig):
     """
     DATA_SOURCE = "ga4gh-example-data"
     DEBUG = True
+    REQUEST_VALIDATION = True
```
jeromekelleher: Why are these being changed?

ga4gh/serverconfig.py
```diff
@@ -59,6 +61,13 @@ class ProductionConfig(BaseConfig):
     DATA_SOURCE = None
 
 
+class SimulatedConfig(BaseConfig):
```
jeromekelleher: This looks like something that should be added to master in a different PR.

ga4gh/sqliteBackend.py
```diff
@@ -0,0 +1,112 @@
...
+    into an array of simple dicts.
+    :param sqliteRows: array of rows returned from fetchall DB call
+    :return:  array of dicts, keyed by the column names.
+    """
+    return map(lambda r: dict(zip(r.keys(), r)), sqliteRows)
+
+
+def sqliteRow2Dict(sqliteRow):
+    """
+    As above, but for a single row
```
jeromekelleher: Either put in a full description or add the function name. (Functions might move around in future)

ga4gh/sqliteBackend.py
```diff
@@ -0,0 +1,112 @@
...
+        start = int(pageToken)
+        end = start + int(pageSize)
+        return " LIMIT {}, {}".format(start, end)
+    else:
+        return ""
+
+
+def _whereClauseSql(**whereClauses):
+    """
+    Takes parsed search query parameters,
```
jeromekelleher: Don't need to indent when breaking docstring descriptions (you do for sphinx params though).

scripts/generate_gff3_db.py
```diff
@@ -0,0 +1,127 @@
+"""
```
jeromekelleher: Presumably this script will need to be rolled into the repo manager, right? We want to say, add-feature-set <gff file> and the repo manager parses the GFF file and creates the SQLlite DB. The repository administrator shouldn't need to know anything about our internal DB format for storing features.

jeromekelleher:
Overall @macieksmuga, this looks absolutely excellent. Very nice, well documented, idiomatic code; thanks! I've made a bunch of nitpicky suggestions throughout; feel free to take on board or ignore as you see fit. High level points:

1. It really would be better to use hanging indents as much as possible. I think we should change to hanging indents anywhere that changing to snake_case would break the alignment. Otherwise it's an aesthetic judgement and I'll stay out of it. (My take is that aesthetics judgements are a waste of time, so I just stick with hanging indents  everywhere.)

2. Do we want to store the .db files in the repo? I guess we need to at the moment, but it would be nice if we could hit some import hooks in the repo manager before running the data driven tests. It's better to leave things as they are for now and tackle this later though, so I think this is just something to bear in mind rather than something to act on.

Unfortunately I've run out of time and I haven't had time to run the code and really get into the nuts and bolts. However, the code looks great so I'm pretty confident this will be ready for use very quickly.

scripts/generate_gff3_db.py
diekhans: Most GFF3 source will most likely require it's own conversion
script.  There is a lot of variation in sources.  This will be
particularly important when we start implementing more complex
name queries.

We need to work towards having a library that can be programmed
against to make developing these programs  straight forward.

ga4gh/datamodel/datasets.py
macieksmuga: That's a good question - it depends on whether we'll decide to host the wiggle data in the same directory, or whether to put them in a parallel structure. Now that I think about it, it may make sense to preserve the current `sequenceAnnotations` directory, and introduce two subdirectories, `features` and `continuous` for the different types of data that will eventually be served up.

ga4gh/datamodel/sequenceAnnotations.py
macieksmuga: If we re-consider the colon, I'd like to explore using some mechanism other than non-alphanumeric characters, as the "native" ID's for features permit (and customarily use) a lot of wacky characters.

ga4gh/datamodel/ontologies.py
macieksmuga: I thought that I inherited the ontologies code as introduced by variant annotations wholesale (that's already in master). I'll check what's going on here.

ga4gh/datamodel/sequenceAnnotations.py
macieksmuga: You're right, should be passed in as a parameter.

scripts/generate_gff3_db.py
jeromekelleher: Ugh. Are GFF files that awful? It seems a cop out to expect the repo manager write their own parser for each GFF file. Anyway, ignore the above comment for now, we should discuss this after the initial drop into the sequence annotations branch.

diekhans: Not parsers per GFF3, but converters.  Common syntax, varying
data semantics.

GFF3 is a lot better than GFF2/GTF, because it added definition
of some data sematics, not just syntax.  There is still a wide
range of data that is encoded

There are also other formats we would want to convert to
features directly, such as BED format, and RepeatMasker, and not
convert to GFF3 just for input.

macieksmuga: Note that the way I wrote it, the GFF3 to DB converter can be used either as a standalone script _or_ as a library within another script. I'm using it in the latter mode in the `prepare_compliance_data.py` script, on line 168.

diekhans: The library portion should end up in a proper package, with generated documentation, etc and not mixed in
with a  program.  The library then be documented and the program serves as an example on writing new programs.
I am not a fan of putting program guts in python packages. 

But not until it's needed.  Please file a ticket to do this, but carry on!

ga4gh/datamodel/sequenceAnnotations.py
diekhans: good point on the @macieksmuga on the colon.  There should be
clear encoding rules for the components of an id, not just
string concatenation.  An element in the id should be able to
have any character in in and then be unambiguously
encode/decoded.

ga4gh/client.py
macieksmuga: it was lonely there, no wonder it was such a sad face ;)

ga4gh/gff3Parser.py
diekhans: gff3Parser.py is import from my code base.  I religiously only
make things `protected' (single underscore) if they are designed
to be overridden.  Overriding methods that are not intended to
be overridden leaves to implicit dependencies that are easy to
break.  These are sometimes behavioral dependencies that having
compiler will not help with.

Inheritance polymorphism is one of the pitfalls of OOP.

ga4gh/gff3Parser.py
diekhans: Is there any other reason to use __slots__?

ga4gh/gff3Parser.py
diekhans: Jerome Kelleher <notifications@github.com> writes:
> Is there a particular reason for making these atttributes public? These should
> probably be private to follow the conventions elsewhere.

Again, donated code.  While I use to use getters/setters, I am trying to follow more
a pythonic style these days.  This is a public attribute of the object.

ga4gh/serverconfig.py
david4096: Here is the simulated config as its own PR https://github.com/ga4gh/server/pull/1017

ga4gh/datamodel/sequenceAnnotations.py
jeromekelleher: 100% disagreement here @diekhans. We've put a lot of work into keeping the codebase as Python3 compatible as possible, and the only reason we're not fully supporting it now is because of Avro. The difference between range and xrange is irrelevant in this case as we're building a list comprehension.

Using 2to3 is a bad idea as that means you have two code bases (and this hasn't been recommended for a good while now). The best practise right now is to have one code base for both Python 2 and 3 and to use packages like six to work out the corner case.

Please don't use xrange.

ga4gh/gff3Parser.py
jeromekelleher: Not that I know of, but it's sufficiently different from the rest of the code to require a note on the point. It's better to be up front with the reader and explain the rationale for something rather than have them try to guess.

ga4gh/gff3Parser.py
jeromekelleher: I don't think it would take particularly long to make this change. Uniformity of style is essential in a large project. All other (non-trivial) classes use protected attributes.

ga4gh/gff3Parser.py
jeromekelleher: OK, well we need make a choice then I think:

1) Use the code here and follow the local coding conventions;
2) Explicitly make your code an external dependency and import it from pypi.

I don't really care that much about private/protected methods, but uniformity of style is essential.

ga4gh/datamodel/sequenceAnnotations.py
dcolligan: I don't mind anything, as long as it's URL-safe!

ga4gh/gff3Parser.py
diekhans: I am not arguing for against anything, just explaining.  You are
free to do what you like.  I am religions about documenting the
non-obvious,  I am just surprise this is non-obvious; but I tend
to read language specifications.

ga4gh/gff3Parser.py
diekhans: @jeromekelleher I am, again, not argument against the change.
Local convention is important, having dealt with code that had
little.  I just wanting to understand the rationale.

I would be concerned if `_' was there to indicate things can be
overridden rather than just communicating data hiding.

A tool that does better checking would save you time. I think
this is especially important for the 2 vs 3, since there are
subtle things.  I have looked at packages that tried to abstract
this away, and it didn't seem elegant enough to pursue.

jeromekelleher: The rationale for ``_`` is 95% data hiding, but also leaving the door open to inheritance in the odd occasions that it turns out to be useful. I used to use ``__`` very systematically in the past but I ended up finding it causing quite a few annoying issues (stuff like, even if a particular method is public, if it uses private variables it will fail in obscure ways), and basically requiring a total replumbing of a class if it turned out I wanted to create a subclass. So, I decided that protected variables were generally more useful and just went with those in the majority of cases.

I don't think there's anything wrong with inheritance, it just has to be used judiciously like any other programming technique. 

As I say, the motivation is almost entirely data hiding by convention, and discouraging the access of anything but simple read-only attributes via ``object.attribute``. I think making attributes public by default leads to all sorts of horrible interdependencies between objects, with object X reaching in and modifying attributes of object Y, making it very hard to figure out where code actually resides. By keeping attributes protected by default we discourage this sort of practice without being overly stringent.

Re Python 2 and 3, there's not all that much to look out for, as we've covered most of the truly nasty gotchas by using the futures imports. I think I've had to make about 5 to 10 comments during the entire history of the project asking people to stop using Python 2 only constructs.

ga4gh/gff3Parser.py
jeromekelleher: I guess I'm just suspicious of premature optimisation. If someone makes a perf optimisation like this, as a code reader I would like to be convinced that it's worthwhile.

diekhans: Jerome Kelleher <notifications@github.com> writes:
> I guess I'm just suspicious of premature optimisation. If someone makes a perf
> optimisation like this, as a code reader I would like to be convinced that it's
> worthwhile.

@macieksmuga, please add a note when you go over this that there can be millions
of features (2.7M in current GENCODE).

ga4gh/datamodel/sequenceAnnotations.py
macieksmuga: roger that :)

macieksmuga:
@jeromekelleher I'm following your suggestion, and introducing sphinx-style docstring params a whole file at a time (so I removed them for now from methods within much larger files that don't yet comply with this standard).

ga4gh/gff3Parser.py
macieksmuga: Put me on the record for agreeing wholeheartedly with @jeromekelleher about premature optimization. We can revisit this once we've hit an actual issue with use.

diekhans: Either delete it or add a comment; I don't see the down side of
having it, but I don't care.  I put in in there because I was
process very larger files.  It was not worth my time to do
performance analysis to see if it mattered, as there were no
downsides to using a standard python feature.

ga4gh/datamodel/sequenceAnnotations.py
macieksmuga: The logic here is a bit touchy, and took me a couple of tries to get it right. I appreciate the suggestion but want to leave it as is for now.

ga4gh/datamodel/datasets.py
macieksmuga: I'll leave it as is for now, unless you guys think this is an immediate issue. Propose changing this when the wiggle (continuous) data endpoints are added.

diekhans: +1 for leaving as-is and moving forward.  Our experience with
deploying larger data sets will inform of the best organization
for the back end.

ga4gh/cli.py
```diff
@@ -735,6 +904,21 @@ def addEffectsArgument(parser):
             """)
 
 
+def addFeatureTypesArgument(parser):
+    parser.add_argument(
+        "--featureTypes", "-n", default="",
```
david4096: How did you choose `-n`?

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,432 @@
...
+
+
+class AbstractFeatureSet(datamodel.DatamodelObject):
+    """
+    A set of sequence features annotations
+    """
+    compoundIdClass = datamodel.FeatureSetCompoundId
+
+    def __init__(
+            self, parentContainer, localId, referenceSetId=None):
```
david4096: It appears there is not currently a way to associate a featureSet with a reference set. This is not unique to sequence annotations, however, and can be left as an issue.

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,432 @@
...
+        :return: string representing ID for the specified GA4GH protocol
+            Feature object in this FeatureSet. If None is passed in,
+            an empty string is passed out, as required by the protocol.
+        """
+        if featureId is None:
+            # This can happen if no feature ID is present in the database,
+            # as can be the case with (absent) parent IDs.
+            return ""
+        compoundId = datamodel.FeatureCompoundId(
+            self.getCompoundId(), _decolonize(featureId))
```
david4096: Although in general I believe encoding identifiers this way is an anti-pattern instead of adding and removing separators you might consider base64 encoding/decoding again. Although it could lead to extremely lengthy identifiers you will have good guarantees your characters are safe.

tests/unit/test_gff3_parser.py
```diff
@@ -0,0 +1,97 @@
+"""
+Data-driven tests for rna quantification.
```
david4096: Typo in comment.

tests/unit/test_simulated_stack.py
```diff
@@ -710,6 +724,89 @@ def testVariantAnnotationsSearch(self):
...
+            request.datasetId = dataset.getId()
+            self.verifySearchMethod(
+                request, path, protocol.SearchFeatureSetsResponse, featureSets,
+                self.verifyFeatureSetsEqual)
+        for badId in self.getBadIds():
+            request = protocol.SearchFeatureSetsRequest()
+            request.datasetId = badId
+            self.verifySearchMethodFails(request, path)
+
+    def testGetFeature(self):
```
david4096: Since these tests use the simulator they don't actually show the query interface at work. I'm not sure it makes sense to mock the entire query interface when we are backed by a database, but it will prove that protocol elements are well formed.

tests/unit/test_simulated_stack.py
```diff
@@ -710,6 +724,89 @@ def testVariantAnnotationsSearch(self):
...
+        dataset = self.dataRepo.getDatasets()[0]
+        featureSet = dataset.getFeatureSets()[0]
+        request = protocol.SearchFeaturesRequest()
+        request.featureSetId = featureSet.getId()
+        request.referenceName = "chr1"
+        request.start = 0
+        request.end = 2**16
+        request.parentId = None
+        request.ontologyTerms = []
```
david4096: Should this be featureTypes?

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -0,0 +1,432 @@
...
+            featureTypes=[], parentId=None, numFeatures=10):
+        """
+        Returns a set number of simulated features.
+
+        :param referenceName: name of reference to "search" on
+        :param start: start coordinate of query
+        :param end: end coordinate of query
+        :param pageToken: None or int
+        :param pageSize: None or int
+        :param ontologyTerms: optional list of ontology terms to limit query
```
david4096: ontologyTerms -> featureTypes

tests/unit/test_simulated_stack.py
diekhans: Yes it should be `feature type'.  ontologyterm is a type, like
`int' is a type, we would have a field names ints in a request.

jeromekelleher:
Where are we with this @macieksmuga? The PR thread is getting a bit hard to follow, so perhaps we should merge it soon and create issues to track any remaining comments that need to be addressed?

ga4gh/cli.py
macieksmuga: It used to be called `--ontologyTerm` and `-o` was already taken. I'm changing it now.

ga4gh/datamodel/sequenceAnnotations.py
macieksmuga: Created #1030 to deal with this.

ga4gh/datamodel/sequenceAnnotations.py
macieksmuga: I agree - in this case, the implementation is encapsulated inside a pair of utility functions, so switching it out to something saner in the near future should be trivial.

tests/unit/test_simulated_stack.py
macieksmuga: Agreed, but I'm leaving them in for now.

tests/unit/test_simulated_stack.py
macieksmuga: Yes it should be. I had simply missed one occurrence in my `s/featureTypes/ontologyTerms/g`. Bless Python and its infinitely liberal attitude towards object attributes.

macieksmuga:
@jeromekelleher I'm in the middle of incorporating @david4096's latest suggestions, after committing those let's merge you suggested.

PR #1005, opened 2016-03-21
Does this break?

dcolligan:
Testing Jerome's theory about CI failing because of non-pinned library


dcolligan:
No, it does not break!

dcolligan:
@jeromekelleher

jeromekelleher:
Weird... Thanks for testing this out @dcolligan.

dcolligan:
Oh, I know why this still works... your error only was caused when you ran the `ga4gh_server` installed package, and we don't do that anywhere in the CI.  We should add that somehow.

Issue #1006, opened 2016-03-21
Pin exact versions of all libraries in setup.py


dcolligan:
See https://github.com/ga4gh/server/issues/1003 and https://github.com/ga4gh/server/pull/1004

reece:
Have you considered a requirements file instead?

To the extent that there's a convention, setup.py (install_requires) is best used to define the minimal package dependencies, typically unversioned. It is these dependencies that one would get with `python setup.py develop`, for example. In rare cases, you might pin as >1.2.3 if you know that there's a bug fixed in a particular version. For hgvs, I tell people to pin as >=0.4, <0.5, which means the'll get the 0.4.x patches, but won't inadvertently upgrade to 0.5 (which will have api changes).

Requirements files are the more conventional place to pin exact versions, often based on functional testing and QC team validation.

Because pinning may have implications for package consumers, it's best to let applications pin versions and to NOT do so in library packages. (Imagine the mess when A depends on B and C, and B and C specify incompatible versions of D.)

https://python-packaging-user-guide.readthedocs.org/en/latest/requirements/

dcolligan:
Fixed in #1010

Issue #1003, opened 2016-03-21
New version of pyjwkest breaks demo


jeromekelleher:
The pyjwkest package has been updated, and has broken the oic version we currently depend on. This breaks our installation:
```
(ga4gh-env)jk@holly$ ga4gh_server                                                                                                                                                                                  
Traceback (most recent call last):
  File "/home/jk/tmp/ga4gh-demo/ga4gh-env/bin/ga4gh_server", line 7, in <module>
    from ga4gh.cli import server_main
  File "/home/jk/tmp/ga4gh-demo/ga4gh-env/local/lib/python2.7/site-packages/ga4gh/cli.py", line 21, in <module>
    import ga4gh.frontend as frontend
  File "/home/jk/tmp/ga4gh-demo/ga4gh-env/local/lib/python2.7/site-packages/ga4gh/frontend.py", line 21, in <module>
    import oic.oauth2
  File "/home/jk/tmp/ga4gh-demo/ga4gh-env/local/lib/python2.7/site-packages/oic/oauth2/__init__.py", line 9, in <module>
    from oic.oauth2.base import PBase
  File "/home/jk/tmp/ga4gh-demo/ga4gh-env/local/lib/python2.7/site-packages/oic/oauth2/base.py", line 7, in <module>
    from oic.utils.keyio import KeyJar
  File "/home/jk/tmp/ga4gh-demo/ga4gh-env/local/lib/python2.7/site-packages/oic/utils/keyio.py", line 18, in <module>
    from jwkest.jwk import rsa_load
  File "/home/jk/tmp/ga4gh-demo/ga4gh-env/local/lib/python2.7/site-packages/jwkest/jwk.py", line 13, in <module>
    from Crypto.PublicKey.RSA import RsaKey
ImportError: cannot import name RsaKey
```

We need to fix this ASAP, as our current version on pypi does not install. I suggest we pin to version
pyjwkest==1.0.1 as this is the currently known working version, and release an updated server.

We should investigate either pinning pyjwkest==1.0.1 or updating oic in master.

jeromekelleher:
Closed in #1007.

PR #1007, opened 2016-03-21
Pin all library versions in setup.py...

dcolligan:
... by reading them from requirements.txt

Issue #1006 

So, turns out we can't pin every library that we have since some of them "require" different versions of certain libraries than the ones we need to run the server successfully.  I found only one case of this: the package `yubico-client` (which we have pinned at 1.9.1 in `requirements.txt`, the most recent version of the package) requires a `requests` library version in the range [2.2.0, 2.3) and we are pinning `requests` version 2.7.  Rolling `requests` back to a version that `yubico-client` likes breaks some of our tests.  Of course, our tests and install do seem to work fine with `requests` version 2.7, so I think what I have here will work.


dcolligan:
(also the latest release of `yubico-client` is in 2014, so I don't anticipate it revving and breaking our code anytime soon)

jeromekelleher:
A problem with this is that we now require all development packages for deployment also (stuff like freezegun, e.g.). Not sure if it's a problem though. Thoughts?

setup.py
```diff
@@ -10,6 +10,22 @@
 with open("README.pypi.rst") as readmeFile:
     long_description = readmeFile.read()
 
+install_requires = []
+with open("requirements.txt") as requirementsFile:
+    lines = requirementsFile.readlines()
+    for line in lines:
```
jeromekelleher: ``for line in requirementsFile:`` is a bit cleaner.

dcolligan:
True.  Would you support separating out those packages that are clearly only for development into a `dev-requirements.txt` and not reading those into `setup.py`? (And then modifying documentation appropriately)

jeromekelleher:
I guess so. I'd like to see what the python packaging authority have to say about this, as it must be a very common problem. Do they have any guidance on how to solve the wayward dependency graph?

dcolligan:
Everybody seems to have a different way of solving things.  This is probably the best thread I've found on the issue, although it's by no means conclusive https://news.ycombinator.com/item?id=11210370 here is some documentation that describes the differences between setup.py and requirements.txt http://python-packaging-user-guide.readthedocs.org/en/latest/requirements/

A lot of people seem to be recommending splitting development.txt into multiple files for dev,prod,etc. and putting different requirements in each... which seems fine until a point -- we want to test, develop and deploy with (at least a subset of) the exact same libraries, so we avoid breakage, right?

One of the things that makes it a bit confusing is that setup.py and requirements.txt provide some of the same functionality.  So, for instance, setup.py has a `tests_require` in which one can specify libraries used only for tests, and get those libraries with `python setup.py test` (although each of these commands has their own wrinkles... for instance, that command downloads the libraries to the current directory (!) not site-packages).

The long and short of it is that what these tools can do is complicated and overlapping, and everyone seems to have just slightly different requirements and setup, so I don't think there are any widely agreed upon best practices.

Here is what I think we should do, generally:

Any library that can be reached from something in setup.py's `entry_points` should be in both requirements.txt and setup.py, pinned to the same version.  These are the libraries that an end user could potentially interact with.  We want to specify these libraries as exactly as possible, and develop against the same versions.

Any library that can not be reached from something in setup.py's `entry_points` (i.e. ones that are only in tests, scripts, other non-deployed utilities, etc.) should be in a dev-requirements.txt (e.g. freezegun).  (2 questions: 1) we have the option of leaving these libraries unpinned, since they won't break end users... should we? 2) should we have the equivalent of an "include requirements.txt" at the start of dev-requirements.txt?  Or should they be disjoint?  This really only means that developers will have to run 1 `pip install -r` command instead of 2)

I think that makes sense, although perhaps the packaging/CI/etc. will surprise us in some way while figuring this out, so it will probably involve some trial and error.

jeromekelleher:
All very sensible @dcolligan. How about this (or something similar that works):

1) `requirements.txt` contain all modules reachable from setup.py's entry_points pinned at a good version. We read in `requirements.txt` in setup.py and use this for ``install_requires``.
2) `dev-requirements.txt` contains freezegun and other stuff we need for development and testing. We install these requirements before running tests during CI.

dcolligan:
This needs to be shipped at the same time as #1008, so I'm rolling this change into there.

reece:
Doh. I just saw that there are kinda two threads (#1006) on this topic. Here's a comment I just added there:

Have you considered a requirements file instead?

To the extent that there's a convention, setup.py (install_requires) is best used to define the minimal package dependencies, typically unversioned. It is these dependencies that one would get with python setup.py develop, for example. In rare cases, you might pin as >1.2.3 if you know that there's a bug fixed in a particular version. For hgvs, I tell people to pin as >=0.4, <0.5, which means the'll get the 0.4.x patches, but won't inadvertently upgrade to 0.5 (which will have api changes).

Requirements files are the more conventional place to pin exact versions, often based on functional testing and QC team validation.

Because pinning may have implications for package consumers, it's best to let applications pin versions and to NOT do so in library packages. (Imagine the mess when A depends on B and C, and B and C specify incompatible versions of D.)

It's tempting to read requirements in setup.py, but they really do have different uses. I would avoid that.

https://python-packaging-user-guide.readthedocs.org/en/latest/requirements/

dcolligan:
Hi @reece , you're forgiven... there are a lot of threads on this bundle of related topics!

This is the problem I am concerned with over all else: that we do not ship executables that, when installed with pip, do not work (and, secondarily, to provide a precisely specified development environment for developers).  This is what happened with #1003 -- pyjwkest revved and broke oic, which in turn broke the server.  From that, I conclude that any version change in any library in our dependency graph has the potential to break our executables.  This is the risk that we want to eliminate, or at least mitigate.

I know that there are conventions regarding how the version numbers of python packages relates to the changes that happen within them.  Unfortunately, I do not want to depend on the good will and judiciousness of every package in our dependency graph not to break anything.  Better to be safe than sorry.

I know that this strategy does have downsides, including not getting the most updated packages, with their bugfixes and features.  However, this is a cost that I am willing to pay, and -- more to the point -- it's not the priority.

Currently, our requirements file and our setup.py specify a different set of packages.  The changes that I have in the pipeline in #1010 read the libraries from the requirements file into setup.py, so that they will be identical.  That change also pins all the libraries (and their dependencies) we use to an exact version, as well as creating a new `dev-requirements.txt` file intended to be used by developers that contains a superset of the libraries in `requirements.txt`.

Regarding library incompatibilities: yes, the scenario you point out could happen, and if it happens we will relax the exact pinning requirement in that particular case.  (Of course, the problems that exact pinning causes are a subset of library incompatibility problems of this type.)  The implications for package consumers are more worrying (since in that case the problem is detected not by us, but by a consumer), but still tolerable, I think.  We're primarily shipping standalone executables, not code that is going to be included as libraries by other software (although I understand this is a shaky assumption).

I did review what literature I could find on the internet about best practices for specifying requirements, including the page that you linked.  Unfortunately, I don't think what it recommends guarantees that our shipped executables won't break.  What I did find was that every project seemed to have slightly different requirements, and used some combination of setup.py/requirement(s) files in whatever way that fit the needs of their project.  (Although I was a bit surprised not to find more, or really any, discussion on how not to have your packages break because of dependency revving.  Surely this is a problem that just about every open source python project faces?)

I am open to being convinced otherwise about my proposed course of action, but right now I think my proposed change is the best for the ga4gh server.  Let me know if you have further thoughts.

jeromekelleher:
Thanks for the input @reece. I started out with your point of view, hoping to pin as little as possible and therefore disrupt users as little as possible. However, we have a real problem here with the deep dependency tree required by OIDC, which pulls in all sorts of Crypto requirements. I agree what we've done in #1010 isn't ideal, but the first priority has to be a working package on pypi at all times. We can look at other ways of solving our specific issues though.

reece:
Thanks for your hard work guys. Packaging is hard. I completely trust whatever you come up with!

Issue #1008, opened 2016-03-21
Invoke installed ga4gh_* packages as part of CI


dcolligan:
@jeromekelleher discovered (https://github.com/ga4gh/server/issues/1003) that it is possible for our CI to pass and the `ga4gh_server` utility to fail to start.  We should have a step at the end of our CI that tests that we can start and invoke the ga4gh_* utilities without any issue.

dcolligan:
(although I am a bit surprised that `tests/end_to_end/test_server_start.py` did not catch that issue... will investigate)

dcolligan:
Fixed in #1010

PR #1010, opened 2016-03-22
Test package-installed server startup

dcolligan:
This new test runs only on travis, normally

Issue #1008

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,15 +9,23 @@ cache:
 
 install:
   - pip install pip --upgrade 
-  - pip install -r requirements.txt
   - python setup.py sdist
   - pip install dist/ga4gh*.tar.gz
+# every installable in setup.py's entry_points should be tested here
+  - ga4gh_client --version
+  - ga4gh_configtest --version
+  - ga4gh_server --version
+  - ga2vcf --version
+  - ga2sam --version
+  - ga4gh_repo --version
 
 before_script:
   # the following two lines are to prevent travis from
   # verbosely complaining about a security problem; see issue 54
   - mkdir $HOME/.python-eggs
   - chmod og-w $HOME/.python-eggs
+  - pip install -r dev-requirements.txt
+
 
 # run_tests.py runs everything under the script: tag so only put commands
 # under it that we want to run (and want to be able to run) as local tests
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,26 +1,45 @@
-Flask==0.10.1
+# This file lists requirements needed to run the executables shipped
+# by ga4gh/server.
+#
+# These requirements are read directly into setup.py, so specify them
+# in the order that they should be in in setup.py
+
+### This section is for dependencies of the libraries ###
+# these libraries are the set listed by pipdeptree -f -w
+# that are dependencies of libraries listed in the next section
+
+Werkzeug==0.11.5
+MarkupSafe==0.23
+itsdangerous==0.24
+six==1.10.0
+pycrypto==2.6.1
+Mako==1.0.4
+Beaker==1.8.0
+funcsigs==0.4
+alabaster==0.7.7
+cryptography==1.3.1
+idna==2.1
+pyasn1==0.1.9
+enum34==1.1.2
+ipaddress==1.0.16
+cffi==1.5.2
+pycparser==2.14
+Jinja2==2.7.3
+future==0.15.2
+pyjwkest==1.0.1
+
+### This section is for the actual libraries ###
+# these libraries are imported in code that can be reached via 
+# an executable that we ship
+
+# Flask must come after all other requirements that have "flask" as a
+# prefix due to a setuptools bug.
 Flask-Cors==2.0.1
+Flask==0.10.1
 avro==1.7.7
-coverage==3.7.1
-flake8==2.3.0  # Set due to conflict with pep8 in version 2.4.0
-freezegun==0.3.6
-guppy==0.1.10
 humanize==0.5.1
-mock==1.0.0
-nose==1.3.7
-pep8==1.6.2
 pysam==0.8.4
-PyVCF==0.6.7
-PyYAML==3.11
 requests==2.7.0
-snakefood==1.4
-Sphinx==1.3.1
 oic==0.7.6
 pyOpenSSL==0.15.1
 lxml==3.4.4
-# Requirements for test OIDC provider.
-cherrypy==3.2.4
-pyaml==15.03.1
-Jinja2==2.7.3
-yubico-client==1.9.1
-pyjwkest==1.0.1
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,0 +1,32 @@
+# This file lists the requirements needed to develop on ga4gh/server.
+
+# Include the requirements.txt file
+-r requirements.txt
+
+### This section is for dependencies of the libraries ###
+# these libraries are the set listed by pipdeptree -f -w
+# that are dependencies of libraries listed in the next section
+# and are not already listed in requirements.txt
+
+python-dateutil==2.5.1
+PyYAML==3.11
+
+### This section is for the actual libraries ###
+# we import these libraries somewhere in the codebase but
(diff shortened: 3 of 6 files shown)


dcolligan:
Actually, maybe we should do the install, then run the travis_only test, then pip install requirements.txt, then do the regular tests (to more realistically simulate a pip install that an end user might be doing) ... what do you think, @jeromekelleher ?

dcolligan:
(also, we should get this PR to a point where it is actually failing, so we know that it reproduces the bug we're trying to fix)

.travis.yml
```diff
@@ -25,6 +25,11 @@ script:
   - flake8 *.py tests ga4gh scripts --exclude=ez_setup.py
   - nosetests --with-coverage --cover-package ga4gh
               --cover-inclusive --cover-min-percentage 85
-              --cover-branches --cover-erase
+              --cover-branches --cover-erase -a '!travis_only'
```
jeromekelleher: What's this for?

jeromekelleher:
I think we'd need to do this in a different travis target that gets run before any of the others. This should be something like

```
$ python setup.py sdist
$ pip install <tarball>
$ ga4gh_server --version
```

This should be enough to shake out the majority of problems. We could then proceed with the rest of the tests as before, perhaps after running pip install -r requirements.txt?

.travis.yml
dcolligan: This means "do not run tests tagged with the attribute `travis_only`".  We do not want to run tests with installed executables in development, since they might be (probably are) from a different code base that the one that one is changing.

jeromekelleher: I don't think we need this, or to touch the test cases. We should be able to provoke the error with

```
pip install <tarball>
ga4gh_server --version
```
Does this work?

dcolligan:
Excellent!  This new version catches the failure (from the Travis CI output):
```
$ ga4gh_client --version
  File "/home/travis/virtualenv/python2.7.9/bin/ga4gh_client", line 7, in <module>
    from ga4gh.cli import client_main
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/ga4gh/cli.py", line 23, in <module>
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/ga4gh/frontend.py", line 21, in <module>
    import oic.oauth2
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/oic/oauth2/__init__.py", line 9, in <module>
    from oic.oauth2.base import PBase
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/oic/oauth2/base.py", line 7, in <module>
    from oic.utils.keyio import KeyJar
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/oic/utils/keyio.py", line 18, in <module>
    from jwkest.jwk import rsa_load
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/jwkest/jwk.py", line 13, in <module>
    from Crypto.PublicKey.RSA import RsaKey
ImportError: cannot import name RsaKey
```

dcolligan:

jeromekelleher:
Excellent, great work @dcolligan! Now, how do we catch the full dependency tree? I think it's enough to use the fancy tool you found to populate  requirements.txt by hand, and to update it periodically. What do you think?

dcolligan:
That sounds good, although specifying and updating the libraries is going to be a bit cumbersome with so much specified.  Here is the list for the project:
```
$ pipdeptree -f 2> /dev/null
flake8==2.3.0
    pyflakes
    mccabe
    pep8==1.6.2
guppy==0.1.10
humanize==0.5.1
nose==1.3.7
lxml==3.4.4
yubico-client==1.9.1
    requests==2.7.0
mock==1.0.0
oic==0.7.6
    requests==2.7.0
    pycrypto==2.6.1
    pyjwkest==1.0.1
      pycrypto==2.6.1
      requests==2.7.0
      six==1.10.0
      future==0.15.2
    Mako==1.0.4
      MarkupSafe==0.23
    Beaker==1.8.0
      funcsigs==0.4
    alabaster==0.7.7
    pyOpenSSL==0.15.1
      six==1.10.0
      cryptography==1.3.1
        idna==2.1
        pyasn1==0.1.9
        six==1.10.0
        setuptools
        enum34==1.1.2
        ipaddress==1.0.16
        cffi==1.5.2
          pycparser==2.14
pyaml==15.03.1
    PyYAML==3.11
wsgiref==0.1.2
coverage==3.7.1
freezegun==0.3.6
    python-dateutil
    six==1.10.0
CherryPy==3.2.4
pysam==0.8.4
PyVCF==0.6.7
    distribute
Flask-Cors==2.0.1
    Flask==0.10.1
      Werkzeug==0.11.5
      Jinja2==2.7.3
        MarkupSafe==0.23
      itsdangerous==0.24
    six==1.10.0
avro==1.7.7
snakefood==1.4
```

jeromekelleher:
OK, we should be able to separate out the deployment and dev packages out of that easily enough. At first glance, we have
```
snakefood==1.4
PyVCF==0.6.7
freezegun==0.3.6
    python-dateutil
    six==1.10.0
coverage==3.7.1
CherryPy==3.2.4
flake8==2.3.0
    pyflakes
    mccabe
    pep8==1.6.2
guppy==0.1.10
nose==1.3.7
```
are all for dev?

dcolligan:
@jeromekelleher right, check the most recent push... I think this one is good to merge

dcolligan:
(and I don't think it's necessary to pin dependencies of libraries in dev-requirements... the main point of this change is to ensure our shipped installables don't break... development environment breaking is of much lower importance)

requirements.txt
```diff
@@ -1,26 +1,46 @@
...
 PyYAML==3.11
 requests==2.7.0
-snakefood==1.4
-Sphinx==1.3.1
 oic==0.7.6
 pyOpenSSL==0.15.1
 lxml==3.4.4
-# Requirements for test OIDC provider.
-cherrypy==3.2.4
+# Requirements for test OIDC provider
```
jeromekelleher: These should be in development only. The OIDC provider is purely for testing, we should absolutely not encourage people to use it in deployments.

jeromekelleher:
Looks good modulo a few extra packages in requirements.txt. I think it's important we pin requirements for development, it would be very frustrating for developers finding their tests running locally but failing in CI (or vice versa) because of some weird dependency issues.

It shouldn't be too hard to do a package refresh every now and again to bring them up to date.

dcolligan:
@jeromekelleher updated

jeromekelleher:
Do we use these in deployment? 

 PyVCF==0.6.7
 PyYAML==3.11
 requests==2.7.0

Sorry, should have seen these a minute ago.

dcolligan:
@jeromekelleher updated

PyVCF and PyYAML are not needed in deployment; I moved them to dev
requests is required by oic, which is needed in deployment

jeromekelleher:
Excellent, thanks @dcolligan. @reece makes great points over in #1007, but we have a particularly troublesome dependency graph here and there's very little we can do about it if we want to support authentication (which we do). Let's try this out for a while and see how it works. We can always have another look before we release.

macieksmuga:
What do the rest of us do when we add packages needed for new functionality?

dcolligan:
Good question.  Here's the algorithm:

Determine if the new library can be reached via a shipped executable.  If so, place it in the bottom section in `development.txt` and use `pipdeptree` to find any dependencies of that library that aren't already listed in `development.txt` and put them in the top section.  If the new library can not be reached via a shipped executable (e.g. only in tests, scripts, etc.), do the same, but in `dev-requirements.txt` (with the exception that no libraries should be added to `dev-requirements.txt` that are already in `requirements.txt` because the former imports the later -- pip will give an error on double-imports anyway).  Every library needs to be pinned to an exact version -- exceptions are only made in case of incompatible version issues.

dcolligan:
Every time anyone makes a change to either `requirements.txt` or `dev-requirements.txt` they will probably need to do this.  There might be a way to automate this.

PR #880, opened 2016-01-26
Updates for pysam 0.8.5.

jeromekelleher:
Uses the local download facility for BAM indexes.

-1: This shouldn't be merged until pysam 0.8.5 is actually released.

PR #1016, opened 2016-03-23
Download compliance data from github

david4096:
This updates the prepare_compliance_data script to download data from the compliance repo. This will improve ease of starting development.

The original default was to add the current directory if no input directory was specified. This changes the default to download from an external source when no input directory is specified, currently `compliance-master`.

### DIFF
--- a/scripts/prepare_compliance_data.py
+++ b/scripts/prepare_compliance_data.py
@@ -8,19 +8,34 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import shutil
 import argparse
 import os
-import sys
+import shutil
 import glob
 import pysam
 import utils
+import tempfile
+import zipfile
 
 
 class ComplianceDataMunger(object):
     def __init__(self, args):
         self.inputDirectory = args.inputDirectory
         self.outputDirectory = args.outputDirectory
+        # If no input directory is specified download from GitHub
+        if args.inputDirectory is None:
+            utils.log("Downloading test data...")
+            self.tempdir = tempfile.mkdtemp()
+            assert(os.path.exists(self.tempdir))
+            url = "https://github.com/ga4gh/compliance/archive/master.zip"
+            filePath = os.path.join(self.tempdir, 'compliance-master.zip')
+            downloader = utils.HttpFileDownloader(url, filePath)
+            downloader.download()
+            utils.log("Extracting test data...")
+            with zipfile.ZipFile(filePath, "r") as z:
+                z.extractall(self.tempdir)
+            self.inputDirectory = os.path.join(
+                self.tempdir, 'compliance-master', 'test-data')
 
         # get all the reference files (they'll be the ones with .fa extension)
         self.referenceFiles = map(
@@ -69,7 +84,7 @@ class ComplianceDataMunger(object):
 
         # Clean out, make and re-populate references directory
         # For now, assume a single, statically-named referenceSet
-        print("Converting references...", file=sys.stderr)
+        utils.log("Converting references...")
         shutil.rmtree(self.refsetsDirectory, ignore_errors=True)
         os.makedirs(self.refsetsDirectory)
         shutil.copy(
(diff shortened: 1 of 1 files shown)


scripts/prepare_compliance_data.py
```diff
@@ -8,19 +8,38 @@
...
+import zipfile
 
 
 class ComplianceDataMunger(object):
     def __init__(self, args):
         self.inputDirectory = args.inputDirectory
         self.outputDirectory = args.outputDirectory
+        # If no input directory is specified download from GitHub
+        if not args.inputDirectory:
+            print("Downloading test data...")
```
dcolligan: I have been using utils.log for statements like this instead of print, which for now just calls print, but the advantage is that we can easily change the stream in the future if we decide to do that.

scripts/prepare_compliance_data.py
```diff
@@ -8,19 +8,38 @@
...
+            self.tempdir = tempfile.mkdtemp()
+            assert(os.path.exists(self.tempdir))
+            url = "https://github.com/ga4gh/compliance/archive/master.zip"
```
dcolligan: We have a `FileDownloader` in utils... use that instead

dcolligan:
generally looks good, +1 after comments addressed

scripts/prepare_compliance_data.py
```diff
@@ -141,6 +160,8 @@ def run(self):
         os.makedirs(sequenceOntologyDir)
         shutil.copy(os.path.join(self.inputDirectory, "sequence_ontology.txt"),
                     os.path.join(sequenceOntologyDir, "sequence_ontology.txt"))
+        if self.tempdir:
```
jeromekelleher: What does this check mean? We should really be doing a try...finally for cleanup whenever we make temporary files in any case.  The easiest way to do it is to put in a ``cleanup`` method on your class (see #880 for an example).

david4096:
Updated @dcolligan @jeromekelleher .

scripts/prepare_compliance_data.py
```diff
@@ -8,19 +8,34 @@
...
+import tempfile
+import zipfile
 
 
 class ComplianceDataMunger(object):
     def __init__(self, args):
         self.inputDirectory = args.inputDirectory
         self.outputDirectory = args.outputDirectory
+        # If no input directory is specified download from GitHub
+        if not args.inputDirectory:
```
jeromekelleher: Should be `` if args.inputDirectory is not None:``, see https://www.python.org/dev/peps/pep-0008/#programming-recommendations point 2.

scripts/prepare_compliance_data.py
```diff
@@ -154,12 +172,15 @@ def main():
...
-        help="Path to local directory containing compliance dataset.",
-        default='.')
+        help="Path to local directory containing compliance dataset. "
+        "If no directory is provided this script will attempt to "
+        "download the compliance test-data from github",
+        default=None)
     parser.add_argument('--verbose', '-v', action='count', default=0)
     args = parser.parse_args()
     cdm = ComplianceDataMunger(args)
     cdm.run()
```
jeromekelleher: Need a try, finally here or we'll spray temp dirs over the file system when errors occur:
```python
try:
    cdm = ComplianceDataMunger(args)
    cdm.run()
finally:
    cdm.cleanup()
```

scripts/prepare_compliance_data.py
david4096: Do you mean `is None`?

jeromekelleher: Yes, probably!

jeromekelleher:
+1 after squash.

dcolligan:
doesn't look like you addressed any of my comments...

david4096:
@dcolligan sorry got lost during the squash, please merge if it looks ok!

dcolligan:
Looks like you have one more comment from Jerome to address and then I can merge it

david4096:
@dcolligan :8ball:

jeromekelleher:
@dcolligan, what do you think of rebase's vs. merges for bringing these PRs up to date with master? I lean towards rebasing always, but I'm happy to go with your opinion.

david4096:
Hey I just rebased it anyhows @jeromekelleher

dcolligan:
@jeromekelleher I prefer rebasing for updating non-long-lived branch PRs

dcolligan:
+1

Issue #596, opened 2015-08-28
SearchReads accepts only one ReadGroup ID, contrary to schema


hjellinek:
The schemas say that `SearchReadsRequest` accepts *at least* one ReadGroup ID, but the ref server accepts *at most* one.

```avro
record SearchReadsRequest {
   /**
   The ReadGroups to search. At least one readGroupId must be specified.
   */
   array<string> readGroupIds;
   ...
}
```

This test case is from a forthcoming compliance test, `ReadsSearchIT.searchReadsWithAllIdsReturnsReadsForEach`:

```java
threeReadGroupIds = getThreeReadGroupIdsSomehow(); // details omitted (obviously :-))
SearchReadsRequest request =
                SearchReadsRequest.newBuilder()
                                  .setReadGroupIds(threeReadGroupIds)
                                  .setStart(0L)
                                  .setEnd(150L)
                                  .setReferenceId(refId)
                                  .build();
SearchReadsResponse response = client.reads.searchReads(request);
```

The call to `searchReads` throws a `GAException`:

`{"errorCode":1151425451,"message":"Exactly one read group id must be specified"}`

with HTTP status `501` (`NOT_IMPLEMENTED`).

This is contrary to the comments in the schema, not to mention that the declaration of the `readGroupIds` field is `array<string>`, not a single `string`.

jeromekelleher:
This is an intentional limitation of the reference server, as this feature is too complicated for us to implement at the moment. We will try to implement it at some point, but it is difficult to do with our stateless paging architecture. 

The reference server's compliance score should be appropriately reduced for this.

david4096:
We've just proposed a change to the schemas that would make this feature "implementable". @diekhans 

https://github.com/ga4gh/schemas/pull/570

diekhans:
There was strong objection to removing the cross-BAM merge-sort of reads from the GA4GH API.  There are very good use cases for this in multi-sample variant calling and it is supported by htslib.

There should be no requirement to scale beyond small number of BAMs for the reference server.

It's very important the primary single BAM be supported very soon.  This is obtaining all readgroups
within a BAM as part of a single requests

dcolligan:
@jeromekelleher can you elaborate on why the limitation was put in place and what the roadmap to remove it would look like?

dcolligan:
From a @diekhans email:

Right now, we don't support the single most important use cases,
which is read all the readgroups in a single BAM.  The reference
server can make the requirement that BAMs are sorted by location
[...]

Issue #347, opened 2015-04-08
Expert review of the read translation code required


jeromekelleher:
There are multiple TODO items in the ``ga4gh/datamodel/reads.py`` file, particularly where we translate from pysam objects into the GA4GH equivalents (see the ``convertReadAlignment`` method). None of the current developers have the expertise required to be certain that the decisions made here are correct. We would very much appreciate some outside help on this.

dcolligan:
There's also the GA4GH -> pysam conversion code in `ga4gh/converters.py`

bioinformed:
I won't claim to be an expert, but I'll try to find time to review the conversion code.

dcolligan:
That would be awesome if you could, @bioinformed

jeromekelleher:
@jmarshall --- It seems you have been having a look over our read translation code (which we are _very_ grateful for!). Do you think we can mark this issue as done?

dcolligan:
I don't think this is close to done.  There are many TODOs in the relevant parts of `ga4gh/datamodel/reads.py`.  I'm sure there are more bugs in there.  On the other hand, given our uncertainty, this may be an otherwise eternally-open bug that serves no use and might be worth closing for that reason.

jmarshall:
So far I've only been reacting to issues that go past in my @ga4gh news feed… but it has occurred to me that I'm risking accidentally volunteering to review _reads.py_ in general :smile:

Okay, will continue to have a look, and I agree that there's still plenty of TODOs to look at.

Is the idea that people can run this server pointed at pretty much any BAM files they want to serve up, or are there constraints on the BAM files that the server likes?

jeromekelleher:
>Is the idea that people can run this server pointed at pretty much any BAM files they want to serve up, or are there constraints on the BAM files that the server likes?

At the moment, yes, we'd like to be able to point to an arbitrary (well formed) bam file and do something sensible. We can introduce restrictions if we need to though --- I think at the moment we don't really understand the range of inputs we can expect, so we haven't made any requirements.

Thanks for looking at this @jmarshall!

PR #1021, opened 2016-03-24
readsSearch multiple readGroupIds

dcolligan:
Issue #596

Depends on #1041


kozbo:
I moved the implementation of the tests to the top to say that they should be done first where ever possible.

dcolligan:
@kozbo btw I don't think assigning people their own PRs in GitHub is helpful... what information is this adding?

dcolligan:
@jeromekelleher I'd like to hear what you think about this POC so far

david4096:
For reference, this is the compliance request that isn't being fulfilled:

```
{
  "readGroupIds": [
    "YnJjYTE6SEcwMDA5NjpTUlIwNjI2MzQ",
    "YnJjYTE6SEcwMDA5NjpTUlIwNjI2MzU",
    "YnJjYTE6SEcwMDA5NjpTUlIwNjI2NDE",
    "YnJjYTE6SEcwMDA5OTpTUlI3NDE0MTE",
    "YnJjYTE6SEcwMDA5OTpTUlI3NDE0MTI",
    "YnJjYTE6SEcwMDEwMTpFUlIyMjk3NzY"
  ],
  "referenceId": "aGczNzpyZWZfYnJjYTE",
  "start": 0,
  "end": 150,
  "pageSize": null,
  "pageToken": null
}
```

kozbo:
@dcolligan - quite right. I did the assignment before I looked to see if this was an issue or a PR. My other comment about doing the tests first was more issue-centric too. apologies for the confusion.

ga4gh/backend.py
```diff
@@ -176,6 +176,20 @@ def _getEnd(cls, readAlignment):
             len(readAlignment.alignedSequence))
 
 
+class ReadGroupSetIntervalIterator(ReadsIntervalIterator):
+    """
+    An interval iterator for reads, for an entire ReadGroupSet
+    """
+    def __init__(self, request, parentContainer, reference):
+        self._reference = reference
```
jeromekelleher: Isn't this identical to ReadsIntervalIterator at the moment? If so, why implement these methods again?

ga4gh/backend.py
```diff
@@ -513,18 +527,41 @@ def readsGenerator(self, request):
...
+            request, readGroup, reference)
+        return intervalIterator
+
+    def _readsGeneratorMultiple(self, request):
+        compoundId = datamodel.ReadGroupCompoundId.parse(
+            request.readGroupIds[0])
+        dataset = self.getDataRepository().getDataset(compoundId.datasetId)
+        readGroupSet = dataset.getReadGroupSet(compoundId.readGroupSetId)
+        readGroupIds = readGroupSet.getReadGroupIds()
+        if (len(readGroupIds) != len(request.readGroupIds) or
```
jeromekelleher: The first clause here is unneccesary, set equality will check this as well.

ga4gh/cli.py
```diff
@@ -550,34 +550,14 @@ def __init__(self, args):
...
     def run(self):
         """
         Iterate passed read group ids, or go through all available read groups
         """
-        if not self._readGroupIds:
-            for referenceGroupId in self.getAllReadGroups():
-                self._run(referenceGroupId)
-        else:
-            for referenceGroupId in self._readGroupIds:
-                self._run(referenceGroupId)
```
jeromekelleher: The implementation isn't great right now, but we're losing some genuinely useful functionality here. Unless it's causing immediate issues, can we leave it as-is, and create an issue to address shortcomings?

ga4gh/datamodel/reads.py
```diff
@@ -289,6 +300,23 @@ def getNumUnalignedReads(self):
...
 
+    def getReadAlignments(self, reference, start, end):
+        """
+        Returns an iterator over all reads in a ReadGroupSet
+        """
+        samFile = self.getFileHandle(self._samFilePath)
+        referenceName = reference.getLocalId().encode()
+        start, end = self.sanitizeAlignmentFileFetch(start, end)
+        readAlignments = samFile.fetch(referenceName, start, end)
+        syntheticReadGroup = HtslibReadGroup(self, 'synthetic')
```
jeromekelleher: Confused by this... Is this to get access to the ``readGroup.convertReadAlignment`` method? If so, I think we should consider refactoring so that both classes have access to this function/method.

jeromekelleher:
Looks good to me @dcolligan, I think it's definitely on the right track.

ga4gh/backend.py
dcolligan: I suppose it is!  I thought that passing in a different object  would have different implications down the line, but I guess not if the read-accessor methods are named the same and have the same prototype.

ga4gh/cli.py
dcolligan: The client right now "fakes" using the multiple-readGroupId functionality by requesting each readGroup individually.  Yes, we can leave it as-is if you don't want the client to take advantage of the functionality I am now implementing.

ga4gh/datamodel/reads.py
dcolligan: Yes, that is the intention.  What refactor do you have in mind?  A mixin class that has getReadAlignments as a method?

ga4gh/cli.py
jeromekelleher: Are we replacing this functionality with the equivalent calls though? I don't want to have to type in 'n' readGroupIds just to get the readGroupSet functionality, that would be really tedious.

ga4gh/datamodel/reads.py
jeromekelleher: Yeah, I think that's probably a good approach. Say, ``AlignmentDataMixin`` which subclasses ``datamodel.PysamDatamodelMixin``? Creating a new readgroup here will complicate things in the long run as we're re-interpreting the header information for each incoming request.

ga4gh/cli.py
dcolligan: Well what do you think we should do?

jeromekelleher: Probably leave it as-is for now and follow up in another PR where we can think it through in isolation.

dcolligan: Roger that.

dcolligan:
@jeromekelleher updated to include mixin.  Merging the two getReadAlignments methods was a bit less pretty than anticipated, but I'm not sure if there's a way to simplify it (perhaps you have some ideas).  Also, what should we do for a readsSearch call with multiple readGroupIds (that is, searching the readGroupSet) for unmapped reads (see the TODO)?

ga4gh/datamodel/reads.py
```diff
@@ -91,6 +91,124 @@ def setFlag(flagAttr, flag):
...
+                if 'RG' in tags:
+                    alignmentReadGroupId = tags['RG']
+                    alignmentReadGroup = self._readGroupLocalIdMap[
+                        alignmentReadGroupId]
+                    yield self.convertReadAlignment(
+                        readAlignment, alignmentReadGroup)
+                else:
+                    # TODO what to do for unmapped reads (i.e. no
+                    # alignment read group context)? skip for now
+                    pass
```
jeromekelleher: What are we using the ReadGroup object for in this case during translation? It mignt be helpful to break up the conversion process into a number of steps that can be composed. It seems that the majority of the translation process only needs the read object. Can we break it up based on that?

dcolligan: In all cases, the readGroup is used in convertReadAlignment to set the read.readGroupId and the read.id (which is a compoundId that, of course, includes the readGroupId in its hierarchy).

To me, it sounds like we should not be setting these id(s) as we currently are in the case of unmapped reads (i.e. reads that have no readGroup).  What should we do then instead?  Keep in mind these ids need to be both unique and stable (i.e. we can't just use a random number).  What about:
```
datasetId:readGroupSetId:unmapped:readId
```

jeromekelleher: +1 on the ID scheme.

jeromekelleher: Er no, hold on. Is being unmapped the same thing as not being in a ReadGroup? I thought being unmapped meant there was no reference and coordinates, and not belonging to a readGroup essentially means that the read doesn't contain an RG tag. If this is true we shouldn't conflate the two ideas.

Do we need the third part of the ID string? If we get ``datasetId:readGroupSetId:readGroupId:readId`` then the read belongs to a read group, and if we get ``datasetId:readGroupSetId:readId`` we know it doesn't? Seems simple enough to tell the two cases apart. I'm probably misunderstanding what's going though.

dcolligan: Yes, you're right about unmapped, sorry.

Good point... I guess we don't actually need the readGroupId for a unique readId.  (However, we do still need to set the standalone read.readGroupId that is returned to the client, if not the readGroupId that is included in the read's compoundId.)

diekhans: Correct.  Every read is an a read group.  This usually
corresponds to which lane of the sequencer the read came
from. Mapped/upmapped has to do it it can be aligned to the
reference genomes.

Very different concepts.

dcolligan: @jeromekelleher I'm going to break the change of ReadAlignmentCompoundIds from `datasetId:readGroupSetId:readGroupId:readId` to `datasetId:readGroupSetId:readId` into a separate PR

jeromekelleher: Good idea.

dcolligan:
@jeromekelleher updated

(ignore the changes in cli.py ... I'll take those out later)

jeromekelleher:
Looks good to me @dcolligan...

dcolligan:
@jeromekelleher ready for final review

ga4gh/datamodel/reads.py
```diff
@@ -91,6 +91,121 @@ def setFlag(flagAttr, flag):
...
+            ret.alignment.position.position = read.reference_start
+            ret.alignment.position.strand = protocol.Strand.POS_STRAND
+            if SamFlags.isFlagSet(read.flag, SamFlags.READ_REVERSE_STRAND):
+                ret.alignment.position.strand = protocol.Strand.NEG_STRAND
+            ret.alignment.cigar = []
+            for operation, length in read.cigar:
+                gaCigarUnit = protocol.CigarUnit()
+                gaCigarUnit.operation = SamCigar.int2ga(operation)
+                gaCigarUnit.operationLength = length
+                gaCigarUnit.referenceSequence = None  # TODO fix this!
```
david4096: Would you please make an issue to go with this TODO? Does it require a lookup on the reference to perform properly?

dcolligan: The general issue is already covered by https://github.com/ga4gh/server/issues/347

ga4gh/backend.py
```diff
@@ -513,18 +513,40 @@ def readsGenerator(self, request):
...
+        compoundId = datamodel.ReadGroupCompoundId.parse(
+            request.readGroupIds[0])
+        dataset = self.getDataRepository().getDataset(compoundId.datasetId)
+        readGroupSet = dataset.getReadGroupSet(compoundId.readGroupSetId)
+        readGroupIds = readGroupSet.getReadGroupIds()
+        if set(readGroupIds) != set(request.readGroupIds):
+            raise exceptions.BadRequestException(
+                "If multiple readGroupIds are specified, "
+                "they must be all of the readGroupIds in a ReadGroup")
+        referenceSet = readGroupSet.getReferenceSet()
```
david4096: What do you think about making this error message visible to the user? This prints to the error log, but I think it would be good a user knew exactly what went wrong in this case.

david4096:
Took this down and ran it and it lets you search across a whole BAM's readgroups, nice! I verified for a few queries that the results were sorted well across pages. This is really useful! A [compliance test](https://github.com/ga4gh/compliance/issues/184) demonstrating this feature would be great. +1

ga4gh/backend.py
jeromekelleher: This should go back to the user; can you open an issue against the error handling code if a BadRequestException message doesn't get propogated back to the user please @david4096?

jeromekelleher:
Thanks @dcolligan, looks great. +1

Can you update and merge please?

Issue #1012, opened 2016-03-22
ID scheme allows non-unique IDs


david4096:
Identifiers are required to be unique, but naming two files the same will cause them to have the same ID.

```
import ga4gh.datamodel.variants as variants
import ga4gh.datamodel.reads as reads
import ga4gh.datamodel.datasets as datasets

ds = datasets.AbstractDataset("bruce")
rgs = reads.AbstractReadGroupSet(ds, "wayne")

rgs.getId()
# 'YnJ1Y2U6d2F5bmU'

vs = variants.AbstractVariantSet(ds, "wayne")

vs.getId()
#'YnJ1Y2U6d2F5bmU'
```

Perhaps we could add a character per type to the compound ID?

jeromekelleher:
Hah, well spotted @david4096! Yes, adding a char for each type is a good solution I think.

dcolligan:
It is a bit ugly that 2 different objects of different type can share the same id, but is it really a problem, strictly speaking?  We don't have any getObjectById function (or something similar) which searches the global id space.  We only have type-specific id spaces, I think.

jeromekelleher:
Hmm, this is true @dcolligan, we can always identify the correct object by the context. Does the spec have anything to say about this?

david4096:
Although I agree that we can function with degenerate IDs it is out of scope of the schemas.
```
* The standard way to programmatically reference an object is via an
  **id** field. The id is server-assigned, and must provide “id-like”
  semantics, including:

  * ids are unique within the scope of the server instance
...
The purpose of IDs is to allow
unique identification of all objects within a single server, such that
no two objects in a given server have the same ID and no object has
more than one ID.
```
https://raw.githubusercontent.com/ga4gh/schemas/master/doc/source/api/apidesign_intro.rst

jeromekelleher:
OK, that sounds conclusive. I think we should just use @david4096's idea and shove in a single char to make the IDs unique so we're following the letter of the spec.

dcolligan:
+1

diekhans:
From a user perspective, it's very confusing to not have ids
globally unique.  For debugging and if one does anything like a
cache.

Please figure out a short-term fix.  Once we get protocol
buffers integrated, it would be good to revisit the id scheme.

macieksmuga:
The following is a sketch of a solution architecture that I'd like to implement at some point in the future, not a reflection on the code that's generating these IDs now.

[...]

PR #1022, opened 2016-03-25
Ensure unique IDs across types

dcolligan:
Issue #1012

### DIFF
--- a/ga4gh/datamodel/__init__.py
+++ b/ga4gh/datamodel/__init__.py
@@ -120,6 +120,21 @@ class CompoundId(object):
     further up the tree. This list is a set of tuples giving the
     name and length of a given prefix forming an identifier.
     """
+    prefix = 'X'
+    """
+    A unique prefix for each type of CompoundId.  Used to ensure
+    any two IDs for different types are not identical.
+    """
+    prefixSeparator = '_'
+    """
+    A delimiter between the prefix and the rest of the ID
+    """
+    obfuscators = []
+    """
+    An array that defines the obfuscators to be used for each id in
+    containerIds.  A None value indicates that the class' own
+    obfuscator should be used for that containerId array index.
+    """
 
     def __init__(self, parentCompoundId, *localIds):
         """
@@ -139,10 +154,13 @@ class CompoundId(object):
         if len(localIds) != len(self.fields) - index:
             raise ValueError(
                 "Incorrect number of fields provided to instantiate ID")
-        for idFieldName, prefix in self.containerIds:
+        for i, (idFieldName, prefix) in enumerate(self.containerIds):
             values = [getattr(self, f) for f in self.fields[:prefix + 1]]
             containerId = self.separator.join(values)
-            obfuscated = self.obfuscate(containerId)
+            obfuscator = self.obfuscators[i]
+            if obfuscator is None:
+                obfuscator = self
+            obfuscated = obfuscator.obfuscate(containerId)
             setattr(self, idFieldName, obfuscated)
 
     def __str__(self):
@@ -165,7 +183,7 @@ class CompoundId(object):
             raise exceptions.BadIdentifierException(compoundIdStr)
         try:
             deobfuscated = cls.deobfuscate(compoundIdStr)
-        except TypeError:
+        except (TypeError, ValueError):
             # When a string that cannot be converted to base64 is passed
             # as an argument, b64decode raises a TypeError. We must treat
             # this as an ID not found error.
@@ -187,6 +205,7 @@ class CompoundId(object):
         fashion. This is not intended for security purposes, but rather to
         dissuade users from depending on our internal ID structures.
         """
+        idStr = cls.prefix + cls.prefixSeparator + idStr
         return base64.urlsafe_b64encode(str(idStr)).replace(b'=', b'')
 
     @classmethod
@@ -196,8 +215,10 @@ class CompoundId(object):
         If an identifier arrives without correct base64 padding this
         function will append it to the end.
         """
-        return base64.urlsafe_b64decode(str((
+        decoded = base64.urlsafe_b64decode(str((
             data + b'A=='[(len(data) - 1) % 4:])))
+        index = decoded.index(cls.prefixSeparator)
+        return decoded[index + 1:]
 
 
 class ReferenceSetCompoundId(CompoundId):
@@ -206,6 +227,8 @@ class ReferenceSetCompoundId(CompoundId):
     """
     fields = ['referenceSet']
     containerIds = [('referenceSetId', 0)]
+    obfuscators = [None]
+    prefix = 'RS'
 
 
 class ReferenceCompoundId(ReferenceSetCompoundId):
@@ -213,6 +236,8 @@ class ReferenceCompoundId(ReferenceSetCompoundId):
     The compound id for a reference
     """
     fields = ReferenceSetCompoundId.fields + ['reference']
+    obfuscators = [ReferenceSetCompoundId]
+    prefix = 'R'
 
 
 class DatasetCompoundId(CompoundId):
@@ -221,6 +246,8 @@ class DatasetCompoundId(CompoundId):
     """
     fields = ['dataset']
     containerIds = [('datasetId', 0)]
+    obfuscators = [None]
+    prefix = 'D'
 
 
 class VariantSetCompoundId(DatasetCompoundId):
@@ -229,6 +256,8 @@ class VariantSetCompoundId(DatasetCompoundId):
     """
     fields = DatasetCompoundId.fields + ['variantSet']
     containerIds = DatasetCompoundId.containerIds + [('variantSetId', 1)]
+    obfuscators = [DatasetCompoundId, None]
+    prefix = 'VS'
 
 
 class VariantAnnotationSetCompoundId(VariantSetCompoundId):
@@ -238,6 +267,8 @@ class VariantAnnotationSetCompoundId(VariantSetCompoundId):
     fields = VariantSetCompoundId.fields + ['variantAnnotationSet']
     containerIds = VariantSetCompoundId.containerIds + [
         ('variantAnnotationSetId', 2)]
+    obfuscators = [DatasetCompoundId, VariantSetCompoundId, None]
+    prefix = 'VAS'
 
 
 class VariantSetMetadataCompoundId(VariantSetCompoundId):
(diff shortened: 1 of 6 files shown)


ga4gh/datamodel/__init__.py
```diff
@@ -229,6 +256,8 @@ class VariantSetCompoundId(DatasetCompoundId):
     """
     fields = DatasetCompoundId.fields + ['variantSet']
     containerIds = DatasetCompoundId.containerIds + [('variantSetId', 1)]
+    obfuscators = [DatasetCompoundId, None]
```
macieksmuga: here and elsewhere, I don't see any overrides on the `obfuscate` method, so why this list of obfuscators (classes, no less!) in every subclass.

macieksmuga:
Thanks @dcolligan for taking this on - I'm confused about the implementation of arrays of obfuscator classes. This looks overly complicated to me, especially since there are no actual overrides of `obfuscate` in this new codebase, so they all end up resolving to the same base method.

ga4gh/datamodel/__init__.py
dcolligan: Each class has a distinct `prefix`.  For instance, a dataset has `D`.  In the `obfuscate` method, this string is concatenated to the id before obfuscating.

dcolligan: We need a list of obfuscators because in the CompoundId's `__init__` method, `obfuscate` is called on every component part of the id to set an attribute in the current compoundId.

diekhans: @dcolligan, what is the rationale for obfuscating the individual components rather
than the id as a whole.

dcolligan: The compoundId object often does contain the ids of other objects within it (for instance, a ReadGroup compoundId contains the ReadGroupSetId of which it belongs to).  In order for the, for instance, readGroupCompoundId.readGroupSetId to match the original readGroupSet id, they must be obfuscated in the same way.

diekhans: Sorry, I don't understand the functional of obfuscating each
component using a different method vs just obfuscating the
entire id.

dcolligan: We are not obfuscating each component.  We are storing the obfuscated ids of other objects in the compositeId -- those other objects are, specifically, the parent objects of the object to which the compositeId corresponds.

diekhans: Maybe I am misusing the term component or confused.  What I am
requesting this the rationale for storing "obfuscated ids of
other objects in the compositeId".  This is confusing, given the
goal of obfuscation is to keep the full id string from being
over-interpreted by client.

dcolligan: We refer to objects by their ids.  All those ids within our system are obfuscated.  If we want to lookup any item by an id, we need the correct (obfuscated) id.

diekhans: I am still confused; Let me ask a different why: isn't there one
obfuscator?

dcolligan: Each type has its own obfuscator.  Obfuscating the same string with two different obfuscators will return two different strings by very virtue of there needing to be globally unique IDs.  CompoundIds store IDs of parent objects as attributes on themselves.  In order to do this, they need access to the correct type obfuscators.

diekhans: The rationale that I don't understand is why each type has an
obfuscator when the goal is to make the string representing the
entire compound id obscure.

dcolligan: The only thing the type-specific obfuscator really does is point to the correct prefix to obfuscate with for that type.

jeromekelleher:
I tend to agree with @diekhans and @macieksmuga here. Is there any way that we can process the prefix/suffix  _before_ deobfuscation? That is, in the ``parse`` method we have something like
```
if not compoundIdStr.startswith(self.prefix):
    raise exceptions.BadIdentifierException(#etc)
idStr = compoundIdStr[len(self.prefix):]
# Use idStr for compoundIdStr in the rest
```
and in ```__str__``` we just dump ``self.prefix`` on to the start of the obfuscated IDs. Does this work?

dcolligan:
Well it's all a question of what properties we want to enforce.  I thought the going assumption was "we want the ids to be completely opaque to the client."  Therefore, having ids like VS_base64gobbledygook for a variant set would violate that principle.  I therefore implemented obfuscation as the very last step.

If we do what you are proposing, clients could build string parsing that relies on this id convention, which will then break when we change our id scheme (or even the prefixes of some compoundId classes!).

diekhans:
First, can we have a problem statement? Given that the id is logically
equivalent to a Unix file path, why can ids collide?  Are container ids
logically equivalent to directories?

jeromekelleher:
This is true @dcolligan. Opinions from others?

I think if we put a non-obvious suffix (say, w, x, y, z) on to the end of things that can possibly collide like variantSets and readGroupSets then the client would be pretty eagle-eyed to spot this pattern and would get what they deserve when if they build code depending on this.

Problem statement is in the linked issue @diekhans.

diekhans:
re problem statement: so the readgroupset id is not
a parent id of the readgroup?  The variantset id is
not a parent id of the variant?

so it's not really analogous to the file path, its
more analogous to someone putting everything in their
home directory???

jeromekelleher:
The readGroupSetId is the parent ID of a readGroup. The problem occurs because datasets logically contain a bunch of different objects, and it is logically a flat name space. If I have a readGroupSet with name = X and a VariantSet with name = X (which is perfectly possible) within dataset with name = D, then the ids of both will be "D:X". 

I guess we could enforce name uniqueness of containers with a dataset, and this might make the problem go away.

diekhans:
My understanding is the user only sees the CompoundId, correct?

If so, I think the whole premises of the bug is incorrect. Given
the definition of CompoundId in the comment, correct use of
CompoundId means it is not possible to generate a duplicate id,
as it will always include it's containing context.

Just like you can have to files of the same name in the same
directory under POSIX.

dcolligan:
@jeromekelleher I don't want to implement any implicit api within our ids, no matter how "eagle-eyed" the user would have to be to (mis)use it.  I think we should stick with base64'ing any craziness that we come up with.

@diekhans  CompoundId is a base class.  The id that the user sees (after this change) is essentially `base64(CompoundIdSubClass.prefix + str(CompoundIdSubClassInstance))`.

If two files in the same directory on a POSIX system can have the same name, that's news to me.  I also don't understand what "ids outside of the CompoundId object" means.  If an object exists in our data model, it has a CompoundId associated with it.

I'm not sure why this change is generating such controversy when our requirements are:
- every id of an object must be unique (per the schema; we do not satisfy that condition right now... that's what this PR is all about)
- we want ids to be opaque
- ids are handled in a modular way

This change satisfies all those requirements.

Mark's comment in the linked issue was:
"From a user perspective, it's very confusing to not have ids
globally unique.  For debugging and if one does anything like a
cache.  Please figure out a short-term fix.  Once we get protocol
buffers integrated, it would be good to revisit the id scheme."

Again, this change satisfies this requirement.

I haven't heard an alternative proposal that satisfies the above requirements.  If we want a simpler id scheme, we either need to relax the requirements (which I don't think we want to do) or redesign our id handling.  Since this issue was conceived as a short term fix, that sounds like a task for another day.

diekhans:
This is generated controversy because it seems over-complex and
developers have indicated the id systems is already confusing
compared to what would would have in a pure database system.

The point was POSIX files can't have the same name and I don't
see how two CompoundIds can have the same name either.  It looks
like use case in the bug report is incorrect.

ids outside == localids

Please explain how, given the usage defined in class CompoundId,
how there is a bug.

dcolligan:
Well, the fact is that two objects in our system, even of different types, can have the same ID right now.  That was why the original issue was filed and what this PR is intending to fix.  The use case in the bug report is not incorrect.  The issue was filed because the schema says that the current ID scheme is invalid.

There are no restrictions on what a localId can be.  So that requirement is already satisfied.

jeromekelleher:
@diekhans, the issue is this (from above):

>  The problem occurs because datasets logically contain a bunch of different objects, and it is logically a flat name space. If I have a readGroupSet with name = X and a VariantSet with name = X (which is perfectly possible) within dataset with name = D, then the ids of both will be "D:X".

> I guess we could enforce name uniqueness of containers with a dataset, and this might make the problem go away.

So, can we just enforce the requirement that names within a dataset must be unique, and walk away from this one? The spec is silent on this issue, so we can do what we like. All we're saying is that you can't have a variant set and a readgroups set with the same name. @dcolligan, what do you think?

dcolligan:
@jeromekelleher I would be against that.  We should solve this issue in the most general way possible, without tying our ID system to our data model.  They should be cross-cutting concerns.  We don't want our IDs to break if, say, we for some reason decide that, I dunno, readGroupSets should be in the same directory with datasets.

jeromekelleher:
Well, our ID system is already entirely defined by the data model and is what allows us to service GET requests, so I don't think that adding a minor requirement on top is going to make any difference. All we're saying is that two root level entities within a dataset can't have the same name (which, incidentally, allows us to put them in the same directory not prohibits it). This is not an unreasonable definition of the semantics of a name.

We're using a lot of time on an esoteric problem that will probably never occur in practise here. Is there any chance we can take the easy solution and move on to other stuff?

diekhans:
According to the documentation on compound id, this is misusing
is it:

class CompoundId(object):
    """
    Base class for an id composed of several different parts, separated
    by a separator. Each compound ID consists of a set of fields, each
    of which corresponds to a local ID in the data hierarchy. For example,
    we might have fields like ["dataset", "variantSet"] for a variantSet.
    These are available as cid.dataset, and cid.variantSet.  The actual IDs
    of the containing objects can be obtained using the corresponding
    like cid.datasetId and cid.variantSetId.
    """

My reading of this indicates the variant local id need only be unique in the
variantSet.

This makes a lot of sense.  A flat dataset space does not make sense.
It's possible to have multiple variantSets within a dataset.  dataset
is a generic container that could have one of 100,000 other types
of sets within it.

dcolligan:
@jeromekelleher I have to disagree with you there.  Our id system does assume a *hierarchical* data model, but there is nothing about datasets, variantSets, etc. in the CompoundId base class or logic.  The only times those concepts come in is in the metadata driving the system -- i.e. the CompoundId subclasses which define attributes.

What exactly is your alternate proposal for the "easy solution" that will guarantee globally unique ids for all objects?

diekhans:
@jeromekelleher just explained the issue.  It's that there is
nothing in the id path representing the object type division.

I think this can be solved just by adding something to
path that reflects the file hierarchy

  <datasetid>:vs:<varientsetid>
  <datasetid>:rs:<readsetid>

No need for addition obfuscatory, since this is not an obfuscatory
issue.

Correct?  However, Jerome is also correct in that this is not
high-priority, as it doesn't currently happen.

dcolligan:
Closing since #1060 seems to have more buy-in

PR #1023, opened 2016-03-25
Compliance downloader: properly check for tempdir

david4096:
Even though it would run successfully, when importing existing data an error would be thrown trying to find the tempdir.

From github:
`python scripts/prepare_compliance_data.py -o ga4gh-example-data`

From compliance test data:
`python scripts/prepare_compliance_data.py -i ~/compliance/test-data -o ga4gh-example-data`


macieksmuga:
+1 Makes sense, and works in testing. Good catch, David.

PR #1028, opened 2016-03-28
Biodata protocol

david4096:
Replaced by https://github.com/ga4gh/server/pull/1235


ga4gh/client.py
```diff
@@ -98,6 +98,24 @@ def _runGetRequest(self, objectName, protocolResponseClass, id_):
         """
         raise NotImplemented()
 
+    def getBioSample(self, bioSampleId):
+        """
+        Perform a get request for the given BioSample.
+        :param bioSampleId:
+        :return: protocol.BioSample
```
jeromekelleher: I think the sphinxdoc should be 

```
:return: The requested BioSample.
:rtype: :class: `ga4gh.protocol.BioSample`
```

ga4gh/client.py
```diff
@@ -98,6 +98,24 @@ def _runGetRequest(self, objectName, protocolResponseClass, id_):
...
+        """
+        return self._runGetRequest(
+            "biosamples", protocol.BioSample, bioSampleId)
+
+    def getIndividual(self, individualId):
+        """
+        Perform a get request for the given Individual.
+        :param individualId:
+        :return: protocol.Individual
```
jeromekelleher: Likewise here. This is user-facing documentation, so we should keep it tidy and consistent.

ga4gh/client.py
```diff
@@ -371,30 +389,71 @@ def searchReferences(
...
-    def searchCallSets(self, variantSetId, name=None):
+    def searchCallSets(self, variantSetId, name=None, bioSampleId=None):
         """
         Returns an iterator over the CallSets fulfilling the specified
         conditions from the specified VariantSet.
 
         :param str name: Only CallSets matching the specified name will
             be returned.
+        :param str bioSampleId: Returns only callsets with the given
+            Id.
```
jeromekelleher: s/Id/bioSampleId

ga4gh/client.py
```diff
@@ -371,30 +389,71 @@ def searchReferences(
...
         return self._runSearchRequest(
             request, "callsets", protocol.SearchCallSetsResponse)
 
-    def searchReadGroupSets(self, datasetId, name=None):
+    def searchBioSamples(self, datasetId, name=None, individualId=None):
+        """
+        Returns an iterator over the BioSamples optionally matching
+        a name or individualId for the given datasetId.
+
+        :param str name: Only BioSamples given this name are returned
```
jeromekelleher: ``Only BioSamples given this name`` reads a bit weirdly. "Only BioSamples with this name" perhaps?

ga4gh/client.py
```diff
@@ -371,30 +389,71 @@ def searchReferences(
...
+        request.pageSize = self._pageSize
+        return self._runSearchRequest(
+            request, "biosamples", protocol.SearchBioSamplesResponse)
+
+    def searchIndividuals(self, datasetId, name=None):
+        """
+        Returns an iterator over the Individuals optionally matching
+        a name for the given datasetId.
+
+        :param str name: Only Individuals given this name are returned
```
jeromekelleher: Likewise with the 'given' here.

ga4gh/client.py
```diff
@@ -371,30 +389,71 @@ def searchReferences(
...
+            request, "individuals", protocol.SearchIndividualsResponse)
+
+    def searchReadGroupSets(self, datasetId, name=None, bioSampleId=None):
         """
         Returns an iterator over the ReadGroupSets fulfilling the specified
         conditions from the specified Dataset.
 
         :param str name: Only ReadGroupSets matching the specified name
             will be returned.
+        :param str bioSampleId: Filters readgroups according to the ID
```
jeromekelleher: Follow the pattern here. "Only ReadGroupSets matching the specified BioSampleId will be returned."

ga4gh/client.py
```diff
@@ -536,7 +596,9 @@ def __init__(self, backend):
             "variants": self._backend.runGetVariant,
             "readgroupsets": self._backend.runGetReadGroupSet,
             "readgroups": self._backend.runGetReadGroup,
-            "variantannotationsets": self._backend.runGetVariantAnnotationSet
+            "variantannotationsets": self._backend.runGetVariantAnnotationSet,
```
jeromekelleher: variantannotationsets should be in here already, right?

ga4gh/datamodel/__init__.py
```diff
@@ -488,3 +505,25 @@ def _getDataFilenames(self, dataDir, patterns):
...
+    def loadSidecar(self, filepath):
+        jsonFilename = os.path.splitext(filepath)[0] + ".json"
+        try:
+            with open(jsonFilename) as data:
+                self._sidecar = json.load(data)
+        except (ValueError, IOError):
+            raise exceptions.FileOpenFailedException(jsonFilename)
+            self._sidecar = {}
+
+    def sidecar(self, key):
```
jeromekelleher: It's not clear what this method does, and I'm not finding the ``sidecar`` term very useful here. I would have thought ``_metadata`` would have been a more descriptive --- metadata could come from anywhere in the future, the fact we're using json files at the moment is an implementation detail.

Do we use this ``sidecar`` method anywhere? If so, I would suggest replacing with ``get_metadata(key)``, which raises a KeyError if the key is not found.

ga4gh/datamodel/biodata.py
```diff
@@ -0,0 +1,226 @@
...
+    Allows a BioSample object to be populated using JSON
+    metadata information by passing a filepath during
+    instantiation.
+    b = JsonBioSample(dataset, name, filepath)
+    """
+    def __init__(self, parentContainer, localId, filePath):
+        super(JsonBioSample, self).__init__(parentContainer, localId)
+        self._filePath = filePath
+        self.loadSidecar(filePath)
+        self.validate()
```
jeromekelleher: I think the mixin is overcomplicating things a bit. How about this
```python
metadata = readJsonMetadata(filePath)
self._description = metadata.get("description", self._description)
# And so on for the rest of the attributes.
```

Then, this class is nice and simple and doesn't need all the overrides.

ga4gh/datamodel/biodata.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+    Allows a Individual object to be populated using JSON
+    metadata information by passing a filepath during
+    instantiation.
+    b = JsonIndividual(dataset, name, filepath)
+    """
+    def __init__(self, parentContainer, localId, filePath):
+        super(JsonIndividual, self).__init__(parentContainer, localId)
+        self._filePath = filePath
+        self.loadSidecar(filePath)
```
jeromekelleher: Ditto for the metadata loading pattern from above. I think it will simplify this  quite a bit.

scripts/1kgenomes_biodata.py
```diff
@@ -0,0 +1,82 @@
...
+                        "sourceName": "http://purl.obolibrary.org/obo",
+                        "sourceVersion": "2016-02-02"
+                    },
+                    "sex": sex,
+                    "createDateTime": datetime.datetime.now().isoformat(),
+                    "updateDateTime": datetime.datetime.now().isoformat(),
+                    "info": info
+                }
+
+                json.dump(biosample, bioSampleOut)
```
jeromekelleher: OK, so we're not actually using ``bioSampleOut`` until here, yeah? Why not
```
filename =  "biodata/biosamples/" + row['Sample'] + ".json"
with open(filename, 'w') as bioSampleOut:
    json.dump(biosample, bioSampleOut)
# ditto for individualOut
```
This means we lose a level of indentation and makes the code easier to break up later if/when it gets more complex.

tests/unit/test_biodata.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class TestIndividuals(unittest.TestCase):
+    """
+    Tests the JSON Individuals class
+    """
+    dataset = datasets.AbstractDataset('dataset1')
+
+    def testToProtocolElement(self):
+        for localId in os.listdir(paths.testIndividualsDataDir):
```
jeromekelleher: This seems more like a data driven test. Would it make sense to change them to fit into that framework?

ga4gh/backend.py
```diff
@@ -391,26 +392,150 @@ def _topLevelObjectGenerator(self, request, numObjects, getByIndexMethod):
...
+            nextPageToken = None
+            if currentIndex < numObjects and recur:
+                request.pageToken = str(currentIndex)
+                for rgs, pageToken in self._readGroupSetObjectGenerator(
+                        request, numObjects, getByIndexMethod, False):
+                    nextPageToken = str(currentIndex)
+                    break
+            if (request.name == "" or request.name) and request.bioSampleId:
+                if request.name == readGroupSet.name:
+                    yield self._filterReadGroupsByBioSampleId(
```
jeromekelleher: This is quite cunning, but I'm suspicious there'll be some nasty gotchas. Can you talk me through this a bit please? How heavily is this paging code tested? We would need to have lots of tests with paging with the various filters turned on and off.

jeromekelleher:
Implementation generally looks great. Comments above.

ga4gh/backend.py
```diff
@@ -391,26 +392,150 @@ def _topLevelObjectGenerator(self, request, numObjects, getByIndexMethod):
...
+                        request.bioSampleId, readGroupSet), nextPageToken
+            elif (request.name == "" or request.name):
```
diekhans: It's confusing what this is checking for.  Using a string as a boolean is a good thing to avoid  to keep code readable. Is this really a check for `is not None' ??

tests/unit/test_biodata.py
david4096: I started down that path, but decided against it when I observed `__init__` for datadriven since it intentionally skips `.json` elsewhere. https://github.com/ga4gh/server/blob/master/tests/datadriven/__init__.py#L45

ga4gh/backend.py
jeromekelleher: Also: I think using 'Falsy' semantics this is equivalent to ``if request.name``, since the empty string evaluates to False in a boolean context. I'm all for being explicit about checking for None though.

ga4gh/backend.py
david4096: Previously read groups and call sets were generating single objects when making a request by name and sent a null pageToken. This was based on the assumption that names were singular. This more general approach uses a look-ahead to determine if the request would generate another object, and if not send a null pageToken.

When a request is made the readGroupSet as selected by the pageToken is being examined. We iterate the currentIndex and perform the same search again having iterated our pageToken. If there are results in the lookahead the nextPageToken is set to the given index.

The schema specifies that `SearchReadGroupSetsResponse` will include only readgroupsets with readgroups that match the bioSampleId, and it will only return readgroups that match the bioSampleId. This means we need to remove non-matching readgroups from a readgroupset if they don't match our request. This is what `filterReadGroupsByBioSampleId` should do.

First, we look at the request to see if it has our query fields present. If both are present we return matches on name, filtered by bioSampleId. If only a name is present we match on name. In the case that only a biosampleId is provided we check to see if the filter leaves behind any readgroups in the readgroupset, and if so returned that filtered object. In the final case, when neither of our query fields are present, simply yield the readgroupset.

ga4gh/datamodel/__init__.py
david4096: The problem is overloading the term metadata, in this case the actual data is represented by JSON so both metadata and sidecar are a bit obtuse. The goal here is to supply any datamodel object with a way to access key/values stored in a JSON. I'll refactor to follow the pattern you've requested.

ga4gh/backend.py
jeromekelleher: OK, sounds good. I like this approach, once we've tested it pretty thoroughly with different page sizes and combinations of the filters on the simulated backend.

docs/configuration.rst
```diff
@@ -325,6 +349,50 @@ Removes a variant set from a given data repository and dataset.
...
     $ ga4gh_repo remove-variantset path/to/datarepo aDataset aVariantSet
 
++++++++++++++++
+add-biosample
++++++++++++++++
+
+Add a BioSample record to the given data repository and dataset. Records
+should be described in JSON format according to the GA4GH protocol. To 
+associate this BioSample with an Individual record add the optional 
+individualName flag.
```
mbaudis: Shouldn't this be ```individualId``` ?

david4096: This is a convenience made possible by the fact that individuals are named uniquely within a dataset. This could be extended to optionally provide either.

mbaudis: But:
* Where is the uniqueness required? (Sorry; I just may have lost track.)
* ID is used for references, never name.
* Even though I understand that "name" in the schema may not mean "Walter White", it hurts a bit to have "name" of "individual"  being a unique reference point.

david4096: This uniqueness is required by the server's architecture, and is not meant to reflect the schemas. This name is also the filename of the record with the extension removed.

diekhans: The uniqueness restrictions is fine for now.  The API doesn't
require the name to be unique, however it is reasonable for
implementation to place reasonable data restrictions.

There was objection to allowing name to be non-unique, although
this was never resolved.

david4096:
For those that haven't seen, there is active discussion on the schemas side that will end in code changes here. https://github.com/ga4gh/schemas/pull/606

Issue #1018, opened 2016-03-24
Using ':' as a separator in CompoundIds is problematic. Replace.


jeromekelleher:
See discussion at https://github.com/ga4gh/server/pull/999#discussion_r57175788

It should be possible to replace easily enough. Suggestions?

diekhans:
Don't try to choose a safe delimter, insead encode each
component so that the chosen delimiter can't collide: For
instance:

   `:' -> `\:'
   `\\' -> `\'

since it's all done in one place, this should be transparent.

macieksmuga:
I agree with Mark, there is no safe delimiter - I'd much rather use a uniform, generic mechanism (bundled with the Compound ID object) for escaping the delimiter.

macieksmuga:
FYI, since the latest refactor of the Sequence Annotation PR #999 - I no longer have a problem with special characters in IDs (as all internal IDs for features have now been replaced with unique long integers). So while I still agree with Mark's general suggestion, this issue is no longer a blocker on any of my current code.

dcolligan:
Fixed in #1042

PR #1029, opened 2016-03-29
Make id separators not collide with other id chars

dcolligan:
Issue #1018

### DIFF
--- a/ga4gh/datamodel/__init__.py
+++ b/ga4gh/datamodel/__init__.py
@@ -135,7 +135,8 @@ class CompoundId(object):
                 setattr(self, field, getattr(parentCompoundId, field))
                 index += 1
         for field, localId in zip(self.fields[index:], localIds):
-            setattr(self, field, str(localId))
+            encodedLocalId = self.encode(str(localId))
+            setattr(self, field, encodedLocalId)
         if len(localIds) != len(self.fields) - index:
             raise ValueError(
                 "Incorrect number of fields provided to instantiate ID")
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Hmm, this is getting frickin complicated! Is there any way we can avoid going down this encoding rabbit hole using a standard language parser? Perhaps we could use the JSON parser to do the decoding for us? e.g. something like,

```python
idStr = '["{}","{}"]'.format(idPart1, idPart2)
# Later
idPart1, idPart2 = json.parse(idStr)
```

Thoughts?

dcolligan:
Well, the whole point of this change is to *absolutely guarantee* that there can never be ambiguity between our id segment scheme and wacky ids (that come from values read out from filenames, bioinformatics file headers, etc.).  Unfortunately, I think I can come up with a wacky id to defeat the scheme that you just proposed.

jeromekelleher:
I don't think it's that easy to break @dcolligan --- once double quotes are escaped, then it should be bullet proof, right? I'm sure your method works nicely, I'm just conscious that this whole CompoundId thing is already quite complex and it would be nice to cut the lines-of-code count down a bit.

dcolligan:
So if I had a filename like:
```
",".txt
```
and the relevant part got read in like:
```
","
```
that would make the formatted id string look like (if Y is another id):
```
'["","","Y"]'
```
so parsing it would give 3 localIds, not 2, which would be an error.

jeromekelleher:
Quotes are escaped pre-jsonification though (``str.replace('"', '\"')``), so this would be 
```
'["\",\"","Y"]'
```

dcolligan:
So we still need an encode and decode step... your objection is just to my `split` method basically?

diekhans:
My confusion was calling it obfuscation when it was making it
unique.

It doesn't seem unnecessary to do this as a function/class in a
list list rather than having the type a parameter when one adds
a new local id.

Having type part of the the naming hierarchy vs each element in
the name hierarchy then being two parts is a bit of a trade off.
What ever works best.

It really doesn't which as long as it's transparent to the code.
The naming spliting/joining is really just a serialization
function and only happen at id parse/format time.

jeromekelleher:
@dcolligan, yeah, basically I worry about writing our own parsing code in internet-facing locations. To be confident that something like this wouldn't present a security hole sometime in the future I'd need to look at it for a long time and write a lot of test cases. So, I would generally look to use code that is hardened against this sort of thing, and a JSON parser seems like a good example.

macieksmuga:
Agree with @jeromekelleher - Also note that in the sequence annotations API coming in, I've replaced formerly weird IDs with simple ints, removing the need for further munging there.

dcolligan:
Closing since #1042 has more buy-in

PR #1037, opened 2016-03-30
Merge latest master (e787eab) into sequence annotations

macieksmuga:


david4096:
Looks OK, kicked the tires and ran prepare compliance data with and without the download flag. +1

Issue #1035, opened 2016-03-30
Thousand genomes variant annotations special case


david4096:
The split ANN field has become more or less a standard for representing annotated variants in VCF, however, the Thousand Genomes annotated variant sets have a `CSQ` field that is arranged differently. If we would like to host the thousand genomes annotated variants we either have to detect this case or munge them into the expected form.

From ALL.chr1.phase3_shapeit2_mvncall_integrated_v5.20130502.sites.annotation.vcf
```
1       13382   .       C       G       100     PASS    AC=1;AF=0.000199681;AN=5008;NS=2504;DP=28817;EAS_AF=0;AMR_AF=0;AFR_AF=0;EUR_AF=0;SAS_AF=0.001;AA=c|||;CSQ=G|ENSG00000223972|ENST00000456328|Transcript|non_coding_transcript_exon_variant&non_coding_transcript_variant|630|||||||1||||||;ERB=G||proximal_2|Regulatory_Feature|proximal_enhancer,G||ctcf_2|Regulatory_Feature|ctcf_insulator;FUNSEQ=0.0000
```

jeromekelleher:
I would vote for  the format. Munging 1KG data is not something to be taken lightly, and probably other people have adopted the convention anyway.

david4096:
This feature is ready for review/merge.

PR #1039, opened 2016-03-31
Thousand genomes annotation support

david4096:
Thousand genomes doesn't use the split ANN field and so is a special case to be handled.

Will close https://github.com/ga4gh/server/issues/1035


jeromekelleher:
Data-driven-test-driven development: I like it @david4096!

david4096:
This is looking complete and adds some 1kgenome annotated VCF into our test data directory.

Also fixes a bug in the VA dateTime key.

ga4gh/datamodel/variants.py
```diff
@@ -787,9 +787,20 @@ def buildMetadata(
...
+
+    def hasAnnField(self, path):
+        pysamreader = self.openFile(
+            glob.glob(os.path.join(path, "*.vcf.gz"))[0])
+        items = [x[0] for x in pysamreader.header.info.items()]
+        return 'ANN' in items
+
+    def hasConsequenceField(self, path):
         pysamreader = self.openFile(
             glob.glob(os.path.join(path, "*.vcf.gz"))[0])
```
diekhans: The code for obtaining an open VCF is repeated multiple times.  Put in function rather than copy and paste.

Is the name of the file already know by this point?  It seems odd to glob for it multiple times.  Also, this if the file is missing this will result in a very confusing IndexError rather than report the file missing.

ga4gh/datamodel/variants.py
```diff
@@ -903,7 +917,7 @@ def convertLocationHgvsP(self, hgvsp):
 
     def addCDSLocation(self, effect, cdnaPos):
         hgvsC = effect.hgvsAnnotation.transcript
-        if hgvsC != '':
+        if hgvsC != '' and hgvsC is not None:
```
diekhans: This idiom of testing for an empty string  and None is repeated multiple places.  It would be good to have a function to do this.

ga4gh/datamodel/variants.py
```diff
@@ -943,6 +957,41 @@ def addLocations(self, effect, protPos, cdnaPos):
...
+            effect.featureId = featureId
+            effect.hgvsAnnotation = protocol.HGVSAnnotation()
+            # These are not present in the data
+            effect.hgvsAnnotation.genomic = None
+            effect.hgvsAnnotation.transcript = None
+            effect.hgvsAnnotation.protein = None
+            self.addLocations(effect, protPos, cdnaPos)
+            effect.id = self.getTranscriptEffectId(effect)
+            effect.analysisResults = []
+            transcriptEffects.append(effect)
```
diekhans: this is a large loop, suggest making a function

ga4gh/datamodel/variants.py
david4096: Python provides the idiom `if hgvsC` which checks both cases, but I changed it elsewhere at your request because it wasn't verbose enough.

ga4gh/datamodel/variants.py
david4096: I moved the repeated code into a separate function. When adding a variant set you are adding a directory, and there are no guarantees for the name. https://github.com/ga4gh/server/blob/master/ga4gh/datamodel/datasets.py#L219

ga4gh/datamodel/variants.py
diekhans: Personally, I have also found that treating non-boolean types as
implicit booleans is unclear.

A function that obviously documents intention over
representation is clearer.  Assuming this is testing for
something being unspecified a function isUnspecified(value)
communicates more than either `if hgvsC ' or
`if hgvsC == "" or hgvsC is None' and is less verbose than the
latter.

david4096:
Added the `isUnspecified` function https://github.com/ga4gh/server/pull/1039/files#diff-f9ce58fdc26b5bb89ea0a4aa47a608faR23 .

ga4gh/datamodel/variants.py
```diff
@@ -20,6 +20,10 @@
 import ga4gh.datamodel as datamodel
 
 
+def isUnspecified(str):
```
jeromekelleher: Docstring please.

ga4gh/datamodel/variants.py
```diff
@@ -787,9 +791,19 @@ def buildMetadata(
...
+    def hasAnnField(self, path):
+        return 'ANN' in self._getHeaderItems(path)
+
+    def hasConsequenceField(self, path):
+        return 'CSQ' in self._getHeaderItems(path)
+
+    def _getHeaderItems(self, path):
         pysamreader = self.openFile(
             glob.glob(os.path.join(path, "*.vcf.gz"))[0])
-        return 'ANN' in [x[0] for x in pysamreader.header.info.items()]
```
jeromekelleher: We shouldn't need to parse the header each time to answer these (which we only use in the constructor anyway?). VCF headers and big and complicated so this will be slow. We should try to read the header information once and store the information sufficient to answer the questions we need to ask during the lifetime of the object as instance variables.

jeromekelleher:
Looks good modulo minor comments above.

david4096:
OK added a docstring and optimized the test!

jeromekelleher:
+1. Thanks, can you update and merge please?

PR #1042, opened 2016-03-31
Refactor CompoundIds to eliminate separators

dcolligan:
Use json/arrays instead

Issue #1018

Alternate to fix in #1029 (see discussion there)

### DIFF
--- a/ga4gh/datamodel/__init__.py
+++ b/ga4gh/datamodel/__init__.py
@@ -6,6 +6,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import json
 import base64
 import collections
 import glob
@@ -99,15 +100,14 @@ fileHandleCache = PysamFileHandleCache()
 
 class CompoundId(object):
     """
-    Base class for an id composed of several different parts, separated
-    by a separator. Each compound ID consists of a set of fields, each
-    of which corresponds to a local ID in the data hierarchy. For example,
-    we might have fields like ["dataset", "variantSet"] for a variantSet.
-    These are available as cid.dataset, and cid.variantSet.  The actual IDs
-    of the containing objects can be obtained using the corresponding
-    like cid.datasetId and cid.variantSetId.
+    Base class for an id composed of several different parts.  Each
+    compound ID consists of a set of fields, each of which corresponds to a
+    local ID in the data hierarchy. For example, we might have fields like
+    ["dataset", "variantSet"] for a variantSet.  These are available as
+    cid.dataset, and cid.variantSet.  The actual IDs of the containing
+    objects can be obtained using the corresponding attributes, e.g.
+    cid.datasetId and cid.variantSetId.
     """
-    separator = ':'
     fields = []
     """
     The fields that the compound ID is composed of. These are parsed and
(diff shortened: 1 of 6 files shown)


ga4gh/datamodel/__init__.py
```diff
@@ -135,22 +136,58 @@ def __init__(self, parentCompoundId, *localIds):
...
 
     @classmethod
+    def join(cls, splits):
+        """
+        Join an array of ids into a compound id string
+        """
+        segments = []
+        for split in splits:
+            segments.append('"{}", '.format(split))
+        if len(segments) != 0:
```
jeromekelleher: I think it's an internal error condition to pass in an empty splits here. Just ``asset(len(segments) > 0)``.

ga4gh/datamodel/__init__.py
```diff
@@ -135,22 +136,58 @@ def __init__(self, parentCompoundId, *localIds):
...
         return self.obfuscate(compoundIdStr)
 
     @classmethod
+    def join(cls, splits):
+        """
+        Join an array of ids into a compound id string
+        """
+        segments = []
+        for split in splits:
+            segments.append('"{}", '.format(split))
```
jeromekelleher: I think drop the trailing space here, it's just inflating the size of IDs, which already have a lot of padding.

jeromekelleher:
I like this approach, seems fairly bullet proof. However, the immediate motivation (wacky name values containing ":" characters has gone away), so are we bringing trouble on ourselves for not a lot of gain? Opinions?

dcolligan:
@jeromekelleher updated

The immediate motivation has gone away, but there's no guarantee that it won't reappear in the future, when someone has wacky filenames/header values, as they inevitably will.

jeromekelleher:
OK, I'm convinced @dcolligan. +1.

We need wider review on something like this though. @macieksmuga, @david4096 can you weigh in here please?

david4096:
> However, the immediate motivation (wacky name values containing ":" characters has gone away), so are we bringing trouble on ourselves for not a lot of gain? Opinions?

@jeromekelleher Thanks for asking! Yes, I believe this is going further down an anti-pattern rabbit hole. Compound IDs are a premature optimization that have hindered the ability to bear out the schemas and iterate quickly. The localId/name confusion, IDs with unspecified length, leaking private data, and degeneracy would be solved by applying a uniform ID scheme. The system would be simpler and more robust if this optimization were removed entirely. For example, look at the logic behind `getVariant`.

dcolligan:
Well it looks to me like we have a few options:

1) Don't address this and have the server error when someone has ':'s in their filenames/headers that are used as part of an id
2) Accept this change -- or another like it -- and fix the issue for the time being
3) Redesign our id system

I'm sure we all have ideas of how the id system could be better, but redesigning it will be a long and involved process.

diekhans:
This seems reasonable and we need to move forward now.

Redesigning the id systems needs to happen in the
context of objectives for the next year or so

jeromekelleher:
It would be good to hear alternatives to the current ID scheme @david4096 --- I agree that this has become complicated and isn't ideal. A reverse lookup table for IDs would be a lot easier to work with, and is something we should try to do when we move to the integrated DB backend. However, we still won't be able to do full DB index of IDs for all (e.g.) variants, so we'll still need some sort of way to encode the container ID within our string. 

I also agree that the LocalId/Name confusion isn't ideal, but this wasn't planned. The schemas don't have a lot to say on a lot of things, and we've had to come up with our own way of doing things. LocalId was brought in to the server before the concept of ``name`` was formalised in the schema (to whatever extent it is formalised). The server has evolved over time in parallell with the schemas, and there are bit left over from older designs that are just plain weird now.

As a full time developer @david4096 your views are very much welcomed on any and all aspects of the design of the server. Please just weigh in on issues as they arise, and if you have ideas for how to improve the design of the server in any way please start a discussion on it. We may not want to embark on hefty refactors immediately (mainly because of the major topic branches that are open, particularly and especially protobuf), but we can discuss and plan refactors for a time when we'll be able to do them with the least disruption to the overall development process.

The design of the server isn't set in stone, and it's certainly not optimal; please help to make it better by starting and participating in design discussions, either here, over email or in chats.

diekhans:
We will be in a good place to evolve id once this milestone is done.

One issue will be how to map ids to UCSC genome browser tables
that do not have primary keys.   The best unique key for
most  of these tables is (chr, start, end, name).

dcolligan:
@david4096 are your concerns addressed?  If so, can I get a +1 and merge on this?

macieksmuga:
+1 here too. This is conceptually cleaner, and the code reads much better to me as a result.

david4096:
@dcolligan yes this works, thanks for taking some time to explain.

jeromekelleher:
OK, looks like consensus. Can you update and merge please @dcolligan?

PR #1043, opened 2016-03-31
Initial Rna api implementation for topic branch

saupchurch:
Initial PR for review and comment


ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,390 @@
...
+        if expressionId is not None:
+            rawId = datamodel.ExpressionLevelCompoundId.deobfuscate(
+                expressionId)
+            expressionId = string.split(rawId, ":")[-1]
+        with self._db as dataSource:
+            expressionsReturned = dataSource.searchExpressionLevelsInDb(
+                rnaQuantID, pageToken=pageToken,
+                pageSize=pageSize, expressionId=expressionId,
+                featureGroupId=featureGroupId, threshold=threshold)
+        return [ExpressionLevel(self, expressionEntry) for
```
jeromekelleher: Do we need to return a list here? Would a generator help, or is the number of expression levels small enough to make this pointless (less than 10, say)? Generally, we want to avoid creating lists if we can, so we could do 
```python
 for expressionEntry in expressionsReturned:
    yield ExpressionLevel(self, expressionEntry)
```
Assuming that searchExpressionLevelsInDb returns an iterator also, this avoids creating lists.

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,390 @@
...
+        rnaQuantId = self.getLocalId()
+        with self._db as dataSource:
+            rnaQuantReturned = dataSource.getRnaQuantificationById(
+                rnaQuantId)
+        if rnaQuantReturned is not None:
+            self.addRnaQuantMetadata(rnaQuantReturned)
+        else:
+            raise exceptions.RnaQuantificationNotFoundException(rnaQuantId)
+
+    def getExpressionLevels(self, rnaQuantID, pageToken=0, pageSize=None,
```
jeromekelleher: Probably better to use hanging indent in situations like this where future code changes will break the alignment, eg.
```python
def getExpressionLevels(
    self, rnaQuantID, pageToken=0, pageSize=None,  expressionId=None, 
    featureGroupId=None,threshold=0.0):
```

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,390 @@
...
+                expressionEntry in expressionsReturned]
+
+    def getExpressionLevel(self, compoundId):
+        rawId = datamodel.ExpressionLevelCompoundId.deobfuscate(compoundId)
+        localExpressionId = string.split(rawId, ":")[-1]
+        with self._db as dataSource:
+            expressionReturned = dataSource.getExpressionLevelById(
+                localExpressionId)
+
+        if expressionReturned is not None:
```
jeromekelleher: I find it helpful when reading code to positively identify error conditions and make it clear that they are not the nominal case (which corresponds to the direct path through the code). Here, I would have written
```python
if expressionReturned is None:
    raise exceptions.ExpressionLevelNotFoundException(compoundId)
return ExpressionLevel(self, expressionReturned)
```
Seeing the ``else`` clause here makes me think, "there are two non-error situations that can occur here", which slows down the process of reading the code.

This is subjective of course, others may argue against this!

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,390 @@
...
+            expressionsReturned = dataSource.searchExpressionLevelsInDb(
+                rnaQuantID, pageToken=pageToken,
+                pageSize=pageSize, expressionId=expressionId,
+                featureGroupId=featureGroupId, threshold=threshold)
+        return [ExpressionLevel(self, expressionEntry) for
+                expressionEntry in expressionsReturned]
+
+    def getExpressionLevel(self, compoundId):
+        rawId = datamodel.ExpressionLevelCompoundId.deobfuscate(compoundId)
+        localExpressionId = string.split(rawId, ":")[-1]
```
jeromekelleher:
```python
parsedId = datamodel.ExpressionLevelCompoundId.parse(compoundId)
localExpressionId = parsedId.expression  # or whatever the name is.
```
We shouldn't be calling deobfuscate and parsing ID strings here, it should be done by the CompoundId machinery.

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,390 @@
...
+            raise exceptions.RnaQuantificationNotFoundException(rnaQuantId)
+
+    def getExpressionLevels(self, rnaQuantID, pageToken=0, pageSize=None,
+                            expressionId=None, featureGroupId=None,
+                            threshold=0.0):
+        """
+        Returns the list of ExpressionLevels in this RNA Quantification.
+        """
+        if expressionId is not None:
+            rawId = datamodel.ExpressionLevelCompoundId.deobfuscate(
```
jeromekelleher: See earlier comment about deobfuscate

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,390 @@
...
+    def getFeatureGroups(self, rnaQuantID, pageToken=0, pageSize=None,
+                         featureGroupId=None):
+        """
+        Returns the list of FeatureGroups in this RNA Quantification.
+
+        for now the feature group data is autogenerated by examining the
+        relevant expression data file
+        """
+        if featureGroupId is not None:
+            rawId = datamodel.FeatureGroupCompoundId.deobfuscate(
```
jeromekelleher: Here also. Basically don't use deobfuscate in this module.

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,390 @@
...
+            featureGroupReturned = dataSource.getFeatureGroupById(
+                localFeatureGroupId)
+
+        if featureGroupReturned is not None:
+            return FeatureGroup(self, featureGroupReturned)
+        else:
+            raise exceptions.FeatureGroupNotFoundException(compoundId)
+
+
+_rnaQuantColumns = [('id', 'TEXT'),
```
jeromekelleher: Pleas use hanging indents for these --- planned snake_case changes will break the alignment.

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,390 @@
...
+        sql = ("SELECT * FROM RNAQUANTIFICATION")
+        sql_args = ()
+        if rnaQuantificationId is not None:
+            sql += " WHERE id = ? "
+            sql_args += (rnaQuantificationId,)
+        sql += sqliteBackend.limitsSql(pageToken, pageSize)
+        query = self._dbconn.execute(sql, sql_args)
+        try:
+            return sqliteBackend.sqliteRows2dicts(query.fetchall())
+        except AttributeError:
```
jeromekelleher: Why return None here, rather than raising an exception directly? This would seem to be the best place to identify where the problem occurs, and it would simplify client code if it didn't have to test for None.

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,390 @@
...
+        :param rnaQuantificationId: the RNA Quantification ID
+        :return: dictionary representing an RnaQuantification object,
+            or None if no match is found.
+        """
+        sql = ("SELECT * FROM RNAQUANTIFICATION WHERE id = ?")
+        query = self._dbconn.execute(sql, (rnaQuantificationId,))
+        try:
+            return sqliteBackend.sqliteRow2Dict(query.fetchone())
+        except AttributeError:
+            return None
```
jeromekelleher: Likewise here: I think it would be better to raise a custom exception here that would clearly identify the problem.

jeromekelleher:
I haven't gone through this fully @saupchurch, but it looks great. Really, very nice code indeed. A few fairly minor general issues popped up:

1. It would be better to use hanging indents in places where code changes (especially changing to camel case) will break the alignment. I've pointed this out in a few places, but it would be great if you could make a pass through the code to look at it.

2. You're slightly misusing compoundIds here. Use the parse method rather than deobfuscating by hand.

3. I think you could streamline things a bit by raising exceptions at the root of problems rather than passing back None as an error indicator --- this is a very C-like way of doing things.

Other than that, awesome.

ga4gh/_protocol_definitions.py
```diff
@@ -746,6 +863,84 @@ def __init__(self, **kwargs):
         """
 
 
+class FeatureGroup(ProtocolElement):
```
macieksmuga: `FeatureGroup` is not a record type in the current Sequence Annotations code. Is this a concept unique to RNA, or is it referring to an older iteration of Sequence Annotations (where `FeatureGroup` was replaced with the `FeatureSet` container type)?

ga4gh/backend.py
```diff
@@ -571,6 +573,78 @@ def callSetsGenerator(self, request):
...
+
+        Currently only supports searching over a specified rnaQuantification
+        """
+        rnaQuantificationId = request.rnaQuantificationId
+        compoundId = datamodel.RnaQuantificationCompoundId.parse(
+            request.rnaQuantificationId)
+        dataset = self.getDataRepository().getDataset(compoundId.datasetId)
+        rnaQuant = dataset.getRnaQuantification(rnaQuantificationId)
+        expressionLevelId = request.expressionLevelId
+        featureGroupId = request.featureGroupId
```
macieksmuga: See my other notes about `FeatureGroup`. I want to make sure you're not relying on entities and mechanisms that aren't there in Sequence Annotations.

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,392 @@
...
+
+        return ExpressionLevel(self, expressionReturned)
+
+    def getFeatureGroups(
+            self, rnaQuantID, pageToken=0, pageSize=None, featureGroupId=None):
+        """
+        Returns the list of FeatureGroups in this RNA Quantification.
+
+        for now the feature group data is autogenerated by examining the
+        relevant expression data file
```
macieksmuga: Ah, I see! Although it's confusing to me that a `FeatureGroup` is effectively an ephemeral entity computed in the RNA code, while `FeatureSet` is a straightforward container entity from Sequence Annotation. Is there a more descriptive name for this?

ga4gh/datamodel/rna_quantification.py
macieksmuga: For future reference, could the note relevant to a line of code be _below_ that line, so that once it's fixed, the note becomes automatically hidden? (Not meaning to pick on you in particular @jeromekelleher, we're all guilty of this)

ga4gh/datamodel/rna_quantification.py
```diff
@@ -0,0 +1,392 @@
...
+               "AND expression >= ? ")
+        sql_args = (rnaQuantId, threshold)
+        if expressionId is not None:
+            sql += "AND id = ? "
+            sql_args += (expressionId,)
+        if featureGroupId is not None:
+            sql += "AND feature_group_id = ? "
+            sql_args += (featureGroupId,)
+        sql += sqliteBackend.limitsSql(pageToken, pageSize)
+        query = self._dbconn.execute(sql, sql_args)
```
macieksmuga: Would it make sense to add an `ORDER BY` clause here?

ga4gh/repo_manager.py
```diff
@@ -79,6 +80,7 @@ class RepoManager(object):
     vcfIndexExtension = '.vcf.gz.tbi'
     bamExtension = '.bam'
     bamIndexExtension = '.bam.bai'
+    dbExtension = '.db'
```
macieksmuga: `.db` files represent both Sequence Annotation and RNA data by this point. We'll want to be careful here.

ga4gh/templates/index.html
```diff
@@ -118,6 +118,17 @@
...
+                <table class="table table-striped">
+                    <th>Name</th>
+                    <th>Id</th>
+                    {% for rnaQuantification in info.getRnaQuantifications(dataset.getId()) %}
+                    <tr>
+                    <td>{{ rnaQuantification.getLocalId() }}</td>
+                    <td>{{ rnaQuantification.getId() }}</td>
+                    </tr>
+                    {% endfor %}
+                </table>
```
macieksmuga: Would it make sense to list `FeatureGroup`s here too?

scripts/encode2ga4gh.py
```diff
@@ -0,0 +1,185 @@
...
+import sys
+import json
+import errno
+import urllib2
+import optparse
+
+import rnaseq2ga
+
+
+class Color(object):
```
macieksmuga: This seems like a nice utility for many other scripts. Perhaps refactor this code to a generally accessible library in the future?

scripts/rnaseq2ga.py
```diff
@@ -0,0 +1,271 @@
...
+                       score real,
+                       units text)''')
+
+    def addRNAQuantification(self, datafields):
+        """
+        Adds an RNAQuantification to the db.  Datafields is a tuple in the
+        order:
+        id, annotation_ids, description, name, read_group_id
+        """
+        self._cursor.execute('INSERT INTO RNAQUANTIFICATION VALUES '
```
macieksmuga: For future reference, I'm thinking of moving my sequence annotation import script to using `executemany` to deal with performance issues.

tests/datadriven/test_rna_quantification.py
```diff
@@ -0,0 +1,149 @@
...
+import ga4gh.datamodel as datamodel
+import ga4gh.datamodel.datasets as datasets
+import ga4gh.protocol as protocol
+import ga4gh.datamodel.rna_quantification as rna_quantification
+import tests.datadriven as datadriven
+import ga4gh.exceptions as exceptions
+
+
+_rnaQuantTestData = {
+    "id": "ZHM6RU5DRkYzMDVMWkI",
```
macieksmuga: I've moved away from quoting actual IDs in these tests: Since the ID generation mechanism is still very much in flux, these _will_ change. Instead, consider similar code from `datadriven/test_sequenceAnnotations.py`:
```
_gencodeV21Set1TestData = {
    "featureSetName": "gencodeV21Set1",
    "referenceName": "chr1",
    "totalFeatures": 543,
    "sampleFeatureId": 4397111632,
    "sampleParentId": 4397111312,
    "sampleStart": 804776,
    "sampleEnd": 804832,
    "sampleStrand": protocol.Strand.POS_STRAND,
    "sampleSiblings": 5,
    "region": [0, 2**32],
    "ontologyRestriction": ["gene", ],
    "featuresWithOntology": 58
}
```
Note that the Ids there are DATABASE IDs, not compoundIds. I generate the compoundIDs on the fly in the test code as follows:
```
        idString = _getFeatureCompoundId(
            _datasetName,
            self._testData["featureSetName"],
            self._testData["sampleFeatureId"])
```

macieksmuga:
I agree with @jeromekelleher - the code here is great! Next step is to merge in the latest sequence annotation code, as it contains all the `Feature` and `Ontology` stuff you'll want.

ga4gh/datamodel/rna_quantification.py
saupchurch: I've been kinda stuck on this one.  Is there a way to get the localId out of the compoundId string?  I see a way to get the localId of a Datamodel object, but it's not clear how to get it from the compoundId directly.

ga4gh/templates/index.html
saupchurch: Probably not as there will be 10's of thousands of them for a given annotation.

ga4gh/datamodel/rna_quantification.py
saupchurch: possibly, though at this point I'd like to get user feedback on the best key to order by

scripts/encode2ga4gh.py
saupchurch: Noted as a possible post-merge issue

saupchurch:
Issues should be address - edited title and initial comment to indicate this is ready for merge with the RNA topic branch.

If all looks well, I'll rebase on master which should bring in the sequence annotation components needed for CI to pass.

ga4gh/_protocol_definitions.py
saupchurch: This is part of RNA and represents an associated group of quantifications - for example a collection of transcripts from the same gene.  If the name is too confusing due to overload of the word 'feature' we can change it in future.

diekhans: Yes, FeatureGroup is somewhat confusing.

david4096:
This is great work, definitely ready to get into the feature branch! Worth noting: the `prepare_compliance_data` script will be updated along with any test data that are added to compliance. The test_simulated_stack tests will be good for top level objects, but I wouldn't worry about mocking things that are backed by the database. For database backed parts you might find a test similar to [this one](https://github.com/ga4gh/server/blob/sequence-annotation/tests/end_to_end/test_sequenceAnnotations.py) helpful, that brings up a datarepo based on the test data files.

ga4gh/datamodel/rna_quantification.py
jeromekelleher: Ah, I didn't realise this! So, if I want to comment on line 275, I click on the '+' on line 276?

ga4gh/datamodel/rna_quantification.py
saupchurch: the length of the list returned here is going to be restricted by the `pageSize` parameter and the backend function that calls this is expected to pass on a list to the next function

scripts/rnaseq2ga.py
macieksmuga: Danny implemented this in #1085 - check out the code pattern for efficient batch imports in that PR.

macieksmuga:
+1 I'm sure we'll find some more things to improve on here, but I'd like to see this code in the topic branch ASAP.

saupchurch:
`rnaseq2ga.py` now does batch inserts to the db

macieksmuga:
Merging, as discussed in the standup from April 19

Issue #1045, opened 2016-04-01
URL support for reads and variants within data repo


jeromekelleher:
{Kevin O: I pasted this initial comment back into the history from my email. Somehow github dropped it!}

We need to provide support for reads and variants being read from a URL. Here is a sketch of one possible implementation:

Within FileSystemDataset, glob for json files. For each JSON file, the localId is the basename of this file. So, for a readGroupSet HG00096, we would have HG00096.json.

The JSON file will look something like this

{
"data_url": "http://example.com/HG00096.bam",
"index_url": "file://ga4gh-example-data/dataset1/reads/HG00096.bam.bai",
}
The idea here is that we abstract back a bit from having the data files and indexes stored in a fixed location. We read the JSON file which stores the URL of the data file and the index file. For now, we assume that the index file is stored locally, but we might change this later to do on-demand fetching and caching. We can also use this JSON file to store metadata about the ReadGroupSet which we obtain from the repo administrator at ingestion time.

Change the signature of reads.HtslibReadGroupSet so that it takes the dataUrl and indexFile as parameters. Use these to open the underlying pysam object.
The rest should follow easily enough. Thoughts?

jeromekelleher:
Another thought: for 1000 genomes, it's probably not such a great idea to have several thousand JSON files lying around inside one directory. We could make a single file to hold all of the URLs/metadata easily enough. Perhaps this should be done as a follow up?

david4096:
Here's an attempt to break this into some more manageable tasks.

- [ ] Repo manager tests
Tests which show the `add-readgroupset repo datasest url` functionality working
Tests which show the `add-removereadgroupset repo datasest rgsName` when a rgs has been added to the repo via url.
- [ ] Reads unit tests
HTSLib Read group set should accept urls
Sending in a url and index file executes the remote rgs code path
- [ ] FileSystemDataset
Datasets will pickup json files as well as BAMs when adding read group sets

jeromekelleher:
Thanks @david4096, I agree we need to split this into more manageable subtasks. However, this was intended as a design discussion rather than an assignment of work. Does anyone have any thoughts on this design, or are we all happy with what I've outlined above?

dcolligan:
Is anyone working on this?  Seems like a prerequisite (or at least something that goes in tandem with) adding remote files via the repo manager.  We should also add it to the milestone, IMHO.

dcolligan:
[...]

PR #1051, opened 2016-04-04
Initial experimental URL code.

jeromekelleher:
-1, for discussion only.

This is intended to illustrate the design discussed in  #1045. Key differences to current code:

1. We only read in the file ``metadata.json`` to find readGroupSets. We no longer depend on globbing files in the directory.
2. The ``name`` field is now stored in the ``metadata.json`` and not derived from the filename. Going forward, we shouldn't assume anything about the structure of filenames (beyond reasonable stuff like them being unique within dataset). We should move towards storing all the stuff we need to know about a readGroupSet here rather than deriving it from the BAM file at run time.
3. Index is supplied separately to allow for reading from URLs.

I would view the ``metadata.json`` as an interrim step towards storing all the metadata in a database, and applying this principle right across the data repo.

Thoughts?


jeromekelleher:
No thoughts at all from anyone on this? I'll assume that everyone thinks this design is perfect in that case.

diekhans:
Looks very sane to me

david4096:
@jeromekelleher Yes, this is a great approach! It places the requirement on the repo manager that it can add and remove elements from the `metadata.json`, and for that reason it may simplify things to make it a map of identifiers to objects as opposed to a list of objects.

Here is the metadata.json I used to test this PR:
```
[
  {"name": "hello",
  "dataUrl": "http://ga4ghstore.blob.core.windows.net/testing/HG00096.bam",
  "indexUrl": "ga4gh-example-data/datasets/1kg-p3-subset/reads/HG00096.bam.bai"}
]
```

I had some issues using pysams remote file access reliably. You can take a look [here](http://nbviewer.jupyter.org/gist/david4096/95c0b187b67a960780f0715863955e48). I found I was getting 500 errors using this PR. Pysam fails with an IO error saying the file is truncated. I'm hoping it's merely my naivete about using pysam properly that caused these issues. I was able to download the file locally and it worked reliably, but when using a remote file I would intermittently receive truncated file errors.

It may be possible that these edited BAMs are a special case that the remote code fails on. I will repeat this exercise with full BAM files to see if the same problem crops up. You don't have to refresh the file handle after a `fetch` or something, do you?

macieksmuga:
I very much like the general direction of this, but I'm wondering - is there some metadata already contained within the BAM headers that we'd be needlessly replicating in the JSON this way?

jeromekelleher:
I think we should just read the BAM headers at import time and store the useful data in a format that's easy to work with. Parsing text headers is slow and complicated, and shouldn't be done at run-time.

I'm closing this PR as it has been superseded by #1079.

Issue #998, opened 2016-03-18
Example data is out of date


gabrielsaldana:
The example-data.tar file in the documentation at http://www.well.ox.ac.uk/~jk/ga4gh-example-data-v3.0.tar is out of date. It is missing the ontologies folder. Also the datasets/brca1/phenotypes folder.

david4096:
The phenotypes folder will need to be folded in with G2P so shouldn't be part of example data, yet. Ontologies, however, does need to be present since it's used by variant annotation.

jeromekelleher:
Do we have some example data to ship for VA @david4096? We should integrate this into the download_data script so it's easy to generate the example data.

david4096:
Yeah there are two annotated VCFs in the compliance test data: https://github.com/ga4gh/compliance/tree/master/test-data

jeromekelleher:
OK, so what we need to do is update download_data script to pull these from some public location, and create the correct directory hierarchy. Any one like to do this?

kozbo:
Assigning to Gabriel as this is right on his path anyway

jeromekelleher:
@kozbo I've invited @gabrielsaldana to join the Global Alliance github group, and added him to the "Global Alliance Contributors" team. He should be on the list of potential assignees once he accepts the invitation.

macieksmuga:
The example data for sequence annotations can already found in the `tests/data/datasets/dataset1/sequenceAnnotations` corner of the server repo, so please queue that for chucking in to the example data package once sequence annotations code is accepted.

david4096:
How about adding the test data to a release @jeromekelleher ? (thanks @diekhans)

https://help.github.com/articles/creating-releases/

jeromekelleher:
What have you in mind @david4096? I don't think we should distribute example data with the pypi package.

gabrielsaldana:
I'm guessing @david4096 is referring to generating downloadable zip files of "releases" from an example-data repository @jeromekelleher instead of the current fixed location mentioned at the top of this thread.

david4096:
You can include large files in a release. You would tag a data release, or make a data branch https://help.github.com/articles/distributing-large-binaries/

jeromekelleher:
I'm perfectly happy with making the data  available somewhere other than the current server. I'm not in favour of rolling the example data in with the distribution tarball on pypi (which is not the same thing as the release tarball created on github). Entirely open to ideas otherwise though.

david4096:
[...]

Issue #940, opened 2016-02-23
Refactor  tests/unit/test_g2p.py


jeromekelleher:
Currently, the tests in  tests/unit/test_g2p.py are a mix of data driven and frontend HTTP tests. These tests should be split out into the data driven tests (which can explicitly depend on the known data) and frontend HTTP tests (which should be put into tests/test_views.py, and follow the patterns in there).

Paging tests should be done in test_simulated_stack.py, following the approach used for ListReferenceBases.

david4096:
This is at least partially fulfilled by the last PR:
Unit tests are [here](https://github.com/ga4gh/server/blob/g2p/tests/unit/test_genotype_phenotype.py) and those tests were moved to [test_views](https://github.com/ga4gh/server/blob/g2p/tests/unit/test_views.py#L121). To really call this done I think there is at least one other unit test to write and the paging tests moved to test_simulated_stack.

bwalsh:
@dcolligan I think these comments have been largely resolved #1345

dcolligan:
Er, maybe @jeromekelleher was talking about `tests/end_to_end/test_g2p.py` ?

jeromekelleher:
Yes, sorry @dcolligan that was a typo. I was talking about ``tests/end_to_end/test_g2p.py``.

dcolligan:
@jeromekelleher has this been resolved to your satisfaction?

kozbo:
@jeromekelleher bump

jeromekelleher:
Thanks yes, looks good.

Issue #939, opened 2016-02-23
G2P needs client JSON tests


jeromekelleher:
Add some tests for g2p endpoints in ``tests/end_to_end/test_client_json.py``. This depends on having a simulator in #936.

dcolligan:
Fixed in #1344

Issue #938, opened 2016-02-23
G2P needs data driven tests


jeromekelleher:
We need to add some data driven tests to ensure that we are correctly translating G2P input data into GA4GH objects.

bwalsh:
@dcolligan I believe this was resolved back in April #1061

PR #1061, opened 2016-04-05
Updated implementation and tests

bwalsh:
Resolves: #998 #940 #939 #938 #937


david4096:
Lots of code, awesome!

It would be great if we could split the update commits from the feature commits. There's enough code in here that github won't print out the diffs for the parts that matter most, so it's hard to do inline comments.

Could the `_toGA4GH` function be decomposed at all?

Although you know how excited I am about the swagger docs, it will complicate the merge to include them at this point. It's still not decided whether they'll be included as part of the package or from some separate location.

In `genotypePhenotypeGenerator` an exception is thrown for a bad request but it looks like the same situation is attempting to be handled on line 305 of `genotype_phenotype.py`, which throws an not implemented. Sorry if I'm mixing cases, just wondering if one is more correct than the other, or perhaps they are on different code paths.

The offsets are being calculated in the beginning of that generator but aren't used (probably to go with the TODO for using SPARQL for paging).

If the PhenotypeAssociationSet is particularly tailored to the cgd database, as suggested by the comment, does it make sense to name it as such? This would make the path to PASets backed by other data more clear.

This is great work! I see there are a couple of problems with the protocol_definitions file, you may want to regenerate this using `scripts/process_schemas`. Does your branch of schemas include VA code? That might be the source of the problem. It seems like it doesn't have the VA stuff fully merged in.

diekhans:
Maybe we need to use a different tool to review.  This is not
showing diffs on 250 line files..  Looks like it's because it
reached an absolute limit, not a per-file limit, so Swagger
and a JS import kill the display.

It's cause a lot of time to reorganize a commit and I am not
convinced it's going to fix this limit.

this looks promising:.
     http://getbarkeep.org/

diekhans:
barkeep did't pass my 10 minute setup threshold.

PR #1067, opened 2016-04-06
Update docs to point at new example data

david4096:
Found a couple of other places where the old example data were being pointed to. Updates the URL to point at a version of the example data with ontologymaps.


dcolligan:
Ah, I was just going to ask about `update_data.py`!

+1

Issue #955, opened 2016-02-29
Support single FASTA ReferenceSets


jeromekelleher:
We need to support ReferenceSets based on a single FASTA file for CRAM support (see #928). Splitting the references by chromosome seems artificial, and will be messy if there are large numbers of contigs in an alignment (perfectly possible in non-model organisms that don't have a good reference).

This may introduce a backward incompatibility with the data repo format currently supported, so we should look carefully at what the best option is:

1) Only support single FASTA ReferenceSets, and provide a script to update old data repo formats to the new one;
2) Only support single FASTA ReferenceSets, and not worry about existing users (there may not be any not using the example data);
3)  Support both single FASTA ReferenceSets and multi-fasta ReferenceSets.

Any thoughts @dcolligan, @diekhans?

dcolligan:
Hm, do we have any idea of which use case is more common?  Or what the preference of the users might be?

diekhans:
A single fasta is most like the more common approach and it's
easy to create a single fasta from multiple fastas.

For less finished assembles, there can be tens of thousands of
scaffolds, so contig per fasta would be a big problem.

macieksmuga:
#829

diekhans:
what is the ETA on finishing this?

macieksmuga:
I'm putting it on my next week (March 28) sprint schedule

jeromekelleher:
Closed by #1166.

Issue #1050, opened 2016-04-02
Sending in unexpected referenceId reads/search request causes 500


david4096:
When sending in requests to `reads/search` if it can't find the reference it fails with an Internal Server Error.

```
➜  server-1 git:(master) ✗ http post kozbo-vm.cloudapp.net/reads/search readGroupIds:='["SEcwMDA5OTpIRzAwMDk5Lm1hcHBlZC5JTExVTUlOQS5id2EuR0JSLmxvd19jb3ZlcmFnZS4yMDEzMDQxNTpTUlI3NDE0MTE"]' referenceId=1  
HTTP/1.0 500 INTERNAL SERVER ERROR
Content-Length: 61
Content-Type: application/json
Date: Sat, 02 Apr 2016 00:32:09 GMT
Server: Werkzeug/0.11.5 Python/2.7.6

{
    "errorCode": 1277500761, 
    "message": "Internal Server Error"
}
```

kozbo:
Here is the output on the server side:
```
ERROR in app [/srv/ga4gh/ga4gh-server-env/local/lib/python2.7/site-packages/flask/app.py:1423]:
Exception on /reads/search [POST]
--------------------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/ga4gh/ga4gh-server-env/local/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/srv/ga4gh/ga4gh-server-env/local/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/srv/ga4gh/ga4gh-server-env/local/lib/python2.7/site-packages/ga4gh-0.2.0a2.dev170+ngcce89bc-py2.7.egg/ga4gh/frontend.py", line 483, in searchReads
    flask.request, app.backend.runSearchReads)
  File "/srv/ga4gh/ga4gh-server-env/local/lib/python2.7/site-packages/ga4gh-0.2.0a2.dev170+ngcce89bc-py2.7.egg/ga4gh/frontend.py", line 407, in handleFlaskPostRequest
    return handleHttpPost(flaskRequest, endpoint)
  File "/srv/ga4gh/ga4gh-server-env/local/lib/python2.7/site-packages/ga4gh-0.2.0a2.dev170+ngcce89bc-py2.7.egg/ga4gh/frontend.py", line 284, in handleHttpPost
[...]

Issue #689, opened 2015-10-08
ReferenceID not set in VariantSet


jeromekelleher:
Currently we have no way of associating a given referenceSet with a VariantSet. Ideally this should be derived from the ``reference`` and ``contig`` tags in the VCF header.

Tixii:
Im working on this

david4096:
bump, this renders the variants interface problematic. There is no way to tell which `referenceName` to send in a `searchVariants` request without knowing ahead of time what is in the VCF.

diekhans:
VariantSet has referenceSetId, this is used to obtain the list
of references that could be in a variant set.

david4096:
@diekhans As far as I can tell `referenceSetId` is pinned to a blank string here:

https://github.com/ga4gh/server/blob/master/ga4gh/datamodel/variants.py#L95

When trying to create a simple script for listing the variants for a variant set I found that one would have to brute force the `referenceName` field of a `SearchVariantsRequest` to find which contigs a variant set refers to.

jeromekelleher:
There is no standard header for defining what reference a VCF uses. I think the only feasible solution here is to require the repository administrator to tell us what it is via the repo manager tool. Ideally, at import time we have an option saying "associate this variant set with this reference set".

diekhans:
ouch.  references set name should be specified when the variant set is added
to by the repo manager.  Given that reference genomes are ubiquitous, I don't
see a reason for it to be optional.

jeromekelleher:
Closed by #1166.

PR #1069, opened 2016-04-07
Sql repo experiment

jeromekelleher:
This is a POC/WIP towards using a SQLite DB as the data repository. After looking at the upcoming requirements for URL based access to data files, metadata improvements to repo manager and scalability requirements for serving 1000 Genomes data, it seemed quicker to just tackle the full DB based repo rather than put more and more bandages on the directory-tree based repo concept.

There are some fairly radical changes here, but I don't think they are too intrusive as a whole. I've tried to minimise the parts of the code that we touch here, and particularly avoid all protobuf-affected code as much as possible. There are several more fairly hefty refactors that we could do after this which would remove a lot of code and complexity (i.e., by using SQL to make paging easier).

## Basic Idea

We get rid of the idea of a hierarchy of files representing the data repository and replace it with a single SQLlite DB. Within this DB we have tables for Datasets, ReadGroupSets, ReadGroups and all the other container objects. These are related to each other via IDs in usual relational manner. Everything that depends on a dataUrl (like a ReadGroupSet) stores the path/URL of the file in question in the DB. The storage/locations of the files is entirely up to the repo administrator. This simplifies many things, particularly the logic in repo manager, as there is no need for copy/move/link semantics any more (effectively, we always link now).

## Transitional Design

The current approach is a transitional design, as we are not fully taking advantage of SQL when running our queries. This is a deliberate approach for now to get things running without breaking too much along the way. However, ultimately we would want to remove the bit where we ``load()`` everything from the DB into memory, and just work directly from the database's model of the world. This would have many advantages such as the possibility of online updating of data, and super-fast server startup times.

## Simplification of repo manager CLI

I have made a start on making changes to the repo manager CLI, which basically involves taking all of the functionality in the current ``repo_manager`` module and spreading this between the immediate CLI interface and the ``datarepo`` module. I think this simplifies adding more functionality to the CLI that we need and keeps the functionality of updating and checking the state of the repository in one place. See the updates to the cli for ``add-referenceset`` as an example. This is rough so far, but the basic idea is there. I know the cli module is getting too big, but we'll be refactoring this into the cli package soon (see #934), and it's good to keep all the CLI related code together.

## Impact for users

Users will have to fully import their data. If there are any, this will be a disruptive change. However, from a user-demo perspective, little will change. We can update the ga4gh-example-data to have the repo DB file in the root directory and contain paths to all the relevant files.

## Repo API

Here's some client code for the repo API:
```python
    repo = datarepo.SqlDataRepository("example-data.db")
    if repo.exists():
        repo.delete()
    
    # Creation and update semantics; TODO create a context manager?
    repo.initialise()
    repo.openDb()
    
    fastaFile = "GRCH38.fa.gz"
    rs1 = references.HtslibReferenceSet("GRCH38", fastaFile)
    repo.insertReferenceSet(rs1)

    ds1 = datasets.Dataset("ds1", "Test dataset 1")
    repo.insertDataset(ds1)

    bamFile = "ga4gh-example-data/datasets/1kg-p3-subset/reads/HG00096.bam"
    readGroupSet = reads.HtslibReadGroupSet(
            ds1, "HG00096", bamFile, bamFile + ".bai")
    repo.insertReadGroupSet(readGroupSet)
    repo.closeDb()

    # Normal read-only usage.
    repo.load()
```

Consider ``HtslibReadGroupSet``. This is created using a name, BAM URL and an index file. It then has two methods, ``populateFromFile`` and ``populateFromRow``. The idea then is that when we are inserting the ReadGroupSet into the DB we call ``populateFromFile`` to read all the headers etc. When we are just creating the object at server run time, we call ``populateFromRow`` which updates the instance variables of the object based on the values from the DB. This means that we _do not_ need to read the data file at run time unless we are actually retrieving read data. This will make a big difference when we're dealing with thousands of remote files in 1000G.

## TODO

1. Finish reference handling, and perform basic checks when importing referenceSets (currently assuming a single FASTA as it's quite a bit simpler, #955)
2. Add in tables for ReadStats and Programs.
3. Add support for Variants (VariantSets, CallSets, VariantAnnotationSets).
4. Fix tests.
5. Update the download_data script to use the new 'repo' API to make creating the example data easier.

In summary, this is an alternative proposal to #1051, and also tackles a number of other issues at the same time. I think I could probably get the basic framework done in another day. As I say, this is a fairly major change so please review what I've done so far and weigh in with what you think (if you think it's rubbish, say so!). Again, apologies for the size of the change but I couldn't see any other way to meet all of our upcoming requirements without tacking this.


david4096:
@jeromekelleher this is stellar!

macieksmuga:
@jeromekelleher +10000 for getting this ball rolling! Will go through the details shortly.

ga4gh/datamodel/reads.py
```diff
@@ -553,17 +573,29 @@ def __init__(self, parentContainer, localId, readGroupHeader=None):
...
+        self._experimentDescription = readGroupHeader.get('DS', None)
+        self._library = readGroupHeader.get('LB', None)
+        self._platformUnit = readGroupHeader.get('PU', None)
+        self._runTime = readGroupHeader.get('DT', None)
+
+    def populateFromRow(self, row):
+        """
+        Populate the instance variables using the specified DB row.
+        """
+        self._sampleId = row[b'sampleId']
```
david4096: Excellent, this is exactly what is needed to be able to handle the variety of linkages that support more complicated queries.

ga4gh/datamodel/reads.py
```diff
@@ -242,7 +256,7 @@ def _referenceSetInit(self, dataRepository, shouldThrowExceptions):
             elif referenceSetName != name:
                 if shouldThrowExceptions:
                     raise exceptions.MultipleReferenceSetsInReadGroupSet(
-                        self._samFilePath, name, referenceSetName)
+                        self._dataUrl, name, referenceSetName)
```
david4096: It makes sense to me to do away with this function (`referenceSetInit`) now that we have a DB row for the readgroupset. Reading the reference set name from the file internals is an implicit assumption we probably shouldn't make. When initializing a readgroupset, providing the referenceSet that it is against would address https://github.com/ga4gh/server/issues/1050 , as well as demonstrate how the other containers would be initialized with a referenceSetId.

david4096:
screen shot 2016-04-07 at 4 02 34 pm

It was helpful to me to see the tables and schemas so we have a clear path to the secondary goal: using queries for paging and querying datamodel objects. I see that you've included `name` in the tables, which is great because our searching-by-name support will improve, same with `sampleId`. The work I've done in the biodata protocol to use the current paging scheme to filter callsets and readgroupsets would be significantly improved if the containers were stored in a table like this.

I believe we should store the referenceSetId alongside the containers for readGroupSets.

With this iteration we can just create a variants table, as annotations and callsets are generated logically from a variantset. 

```
VariantSet
id 			TEXT,
datasetId		TEXT
name 		TEXT,
referenceSetId	TEXT
description 	TEXT,
dataDir 		TEXT,
created		TEXT,
updated		TEXT
callSetIds		TEXT
```

Looking forward to updates!

ga4gh/datamodel/reads.py
jeromekelleher: We definitely want to make this more robust. It would be nice if we could detect the reference set by looking at the md5s at import time automatically (it would save the user some effort), but this should probably be pushed up the stack closer to the repo manager UI.

jeromekelleher:
Thanks @macieksmuga and @david4096, it's great to hear you like the idea! OK, I'd better finish it up then...

jeromekelleher:
Update:

- Variants, References, Reads queries all working in the server.

TODO:

- VariantAnnotations, OntologyMaps
- Update repo manager to use the new API
- Update download data script to use new API
- Fix the rest of the tests.
- Add back in support for ReadStats and Experiment.

ga4gh/datarepo.py
```diff
@@ -179,9 +200,457 @@ def __init__(
...
+            metadataJson, urlMapJson))
+        for callSet in variantSet.getCallSets():
+            self.insertCallSet(callSet)
+
+    def _readVariantSetTable(self, cursor):
+        cursor.row_factory = sqlite3.Row
+        cursor.execute("SELECT * FROM VariantSet;")
+        for row in cursor:
+            dataset = self.getDataset(row[b'datasetId'])
+            referenceSet = self.getReferenceSet(row[b'referenceSetId'])
```
david4096: Perfect, solves this long standing issue https://github.com/ga4gh/server/issues/689.

jeromekelleher:
Update:

- ReferenceSets based on single FASTA.
- download_data script uses the repo API, and has creates a flat file structure
- Partial refactor of VariantAnnotations (was hitting a lot of problems I didn't understand, so seemed better to move on). Disabled test cases for VA while it's broken.
- Partial refactor of repo manager CLI code. Disabled on repo manager while work is continuing.

jeromekelleher:
Any feedback here @dcolligan? I'm hoping to merge this into a topic branch in the next day or two (will be closing this PR as it won't be ready to merge to master), so it would be good to get your thoughts.

diekhans:
Overall, I like this a lot.  Random thoughts on SQL backend

 1) registry problem (aka data outside of transactions)Since
    some data is still in files, it would be to have the repo
    manager have a verify function to make sure the

 2) distributing datasets - it will be really handy to be able
    to put out standard data sets, like reference genomes,
    GENCODE, RefSeq as single database file (or database +
    fasta).

 3) hooking into ucsc browser database - this will still require
    some thinking

jeromekelleher:
@diekhans: great! We should try and push this in as soon as we can I think, as I think this will really speed up our progress towards server 1000G and generally having a useful server. I've created a topic branch and opened a PR against it in #1079. Therefore, closing this PR and moving discussion over there.

ga4gh/cli.py
```diff
@@ -1493,17 +1495,17 @@ def configtest_main(parser=None):
 ##############################################################################
 
 
-def getRawInput(displayString):
```
dcolligan: keep this function... it's mocked in some tests

ga4gh/cli.py
```diff
@@ -1803,6 +1854,7 @@ def getRepoParser():
         subparsers, "init", "Initialize a data repository")
     initParser.set_defaults(runner=InitRunner)
     addRepoArgument(initParser)
+    addForceArgument(initParser)
```
dcolligan: good idea

ga4gh/datarepo.py
```diff
@@ -179,9 +256,536 @@ def __init__(
...
+    def open(self, mode="r"):
+        """
+        Opens this repo in the specified mode.
+
+        TODO: figure out the correct semantics of this and document
+        the intended future behaviour as well as the current
+        transitional behaviour.
+        """
+        if mode not in [MODE_READ, MODE_WRITE]:
+            raise ValueError("Open mode must be 'r' or 'w'")
```
dcolligan: interpolate the r and w strings here into the msg from MODE_READ and MODE_WRITE

tests/datadriven/test_variant_annotations.py
```diff
@@ -13,6 +13,8 @@
 import ga4gh.datamodel as datamodel
 import ga4gh.datamodel.datasets as datasets
 import ga4gh.datamodel.variants as variants
+import ga4gh.datamodel.references as references
+# import ga4gh.datamodel.ontologies as ontologies
```
dcolligan: take out

ga4gh/datamodel/datasets.py
```diff
@@ -6,25 +6,26 @@
...
 import ga4gh.datamodel.reads as reads
 import ga4gh.datamodel.sequenceAnnotations as sequenceAnnotations
 import ga4gh.datamodel.variants as variants
+import ga4gh.datamodel.ontologies as ontologies
 import ga4gh.exceptions as exceptions
 import ga4gh.protocol as protocol
 
 
-class AbstractDataset(datamodel.DatamodelObject):
+class Dataset(datamodel.DatamodelObject):
```
dcolligan: why the name change?

dcolligan:
Definitely something we need going forward.  Commenting all those tests out does worry me, but I since this is a WIP I take it those will be resolved before this is committed...

ga4gh/datamodel/datasets.py
jeromekelleher: I think FileSystemDataset should be removed, and the SimulatedDataset doesn't do anything that a regular Dataset does, other than create a bunch of objects that it contains. Therefore, I think we should get rid of this polymorphism and just use plan Dataset everywhere, and decide what it contains at run-time.

Having said that, I do regret changing the name right now. It has led to a lot of diff noise!

ga4gh/datarepo.py
jeromekelleher: Good spot, done.

jeromekelleher:
Thanks @dcolligan, good to hear you like it. I agree about disabling the tests, but you have to break some eggs some times. These will all be fixed before we merge to master. 

Updates over at #1079.

PR #1074, opened 2016-04-08
1kg handle multiple transcript effects

david4096:
The annotated VCF for 1kg can have multiple ontology terms per transcript, and multiple transcripts per CSQ field. This handles that case properly and adds another test file where this case is shown.


ga4gh/datamodel/variants.py
```diff
@@ -1096,8 +1096,11 @@ def convertVariantAnnotation(self, record, transcriptConverter):
                 annotations, variant, hgvsG, transcriptConverter)
         else:
             annotations = record.info.get('CSQ'.encode())
-            transcriptEffects = self.convertTranscriptEffectCSQ(
-                annotations[0], hgvsG)
+            transcriptEffects = []
+            for ann in annotations:
+                transcriptEffects += \
```
jeromekelleher: Use .append() here rather than creating new lists each time with +=

jeromekelleher:
Looks good, +1 modulo minor comment.

ga4gh/datamodel/variants.py
david4096: Thanks @jeromekelleher the called function returns a list of transcript effects so I've used the extend method, which I hope is more clear.

david4096:
@jeromekelleher updated thanks!

ga4gh/datamodel/variants.py
jeromekelleher: Yes, sorry, my bad. The advantage of using .extend rather than += here is that we can use an arbitrary iterable as the argument. += will choke if the other argument isn't a list.

PR #1076, opened 2016-04-11
Add ExpectException decorator

dcolligan:
Decorator to improve test methods that are written with the purpose of expecting one specific exception to be thrown

### DIFF
--- a/tests/datadriven/test_sequenceAnnotations.py
+++ b/tests/datadriven/test_sequenceAnnotations.py
@@ -11,7 +11,6 @@ import ga4gh.datamodel.datasets as datasets
 import ga4gh.datamodel.sequenceAnnotations as sequenceAnnotations
 import ga4gh.protocol as protocol
 import tests.datadriven as datadriven
-import ga4gh.exceptions as exceptions
 
 _datasetName = "ds"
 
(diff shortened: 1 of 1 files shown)


tests/datadriven/test_sequenceAnnotations.py
```diff
@@ -150,14 +151,11 @@ def testGetFeatureById(self):
...
+    @utils.ExpectException(exceptions.ObjectWithIdNotFoundException)
     def testGetFeatureFailsWithBadId(self):
         idString = str(self._testData["sampleFeatureId"]) + "W00t"
-        try:
-            compoundId = datamodel.FeatureCompoundId.parse(idString)
-            self._gaObject.getFeature(compoundId)
-            self.assertFalse("Exception should be thrown by this point")
-        except exceptions.ObjectWithIdNotFoundException:
-            pass
+        compoundId = datamodel.FeatureCompoundId.parse(idString)
```
jeromekelleher: What's wrong with ``assertRaises``?

dcolligan: It didn't allow me to write a fancy decorator, that's what!

But I agree assertRaises is better because it scopes the line on which one expects the exception to be thrown, rather than the whole function.

dcolligan:
@jeromekelleher updated (although I have half a mind to delete that test entirely, since it seems to be just testing things that are already tested in `test_compound_ids.py`)

jeromekelleher:
Cool, thanks @dcolligan. Delete or not delete, it's up to you: I'm +1 either way.

macieksmuga:
I'd vote to delete the redundant tests if possible. It would be great to have a clearer idea of which kinds of tests go in which modules.

dcolligan:
Decided to just delete the test

PR #1077, opened 2016-04-11
Rename test file to use conventional snakecase

dcolligan:
Proposal: also rename these files to use snake case:

```
ga4gh/datamodel/sequenceAnnotations.py
ga4gh/gff3Parser.py
ga4gh/sqliteBackend.py
tests/end_to_end/test_sequenceAnnotations.py
```


jeromekelleher:
I agree we should change these. What about:
```
ga4gh/datamodel/seqann.py
ga4gh/gff3.py
ga4gh/sqlite_container.py
tests/end_to_end/test_seqann.py
```
I would like to change the module ``sequenceAnnotations`` to ``seqann`` as this is much shorter to type (and takes up less line space) when doing stuff like ``seqann.FeatureSet(...)``. This isn't a strong opinion though, and I'm happy to keep it as something more verbose if people feel strongly about it...

In any case, can we postpone these renamings until either the sql repo is merged or dropped? No point in adding any more merge conflicts.

dcolligan:
I generally dislike truncating names since it makes them less descriptive / harder to read.  And your editor can probably save you the typing, e.g. http://vim.wikia.com/wiki/Any_word_completion

How about:
```
ga4gh/datamodel/sequence_annotations.py
ga4gh/gff3.py
ga4gh/sql.py
tests/end_to_end/test_sequence_annotations.py
```

I suggest `sql.py` on the theory that in principle we should be presenting an interface to any sql database we want to use, and because it's short.

jeromekelleher:
Yeah, OK, fair enough. It's not so much the typing as the line-length issues that I don't like, but you're right, readability is more important.

macieksmuga:
I'm also in favor of `tests/end_to_end/test_sequence_annotations.py`, because by gosh and by golly, I come from Objective-C, a language in which there is no such thing as an identifier that's "too long". :)

dcolligan:
Ok cool sounds like we'll use the names that I proposed, but wait until Jerome has merged the sql repo

dcolligan:
@jeromekelleher updated

jeromekelleher:
I'm not so sure about ``sql`` now, it's quite a natural local variable name. Maybe we should just underscore the name for now and come back to this later when we're refactoring all the SQL'y stuff together? This'll be happening fairly soon I'd imagine.

dcolligan:
@jeromekelleher updated

jeromekelleher:
I agree with this change, but it's cutting across changes I've already made in the SQL Alchemy refactor so I'm not all that enthusiastic about it. However, feel free to push ahead if you think it's important.

david4096:
+1 thanks for doing this @dcolligan

Issue #1064, opened 2016-04-06
Repo_manager assumes that samtools is in PATH


jeromekelleher:
Depending on external executables is brittle and problematic. We should just exit with a helpful error message like "FASTA not indexed. Please run samtools faidx {}" if the FASTA file isn't indexed.

jeromekelleher:
Also remove dependency on tabix.

kozbo:
+1, I like this approach

david4096:
Nice, would also close https://github.com/ga4gh/server/issues/1052 (I believe). @dcolligan what do you think about limiting the scope of the repo manager to expect indexed files already created?

jeromekelleher:
I'd support expecting indexes to exist already for now. I think there are other directions we need to flesh repo manager out in (mostly, metadata) before we bring this functionality in.

dcolligan:
Hm, I'm not so sure about this.  It may be true that depending on outside executables is brittle, but we can do a more involved check for them on a repo manager invocation that might use them.  My concern is that the repo manager should be as easy as possible for people to use -- the user should just be able to point it at a BAM (e.g.) file and have it work.  Creating indexes seems like something that can and should be done automatically.

jmarshall:
Does pysam provide access to htslib's `fai_build()`?  If so, in this case you can build the _.fai_ index natively via the API you're already depending on rather than via an additional dependency on an external executable.

macieksmuga:
The `scripts/prepare_compliance_data.py` script already does this by using equivalent samtools-equivalent `pysam` methods (and its dependencies are identical to what's required of the server code in the first place). Relevant snippet:
```
            pysam.tabix_compress(destFastaFilename, destFastaFilename + ".gz")
            refFasta = pysam.FastaFile(destFastaFilename + ".gz")
            refFasta.close()
```

jeromekelleher:
Closed by #1166.

Issue #1052, opened 2016-04-04
Repo manager leaves files on fail


david4096:
When attempting to add the apparently corrupted file `HG00101.mapped.ILLUMINA.bwa.GBR.low_coverage.20130415.bam` I found the repo manager left behind the linked file, even though the process apparently failed.

```
(server-1)➜  server-1 git:(master) ✗ python repo_dev.py add-readgroupset ga4gh-example-data 1kg-p3-subset ../1000genomes/HG00101.mapped.ILLUMINA.bwa.GBR.low_coverage.20130415.bam 
[W::bam_hdr_read] EOF marker is absent. The input is probably truncated.
Traceback (most recent call last):
  File "repo_dev.py", line 7, in <module>
    ga4gh.cli.repo_main()
  File "/Users/david/server-1/ga4gh/cli.py", line 1672, in repo_main
    runner.run()
  File "/Users/david/server-1/ga4gh/cli.py", line 1468, in run
    self.datasetName, self.filePath, self.moveMode)
  File "/Users/david/server-1/ga4gh/repo_manager.py", line 427, in addReadGroupSet
    pysam.index(destPath.encode('utf-8'))
  File "/Users/david/server-1/server-1/lib/python2.7/site-packages/pysam/utils.py", line 65, in __call__
    "\n".join(stderr)))
pysam.utils.SamtoolsError: 'samtools returned with error 1: stdout=, stderr=samtools index: "ga4gh-example-data/datasets/1kg-p3-subset/reads/HG00101.mapped.ILLUMINA.bwa.GBR.low_coverage.20130415.bam" is corrupted or unsorted\n'

(server-1)➜  server-1 git:(master) ✗ ls ga4gh-example-data/datasets/1kg-p3-subset/reads/HG00101.mapped.ILLUMINA.bwa.GBR.low_coverage.20130415.bam
ga4gh-example-data/datasets/1kg-p3-subset/reads/HG00101.mapped.ILLUMINA.bwa.GBR.low_coverage.20130415.bam
```

dcolligan:
So in the case (like here) where we are doing a processing step that may fail, what do you propose the recovery process should be?  I think:

0) delete all index files that pysam may leave lying around, in any case
1a) if the moveMode is link, delete the link
1b) if the moveMode is copy, delete the file
1c) if the moveMode is move, move the file back to where it came from

Sound good?

david4096:
[...]

Issue #1047, opened 2016-04-01
Refactor repo manager to allow user-specified metadata for reference sets


macieksmuga:
While working on #955 I noticed that the Repo manager currently assumes various metadata values when generating companion JSON files to references. These should instead be optionally specified by the user on the command line or through a user-supplied metadata file. 

Related to #1040

dcolligan:
Please list the metadata values that you want to make user-settable.

jeromekelleher:
Closed by #1166.

Issue #1040, opened 2016-03-31
NCBI Taxon Id 9606 (Homo Sapiens) is hard-coded into the repo manager


macieksmuga:
Make this a default, but allow the user to provide a different value.

jeromekelleher:
Resolved in #1166.

PR #1079, opened 2016-04-11
Sql repo initial work

jeromekelleher:
This is a proposal for changing the DataRepository from a file system derived structure to an explicit SQLite database, containing tables for all container objects. See #1069 for an overview of the design and initial comments and feedback. This PR is against the ``sql_repo`` topic branch so that work can proceed on it until it's ready to merge into master.

I plan on adding a few further updates on the repo manager side, and will summarise the changes in a comment once this is done. In the mean time, any review would be much appreciated.


jeromekelleher:
Status:

1. Variant annotations work needs to be completed and the tests re-enabled.
2. Repo manager CLI needs to be built out and tested following pattern for add-dataset.
3. Test data needs to be converted into a SQL repo so that we can fully remove the FileSystemRepository and FileSystemDataset.

I propose making these changes after merging this initial PR to the topic branch, and tackling these issues as follow up PRs. Then, once we're happy that the code in the topic branch we can merge it into master. Does this sound like a sensible plan?

In any case, this is ready for review (with the caveats above). Please weigh in @macieksmuga, @david4096 , @dcolligan, @diekhans, @kozbo or anyone else that's interested!

ga4gh/datamodel/datasets.py
```diff
@@ -237,63 +252,80 @@ def __init__(
...
         readGroupSetDir = os.path.join(dataDir, self.readsDirName)
         for filename in os.listdir(readGroupSetDir):
             if fnmatch.fnmatch(filename, '*.bam'):
                 localId, _ = os.path.splitext(filename)
                 bamPath = os.path.join(readGroupSetDir, filename)
-                readGroupSet = reads.HtslibReadGroupSet(
-                    self, localId, bamPath, dataRepository)
+                readGroupSet = reads.HtslibReadGroupSet(self, localId)
+                readGroupSet.populateFromFile(bamPath, bamPath + ".bai")
+                referenceSet = dataRepository.getReferenceSetByName(
```
david4096: Nice. Although I would prefer that the reference be associated with the BAM during repo load, this makes it clear how that connection is made.

ga4gh/datamodel/reads.py
```diff
@@ -335,66 +343,73 @@ class HtslibReadGroupSet(AlignmentDataMixin, AbstractReadGroupSet):
...
+        return self._bamHeaderReferenceSetName
+
+    def populateFromRow(self, row):
+        """
+        Populates the instance variables of this ReadGroupSet from the
+        specified database row.
+        """
+        self._dataUrl = row[b'dataUrl']
+        self._indexFile = row[b'indexFile']
+        self._programs = []
```
david4096: Should the referenceSetId be in here?

david4096:
+1 on getting this into a feature branch.

ga4gh/datamodel/datasets.py
jeromekelleher: This is temporary code just to keep our test cases going for now, and is basically just a hack to keep things working. This will all be deleted once we have a SQL repo describing the test data.

ga4gh/datamodel/reads.py
jeromekelleher: I don't think so because we want a referenceSet instance not the ID. I guess we could pass the repo instance through to ``populateFromRow`` and get the referenceSet from there, but this has it's own issues. I'm not overly happy with this ``populateFromRow`` design in any case because we're splitting the knowledge about the SQL schema between several modules. Hopefully we can think of something neater when we're redesigning this container layer.

jeromekelleher:
For the record, here are issues either fully or partially solved here

- #1064 --- we don't try to index anything in repo manager
- #1052 --- repo manager is much simpler
- #1050 --- ReadGroupSet always has a reference set specified at import time
- #1047 --- some initial metadata added, others very simple to add 
- #1045, #970 --- Full remote access for BAMs and VCFs implemented.
- #1040 
- #1030, #751 --- FeatureSets and VariantSets have a mandatory ReferenceSet specified at import time
- #1019 --- We are a long way towards supporting 1000 Genomes
- #955 --- Single FASTA referenceSet fully implemented.

jeromekelleher:
@dcolligan, can you review and weigh in here please? It would be good to get this merged into the feature branch soon.

jeromekelleher:
@david4096, I've done a refactor on the VA code which tidies things up quite a bit I think. I haven't got around to turning the tests on again, but I'm hoping this will be relatively straightforward. Would you mind taking a look please?

dcolligan:
This looks similar enough to what I already reviewed... +1

Reviewing more manageable chunks after this will be much easier

ga4gh/datamodel/datasets.py
```diff
@@ -213,15 +226,18 @@ def __init__(
             numAlignments=1, numFeatureSets=1):
         super(SimulatedDataset, self).__init__(localId)
         self._description = "Simulated dataset {}".format(localId)
+        sequenceOntology = ontologies.Ontology("sequence_ontology")
```
macieksmuga: I think there was a renaming `ontologies` -> `ontologyMaps` that didn't get captured here.

ga4gh/datamodel/datasets.py
```diff
@@ -251,49 +273,51 @@ class FileSystemDataset(AbstractDataset):
     def __init__(self, localId, dataDir, dataRepository):
         super(FileSystemDataset, self).__init__(localId)
         self._dataDir = dataDir
-        self._setMetadata()
 
         # Variants
         variantSetDir = os.path.join(dataDir, self.variantsDirName)
+        # We need a referenceSet instance for variants and features.
+        # To make this work we just pick the first reference set.
+        # This is NOT a good idea!!
```
macieksmuga: Question for @diekhans and others knowledgeable in the use of computational genomics tools: Would it be reasonable to assume that any given dataset is associated with precisely one reference set? If so, this association could be established in the dataset's metadata, and the rest of the code here would be pretty simple.

ga4gh/datamodel/ontologies.py
```diff
@@ -5,25 +5,38 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import fnmatch
-import os
 import ga4gh.protocol as protocol
 
 
-class FileSystemOntology(object):
+class Ontology(object):
```
macieksmuga: Again, `Ontology` -> `OntologyMap` or is it OK as is?

ga4gh/datamodel/datasets.py
diekhans: The GA4GH dataset is an arbitrary container, we shouldn't put
restriction on it.

ga4gh/datamodel/ontologies.py
diekhans: These are not ontologies, they are maps to links to ontologies;
they should not be called Ontology.

ga4gh/datamodel/variants.py
```diff
@@ -619,11 +460,80 @@ def _addDataFile(self, filename):
...
+    def _updateVariantAnnotationSets(self, variantFile):
+        """
+        Updates the variant annotation set associated with this variant using
+        information in the specified pysam variantFile.
+        """
+        # TODO check the consistency of this between VCF files.
+        if self._variantAnnotationSet is None:
+            annotationType = None
+            infoKeys = variantFile.header.info.keys()
+            if 'CSQ' in infoKeys:
```
david4096: This isn't correct, there are three cases at the moment. 1kg uses CSQ, and SNPeff and VEP both use ANN, just formatted differently. It's OK to merge with this, just leaving a note.

david4096:
@jeromekelleher I can see the tests are turned off, and although the implementation is not quite correct, we can adapt the generalization you are introducing for the various formats and work from there. +1

ga4gh/datamodel/ontologies.py
jeromekelleher: I'm confused about this @diekhans --- we have a text file that contains an Ontology (which,  at the moment corresponds to a mapping between IDs and terms), right? Which we call (say) "sequence_ontology". Where does the 'map' come into it? I'm happy to change it back, I'm just confused.

ga4gh/datamodel/variants.py
jeromekelleher: Ah good, thanks. I was wondering about this. How do we we tell the difference between SNPEff and VEP, is there a header to look for?

ga4gh/datamodel/ontologies.py
diekhans: No, it's actually not the ontology, a lot of information is
missing. The ontology has a hierarchal structure.  The ontology
is on a format like OBO:

   http://www.sequenceontology.org/resources/ontology.html

This just gives the ability to take a term name in the variant
annotations inside of a VCF and map it to the full ontology term
which provides a reference that can be use to query an ontology
server.

don't worry, we haven't seen that last of ontologies.

jeromekelleher: Ah, I get it. What about ``OntologyTermMap`` then? This seems to capture what it is we're using fairly well --- we import an OntologyTermMap into the server to allow us to map from term names to full ontology term IDs?

I'll fix this in a follow-up PR (either way) and make sure that it's all consistent.

jeromekelleher:
OK, I've finished the refactor of the VA code. I've changed the generation program to look at the headers, and it looks like we support SnpEff, VEP v77 or VEP v82. This is probably very restrictive, but we'll need to some more VEP output to see how to generalise over different versions. I've left a few TODOs lying around where I think we can improve things,  and we should make issues for these when this is all merged into master. 

@david4096, can you take a look at the last commit please,  and merge the PR into the topic branch if you're happy with it? I'll tackle the remaining issues (bad nomenclature for our handling of ontologies, and lack of repo manager in follow up PRs against the topic branch).

david4096:
Looks good, there are some fixes to how the odd "CSQ" case are handled that were merged in [here](https://github.com/ga4gh/server/pull/1074) after this branch was started. Merging

PR #1081, opened 2016-04-12
Fix 500 if readGroupSet not mapped to referenceSet

dcolligan:
Issue #1050


jeromekelleher:
+1.

dcolligan:
@david4096 could I get your review here?

david4096:
Thanks, I pulled it down and confirmed that it properly catches the 500 from #1050! +1

Issue #1083, opened 2016-04-12
files in scripts directory importing ga4gh


dcolligan:
Some files in the `scripts` directory are importing the `ga4gh` package, which they do not have access to (if it is not installed):
```
$ python scripts/prepare_compliance_data.py -o ga4gh-example-data
Traceback (most recent call last):
  File "scripts/prepare_compliance_data.py", line 18, in <module>
    import generate_gff3_db
  File "/Users/dannyc/ga4gh/server/scripts/generate_gff3_db.py", line 20, in <module>
    import ga4gh.gff3Parser as gff3
ImportError: No module named ga4gh.gff3Parser
```
We need to either remove references to the `ga4gh` package within the `scripts` directory or figure out some import magic to ensure that the scripts are using the development ga4gh rather than the installed ga4gh.

dcolligan:
Right now the only offenders are `generate_gff3_db.py` and `demo_example.py`

dcolligan:
Fixed in #1084

PR #1084, opened 2016-04-12
Add import glue to scripts that import ga4gh

dcolligan:
Issue #1083

### DIFF
--- a/scripts/prepare_compliance_data.py
+++ b/scripts/prepare_compliance_data.py
@@ -196,6 +196,7 @@ class ComplianceDataMunger(object):
         if self.tempdir is not None:
             shutil.rmtree(self.tempdir)
         utils.log("Done converting compliance data.")
+        utils.log("Result in '{}'".format(self.outputDirectory))
 
 
 @utils.Timed()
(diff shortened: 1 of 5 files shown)


macieksmuga:
Looks like a good fix for now, especially as I'd like to re-think our directory structure anyway (move the API-able parts of scripts off into a module or modules within GA4GH, then have the code in `scripts` just be the front-end calls to these APIs.

scripts/prepare_compliance_data.py
```diff
@@ -214,11 +214,13 @@ def main():
         default=None)
     parser.add_argument('--verbose', '-v', action='count', default=0)
     args = parser.parse_args()
+    cdm = None
```
jeromekelleher: The right way to do this is 
```python
cdm = ComplianceDataMunger(args.inputDirectory, args.outputDirectory)
try:
     cdm.run()
finally:
     cdm.cleanup()
 ```

jeromekelleher:
Very nice, +1, modulo comment above.

dcolligan:
Addressed the comments... merging...

PR #1086, opened 2016-04-12
Fixed reference name parsing bug

macieksmuga:
References with dots in their name were crashing the server.

### DIFF
--- a/ga4gh/datamodel/references.py
+++ b/ga4gh/datamodel/references.py
@@ -384,7 +384,7 @@ class HtslibReferenceSet(datamodel.PysamDatamodelMixin, AbstractReferenceSet):
 
     def _addDataFile(self, path):
         dirname, filename = os.path.split(path)
-        localId = filename.split(".")[0]
+        localId = ".".join(filename.split(".")[:-2])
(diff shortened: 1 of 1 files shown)


kozbo:
+1 I tried out this change and it fixed the problem for my server using data from the 1000 Genome data reference set.

dcolligan:
-1 we should solve this at a more general level

see the current repo manager's methods ... perhaps some of those can be refactored out into a path handling library

macieksmuga:
@dcolligan this will be moot once we move to SQL-based server resource management, which is the correct fix. Until then, this is a trivial 1-line fix that actually prevents a server crash on the full GRCh37 reference. Please reconsider.

diekhans:
we need to solve this on a pragmatic level *now*

dcolligan:
Alright, I'll rescind my -1 as long as a bug gets created for the more general issue (some intelligent solution to path/extension handling)

macieksmuga:
Thanks @dcolligan! Created: #1087 to capture this.

jeromekelleher:
+1 --- I agree it's a hacky fix, but the general solution is on its way.

PR #1093, opened 2016-04-13
pull in sequence annotation update to RNA topic branch

saupchurch:
This is a maintenance update only to the RNA topic branch to bring it in line with master.  Critically, this update pulls in the sequence annotation code that the RNA topic branch has a dependence on


macieksmuga:
+1 This is a necessary update to get RNA on the same footing as the rest of the server code. Let's do this now, then sort out any minor resulting issues in separate smaller PRs.

david4096:
+1 merging

PR #1095, opened 2016-04-13
Add landing page snippet

### LINKED ISSUE #1088: Landing page -- support
The index page of the server that prints configuration about the server could include an optional HTML fragment so we can add support blurbs for a deployed server.

This could be a HTML snippet that gets injected into the template if it is present. It would provide a support email for our next deployment.

david4096:
By setting the `LANDING_MESSAGE_HTML` configuration variable to point at an HTML snippet one can modify the landing page without editing the template directly. A default message has been added in the template directory, `landing_message.html`, which links to the reference server github repo.

I refactored the OIDC test, which was examining strings in HTML tags and was brittle to the inclusion of an HTML comment in the landing page. It now selects the session token tag by class.

Closes #1088

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -410,6 +410,12 @@ RESPONSE_VALIDATION
     that they conform to the protocol. This should only be used for development
     purposes.
 
+LANDING_MESSAGE_HTML
+    The server provides a simple landing page at its root. By setting this 
+    value to point at an HTML snippet it is possible to customize the landing 
+    page. This can be helpful to provide support links or details about the 
+    hosted datasets.
+
 OIDC_PROVIDER
     If this value is provided, then OIDC is configured and SSL is used. It is
     the URI of the OpenID Connect provider, which should return an OIDC
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -412,9 +412,9 @@ RESPONSE_VALIDATION
 
 LANDING_MESSAGE_HTML
     The server provides a simple landing page at its root. By setting this 
-    value to point at an HTML snippet it is possible to customize the landing 
-    page. This can be helpful to provide support links or details about the 
-    hosted datasets.
+    value to point at a file containing an HTML block element it is possible to
+    customize the landing page. This can be helpful to provide support links
+    or details about the hosted datasets.
 
 OIDC_PROVIDER
     If this value is provided, then OIDC is configured and SSL is used. It is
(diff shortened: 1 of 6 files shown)


docs/configuration.rst
```diff
@@ -410,6 +410,12 @@ RESPONSE_VALIDATION
     that they conform to the protocol. This should only be used for development
     purposes.
 
+LANDING_MESSAGE_HTML
+    The server provides a simple landing page at its root. By setting this 
+    value to point at an HTML snippet it is possible to customize the landing
```
macieksmuga: at a file containing an HTML snippet

ga4gh/templates/landing_message.html
```diff
@@ -0,0 +1,11 @@
+<!-- //
+    If you would like a different landing message to appear you can modify
+    this file, or set the `LANDING_MESSAGE_HTML` configuration value to
+    point to an HTML snippet.
+//-->
+<div>
+    Welcome to the GA4GH reference server landing page! This page describes
+    the available operations, configuration, and datasets made available by
+    this server instance. For more information please visit
+    <a href="https://github.com/ga4gh/server">our github repository.</a>
```
macieksmuga: Do we want to add the default support address here (ga4gh-server-support@soe.ucsc.edu)?

macieksmuga:
+1 looks great!

docs/configuration.rst
diekhans: This should be defined as an html `block element' (paragraph,
div, etc), not a snippet, which has no formal definition.  The
resulting page should be valid HTML and if the template is
inserted in a place where a block is valid and then block is
valid, it should be valid.

macieksmuga:
Tested the functionality with `DevelopmentConfig`, works like a charm.

jeromekelleher:
Very nice, +1

docs/configuration.rst
```diff
@@ -410,6 +410,12 @@ RESPONSE_VALIDATION
     that they conform to the protocol. This should only be used for development
     purposes.
 
+LANDING_MESSAGE_HTML
+    The server provides a simple landing page at its root. By setting this 
+    value to point at a file containing an HTML snippet it is possible to
```
diekhans: update doc to use correct HTML terminology, not `snippet'

david4096:
Updated the documentation to rigorously request a block element be in the HTML code, as opposed to saying just: "snippet".

PR #1096, opened 2016-04-13
Version bump example data

### LINKED ISSUE #1091: Remove DS_Store from example data
Made on a mac (example-3.2), so hidden files are in there and shouldn't be. Although they don't cause problems when in the root directory, for the sake of cleanliness the tar file should be regenerated without DS_Store files.

david4096:
Housekeeping for the example data, need to bump the version for the `tar` command. Also, the example data in the data release has been updated so we can close #1091.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -63,7 +63,7 @@ Download and unpack the example data:
 .. code-block:: bash
 
   $ wget https://github.com/ga4gh/server/releases/download/data/ga4gh-example-data-v3.2.tar
-  $ tar -xf ga4gh-example-data-v3.1.tar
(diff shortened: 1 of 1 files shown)


jeromekelleher:
+1.

Suggestion: when making the data release tag, should we include the v3.2 there as well?

dcolligan:
+1

david4096:
@jeromekelleher we certainly could do that. I didn't want to add any extra machinery at this stage so am simply adding them to the same release. I think it makes sense to include example data attached to our regular releases, then we can version them with the server release version.

jeromekelleher:
Didn't realise you could do that; much better!

PR #1097, opened 2016-04-13
Moved config errant test to unit tests directory

macieksmuga:
Moved ga4gh.configtest to tests.unit.test_config


macieksmuga:
Needs a bit more work to untangle the config tests from the application code.

PR #1100, opened 2016-04-14
Removed check for Debug so that we always log exceptions.

### LINKED ISSUE #1099: logging exceptions should not be controlled by DEBUG and should default to enabled
It would be really good for server to use the syslog (also python logging) conceptual model.  In particular, logging uncaught exceptions would not be controlled by a option called DEBUG.  This is at least level
`Error'.  This should default to `on'.

see https://en.wikipedia.org/wiki/Syslog#Severity_level

Since logging exceptions is the only thing controlled by `DEBUG',  please change this to default to True until it can be made consistent with syslog.

kozbo:
some nosetests are still failing.

Will close #1099

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -28,6 +28,8 @@ import ga4gh.datamodel as datamodel
 import ga4gh.protocol as protocol
 import ga4gh.exceptions as exceptions
 import ga4gh.datarepo as datarepo
+import logging
+from logging import StreamHandler
 
 
 MIMETYPE = "application/json"
@@ -188,6 +190,9 @@ def configure(configFile=None, baseConfig="ProductionConfig",
     TODO Document this critical function! What does it do? What does
     it assume?
     """
+    file_handler = StreamHandler()
+    file_handler.setLevel(logging.WARNING)
+    app.logger.addHandler(file_handler)
     configStr = 'ga4gh.serverconfig:{0}'.format(baseConfig)
     app.config.from_object(configStr)
     if os.environ.get('GA4GH_CONFIGURATION') is not None:
@@ -340,7 +345,7 @@ def handleException(exception):
     Handles an exception that occurs somewhere in the process of handling
     a request.
     """
-    if app.config['DEBUG']:
+    with app.test_request_context():
         app.log_exception(exception)
     serverException = exception
     if not isinstance(exception, exceptions.BaseServerException):


diekhans:
changing the default for DEBUG might be a better short-term approach.

jeromekelleher:
I'm fine with this as a short-term workaround while we implement proper logging. Not sure why the tests are failing, looks pretty strange. If you like you can put in a check for testing (I'm sure the Flask server sets this somewhere in the config) to suppress the stack dump in this situation (again, as a workaround).

kozbo:
Tests now pass. The earlier failures were due to the test code generating request exceptions without going through the request pathway, thus they were out of context.
... but 
It turns out that app.log_exception() doesn't generate any output if the flask debug flag is off. So my earlier change was insufficient. Added basic logging to stderr and now we see the output again.  There are some tradeoffs to this approach:

- Exceptions generated when running in the development mode will be printed to the console twice, once by flask and once by us. 
- The nose tests will generate a lot of exception output which will give you a scare until you realize that many of these are negative tests. All tests run OK in the end.

Someone might have an idea on how to avoid the first drawback above. If so I would love  to hear it :-)

diekhans:
thoughts:

for #1, on this page, it looks like one is not suppose to mess with the logger
if debugging is enabled and set it if it is not:
   http://flask.pocoo.org/docs/0.10/errorhandling/
there maybe subtleties with debugging turned on and handling
exceptions.  Probably following this paradigm is good.

For #2, we could just setup a logger that sends the output to a
file instead of stderr, but I don't know how nose works so I
don't know any drawbacks.

diekhans:
I looked at the flask code and debug mode adds it's own logger,
hence the prescription to only add a logger if debug is not
enable.

jeromekelleher:
Hmm, this is tricky. Perhaps we could just duplicate the code that Flask uses to add the logger in DEBUG mode, without turning on DEBUG? However, the present workaround is fine if you want to get something in quickly --- we can worry about spamming the output of nosetests later.

diekhans:
The flask documentation gives clear guidance on how setup logging
and app.debug 

    http://flask.pocoo.org/docs/0.10/errorhandling/

jeromekelleher:
+1.

dcolligan:
on the call we decided to merge this (before the release)... merging now...

Issue #1101, opened 2016-04-14
deploy/config.py does not document meaning off DEBUG variable


diekhans:
the documentation on the DEBUG variable just repeats what is obvious from the code and doesn't explain what it does and the fact that it is an dangerous Flask configuration option.

david4096:
Closed with #1103

PR #1103, opened 2016-04-14
increased level of alarm about Flask DEBUG

diekhans:
this is documentation to address user misunderstanding that results in https://github.com/ga4gh/server/issues/1101


david4096:
+1 rebase against upstream/master

PR #1104, opened 2016-04-15
Updates on repo manager

jeromekelleher:
This is a PR updating the repo manager. Changes:

1. Changed the SQL schema so that it enforces the integrity of the data model at a schema level using foreign key contraints.
2. Added default name for objects added in the CLI to be derived from the input file.
3. add-readgroupset now supports both local files and remote URLs, along with optionally specifying an index file. The index file is provided via an optional positional argument, so we do ``add-readgroupset repoFile datasetName bam_url index_file``. When a URL is provided, the index is mandatory.
4. Changed the ``check`` command to ``verify``, and reimplemented it. Since the DB schema now guarantees the integrity of the data model, the focus of this command is changed to making sure that the various containers we're pointing at work as they are supposed to.  At the moment it goes through all of the various containers and makes sure that we can actually read some data back from them. This is an essential debug tool I think. The current implementation is crude, but it is still useful. 
5. Changed error handling in the CLI to throw exceptions of the relevant type all the way back to the top level, where they are caught. For example when a ReadGroupSet cannot be found with a specific name, this exception is thrown, and we get a message like "Error: ReadGroupSet with name 'HG000sd' not found". This is printed to stderr and we exit with status 1. In general, we should aim to handle errors in the repo manager CLI like this, where we just follow the nominal case and expect the exception handling to do the right thing.
6. Re-enabled some repo manager unit tests, and added unit tests for the parts of the CLI that have been implemented.

Dataset, ReferenceSet and ReadGroupSet handling is now in quite good shape. Follow up PRs will add VariantSet, VariantAnnotationSet, and FeatureSet.

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -1636,6 +1636,14 @@ class RepoManager(object):
         readGroupSet.setReferenceSet(referenceSet)
         self._updateRepo(self._repo.insertReadGroupSet, readGroupSet)
 
+    def removeDataset(self):
+        """
+        Removes a dataset from the repo.
+        """
+        self._openRepo()
+        dataset = self._repo.getDatasetByName(self._args.datasetName)
+        self._updateRepo(self._repo.removeDataset, dataset)
+
     #
     # Methods to simplify adding common arguments to the parser.
     #
@@ -1728,12 +1736,6 @@ class RepoManager(object):
             subparsers, "list", "List the contents of the repo")
         listParser.set_defaults(runner="list")
         cls.addRepoArgument(listParser)
-
-        destroyParser = addSubparser(
-            subparsers, "destroy", "Destroy the repo")
-        destroyParser.set_defaults(runner="destroy")
-        cls.addRepoArgument(destroyParser)
-        cls.addForceOption(destroyParser)
 
         addDatasetParser = addSubparser(
             subparsers, "add-dataset", "Add a dataset to the data repo")
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -10,6 +10,7 @@ from __future__ import unicode_literals
 import argparse
 import logging
 import operator
+import os.path
 import sys
 import unittest
 import unittest.loader
@@ -1496,6 +1497,21 @@ def configtest_main(parser=None):
 ##############################################################################
 
 
+def getNameFromPath(filePath):
+    """
+    Returns the filename of the specified path without its extensions.
+    This is usually how we derive the default name for a given object.
+    """
+    if len(filePath) == 0:
+        raise ValueError("Cannot have empty path for name")
+    fileName = os.path.split(filePath)[1]
+    # We need to handle things like .fa.gz, so we can't use
+    # os.path.splitext
+    ret = fileName.split(".")[0]
+    assert ret != ""
+    return ret
+
+
 class RepoManager(object):
     """
     Class that provide command line functionality to manage a
@@ -1565,16 +1581,6 @@ class RepoManager(object):
         Adds a new dataset into this repo.
         """
         self._openRepo()
-        name = self._args.datasetName
-        # TODO need a better API for this...
-        datasetExists = True
-        try:
-            self._repo.getDatasetByName(name)
-        except exceptions.DatasetNameNotFoundException:
-            datasetExists = False
-        if datasetExists:
-            raise exceptions.RepoManagerException(
-                "dataset with name {} already exists".format(name))
         dataset = datasets.Dataset(self._args.datasetName)
         dataset.setDescription(self._args.description)
         self._updateRepo(self._repo.insertDataset, dataset)
@@ -1585,15 +1591,8 @@ class RepoManager(object):
         """
         self._openRepo()
         name = self._args.name
-        # TODO need a better API for this...
-        referenceSetExists = True
-        try:
-            self._repo.getReferenceSetByName(name)
-        except exceptions.ReferenceSetNameNotFoundException:
-            referenceSetExists = False
-        if referenceSetExists:
-            raise exceptions.RepoManagerException(
-                "referenceSet with name {} already exists".format(name))
+        if name is None:
+            name = getNameFromPath(self._args.filePath)
         referenceSet = references.HtslibReferenceSet(name)
         referenceSet.populateFromFile(self._args.filePath)
         referenceSet.setDescription(self._args.description)
@@ -1605,19 +1604,12 @@ class RepoManager(object):
         Adds a new ReadGroupSet into this repo.
         """
         self._openRepo()
-        # TODO put this into a method which we can reuse.
-        try:
-            dataset = self._repo.getDatasetByName(self._args.datasetName)
-        except exceptions.DatasetNameNotFoundException:
-            raise exceptions.RepoManagerException(
-                "Cannot find dataset with name '{}'".format(
-                    self._args.datasetName))
+        dataset = self._repo.getDatasetByName(self._args.datasetName)
         dataUrl = self._args.filePath
         indexFile = dataUrl + ".bai"
         name = self._args.name
         if self._args.name is None:
-            # TODO derive the name from the input URL.
-            name = "TODO"
+            name = getNameFromPath(dataUrl)
         readGroupSet = reads.HtslibReadGroupSet(dataset, name)
         readGroupSet.populateFromFile(dataUrl, indexFile)
         referenceSetName = self._args.referenceSetName
@@ -1626,7 +1618,7 @@ class RepoManager(object):
             referenceSetName = readGroupSet.getBamHeaderReferenceSetName()
         referenceSet = self._repo.getReferenceSetByName(referenceSetName)
         readGroupSet.setReferenceSet(referenceSet)
-        self.updateRepo(self._repo.insertReadGroupSet, readGroupSet)
+        self._updateRepo(self._repo.insertReadGroupSet, readGroupSet)
 
     #
     # Methods to simplify adding common arguments to the parser.
@@ -1841,8 +1833,21 @@ class RepoManager(object):
         runMethod()
 
 
+def repoExitError(message):
+    """
+    Exits the repo manager with error status.
+    """
+    sys.exit("Error: {}".format(message))
+
+
 def repo_main(args=None):
     try:
         RepoManager.runCommand(args)
-    except exceptions.RepoManagerException as rme:
-        print("Error: ", str(rme), file=sys.stderr)
+    except exceptions.RepoManagerException as e:
+        # These are exceptions that happen throughout the CLI, and are
+        # used to communicate back to the user
+        repoExitError(str(e))
+    except exceptions.NotFoundException as e:
+        # We expect NotFoundExceptions to occur when we're looking for
+        # datasets, readGroupsets, etc.
+        repoExitError(str(e))
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -1625,7 +1625,9 @@ class RepoManager(object):
         dataUrl = self._args.dataFile
         indexFile = self._args.indexFile
         parsed = urlparse.urlparse(dataUrl)
-        if parsed.scheme == 'http':
+        # TODO, add https support and others when they have been
+        # tested.
+        if parsed.scheme in ['http', 'ftp']:
             if indexFile is None:
                 raise exceptions.MissingIndexException(dataUrl)
         else:
@@ -1824,7 +1826,7 @@ class RepoManager(object):
             "indexFile", nargs='?', default=None,
             help=(
                 "The file path of the BAM index for this ReadGroupSet. "
-                "If the dataFile is a argument is a local file, this will "
+                "If the dataFile argument is a local file, this will "
                 "be automatically inferred by appending '.bai' to the "
                 "file name. If the dataFile is a remote URL the path to "
                 "a local file containing the BAM index must be provided"))
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -15,6 +15,7 @@ import sys
 import unittest
 import unittest.loader
 import unittest.suite
+import urlparse
 
 import requests
 
@@ -1576,6 +1577,14 @@ class RepoManager(object):
         # TODO this is _very_ crude. We need much more options and detail here.
         self._repo.printSummary()
 
+    def verify(self):
+        """
+        Checks that the data pointed to in the repository works and
+        we don't have any broken URLs, missing files, etc.
+        """
+        self._openRepo()
+        self._repo.verify()
+
     def addDataset(self):
         """
         Adds a new dataset into this repo.
@@ -1605,8 +1614,15 @@ class RepoManager(object):
         """
         self._openRepo()
         dataset = self._repo.getDatasetByName(self._args.datasetName)
-        dataUrl = self._args.filePath
-        indexFile = dataUrl + ".bai"
+        dataUrl = self._args.dataFile
+        indexFile = self._args.indexFile
+        parsed = urlparse.urlparse(dataUrl)
+        if parsed.scheme == 'http':
+            if indexFile is None:
+                raise exceptions.MissingIndexException(dataUrl)
+        else:
+            if indexFile is None:
+                indexFile = dataUrl + ".bai"
         name = self._args.name
         if self._args.name is None:
             name = getNameFromPath(dataUrl)
@@ -1652,7 +1668,7 @@ class RepoManager(object):
             "the name of the reference set to associate with this {}"
         ).format(objectType)
         subparser.add_argument(
-            "--referenceSetName", default=None, help=helpText)
+            "-R", "--referenceSetName", default=None, help=helpText)
 
     @classmethod
     def addOntologyNameArgument(cls, subparser):
@@ -1702,10 +1718,11 @@ class RepoManager(object):
         cls.addRepoArgument(initParser)
         cls.addForceOption(initParser)
 
-        checkParser = addSubparser(
-            subparsers, "check", "Check to see if repo is well-formed")
-        checkParser.set_defaults(runner="check")
-        cls.addRepoArgument(checkParser)
+        verifyParser = addSubparser(
+            subparsers, "verify",
+            "Verifies the repository by examing all data files")
+        verifyParser.set_defaults(runner="verify")
+        cls.addRepoArgument(verifyParser)
 
         listParser = addSubparser(
             subparsers, "list", "List the contents of the repo")
@@ -1762,7 +1779,17 @@ class RepoManager(object):
         cls.addDatasetNameArgument(addReadGroupSetParser)
         cls.addNameOption(addReadGroupSetParser, objectType)
         cls.addReferenceSetNameOption(addReadGroupSetParser, "ReadGroupSet")
-        cls.addFilePathArgument(addReadGroupSetParser)
+        addReadGroupSetParser.add_argument(
+            "dataFile",
+            help="The file path or URL of the BAM file for this ReadGroupSet")
+        addReadGroupSetParser.add_argument(
+            "indexFile", nargs='?', default=None,
+            help=(
+                "The file path of the BAM index for this ReadGroupSet. "
+                "If the dataFile is a argument is a local file, this will "
+                "be automatically inferred by appending '.bai' to the "
+                "file name. If the dataFile is a remote URL the path to "
+                "a local file containing the BAM index must be provided"))
 
         addOntologyMapParser = addSubparser(
             subparsers, "add-ontologymap",
@@ -1850,4 +1877,8 @@ def repo_main(args=None):
     except exceptions.NotFoundException as e:
         # We expect NotFoundExceptions to occur when we're looking for
         # datasets, readGroupsets, etc.
+        repoExitError(str(e))
+    except exceptions.DataException as e:
+        # We expect DataExceptions to occur when a file open fails,
+        # a URL cannot be reached, etc.
         repoExitError(str(e))
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -1513,6 +1513,14 @@ def getNameFromPath(filePath):
     return ret
 
 
+def getRawInput(display):
+    """
+    Wrapper around raw_input; put into separate function so that it
+    can be easily mocked for tests.
+    """
+    return raw_input(display)
+
+
 class RepoManager(object):
     """
     Class that provide command line functionality to manage a
@@ -1523,14 +1531,14 @@ class RepoManager(object):
         self._repoPath = args.repoPath
         self._repo = datarepo.SqlDataRepository(self._repoPath)
 
-    def _confirmRun(self, func, deleteString):
+    def _confirmDelete(self, objectType, name, func):
         if self._args.force:
             func()
         else:
             displayString = (
-                "Are you sure you want to delete data in {}? "
-                "[y|N] ".format(deleteString))
-            userResponse = raw_input(displayString)
+                "Are you sure you want to delete the {} '{}'? "
+                "[y|N] ".format(objectType, name))
+            userResponse = getRawInput(displayString)
             if userResponse.strip() == 'y':
                 func()
             else:
@@ -1636,13 +1644,28 @@ class RepoManager(object):
         readGroupSet.setReferenceSet(referenceSet)
         self._updateRepo(self._repo.insertReadGroupSet, readGroupSet)
 
+    def removeReferenceSet(self):
+        """
+        Removes a referenceSet from the repo.
+        """
+        self._openRepo()
+        referenceSet = self._repo.getReferenceSetByName(
+            self._args.referenceSetName)
+
+        def func():
+            self._updateRepo(self._repo.removeReferenceSet, referenceSet)
+        self._confirmDelete("ReferenceSet", referenceSet.getLocalId(), func)
+
     def removeDataset(self):
         """
         Removes a dataset from the repo.
         """
         self._openRepo()
         dataset = self._repo.getDatasetByName(self._args.datasetName)
-        self._updateRepo(self._repo.removeDataset, dataset)
+
+        def func():
+            self._updateRepo(self._repo.removeDataset, dataset)
+        self._confirmDelete("Dataset", dataset.getLocalId(), func)
 
     #
     # Methods to simplify adding common arguments to the parser.
@@ -1803,7 +1826,7 @@ class RepoManager(object):
         removeOntologyMapParser = addSubparser(
             subparsers, "remove-ontologymap",
             "Remove an ontology map from the repo")
-        removeOntologyMapParser.set_defaults(runner="removeOntology")
+        removeOntologyMapParser.set_defaults(runner="removeOntologyMap")
         cls.addRepoArgument(removeOntologyMapParser)
         cls.addOntologyNameArgument(removeOntologyMapParser)
         cls.addForceOption(removeOntologyMapParser)
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -1656,6 +1656,19 @@ class RepoManager(object):
             self._updateRepo(self._repo.removeReferenceSet, referenceSet)
         self._confirmDelete("ReferenceSet", referenceSet.getLocalId(), func)
 
+    def removeReadGroupSet(self):
+        """
+        Removes a readGroupSet from the repo.
+        """
+        self._openRepo()
+        dataset = self._repo.getDatasetByName(self._args.datasetName)
+        readGroupSet = dataset.getReadGroupSetByName(
+            self._args.readGroupSetName)
+
+        def func():
+            self._updateRepo(self._repo.removeReadGroupSet, readGroupSet)
+        self._confirmDelete("ReadGroupSet", readGroupSet.getLocalId(), func)
+
     def removeDataset(self):
         """
         Removes a dataset from the repo.
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -8,6 +8,7 @@ from __future__ import unicode_literals
 
 import datetime
 import json
+import os.path
 import random
 
 import pysam
@@ -204,6 +205,10 @@ class AlignmentDataMixin(datamodel.PysamDatamodelMixin):
         return ret
 
     def openFile(self, dataFile):
+        # We need to check to see if the path exists here as pysam does
+        # not throw an error if the index is missing.
+        if not os.path.exists(self._indexFile):
+            raise exceptions.FileOpenFailedException(self._indexFile)
         return pysam.AlignmentFile(
             self._dataUrl, filepath_index=self._indexFile)
 
--- a/ga4gh/datarepo.py
+++ b/ga4gh/datarepo.py
@@ -320,6 +320,45 @@ class SqlDataRepository(AbstractDataRepository):
         self._dbConnection.close()
         self._dbConnection = None
 
+    def verify(self):
+        """
+        Verifies that the data in the repository is consistent.
+        """
+        # TODO this should emit to a log that we can configure so we can
+        # have verbosity levels. We should provide a way to configure
+        # where we look at various chromosomes and so on. This will be
+        # an important debug tool for administrators.
+        for referenceSet in self.getReferenceSets():
+            print(
+                "Verifying ReferenceSet", referenceSet.getLocalId(),
+                "@", referenceSet.getDataUrl())
+            for reference in referenceSet.getReferences():
+                length = min(reference.getLength(), 1000)
+                bases = reference.getBases(0, length)
+                assert len(bases) == length
+                print(
+                    "\tReading", length, "bases from",
+                    reference.getLocalId())
+        for dataset in self.getDatasets():
+            print("Verifying Dataset", dataset.getLocalId())
+            for readGroupSet in dataset.getReadGroupSets():
+                print(
+                    "\tVerifying ReadGroupSet", readGroupSet.getLocalId(),
+                    "@", readGroupSet.getDataUrl())
+                references = readGroupSet.getReferenceSet().getReferences()
+                # TODO should we cycle through the references? Should probably
+                # be an option.
+                reference = references[0]
+                max_alignments = 10
+                for readGroup in readGroupSet.getReadGroups():
+                    alignments = readGroup.getReadAlignments(reference)
+                    for i, alignment in enumerate(alignments):
+                        if i == max_alignments:
+                            break
+                    print(
+                        "\t\tRead", i, "alignments from",
+                        readGroup.getLocalId())
+
     def _createSystemTable(self, cursor):
         sql = """
             CREATE TABLE System (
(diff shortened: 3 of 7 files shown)


jeromekelleher:
Here's the current schema for the SQLites among you (see what I did there?):
```sql
CREATE TABLE System (
                key TEXT NOT NULL PRIMARY KEY,
                value TEXT NOT NULL
            );
CREATE TABLE Ontology (
                name TEXT NOT NULL PRIMARY KEY,
                dataUrl TEXT NOT NULL
            );
CREATE TABLE ReferenceSet (
                id TEXT NOT NULL PRIMARY KEY,
                name TEXT NOT NULL,
                description TEXT,
                assemblyId TEXT,
                isDerived INTEGER,
                md5checksum TEXT,
                ncbiTaxonId TEXT,
                sourceAccessions TEXT,
                sourceUri TEXT,
                dataUrl TEXT NOT NULL,
                UNIQUE (name)
            );
CREATE TABLE Reference (
                id TEXT PRIMARY KEY,
                name TEXT NOT NULL,
                referenceSetId TEXT NOT NULL,
                length INTEGER,
                isDerived INTEGER,
                md5checksum TEXT,
                ncbiTaxonId TEXT,
                sourceAccessions TEXT,
                sourceDivergence REAL,
                sourceUri TEXT,
                UNIQUE (referenceSetId, name),
                FOREIGN KEY(referenceSetId) REFERENCES ReferenceSet(id)
                    ON DELETE CASCADE
            );
CREATE TABLE Dataset (
                id TEXT NOT NULL PRIMARY KEY,
                name TEXT NOT NULL,
                description TEXT,
                UNIQUE (name)
            );
CREATE TABLE ReadGroupSet (
                id TEXT NOT NULL PRIMARY KEY,
                name TEXT NOT NULL,
                datasetId TEXT NOT NULL,
                referenceSetId TEXT NOT NULL,
                programs TEXT,
                dataUrl TEXT NOT NULL,
                indexFile TEXT NOT NULL,
                UNIQUE (datasetId, name),
                FOREIGN KEY(datasetId) REFERENCES Dataset(id)
                    ON DELETE CASCADE,
                FOREIGN KEY(referenceSetId) REFERENCES ReferenceSet(id)
            );
CREATE TABLE ReadGroup (
                id TEXT NOT NULL PRIMARY KEY,
                name TEXT NOT NULL,
                readGroupSetId TEXT NOT NULL,
                predictedInsertSize INTEGER,
                sampleId TEXT,
                description TEXT,
                created TEXT,
                updated TEXT,
                UNIQUE (readGroupSetId, name),
                FOREIGN KEY(readGroupSetId) REFERENCES ReadGroupSet(id)
                    ON DELETE CASCADE
            );
CREATE TABLE CallSet (
                id TEXT NOT NULL PRIMARY KEY,
                name TEXT NOT NULL,
                variantSetId TEXT NOT NULL,
                UNIQUE (variantSetId, name),
                FOREIGN KEY(variantSetId) REFERENCES VariantSet(id)
                    ON DELETE CASCADE
            );
CREATE TABLE VariantSet (
                id TEXT NOT NULL PRIMARY KEY,
                name TEXT NOT NULL,
                datasetId TEXT NOT NULL,
                referenceSetId TEXT NOT NULL,
                created TEXT,
                updated TEXT,
                metadata TEXT,
                dataUrlIndexMap TEXT NOT NULL,
                UNIQUE (datasetID, name),
                FOREIGN KEY(datasetId) REFERENCES Dataset(id)
                    ON DELETE CASCADE,
                FOREIGN KEY(referenceSetId) REFERENCES ReferenceSet(id)
            );
CREATE TABLE VariantAnnotationSet (
                id TEXT NOT NULL PRIMARY KEY,
                name TEXT NOT NULL,
                variantSetId TEXT NOT NULL,
                analysis TEXT,
                annotationType TEXT,
                UNIQUE (variantSetId, name),
                FOREIGN KEY(variantSetId) REFERENCES VariantSet(id)
                    ON DELETE CASCADE
            );
CREATE TABLE FeatureSet (
                id TEXT NOT NULL PRIMARY KEY,
                name TEXT NOT NULL,
                datasetId TEXT NOT NULL,
                referenceSetId TEXT NOT NULL,
                info TEXT,
                sourceUri TEXT,
                dataUrl TEXT NOT NULL,
                UNIQUE (datasetId, name),
                FOREIGN KEY(datasetId) REFERENCES Dataset(id)
                    ON DELETE CASCADE,
                FOREIGN KEY(referenceSetId) REFERENCES ReferenceSet(id)
            );
```
What do people think of using the schema to enforce repo integrity?

diekhans:
I am all for foreign key constraints to enforce integrity.

There is a reason SQL still runs the world.

jeromekelleher:
Can I get a review please @dcolligan, @macieksmuga, @david4096?

ga4gh/cli.py
```diff
@@ -1496,6 +1498,29 @@ def configtest_main(parser=None):
...
+    Returns the filename of the specified path without its extensions.
+    This is usually how we derive the default name for a given object.
+    """
+    if len(filePath) == 0:
+        raise ValueError("Cannot have empty path for name")
+    fileName = os.path.split(filePath)[1]
+    # We need to handle things like .fa.gz, so we can't use
+    # os.path.splitext
+    ret = fileName.split(".")[0]
+    assert ret != ""
```
dcolligan: probably good to turn this into a better exception

ga4gh/cli.py
```diff
@@ -1496,6 +1498,29 @@ def configtest_main(parser=None):
...
+    return ret
+
+
+def getRawInput(display):
```
dcolligan: yay it's back

ga4gh/cli.py
```diff
@@ -1605,19 +1621,19 @@ def addReadGroupSet(self):
...
-            raise exceptions.RepoManagerException(
-                "Cannot find dataset with name '{}'".format(
-                    self._args.datasetName))
-        dataUrl = self._args.filePath
-        indexFile = dataUrl + ".bai"
+        dataset = self._repo.getDatasetByName(self._args.datasetName)
+        dataUrl = self._args.dataFile
+        indexFile = self._args.indexFile
+        parsed = urlparse.urlparse(dataUrl)
+        if parsed.scheme == 'http':
```
dcolligan: what about https?

ga4gh/cli.py
```diff
@@ -1710,22 +1762,17 @@ def getParser(cls):
...
+            "Verifies the repository by examing all data files")
+        verifyParser.set_defaults(runner="verify")
+        cls.addRepoArgument(verifyParser)
 
         listParser = addSubparser(
             subparsers, "list", "List the contents of the repo")
         listParser.set_defaults(runner="list")
         cls.addRepoArgument(listParser)
 
-        destroyParser = addSubparser(
```
dcolligan: we're removing the destroy method?

ga4gh/datamodel/reads.py
```diff
@@ -204,6 +205,10 @@ def convertReadAlignment(self, read, readGroupSet, readGroup):
         return ret
 
     def openFile(self, dataFile):
+        # We need to check to see if the path exists here as pysam does
+        # not throw an error if the index is missing.
```
dcolligan: grumble grumble

ga4gh/datarepo.py
```diff
@@ -320,28 +324,73 @@ def close(self):
...
+        # where we look at various chromosomes and so on. This will be
+        # an important debug tool for administrators.
+        for referenceSet in self.getReferenceSets():
+            print(
+                "Verifying ReferenceSet", referenceSet.getLocalId(),
+                "@", referenceSet.getDataUrl())
+            for reference in referenceSet.getReferences():
+                length = min(reference.getLength(), 1000)
+                bases = reference.getBases(0, length)
+                assert len(bases) == length
```
dcolligan: why is this check in here?

ga4gh/datarepo.py
```diff
@@ -320,28 +324,73 @@ def close(self):
...
+            for readGroupSet in dataset.getReadGroupSets():
+                print(
+                    "\tVerifying ReadGroupSet", readGroupSet.getLocalId(),
+                    "@", readGroupSet.getDataUrl())
+                references = readGroupSet.getReferenceSet().getReferences()
+                # TODO should we cycle through the references? Should probably
+                # be an option.
+                reference = references[0]
+                max_alignments = 10
+                for readGroup in readGroupSet.getReadGroups():
```
dcolligan: what is this check doing?

ga4gh/datarepo.py
```diff
@@ -320,28 +324,73 @@ def close(self):
...
+                    print(
+                        "\t\tRead", i, "alignments from",
+                        readGroup.getLocalId())
+
     def _createSystemTable(self, cursor):
         sql = """
             CREATE TABLE System (
-                repositoryVersion TEXT,
-                creationTimeStamp TEXT
+                key TEXT NOT NULL PRIMARY KEY,
```
dcolligan: maybe at some point in the future we could add creation / modification timestamps to objects... that could be useful in some cases...

ga4gh/datarepo.py
```diff
@@ -320,28 +324,73 @@ def close(self):
...
+                value TEXT NOT NULL
             );
         """
         cursor.execute(sql)
-        cursor.execute("INSERT INTO System VALUES (0.1, datetime('now'));")
+        cursor.execute(
+            "INSERT INTO System VALUES ('repositoryVersion', '0.1')")
```
dcolligan: what does 0.1 mean?  what does it refer to?

dcolligan:
Agree with Mark: it's great to have the dependencies between objects made explicit via key constraints.

Generally looks good.

ga4gh/cli.py
jeromekelleher: It doesn't seem to work in pysam right now. I should add FTP though, will change.

ga4gh/cli.py
jeromekelleher: Yeah, it seemed bit pointless since ``ga4gh_repo destroy repo.db == rm repo.db``. We're not storing any data inside the repo anymore, so if the user deletes the file then that's their own business.

ga4gh/datarepo.py
jeromekelleher: Hah, basically just to keep flake8 from complaining! I can get rid of it if it seems confusing, I just thought it was a useful piece of paranoia.

ga4gh/datarepo.py
jeromekelleher: Basically checking that the remote file exists and is readable, and the index exists and works. Actually reading down some data and processing it seems to me the only reliable way of testing the integrity of the data files.

ga4gh/datarepo.py
jeromekelleher: There are in the schema where they are defined (e.g., variant set, etc). This is just for the repository itself.

ga4gh/datarepo.py
jeromekelleher: This is basically just the stub for doing repository versioning properly. In the future we'll need to worry about repository version compatibility when we do schema changes. This 0.1 is just signifying that this is a very early version of the repo schema.

dcolligan: https://github.com/ga4gh/server/issues/1116

dcolligan:
This is a against a long running branch, so I'll +1 it

ga4gh/cli.py
jeromekelleher: The assert? I think this classifies as a programming error if we accept the input and the output is an empty string. This is just trying to be defensive, putting in asserts for stuff that shouldn't happen, but might.

ga4gh/cli.py
```diff
@@ -1770,7 +1819,17 @@ def getParser(cls):
...
         cls.addReferenceSetNameOption(addReadGroupSetParser, "ReadGroupSet")
-        cls.addFilePathArgument(addReadGroupSetParser)
+        addReadGroupSetParser.add_argument(
+            "dataFile",
+            help="The file path or URL of the BAM file for this ReadGroupSet")
+        addReadGroupSetParser.add_argument(
+            "indexFile", nargs='?', default=None,
+            help=(
+                "The file path of the BAM index for this ReadGroupSet. "
+                "If the dataFile is a argument is a local file, this will "
```
david4096: A typo I think

ga4gh/datarepo.py
```diff
@@ -442,15 +495,19 @@ def insertReferenceSet(self, referenceSet):
...
-            json.dumps(referenceSet.getSourceAccessions()),
-            referenceSet.getSourceUri(), referenceSet.getDataUrl()))
+        try:
+            cursor.execute(sql, (
+                referenceSet.getId(), referenceSet.getLocalId(),
+                referenceSet.getDescription(), referenceSet.getAssemblyId(),
+                referenceSet.getIsDerived(), referenceSet.getMd5Checksum(),
+                referenceSet.getNcbiTaxonId(),
+                # We store the list of sourceAccessions as a JSON string.
+                # Perhaps this should be another table?
```
david4096: Yes, considering we want to support searching on this column, we'll probably want to move it to its own table.

ga4gh/exceptions.py
```diff
@@ -595,3 +552,25 @@ class RepoManagerException(Exception):
...
+class DuplicateNameException(RepoManagerException):
+    """
+    The user has tried to insert an object with the same name as
+    an existing object.
+    """
+    def __init__(self, objectName, containerName=None):
+        msg = "Name '{}' already exists".format(objectName)
+        if containerName is not None:
+            msg += " in '{}'".format(containerName)
+        super(DuplicateNameException, self).__init__(msg)
```
david4096: One of the most interesting parts of the SQL repo is removing the requirement that all containers have unique names. Why are we adding that requirement back in? I believe this is the path to doing away with the in-memory `IdNameMap` objects in `Datasets`. If it is important to maintain that restriction it should be a requirement of the (to be deprecated) datasets module, not the repo.

tests/unit/test_cli.py
```diff
@@ -327,31 +313,27 @@ def testAddReferenceSet(self):
...
+        cliInput = "add-readgroupset {} {} {} ".format(
             self.repoPath, self.datasetName, self.filePath)
         args = self.parser.parse_args(cliInput.split())
         self.assertEquals(args.repoPath, self.repoPath)
         self.assertEquals(args.datasetName, self.datasetName)
-        self.assertEquals(args.filePath, self.filePath)
-        self.assertEquals(args.moveMode, "copy")
-        self.assertEquals(args.runner, cli.AddReadGroupSetRunner)
+        self.assertEquals(args.dataFile, self.filePath)
+        self.assertEquals(args.runner, "addReadGroupSet")
```
david4096: Added a new argument to test for here, indexfile.

tests/unit/test_repo_manager.py
```diff
@@ -34,179 +62,253 @@ def readRepo(self):
...
+        super(TestAddReferenceSet, self).setUp()
+        self.init()
+
+    def testDefaults(self):
+        fastaFile = paths.ncbi37FaPath
+        name = os.path.split(fastaFile)[1].split(".")[0]
+        self.runCommand("add-referenceset {} {}".format(
+            self._repoPath, fastaFile))
+        repo = self.readRepo()
+        referenceSet = repo.getReferenceSetByName(name)
```
david4096: Along the same lines as another comment, it seems like we shouldn't be adding extra logic that assumes names are singular. `getByName` returns a single value, which it in fact should not. Although one could argue that references are a special case where one would never have two of the same name, names are in fact optional according to the [schemas for reference sets](https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/references.avdl#L84). We would do best to clear the way for the code to be true to the schemas, which to me would mean this function returns a list, not a single value.

ga4gh/cli.py
macieksmuga: Can we add an optional argument to `getNameFromPath` that would allow you to provide the possible extensions to expect, and remove only those if specified? So for example:
```
getNameFromPath('GL000248.1.fa.gz.fai', ['fa.gz','fa.gz.fai','fa.gz.gzi', 'json'])
```
would correctly return `GL000248.1` as the filename. (this is a real filename from the GRCh37 reference set).

david4096:
Looking good! The SQL foreign key usage is clever and helps us avoid a lot of unnecessary code in maintaining the repo. We could put in blocks for removing reference sets that are currently being used, a nice feature.

As stated in my comments above, I believe the name restrictions being put on the repo are an attempt to fit to the current in-memory dataset and should be a part of that code, which will soon be deprecated. The reasoning for this is that we should be able to support degenerate naming schemes and the SQL repo could quite easily provide identifiers at load time, instead of using compound IDs. In fact, I believe since you've set up the tables with foreign keys, the path is clear to remove that bit of code for most container objects.

I would prefer to see the repo not take on the technical debt of the in memory dataset. Leaving those restrictions in place made sense when we were simply trying to make do with the filesystemdataset, but let's take advantage of this architecture change by not making the same assumptions.

macieksmuga:
My one quibble has to do with the primary keys in all the tables. Why are they text and not some auto-generated number? Where are they coming from? Which part of the codebase ensures they're actually unique, and in cases when they're not, why is that (are we really trying to insert the same thing twice by accident, or is there just some name collision somewhere)?

diekhans:
If they are true primary keys (sql unique index), the database
will generate an error on duplicate insert.

ga4gh/cli.py
jeromekelleher: I'm not sure why we'd need this @macieksmuga, the function already gives the output you want:
```
>>> cli.getNameFromPath('GL000248.1.fa.gz.fai')
u'GL000248'
```

ga4gh/cli.py
jeromekelleher: Well spotted, thanks.

ga4gh/datarepo.py
jeromekelleher: OK, good, let's do this in a follow up.

ga4gh/exceptions.py
jeromekelleher: Well, names are still required to be unique _within_ a container, but they don't have to be unique between containers. This is a generic exception that we use to report when someone tries to insert a (say) dataset with the same name as an existing dataset.  I've added a comment to clarify this point.

tests/unit/test_cli.py
jeromekelleher: Nice one. I've added a new test case to test for this.

tests/unit/test_repo_manager.py
jeromekelleher: I disagree here I'm afraid. I don't see what we would have to gain by allowing for duplicate names within a container. The schema _allows_ for duplicate names (bizarrely, IMO), but it doesn't _require_ it. Our lives are made a lot easier by regarding names as within-container, user specified IDs, and we're free to do that. Maybe I'm misunderstanding the point you're making here though? Feel free to open an issue to discuss it in more detail.

jeromekelleher:
Awesome, thanks for the review guys, I really appreciate it. More eyes makes better code! I've added some fixes in response to the stuff you've spotted. For the bigger stuff:

- I genuinely don't see why we'd want to ever allow for duplicate names within a container @david4096 --- what would we gain from this? Seeing names as user specified within-container IDs seems like a genuinely useful construct to me, whereas seeing them as essentially a free-text ``description`` field seems ... I don't know, perverse? This assumption is baked into the current code pretty deeply (look at how the CLI is structured for a start), so I'd need some pretty compelling reasons to make me want to change it at this point. Like I said though, maybe I'm missing some important point here.

- @macieksmuga, excellent point. The IDs are  exactly as they were before, which allows us to continue the (soon to be overhauled) CompoundId infrastructure. Once we have everything commited in and we're all working off the SQL repo, I think we should (probably) change over to an integer ID for all the container types. I think the right strategy here is to generate each ID randomly so there's no risk of information leakage through ID values of containers. This is a discussion for another day though!

tests/unit/test_repo_manager.py
diekhans: The allowance for duplicate names was somewhat contentious. I
believe it came from google wanting to not restrict it's user's
data in a programming API.

There is no requirement for the server to be able to host data
with duplicate names.  There might be some annoying requirements
for in some dataset.

For instance, genes in the PAR regions are given the same gene
identifier because they are annotating the exact same DNA
clones.  The PAR is a weird case were the haploid reference
sequence is pseudo-diploid.  Annotations in these regions break
lots of software assumptions.

The semantics of `name' are so weakly defined that it's hard to say what is
right.

I suspect we will need to allow duplicate names in some
containers, but not all.

jeromekelleher: That seems like a pretty remote corner case to me @diekhans! In any case, we can deal with the situation where we need to support duplicate names within a specific container when we get to it. Right now, do you see any reason to allow for (say) duplicate dataset names?

david4096:
The main reason why I think we shouldn't write code that requires names to be singular is that there is nothing in the schema that requires it. We added the extra requirement that names are singular when backed by a filesystem, but now we have a more general approach which can better satisfy the schemas.

>The name is user-defined, isn’t programmatically required to be unique in any given scope (although in practice data owners will often choose unique names), and is intended to be human readable. This can be thought of as a display name.

This SQL repo is an architectural change and I understand we can't fix everything at once, but I don't think the fact that the CLI requires names to be singular as a good case for a database to. The CLI doesn't serve to prove the schemas are correct or not. Proper separation of concerns would allow the pieces of code fragile to name collisions (filesystemdataset) free to handle exceptions that will affect it. 

Given the schemas explicitly state that names don't have to be unique, we are in a position of needing to justify the reason that they are enforced to be unique for the reference server. Being backed by files directly was a good case for that, but I'm struggling to see it now. The SQL repo is the decoupling we need to adhere to the schemas so adding code to it that steps us back from that seems like a mistake to me.

If singularity of names remains a requirement during this stage that's ok, but it ought to be clear to someone arriving at the code why that requirement exists and moving those checks to `filesystemdataset` make more sense to me than at the repo level.

jeromekelleher:
OK, let's consider this from the perspective of a repo administrator. I have a ten thousand BAM files and a few hundred VCFs, spread across  a dozen datasets. I want to add a new dataset called 'study_xyz', so I run
```bash
$ ga4gh_repo add-dataset repo.db study_xyy
```
As it happens, I made a mistake and I typed 'study_xyy' instead, which is the name of an already existing dataset. What is the expected behaviour here? I would say, overwhelmingly,
```bash
$ ga4gh_repo add-dataset repo.db study_xyy
Error: Name 'study_xyy' already exists
```
What would be the purpose of allowing the administrator create a dataset with the same name as an existing dataset? Why would an administrator want to do this? This is surely an error, or there is no value in the concept of a dataset ``name`` at all.  

You might argue that we should use IDs to refer to datasets instead, and that this will allow us to avoid this problem. OK, but I think this gives makes the lives of our repository administrator really hard. For example,
```bash
$ ga4gh_repo add-dataset repo.db study_xyy
Created dataset with ID '123423542134523'
$ ga4gh_repo add-dataset repo.db study_xyy
Created dataset with ID '123423532134523'
```
Now, which dataset is which? How do I remember which dataset I want to put a particular BAM file into?  To me, the whole purpose of names is to provide the repository administrator a way of attaching meaningful (local) identifiers to objects. Otherwise, I think trying to manage a large data repository would be a nightmare.

I've concentrated on datasets here, but this generalises. For example, suppose I run the following commands:
```bash
$  ga4gh_repo add-readgroupset repo.db study_xyz ga4gh-example-data/HG00096.bam
$  ga4gh_repo add-readgroupset repo.db study_xyz ga4gh-example-data/HG00096.bam
```
What's the expected behaviour here? Again, I would argue that the user expectations would be, overwhelmingly, that we exit with error status. Why would I want to add the same ReadGroupSet to a dataset? OK, if I really want to I can do
```bash
$  ga4gh_repo add-readgroupset repo.db study_xyz ga4gh-example-data/HG00096.bam --name=HG00096_copy
```
but this just emphasises the point that we're doing something odd. Unless we use name as a local identifier, we have no way of catching these common user errors beyond hashing the VCF or BAM file. 

This sort of thing will really matter to an admin who is managing a large repository. Allowing for duplicates names opens up all sorts of possibilities for weird behaviour for zero benefit. Can you think of a concrete case where allowing a duplicate name for a Dataset, ReadGroupSet, ReadGroup, ReferenceSet, Reference, FeatureSet, VariantSet, VariantAnnotationSet or CallSet would be useful? The schema is there to allow a variety of implementations, and some have thought it important that allowing duplicate names will allow freedom of implementation. Fine. But we shouldn't cripple our implementation just for the sake of following the full possible generality of the schema. Our users would not thank us for it.

jeromekelleher:
@macieksmuga, @david4096, if you are happy with this PR (modulo ongoing dicsussion about duplicate names) can you +1 and merge please?

diekhans:
Completely agree that containers with the same name is a
disaster.  I would suggest this be the restriction.

It's gets a bit more complex for the contents of some containers
for items that don't naturally have one name.  For instance,
repeat annotations.

jeromekelleher:
Good distinction @diekhans --- a _container_ name is a local ID used to organise a data repository. The meaning of the "name" of a data object is a different story, and may need complex semantics. Luckily, we're not touching any data objects here in this PR...

macieksmuga:
[GA4GH Server Repo SQL v0.1.pdf](https://github.com/ga4gh/server/files/226378/GA4GH.Server.Repo.SQL.v0.1.pdf)

diekhans:
Very useful diagram.  We should check the source file into the tree.

ga4gh/cli.py
macieksmuga: But... `GL000248` is incorrect, which was my whole point here.

jeromekelleher:
Yes, this is excellent, thanks @macieksmuga!

ga4gh/cli.py
jeromekelleher: Ahh, now I get you! To be honest, I think this is a corner case  we can ignore for now. This specific case (which is a problem in the current server) will go away because we're not trying to read in files for  each reference any more. I think we can safely say now that if the automatic name generation fails in these obscure cases then it's up to the user to step in and use the ``--name=GL000248.1`` command line option.

Feel free to open an issue on this if you disagree though.

macieksmuga: @jeromekelleher works for me! 😄

jeromekelleher:
@macieksmuga, @david4096, can you +1 and merge if you're happy with this please? PR #1117 is waiting for review, and there's another in the pipeline.

macieksmuga:
+1 There are some outstanding issues but I'm comfortable hashing them out in the `ga4gh/repo_sql` branch.

jeromekelleher:
@david4096?

dcolligan:
+1

There's a lot that depends on this topic branch change and there's been enough time to review it... I'm going to merge.

david4096:
@jeromekelleher Thanks for trying to keep me in the loop on this one! The reason I favor the "return an ID" approach is that it always does just what I ask, a server is not a filesystem. It doesn't make sense to me to make decisions outside the scope of the schemas unless we're certain they're wrong. And if they are wrong the ref server should suss it out by implementing the functionality in good faith and finding the error.

Imagine for a moment I have a directory of alignments, each *directory* named for the specimen. However, the files within are just called `reads.bam`. In order to add these reads to the system the manager would have to rename or give an optional name for each, when names, as specified by the schemas, are just human readable strings and should not be given extra importance. They should all be able to be "reads.bam" and sorted out via metadata and search requests.

We could not guarantee the behavior of the add action when we were filling a directory with the file straight, but now we have a mechanism to properly apply a pattern that is familiar from other REST APIs (returning an ID when an add is successful). You can have as many friends named Sam as you like, it's the underlying characteristics that allow you to disambiguate them.

It is too early to optimize the experience of a repo administrator, it will be software that handles IDs making differentiation solved clearly via code. We should be focused on staying as close to the schemas as possible so that we can prove their correctness. If there is a technical reason why we can't follow the schemas, that needs to be fixed in the schemas. Inferring names from files overcomplicates the code.

In fact, if we followed the schemas on this one I believe we would enable API features. For example, de-duplication can be a difficult task in the world of files. If we allowed duplicate names I could easily add a bag of BAMs and craft search requests that would be able to disambiguate them. Another interesting case would be adding multiple of the same BAM to allow an administrator to compare how two different cohorts accessed the data. A third example would be if I wanted to be able to collate reads within a dataset. `Name` is currently the human readable, hand-edited portion that we can use to do this. Then I can search by name to return all BAMs that were aligned using this or that method. Requiring our implementation to always return a single result using search-by-name seems like a hold over from a filesystem approach to the problem.

diekhans:
The semantics of name are weakly defined in the API specification and the lack of a uniqueness requirement within a dataset is controversial.  The use cases have not been thought through.  

For containers, there is be an expectation that this would be the external identifier, such as the accession.  If a container does not have a unique, human readable name withing a dataset it seems very problematic to manage.  How does a human identify it outside of the server-generated id?  

That said, inferring the name from he file is problematic and while it might save typing, it is prone to error, with the user having to remember the inference rules and perhaps rename files.  Explicitly supplying a name would be far better.  

A restriction where datasets are uniquely named and containers names within data sets are uniquely name seems very reasonable and allowed by the schema.

jeromekelleher:
Let's look at your examples in turn @david4096:

> Imagine for a moment I have a directory of alignments, each directory named for the specimen. However, the files within are just called reads.bam. In order to add these reads to the system the manager would have to rename or give an optional name for each, when names, as specified by the schemas, are just human readable strings and should not be given extra importance. They should all be able to be "reads.bam" and sorted out via metadata and search requests.

My expectation here is that the admin would used the ``--name`` option and call the ReadGroupSet after the directory. If there are a bunch of files in the directory (which we'll call ``sample_x`` for arguments sake) called ``reads_chr1.bam``, ``reads_chr2.bam``, ``unmapped.bam``), the admin would use the ``--name`` argument to call the ReadGroupSets ``sample_x_chr1``, `sample_x_unmapped``, etc. 

The vital point here is that we are allowing the admin to organise their repository in whatever way they want. The ``name`` is the human readable identifier that allows the admin _remember_ which ReadGroupSet is which. Humans aren't good at remembering long auto generated IDs for things. We need help in organising data. Bioinformatics data is already hideously messy: why make it worse by not providing memorable names for things?

What possible good would it be to the administrator to have 1000 ReadGroupSets called ``reads``? I accept that it is _possible_ to disambiguate them using queries, but that's not the point. I should be able to give these ReadGroupSets memorable names so that I'll be able to remember what they are and get them back without having to grep the descriptions of all ReadGroupSets for some string, and then get the ID, and then paste this ID into my ``remove-readgroupset`` command line. 

> We could not guarantee the behavior of the add action when we were filling a directory with the file straight, but now we have a mechanism to properly apply a pattern that is familiar from other REST APIs (returning an ID when an add is successful). You can have as many friends named Sam as you like, it's the underlying characteristics that allow you to disambiguate them.

Yes, OK, but how do you _remember_ which IDs are which? Say I had 20 slightly different versions of GRCh38, and a bunch of BAMs aligned to them. How would I look at the ReadGroupSet and know which of these GRCh38 version it was? There's no way you can expect admins to remember long integer IDs. Do we include the ``description`` of the ReferenceSet in with the ReadGroupSet (when we're displaying for listing purposes)? OK, but then what is the point of a ``name``? There is huge scope for error here, for no gain.

> It is too early to optimize the experience of a repo administrator, it will be software that handles IDs making differentiation solved clearly via code. We should be focused on staying as close to the schemas as possible so that we can prove their correctness. If there is a technical reason why we can't follow the schemas, that needs to be fixed in the schemas. Inferring names from files overcomplicates the code.

This isn't optimising for administrator experience, this is trying to make it _possible_ to administer a large repo. We _do_ implement the schema here because it doesn't require names be non-unique. We are not the only implementer of the schema, and we shouldn't expect the schema to fit only our needs. We only infer names from filenames as a user convenience, as I think this is often what people will expect. There is a ``--name`` option for the admin to specify their own names this is always available.

> If we allowed duplicate names I could easily add a bag of BAMs and craft search requests that would be able to disambiguate them.

OK, this is an application for duplicate names. However, I think it's important to distinguish between our two classes of user here: administrators and clients. Administrators are responsible for running the server and setting up the data repository so that it represents some data in a coherent way (for example, they are associating the correct ReferenceSets with ReadGroupSets). Clients, on the other hand consume this coherent model of the data on a server. All they know about the data is what is available through the API. Now, why would it be a good thing from a client's perspective to have a bunch of ReadGroupSets with the same name? The client should expect that the administrator has carefully curated the data and that meaningful names have been attached to objects. This  process of "sort out a bag of BAMs" should  be done before they are added to the repo, and can be done with any number of existing tools. This is better for both the client experience and the administrators sanity.

> Another interesting case would be adding multiple of the same BAM to allow an administrator to compare how two different cohorts accessed the data.

I'm not sure I follow this one. How would this not be achieved by adding the same BAM multiple times under different names?

 > A third example would be if I wanted to be able to collate reads within a dataset. Name is currently the human readable, hand-edited portion that we can use to do this. Then I can search by name to return all BAMs that were aligned using this or that method. Requiring our implementation to always return a single result using search-by-name seems like a hold over from a filesystem approach to the problem.

We should  not build functionality into the API by using non-standard extensions requiring particular formats within the ``name`` field. This is not the way towards an interoperable standard. The current search API is next to useless. It needs way more flexibility and the ability to search on all sorts of things. We need to identify useful things that the API should be able to do (like what you've just said) and build these into the API properly by adding the fields and structures that we need, so that it can be done in an interoperable way. I hope that we can start feeding this functionality back into the API by deploying what we have now on 1000 genomes and other large datasets, and by demonstrating how lame the API is trying to run obvious queries. 

The schema should not be treated as a inviolable source of truth here. It's a work in progress, that hopefully will begin to mature soon when people start to use it. 

I don't think there's much to be gained from debating this any more at this point. I'm pretty convinced that there is a lot more to be lost from having duplicate names (at least for the objects that we currently have) than there is to be gained. I would encourage you to try out the repo management CLI, and to actually build a large repo with it (with lots of different datasets and references, etc), and then see how easy it is to manage without using the names. In the mean time, let's get back to making exciting progress!

jeromekelleher:
@diekhans:

> That said, inferring the name from he file is problematic and while it might save typing, it is prone to error, with the user having to remember the inference rules and perhaps rename files. Explicitly supplying a name would be far better.

This is the default, but can be overridden by the user with the ``--name`` argument. The worst that happens here is that we infer a crappy name and the admin has to change it afterwards. They will never have to change the names of their files.

We can change this to a mandatory positional argument if you like, but I think this would be detrimental. We can have this discussion once everybody has had a chance to use the code and build and manage some repos.

diekhans:
As long as the rules for file name to set name mapping are clear
and precise, it's might save more time than create hassle.  The
previous parsing was too simplistic.

Do the names get printed after load so one knows what happened.

Yesterday, I revived some files names *.fasta (not .fa).  Would
one get an error if it doesn't recognize the name as a common one
for the supported type?

jeromekelleher:
We don't print the name now, which isn't good, I agree. We should probably print out a message summarising what we've just added (which can be silenced with -q). Totally happy to rewrite the parsing rules into whatever we think makes sense.

``.fasta`` files will be handled happily --- at the moment we're just taking ``filename.split(".")[0]``, modulo a few complications.

diekhans:
filename.split(".")[0] is easy to describe, however its
problematic because it's fairly common to have `'' in the name
vs the extension for version number and other reasons.

file type dependent removal of suffixes seems more likely to
produce the correct answer, although more work to document and
maintain.

PR #1107, opened 2016-04-15
Swagger Docs

david4096:
[screen shot 2016-04-15 at 3 22 39 pm](http://ga4gh-david.cloudapp.net:8000/)

After discussing with the group the value of this work and the desire to keep the source tree clean I have taken the following approach.

A new endpoint has been added `/swagger`. At this endpoint the `swagger.json` is served, so any swagger-ui can consume this endpoint to generate the interactive docs. The `swagger.json` included with this PR is not valid JSON, but a jinja2 template that allows us to dynamically set the host and version. This could be extended to dynamically generate other parts of the documentation as well.

I've created a new repo for the swagger docs [ga4gh-swagger-ui](https://github.com/david4096/ga4gh-swagger-ui). This does not contain the `swagger.json` and is a simple reskinning of the swagger-ui.

If you would like to evaluate this setup I've created the following demonstration.

A ref server running this branch (david4096/swagger_json) is running [here](http://ga4gh-david.cloudapp.net). If you hit the [swagger](http://ga4gh-david.cloudapp.net/swagger) endpoint you should be returned a well-formed swagger.json.

At a different port a [simple HTTP server is running](http://ga4gh-david.cloudapp.net:8000) hosting the ga4gh-swagger-ui. I have done this for simplicity and a proper apache configuration could allow both to be served on :80. If you visit the swagger UI by default it will populate using the swagger JSON generated using the code in this PR.

This differs from what I had suggested in the previous comment in that we are not serving a static JSON, but dynamically injecting the host and version. This allows the UI to properly route requests to the proper host once it has consumed the swagger.json, without any javascript voodoo.

@diekhans Swagger-codegen can be used to generate static docs as well. You can view the raw output of `swagger-codegen -l html` [here](https://rawgit.com/david4096/8e4f339f83f35b5b704c38d5f100fd81/raw/36efe0527f5c01e367a80953d31f9a54c16be06c/ga4gh-swagger.html).

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -619,6 +619,19 @@ def getFeature(id):
         id, flask.request, app.backend.runGetFeature)
 
 
+@DisplayedRoute(
+    '/swagger',
+    pathDisplay='/swagger')
+def swaggerDocs():
+    info = {}
+    info['host'] = flask.request.host
+    info['info'] = app.serverStatus
+    return flask.Response(
+        flask.render_template(
+            'swagger.json', info=info),
+        mimetype="application/json")
+
+
 @app.route('/oauth2callback', methods=['GET'])
 def oidcCallback():
     """
--- a/ga4gh/templates/swagger.json
+++ b/ga4gh/templates/swagger.json
@@ -1,0 +1,2460 @@
+{
+    "swagger": "2.0",
+    "info": {
+        "description": "The Genomics API developed by the Global Alliance enables DNA data providers and consumers to better share information and work together on a global scale, advancing genome research and clinical application.",
+        "version": "{{info.info.getServerVersion()}}",
+        "title": "GA4GH Reference Server",
+        "contact": {
+            "email": "ga4gh-server-support@soe.ucsc.edu"
+        },
+        "license": {
+            "name": "Apache 2.0",
+            "url": "http://www.apache.org/licenses/LICENSE-2.0.html"
+        }
+    },
+    "host": "{{ info.host }}",
+    "tags": [
+        {
+            "name": "Variants",
+            "description": "Variant endpoints",
+            "externalDocs": {
+                "description": "Find out more",
+                "url": "https://github.com/ga4gh/schemas"
+            }
+        },
+        {
+            "name": "Reads",
+            "description": "Reads endpoints",
+            "externalDocs": {
+                "description": "Find out more",
+                "url": "https://github.com/ga4gh/schemas"
+            }
+        },
+        {
+            "name": "References",
+            "description": "Reference endpoints",
+            "externalDocs": {
+                "description": "Find out more",
+                "url": "https://github.com/ga4gh/schemas"
+            }
+        },
+        {
+            "name": "Datasets",
+            "description": "Dataset endpoints",
+            "externalDocs": {
+                "description": "Find out more",
+                "url": "https://github.com/ga4gh/schemas"
+            }
+        },
+        {
+            "name": "Sequence Annotation",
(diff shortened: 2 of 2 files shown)


macieksmuga:
+1 this looks like a clean approach to me. My only question is, is there better value in maintaining the swagger json in a separate repo altogether?

jeromekelleher:
This is beautiful stuff, and we should get this running against lots of servers ASAP. However, I agree with @macieksmuga, and I think we should push this into it's own repo (say, ``api-explorer`` ?). Here are my reasons:

1. We should make it easy for people to use this on any GA4GH server, not just the reference server. Embedding it into the reference server code makes it difficult for other server providers to use it. By putting this into it's own repo we can add nice simple instructions to deploy this by pointing it at an arbitrary server.
2. This isn't really part of the server per se, just something really cool and useful to put on top of it. We are increasing our attack surface, and the code we need to maintain within the server.
3. By pushing this into it's own repo we can give proper credit to the people that developed it, and not have their contributions overwhelmed within the general server stuff.

@skeenan, what do you think? Are you happy with us creating a new repo for the API explorer?

david4096:
@jeromekelleher 

1. I agree, but other server developers are going to have to make modifications to the documentation to their purposes (versioning, detailed messages, and authentication). The swagger.json is an attempt to be a faithful port of the schemas and so some version could probably live there. However, we make modifications for our usage of it, and so it makes sense to me to serve the documentation for the running server with it.

2. The addition of this PR does add an endpoint that simply returns a JSON. Eventually this file would be auto-generated. You may have seen [examples](https://github.com/gangverk/flask-swagger) [of](https://github.com/rantav/flask-restful-swagger) [servers](https://github.com/kogosoftwarellc/express-openapi) that self-document, they generate a description of the API based on the code. This is a fruitful direction as well and is captured in the spirit of the `DisplayedRoute` decorator. In the meantime grabbing this file during `process_schemas` and adding in the details to our installation seems like it would avoid the added maintenance.

3. It makes sense to me to host a branded version of the `swagger-ui` [separately](https://github.com/david4096/ga4gh-swagger-ui) from the server. If we are determined to not host the swagger.json from the server itself we could add a feature to swagger-ui to grab a description from one place and point requests at a separate host.

jeromekelleher:
I'm worried that we are embedded documentation within the server that is effectively generated somewhere else. This will inevitably grow stale and out of date. It would be better IMO to have a separate repository that is responsible for converting the Avro/protobuf schema into the swagger description, and releasing the authoritative swagger.json. 

How will this be affected by the protobuf conversion?

What's the killer reason for including this in the server itself in your view? It doesn't seem like that much extra work to have an optional "if you want to also install the API explorer UI, follow these instructions" bit in the documentation. For everyday development, we can add an option into the CLI with a path to a swagger.json file to serve.

david4096:
@jeromekelleher The schemas are where these should be generated from. Unfortunately the protobuf conversion complicates the picture since the JSON on the wire is different from the IDL. Perhaps not significantly, but it will not be a 1-1 conversion as it is at the moment.

However, there are [existing tools](https://github.com/gengo/grpc-gateway) for generating swagger.json from gRPC definitions (which the protobuf schemas present). 

I've made a [separate repo](https://github.com/david4096/ga4gh-swagger-ui) for the swagger-ui, if we want to further modify it for our purposes. I believe that it is a valuable addition to make the API self-documenting. Your flask endpoints can be decorated such that API documentation are [generated automatically](https://github.com/gangverk/flask-swagger). That way your API description is next to the code, lowering the chance that you will lose sync. The spirit of this was captured with the `@DisplayedRoute` decorator.

PR #1108, opened 2016-04-15
Point at path for landing message in templates dir

david4096:
While testing deployment @macieksmuga found that the running server wouldn't necessarily be able to reach the directory containing the default landing message. This caused an internal server error.

This PR fixes that case by using flask's render_template to guarantee we are looking at the proper path.


macieksmuga:
+1 Better code, plus it works :)

jeromekelleher:
+1. Merging.

Issue #1111, opened 2016-04-18
Port onologies to SQL repo, and fix nomenclature


jeromekelleher:
Currently `ontology` objects are confused and inconsistent. Change the name to OntologyTermMap throughout and implement add/remove/verify functionality.

jeromekelleher:
Done in #1117.
￼
￼￼￼

Issue #1113, opened 2016-04-18
OntologyTermMap should log missed terms


jeromekelleher:
Currently the OntologyTermMap silently returns the empty string when a term name cannot be found in the map. This is very problematic behaviour for an administrator, as they will not know when their map is incomplete.

The behaviour int this case should be configurable (perhaps there are a number of different things an admin might want to do here?), but certainly there should be an option to log these misses.

Logging low-level problems like these should be taken into account when we're designing the logging infrastructure.

PR #1117, opened 2016-04-18
Ontologies update

jeromekelleher:
Depends on #1104.

Makes two changes:

1. Change the Ontology/OntologyMap object to  OntologyTermMap internally. This is supposed to model what it's doing a bit better. The current setup is rather brittle and will need to be improved, but this at least makes it a bit clearer what's going on (I hope).
2. Ported the CLI support for ontologies to the SQL repo.  The SQL schema and CLI used the term Ontology rather than OntologyTermMap because it seems that the current approach is quite
limited and will need to be changed. This seems more forward compatible, since we don't want to affect uses by making them change CLI syntax or make backwards incompatible changes to the schema.

I would imagine that we would want to update the CLI ``add-ontology`` command to download the map from some authoritative source, and store the terms in the DB. Asking the repo administrator to create a text file mapping the term names to IDs is pretty user-unfriendly. I don't think we should tackle this now, I'm just giving my reasoning for not fully pushing the OntologyTermMap nomenclature into the repo mangement UI.

Issue #1111.

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -32,6 +32,7 @@ import ga4gh.datamodel.reads as reads
 import ga4gh.datamodel.references as references
 import ga4gh.datamodel.sequenceAnnotations as sequenceAnnotations
 import ga4gh.datamodel.datasets as datasets
+import ga4gh.datamodel.ontologies as ontologies
 
 
 # the maximum value of a long type in avro = 2**63 - 1
@@ -1594,6 +1595,18 @@ class RepoManager(object):
         self._openRepo()
         self._repo.verify()
 
+    def addOntology(self):
+        """
+        Adds a new Ontology to this repo.
+        """
+        self._openRepo()
+        name = self._args.name
+        if name is None:
+            name = getNameFromPath(self._args.filePath)
+        ontologyTermMap = ontologies.OntologyTermMap(name)
+        ontologyTermMap.populateFromFile(self._args.filePath)
+        self._updateRepo(self._repo.insertOntologyTermMap, ontologyTermMap)
+
     def addDataset(self):
         """
         Adds a new dataset into this repo.
@@ -1713,6 +1726,19 @@ class RepoManager(object):
             self._updateRepo(self._repo.removeFeatureSet, featureSet)
         self._confirmDelete("FeatureSet", featureSet.getLocalId(), func)
 
+    def removeOntology(self):
+        """
+        Removes an ontology from the repo.
+        """
+        self._openRepo()
+        ontologyTermMap = self._repo.getOntologyTermMapByName(
+            self._args.ontologyName)
+
+        def func():
+            self._updateRepo(
+                self._repo.removeOntologyTermMap, ontologyTermMap)
+        self._confirmDelete("Ontology", ontologyTermMap.getLocalId(), func)
+
     #
     # Methods to simplify adding common arguments to the parser.
     #
@@ -1750,8 +1776,8 @@ class RepoManager(object):
     @classmethod
     def addOntologyNameArgument(cls, subparser):
         subparser.add_argument(
-            "ontologyMapName",
-            help="the name of the ontology map to create/modify")
+            "ontologyName",
+            help="the name of the ontology")
 
     @classmethod
     def addReadGroupSetNameArgument(cls, subparser):
@@ -1772,9 +1798,8 @@ class RepoManager(object):
             help="the name of the variant set")
 
     @classmethod
-    def addFilePathArgument(cls, subparser):
-        subparser.add_argument(
-            "filePath", help="the path of the file to be moved into the repo")
+    def addFilePathArgument(cls, subparser, helpText):
+        subparser.add_argument("filePath", help=helpText)
 
     @classmethod
     def addNameOption(cls, parser, objectType):
@@ -1827,7 +1852,10 @@ class RepoManager(object):
             "Add a reference set to the data repo")
         addReferenceSetParser.set_defaults(runner="addReferenceSet")
         cls.addRepoArgument(addReferenceSetParser)
-        cls.addFilePathArgument(addReferenceSetParser)
+        cls.addFilePathArgument(
+            addReferenceSetParser,
+            "The path of the FASTA file to use as a reference set. This "
+            "file must be bgzipped and indexed.")
         cls.addNameOption(addReferenceSetParser, objectType)
         cls.addDescriptionOption(addReferenceSetParser)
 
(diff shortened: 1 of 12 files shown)


ga4gh/cli.py
```diff
@@ -1496,6 +1499,29 @@ def configtest_main(parser=None):
...
+    """
+    Returns the filename of the specified path without its extensions.
+    This is usually how we derive the default name for a given object.
+    """
+    if len(filePath) == 0:
+        raise ValueError("Cannot have empty path for name")
+    fileName = os.path.split(filePath)[1]
+    # We need to handle things like .fa.gz, so we can't use
+    # os.path.splitext
+    ret = fileName.split(".")[0]
```
macieksmuga: Except that this is exactly what causes server failure on fairly common cases where files have `.`s in their name. I'd propose instead something like:
```
def getNameFromPath(filePath, extensions=[]):
```
where `extensions` would be an array of possible extensions to expect, ex: `['.fa','.fa.gz']` and the function would _only_ strip out expected extensions, leaving any additional `.`s in the name intact.

diekhans:
The mapping of ontology names to full ontology terms is a
different issue than having a full support for ontologies.

The current ontologyTermMap functionality is only required in
the server because of variant annotations needed to translate
names in the VCF info column to full ontology terms at run time.
This could be done at load time, as is done with sequence
annotations, at the cost of pulling a lot of data out VCFs.

We will need to implement some kind of ontology-based queries,
which could be done with an external ontology lookup service
(with the refserver doing caching) or internally loading the
ontologies.

We need help on this from someone who has many more clues about
ontology software than I have.

ga4gh/cli.py
jeromekelleher: This is from the earlier PR @macieksmuga, can we continue the discussion over there?

jeromekelleher:
This is complicate @diekhans, I agree. Are you happy within this PR as it is though?

diekhans:
I haven't looked at the code, but it does the job we need
today, which is a great criteria for a +1.

My main point was really there is future complexity coming with
ontologies, so don't worry about this being incomplete.

jeromekelleher:
Excellent, thanks @diekhans.

jeromekelleher:
Can I get some review here please @dcolligan, @macieksmuga, @david4096? This should be easy to follow once #1104 is merged.

dcolligan:
@jeromekelleher could you rebase this so we can see only the changes that aren't in #1104 ?

jeromekelleher:
Done -- thanks for the heads up @dcolligan.

jeromekelleher:
Review please... @macieksmuga, @david4096, @dcolligan.

dcolligan:
+1

jeromekelleher:
OK, last chance to review here... Merging this tomorrow if I don't hear otherwise.

ga4gh/datamodel/ontologies.py
```diff
@@ -8,36 +8,45 @@
...
     def populateFromFile(self, dataUrl):
         """
         Populates this ontology map from the specified dataUrl.
+        This reads the ontology term name and ID pairs from the
+        specified file.
         """
         self._dataUrl = dataUrl
-        self.readOntology(self._dataUrl)
+        # TODO shouldn't sourceName be the name of the ontology, (e.g,
+        # sequence_ontology?)
```
david4096: Yes, I think so. This does need to be specified better, using the user provided name makes enough sense.

To do proper relational management we should probably give unique identifiers to the ontology maps. Perhaps we should provide a mechanism to specify the ontology map for a given variant set?

ga4gh/datarepo.py
```diff
@@ -332,6 +339,11 @@ def verify(self):
         # have verbosity levels. We should provide a way to configure
         # where we look at various chromosomes and so on. This will be
         # an important debug tool for administrators.
+        for ontologyTermMap in self.getOntologyTermMaps():
+            print(
+                "Verifying OntologyTermMap", ontologyTermMap.getLocalId(),
+                "@", ontologyTermMap.getDataUrl())
+            # TODO how do we verify this? Check some well-know SO terms?
```
david4096: I imagine we would fail somewhere earlier if the file is not formatted correctly. Knowing that for each key in the ontology term map there exists a term would probably be sufficient for now.

ga4gh/datarepo.py
```diff
@@ -400,33 +412,50 @@ def _readSystemTable(self, cursor):
         self._creationTimeStamp = config["creationTimeStamp"]
 
     def _createOntologyTable(self, cursor):
+        # Right now we support a crude ontology term name-ID bidirectional
+        # map. However, in the future we will want to have better ontology
+        # support. Therefore we're not making the SQL schema too specific
+        # so that we can make this transition without needing backwards
+        # incompatible schema changes.
         sql = """
```
david4096: We could safely store the sourceVersion here as well (I think).

ga4gh/datarepo.py
```diff
@@ -400,33 +412,50 @@ def _readSystemTable(self, cursor):
...
+            ontologyTermMap.populateFromRow(row)
+            self.addOntologyTermMap(ontologyTermMap)
+
+    def removeOntologyTermMap(self, ontologyTermMap):
+        """
+        Removes the specified ontology term map from this repository.
+        """
+        sql = "DELETE FROM Ontology WHERE name=?"
+        cursor = self._dbConnection.cursor()
+        cursor.execute(sql, (ontologyTermMap.getLocalId(),))
```
david4096: Could we do some foreign key integrity checking here like you've done for references and datasets? If so, you may consider using something other than name for the primary key.

ga4gh/datarepo.py
```diff
@@ -110,15 +110,17 @@ def getNumReferenceSets(self):
...
 
-    def getOntology(self, name):
+    def getOntologyTermMapByName(self, name):
         """
-        Returns an ontology map by name
+        Returns an ontologyTermMap by name
         """
-        return self._ontologyNameMap.get(name, None)
+        if name not in self._ontologyNameMap:
+            raise exceptions.OntologyNameNotFoundException(name)
```
david4096: In the variant annotation files I was working with, the transcript effect terms did not always map directly to the SO terms. Sometimes synonyms are used, or the developer adds terms that are not present. For that reason, we'll want to be sure this exception gets handled and doesn't throw to the user. When working directly with VCF there are not enough guarantees that the terms will map.

Forcing users to specify a specific SO version when adding a variant annotation set could help mitigate this, however, it is doubtful they will know unless they are the creators of the tool used to generate the annotations.

david4096:
+1 thanks for pushing this part of the code forward. The ontology term map stuff was done in a rush for variant annotation and this keeps it in parallel with the internal server changes without modifying it greatly.

ga4gh/datamodel/ontologies.py
jeromekelleher: Updated to remove the comments and make the sourceName == name. We definitely need the think how we handle ontologies through again, and allowing the user specify the map used is a good start. I think we should do this after merging into master though.

ga4gh/datarepo.py
jeromekelleher: Good idea. I think we should add this as a follow up issue.

ga4gh/datarepo.py
jeromekelleher: Where do we get the sourceVersion from though?

ga4gh/datarepo.py
jeromekelleher: Yeah, I hope we can. I think we have some basic questions we need to clarify first though, like, where exactly is the admin supposed to get this mapping file from?

ga4gh/datamodel/ontologies.py
diekhans: We need to require specifying which ontology to use a load time.
As you say, It might not be entirely obvious, however its hard
for us to guess.

ga4gh/datarepo.py
jeromekelleher: This exception is thrown when we can't find the OntologyTermMap itself (when, say, ``sequence_ontology`` isn't there), not when a term mapping failure occurs. Term mapping failures map silently to the empty string as before, which is probably all we can do (although, I think these failures should be logged; see #1113).

jeromekelleher:
Excellent, thanks for the review @david4096. I think what we have here is good enough to keep us going for now, but we definitely need to look at this ontology stuff again. I think it would be great if we could take the burden of creating term maps away from the admin. One idea would be to  build them by downloading directly from some authorative source and store the map directly in the DB, without the intermediate text file faffing.

ga4gh/datamodel/ontologies.py
diekhans: +100 on moving forward with this ontology map changes.  There
will be much more work with ontologies in the future.

ga4gh/datarepo.py
david4096: As far as I can tell there are release numbers [here](http://www.sequenceontology.org/browser/obob.cgi). Perhaps an argument when adding via repo manager? Let's avoid adding anything else to this custom filetype, like version headers.

ga4gh/datarepo.py
david4096: Yes, good point. We should at the very least offer the script that was used to generate it... @pcingola was the original contributor of this code, perhaps he has it sitting around somewhere.

ga4gh/datarepo.py
david4096: Thanks, I realized that but wanted to make sure this point was not lost. We may need to be able to work with multiple sequence ontology mappings.

jeromekelleher:
OK, merging this. Thanks again for the review @david4096, I'll open up issue(s) now for the dangling points.

Issue #718, opened 2015-10-11
Create interesting client API example touching all parts of the API


jeromekelleher:
We need an example of the API working across all parts of the data. We need to show how we can link up variants, references and reads with simple Python code. 

The API docs are [here](http://ga4gh-reference-implementation.readthedocs.org/en/develop/api.html).

The existing demo for using the [variants API](http://ga4gh-reference-implementation.readthedocs.org/en/develop/demo.html) is a good place to start.

david4096:
Some WIP towards fulfilling this, examples with the variant API.

https://github.com/david4096/bioapi-examples

jeromekelleher:
Looks good @david4096!

PR #1118, opened 2016-04-18
Update apache docs

### LINKED ISSUE #821: Update Apache installation documentation
Update installation docs for Apache to reflect @macieksmuga's recent experience on an instance running Ubuntu 14.04:

Following the instructions as written in the installation documentation at
http://ga4gh-reference-implementation.readthedocs.org/en/develop/installation.html

A couple of changes needed:

—pre tag didn’t work on pip install ga4gh. I performed the pip install without that tag, and it succeeded.

The Apache config code snippet given must be for Apache 2.4 or greater.
The Apache 2.2 that comes for Ubuntu 14.04 needed a couple of minor changes:

I needed to edit
/etc/apache2/sites-enabled/000-default

instead of the recommended
/etc/apache2/sites-enabled/000-default.conf

And while in that file, I needed to set permissions for the directory to
Allow from all

instead of:
Require all granted

(see http://stackoverflow.com/questions/21265191/apache-authtype-not-set-500-error)

david4096:
Minor changes:

* Editing the sites-available and then using apache2ctl to move from sites-available to sites-enabled ([more](http://serverfault.com/questions/83508/purpose-of-debian-sites-available-and-sites-enabled-dirs)). 
* The [apache2ctl](https://httpd.apache.org/docs/current/programs/apachectl.html) tool is a better interface for working with apache than service. Removed the line that used links, this is confusing for many as links is not installed by default. 
* Removed the TODOs, one had been done and the other was too broad to make sense for this document. The general gist, "more example client code," has been captured [elsewhere](https://github.com/ga4gh/server/issues/718).
* Added installation of libcurl headers to apt-get

Aspiring to close #821

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -22,7 +22,7 @@ First, we install some basic pre-requisite packages:
 
 .. code-block:: bash
 
-  $ sudo apt-get install python-dev python-virtualenv zlib1g-dev libxslt1-dev libffi-dev libssl-dev
+  $ sudo apt-get install python-dev python-virtualenv zlib1g-dev libxslt1-dev libffi-dev libssl-dev libcurl4-openssl-dev
 
 Install Apache and mod_wsgi, and enable mod_wsgi:
 
@@ -55,7 +55,7 @@ Make a virtualenv, and install the ga4gh package:
 
   $ virtualenv ga4gh-server-env
   $ source ga4gh-server-env/bin/activate
-  (ga4gh-server-env) $ pip install --pre ga4gh  # We need the --pre because ga4gh is pre-release
+  (ga4gh-server-env) $ pip install ga4gh
   (ga4gh-server-env) $ deactivate
 
 Download and unpack the example data:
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -105,18 +105,21 @@ Restart Apache:
 
 .. code-block:: bash
 
-  $ sudo service apache2 restart
+  $ sudo apache2ctl restart
```
jeromekelleher: Why use ``apache2ctl`` rather than ``service``? I think ``service`` is the recommended way of interacting with system daemons on Debian (and it uses apache2ctl under the hood). Unless there's a good reason to bypass this, I think we should leave it as ``service``.

jeromekelleher:
Looks good, thanks @david4096. +1 modulo query about service/apache2ctl

david4096:
Got it @jeromekelleher, rolled that one back!

jeromekelleher:
Thanks @david4096, merging.

Issue #1109, opened 2016-04-18
Port FeatureSet functionality to SQL repo


jeromekelleher:
Update the SQL repo branch to include  add/remove/verify functionality for FeatureSets. This should follow the testing pattern given in #1104.

dcolligan:
Fixed in #1125

PR #1125, opened 2016-04-20
Add feature set add / delete to repo manager

dcolligan:
Issue #1109

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -10,7 +10,7 @@ from __future__ import unicode_literals
 import argparse
 import logging
 import operator
-import os.path
+import os
 import sys
 import unittest
 import unittest.loader
@@ -30,6 +30,7 @@ import ga4gh.datarepo as datarepo
 import ga4gh.protocol as protocol
 import ga4gh.datamodel.reads as reads
 import ga4gh.datamodel.references as references
+import ga4gh.datamodel.sequenceAnnotations as sequenceAnnotations
 import ga4gh.datamodel.datasets as datasets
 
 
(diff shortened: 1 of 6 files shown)


jeromekelleher:
Looks good @dcolligan, thanks. I can't figure out how this relates to PRs #1104 and #1107 though... what have you done here?

Verify functionality would be good here. Just iterate over each FeatureSet and pull a few features out of it to confirm it's all good.

dcolligan:
Doesn't relate at all to #1107

Pulls a few functions out of #1104

jeromekelleher:
We really need to merge #1104 ASAP...

dcolligan:
I agree, but it's not my +1 that you're waiting for!

dcolligan:
@jeromekelleher updated

ga4gh/cli.py
```diff
@@ -1531,6 +1532,15 @@ def __init__(self, args):
         self._repoPath = args.repoPath
         self._repo = datarepo.SqlDataRepository(self._repoPath)
 
+    def _getDatasetFromArgs(self):
+        try:
```
jeromekelleher: This isn't necessary --- the exception should be propagated up automatically and caught at the CLI. This should be working nicely for (say) ReadGroupSets in the current sql repo code.

ga4gh/cli.py
```diff
@@ -1682,6 +1692,33 @@ def func():
...
+        """
+        Adds a new feature set into this repo
+        """
+        self._openRepo()
+        dataset = self._getDatasetFromArgs()
+        name = getNameFromPath(self._args.filePath)
+        featureSet = sequenceAnnotations.Gff3DbFeatureSet(
+            dataset, name)
+        referenceSet = self._repo.getReferenceSetByName(
+            self._args.referenceSetName)
```
jeromekelleher: What happens when ``referenceSetName`` is None? Have we any hope of guessing the correct reference set from a given FeatureSet? We should output an explicit error message here anyway.

jeromekelleher:
Looks good @dcolligan, some minor comments above. We should make sure we error out nicely in all possible situations now; this is really important for a good user experience when dealing with confusing and difficult data.

dcolligan:
@jeromekelleher updated

ga4gh/datarepo.py
```diff
@@ -345,6 +345,14 @@ def verify(self):
                     reference.getLocalId())
         for dataset in self.getDatasets():
             print("Verifying Dataset", dataset.getLocalId())
+            reference = self.getReferenceSets()[0].getReferences()[0]
```
jeromekelleher: This is broken --- you need to get the referenceSet for each of the FeatureSets. No reason to think that they'd all be the same.

jeromekelleher:
Looks good, +1 after the reference issue in ``verify`` is fixed.

dcolligan:
@jeromekelleher updated

jeromekelleher:
Thanks, merging.

Issue #1116, opened 2016-04-18
Versioning scheme for sql data repo


dcolligan:
Right now we are just inserting a literal arbitrary value into the `System` table.  We need to come up with a better scheme than this.

macieksmuga:
I'm not sure what you're thinking here @dcolligan - Having a literal schema version satisfies the need for a migration path as the schema evolves (application checks schema version in the DB against the value in its codebase, performs any needed actions on DB to update it). Are there any other requirements here that need to be satisfied?

jeromekelleher:
I agree with @macieksmuga  here. Initially I was thinking just a simple ``major.minor`` version scheme using semantic versioning: we increment ``minor`` each time we make a non-breaking change, and ``major`` each time we make a breaking change. What do you think @dcolligan?

dcolligan:
If it's as easy as that and adding an `if` to make sure the version is correct, fine... just as long as we have something!

jeromekelleher:
Sounds like you'd like to implement this @dcolligan!

dcolligan:
Fixed in #1128

PR #1128, opened 2016-04-21
Add version check to data repo

dcolligan:
Issue #1116

### DIFF
--- a/ga4gh/datarepo.py
+++ b/ga4gh/datarepo.py
@@ -267,6 +267,23 @@ class SqlDataRepository(AbstractDataRepository):
     """
     A data repository based on a SQL database.
     """
+    class SchemaVersion(object):
+        """
+        The version of the data repository SQL schema
+        """
+        def __init__(self, versionString):
+            splits = versionString.split('.')
+            assert len(splits) == 2
+            self.major = splits[0]
+            self.minor = splits[1]
+
+        def __str__(self):
+            return "{}.{}".format(self.major, self.minor)
+
+    version = SchemaVersion("0.1")
+    systemKeySchemaVersion = "schemaVersion"
+    systemKeyCreationTimeStamp = "creationTimeStamp"
+
     def __init__(self, fileName):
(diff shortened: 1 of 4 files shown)


ga4gh/datarepo.py
```diff
@@ -385,6 +388,11 @@ def _readSystemTable(self, cursor):
         row = cursor.fetchone()
         self._repositoryVersion = config["repositoryVersion"]
         self._creationTimeStamp = config["creationTimeStamp"]
+        if self._repositoryVersion != self.version:
+            raise exceptions.ConfigurationException(
```
jeromekelleher: Is this caught by the CLI, or does it fall through as a stack trace? We should aim to give friendly error messages from now on, and only dump a stack trace in a genuine bug situation.

ga4gh/datarepo.py
```diff
@@ -385,6 +388,11 @@ def _readSystemTable(self, cursor):
         row = cursor.fetchone()
         self._repositoryVersion = config["repositoryVersion"]
```
jeromekelleher: I was thinking about this, perhaps it should be ``schemaVersion``? Repository version seems to imply that we're somehow tracking all the data that's going in and out. What do you think?

ga4gh/datarepo.py
```diff
@@ -385,6 +388,11 @@ def _readSystemTable(self, cursor):
...
         self._creationTimeStamp = config["creationTimeStamp"]
+        if self._repositoryVersion != self.version:
```
jeromekelleher: We may as well do semver checking here from the start. If minor != current minor, fine. If major != current major, throw exception.

ga4gh/datarepo.py
diekhans: yes to schemaVersion

dcolligan:
@jeromekelleher updated

jeromekelleher:
Looks good, +1.

jeromekelleher:
Can you rebase and merge this please @dcolligan?

PR #1133, opened 2016-04-21
Rna api feature group search fix

macieksmuga:
Fixes typos in implementation of `featuregroup/search`


saupchurch:
good catch - thanks! +1

Issue #1136, opened 2016-04-21
add robot.txt to production server files


kozbo:
Place the file next to the application.wsgi file in the root of the server. 

Suggested content is

```
User-agent: *
Disallow: /
User-agent: Googlebot
Disallow:
```

diekhans:
Is the googlebot entry for allowing indexing of static pages?
We want to make sure it doesn't hit the API end points.

kozbo:
@diekhans Yes, the googlebot entry will only index static pages, it will not hit the API endpoints.

dcolligan:
Fixed in 1136_robots_txt

PR #1137, opened 2016-04-22
Add robots.txt

dcolligan:
Issue #1136

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -478,6 +478,13 @@ def index():
     return flask.render_template('index.html', info=app.serverStatus)
 
 
+@app.route('/favicon.ico')
+@app.route('/robots.txt')
+def robots():
+    return flask.send_from_directory(
+        app.static_folder, flask.request.path[1:])
+
+
 @DisplayedRoute('/references/<id>')
 def getReference(id):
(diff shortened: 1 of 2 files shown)


kozbo:
You beat me to it @dcolligan :-) Were you able to verify that my configuration file allowed the googlebot through?  I found this test tool that I was using to verify access: https://support.google.com/webmasters/answer/6062598?hl=en

dcolligan:
@kozbo that tool only works on publicly-accessible robots.txt files

kozbo:
@dcolligan ok. Have you found a way to verify that the robots.txt file is doing what it is supposed to do?

david4096:
I verified that it serves the robot.txt from the expected location. I think we want to remove the lines about GoogleBot, the disallow all User agents is sufficient. +1

dcolligan:
@kozbo  I think @david4096 is right here... someone that deploys the server probably doesn't want google indexing it ... the syntax in this latest updates disallows all robots

http://www.robotstxt.org/robotstxt.html

dcolligan:
@kozbo can I get your +1 here?

kozbo:
yes, @dcolligan, sorry for the delay. Looks good, +1

PR #1143, opened 2016-04-25
fixes declaration of threshold

saupchurch:
fixes a problem where the default value wasn't being used by the server when sending JSON directly using curl


david4096:
Ah, I see, the default value needs to map to the default type. Neat feature!

@dcolligan this schema change is not passing on schemas. I think we might need to change the test. Where did the "schema union assumptions" come from?

https://github.com/ga4gh/schemas/blame/rna/tests/test_protocol.py

macieksmuga:
Out of curiosity, why not just 
```
float threshold = 0.0
```
i.e. What's the reason for an optional null here? If truly necessary, how will it look in protobuf (which I believe doesn't support such unions with `null`)?

saupchurch:
`float threshold = 0.0` was the first thing I tried.  Unless I did something wrong, when doing that threshold was still required to be in the POST JSON.  I can check it again to be sure.

macieksmuga:
leaving aside the schemas issue as pointed out above, this works. +1 and merging.

dcolligan:
@david4096 the assumption is that for nullable fields the default is null.  I forget where it came from... might be an avro thing; might be a enforced convention of our schema?

Issue #1141, opened 2016-04-25
Opening non-existent DB results in stack trace


jeromekelleher:
If I try to read a non-existent repo DB I get a ``sqlite3.OperationalError: no such table: System`` rather than a sensible error message. We should at least test for existence of the file and throw a good exception here.

dcolligan:
Fixed in #1148

PR #1148, opened 2016-04-26
Throw exception if DB does not exist

dcolligan:
Issue #1141

### DIFF
--- a/ga4gh/datarepo.py
+++ b/ga4gh/datarepo.py
@@ -319,7 +319,9 @@ class SqlDataRepository(AbstractDataRepository):
                 MODE_READ, MODE_WRITE)
             raise ValueError(error)
         self._openMode = mode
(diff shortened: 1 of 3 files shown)


jeromekelleher:
Looks good. Some stuff we need:

1. Test cases in the CLI. Verify that the correct exception (RepoNotFoundException?) is thrown when we pass a bad argument to the various commands.
2. Should we check the DB schema here as well? We should try to load (a) some text file, and throw a BadRepoFormatException(),  (b) an empty SQLite DB, (c) a directory. We want to make sure that a sensible error message is sent back to the user for every possible class of stupid input here.

dcolligan:
@jeromekelleher updated

jeromekelleher:
Awesome, +1.

jeromekelleher:
Can you rebase and merge please @dcolligan?

PR #1153, opened 2016-04-27
Update sql repo

jeromekelleher:
This brings the sql repo up to date with changes in master.

@dcolligan, this was a fairly straightforward merge, so would you mind having a quick scan and merging if you think it's OK? I'm keen to get us to the point where we can merge into master, so getting ourselves up to date is important.


dcolligan:
+1

PR #1156, opened 2016-04-27
Updated download example data script.

jeromekelleher:
Minor mods to example data script to use import glue. Confirm that this produces working example data.

### DIFF
--- a/scripts/download_example_data.py
+++ b/scripts/download_example_data.py
@@ -17,11 +17,15 @@ import urllib2
 
 import pysam
 
-import ga4gh.datarepo as datarepo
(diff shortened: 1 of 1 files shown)


jeromekelleher:
I've used this script to create ga4gh-example-data-v4.0.tar, which I uploaded to the usual location.

We'll want to add the Features and VariantAnnotation data to this soon, but we should do this by updating the generation script, not by updating the tarball manually.

dcolligan:
+1

david4096:
+1 downloaded using the script and ran the variants and reads through a couple of queries.

PR #1158, opened 2016-04-27
Updated documentation for SQL repo.

jeromekelleher:
- Removed inaccurate information about file hierarchy
- Updated links to ga4gh-example-data.

We'll want to make a much better job of documenting the repo manager soon (with a nice tutorial on how to set up a repo). This change basically means the information is not inaccurate, although it is quite incomplete.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -5,214 +5,39 @@ Configuration
 *************
 
 The GA4GH reference server has two basic elements to its configuration:
-the `Data repository`_ and the `Configuration file`_.  The repository is most easily configured via the `Repository manager`_ command line tool.
+the `Data repository`_ and the `Configuration file`_.
 
 ---------------
 Data repository
 ---------------
 
-Data is input to the GA4GH server as a directory hierarchy, in which
-the structure of data to be served is represented by the file system.
-At the top level of the data hierarchy there are two required
-directories to hold the top level container types: ``referenceSets`` and
-``datasets``.
+The repository in the GA4GH reference server defines how your data is organised. The
+repository itself is a SQLite database, which contains information about your
+datasets, reference sets and so on. Bulk data (such as variants and reads)
+is not stored in database, but instead accessed directly from the primary
+data files at run time. The locations of these data files is entirely up
+to the administrator.
 
-.. todo:: We need to link to the high-level API documentation for descriptions
-    of what the various objects here mean.
-
-+++++++++++++
-ReferenceSets
-+++++++++++++
-
-Within the data directory there must be a directory called ``referenceSets``.
-Within this directory, each directory is interpreted as containing a
-``ReferenceSet`` with the directory name mapped to the name of the
-reference set.  Here is an example of how reference data should be arranged::
-
-    references/
-        GRCh37.json
-        GRCh37/
-            1.fa.gz
-            1.fa.gz.fai
-            1.json
-            2.fa.gz
-            2.fa.gz.fai
-            2.json
-            # More references
-        GRCh38.json
-        GRCh38/
-            1.fa.gz
-            1.fa.gz.fai
-            1.json
-            2.fa.gz
-            2.fa.gz.fai
-            2.json
-            # More references
-
-In this example we have two reference sets, with names ``GRCh37`` and ``GRCh38``.
-Each reference set directory must be accompanied by a file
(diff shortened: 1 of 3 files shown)


dcolligan:
+1

docs/installation.rst
```diff
@@ -62,8 +62,8 @@ Download and unpack the example data:
 
 .. code-block:: bash
 
-  $ wget https://github.com/ga4gh/server/releases/download/data/ga4gh-example-data-v3.2.tar
-  $ tar -xf ga4gh-example-data-v3.2.tar
+  $ wget https://github.com/ga4gh/server/releases/download/data/ga4gh-example-data-v4.0.tar
```
david4096: woooh release point!

docs/configuration.rst
```diff
@@ -357,12 +172,12 @@ The GA4GH reference server is a `Flask application <http://flask.pocoo.org/>`_
...
 <http://flask.pocoo.org/docs/0.10/config/>`_.
 Many configuration files will be very simple, and will consist of just
-one directive instructing the server where to look for data; for
+one directive instructing the server where to find the data repository;
 example, we might have
 
 .. code-block:: python
 
-    DATA_SOURCE = "/path/to/data/root"
+    DATA_SOURCE = "/path/to/ga4gh-repo.db"
```
david4096: I think the server needs read and write access to the directory since it writes .shm and .wal files. https://www.sqlite.org/tempfiles.html

docs/configuration.rst
```diff
@@ -284,10 +98,10 @@ Removes a given reference set from a given data repository.
     $ ga4gh_repo remove-referenceset path/to/datarepo aReferenceSet
 
 ++++++++++++++++
-add-ontologymap
+add-ontology
```
david4096: Not sure where we're at on the naming of this, but this looks like a roll back?

docs/configuration.rst
```diff
@@ -5,222 +5,47 @@ Configuration
...
+verify
 +++++++
 
 Performs some consistency checks on the given data repository to ensure it is
 well-formed.
 
 .. code-block:: bash
 
-    $ ga4gh_repo check path/to/datarepo
+    $ ga4gh_repo verify path/to/datarepo
```
david4096: Does this need to point to the .db file or a directory?

docs/configuration.rst
jeromekelleher: Excellent spot @david4096, opening an issue on this.

docs/configuration.rst
jeromekelleher: Yeah, I've changed it to ``ontology`` at this level, in an attempt to future-proof the command a bit. See #1117.

docs/configuration.rst
jeromekelleher: It's always a .db file. Does this make it look like it should be a directory? I didn't bother changing this much as it's going to need a thorough overhaul to make it properly user friendly. I'm happy to change the path placeholder if you have a replacement suggestion though.

jeromekelleher: Yeah, this is confusing. I've change it to ``/path/to/repo.db``; what do you think?

docs/configuration.rst
diekhans: +1 if we are adding a true ontology, not just the mappings

docs/configuration.rst
```diff
@@ -5,222 +5,47 @@ Configuration
...
-+++++
-Reads
-+++++
-
-A dataset can contain many ReadGroupSets, and each ReadGroupSet contains
-a number of ReadGroups. The ``reads`` directory contains a number of BAM
-files, each of which corresponds to a single ReadGroupSet. ReadGroups are
-then mapped to the ReadGroups that we find within the BAM file.
+The repository in the GA4GH reference defines how your data is organised. The
+repository itself is a SQLite database, which contains information about your
```
macieksmuga: This is no longer a repository, but a registry. By continuing to use this term, we're just baking in the potential for misunderstanding.

jeromekelleher: We can't use the word "registry" @macieksmuga, it has been forever tainted by it's association with Windows! We can have a discussion about renaming if you like, but we should be consistent with how we describe things. Repository is what we use everywhere else, so we should also use it here in the docs.

PR #1163, opened 2016-04-27
Decode utf8 in landing message html

### LINKED ISSUE #1121: Landing message utf8 bug
When trying to parse a unicode byte in the landing message HTML flask's template rendering fails. We should coerce the file to ascii before placing in the template. Python version 2.7.6

david4096:
Fixes a bug that occurs when the landing message HTML contains unicode characters by adding `.decode('utf-8')`.

Closes #1121 

http://stackoverflow.com/questions/24566538/flask-rendering-unicode-characters-in-template


macieksmuga:
+1 tested & works as advertised on my system.

jeromekelleher:
+1.

PR #1165, opened 2016-04-28
Fix for issue 1164, ensures isDerived is boolean.

jeromekelleher:
This fixes the compliance issue, and ensures that isDerived is the correct type.

Does not fix the underlying problem that this was not detected by the unit tests.


dcolligan:
+1 rebase and merge

Issue #1150, opened 2016-04-26
add-referenceset in repo manager needs supplied metadata fields


macieksmuga:
When I issue a request to add a reference set:
```
python repo_dev.py add-referenceset -n hg37 ga4gh-example-data/repo.db ../maciek-ga4gh-compliance/test-data/hg37.fa
```
The resulting entity is missing several important fields:
```
{"name": "hg37", "sourceURI": null, "assemblyId": null, "sourceAccessions": [], "ncbiTaxonId": null, "isDerived": 0, "id": "WyJoZzM3Il0", "md5checksum": "12827085103f42e97428dc05a7f26fec", "description": ""}
```
most importantly the `assemblyId`, which can be searched by.

diekhans:
The schema does allow these fields to be omitted.  is there no way to set them
with a metadata json file?

jeromekelleher:
We need to add CLI options for these metadata fields. It's a 10 minute job.

dcolligan:
Fixed in #1184

Issue #1149, opened 2016-04-26
FeatureSet record has empty ReferenceSet field


macieksmuga:
The current implementation of `FeatureSet` is not properly initialized with a valid `ReferenceSetId`, and the corresponding field in responses to `featuresets/search` and `featuresets/<id>` is currently left blank.

jeromekelleher:
This is resolved in the sql repo branch.

macieksmuga:
Not sure I see how - I've been testing the `jeromekelleher/remove-filesystemdataset` branch today (which I believe is the latest SQL repo manager codebase), and `addFeatureSet` (as well as `removeFeatureSet`) is not implemented in the `RepoManager` object there yet. Unless there's some other place I should be looking, I'll supply the needed code.

jeromekelleher:
It looks like something weird has happened with that branch @macieksmuga, sorry about that. It's turned into a bit of a mess trying to keep all these PRs going at the same time. Here's what happens with the code in the ``sql_repo`` branch now:
```
jk@holly$ python repo_dev.py init test.db  -f
jk@holly$ python repo_dev.py add-dataset test.db ds
jk@holly$ python repo_dev.py add-featureset test.db ds tests/data/datasets/dataset1/sequenceAnnotations/gencodeV21Set1.db
repo_dev.py: error: a reference set name must be provided
jk@holly$ python repo_dev.py add-featureset test.db ds tests/data/datasets/dataset1/sequenceAnnotations/gencodeV21Set1.db -R NCBI37
repo_dev.py: error: ReferenceSet with name 'NCBI37' not found
jk@holly$ python repo_dev.py add-referenceset test.db tests/data/referenceSets/NCBI37.fa.gz
jk@holly$ python repo_dev.py add-featureset test.db ds tests/data/datasets/dataset1/sequenceAnnotations/gencodeV21Set1.db -R NCBI37
```

jeromekelleher:
Closed by #1166.

Issue #1127, opened 2016-04-20
Repo manager move mode cross-device error


david4096:
I received this error when trying to move a file from one drive to another. GA4GH is installed on /srv, while the file is in /run/shm.

```
(newserverenv)ga4gh-admin@ga4gh-demo:/srv/ga4gh$ ga4gh_repo add-featureset ga4gh-example-data demo /run/shm/1000Genomes/gencode_v24lift37.db --moveMode=move
Traceback (most recent call last):
  File "/srv/ga4gh/server/newserverenv/bin/ga4gh_repo", line 9, in <module>
    load_entry_point('ga4gh==0.2.3.dev4+nge0b07f3', 'console_scripts', 'ga4gh_repo')()
  File "/srv/ga4gh/server/newserverenv/local/lib/python2.7/site-packages/ga4gh-0.2.3.dev4+nge0b07f3-py2.7.egg/ga4gh/cli.py", line 1953, in repo_main
    runner.run()
  File "/srv/ga4gh/server/newserverenv/local/lib/python2.7/site-packages/ga4gh-0.2.3.dev4+nge0b07f3-py2.7.egg/ga4gh/cli.py", line 1728, in run
    self.datasetName, self.filePath, self.moveMode)
  File "/srv/ga4gh/server/newserverenv/local/lib/python2.7/site-packages/ga4gh-0.2.3.dev4+nge0b07f3-py2.7.egg/ga4gh/repo_manager.py", line 526, in addFeatureSet
    self._moveFile(filePath, destPath, moveMode)
  File "/srv/ga4gh/server/newserverenv/local/lib/python2.7/site-packages/ga4gh-0.2.3.dev4+nge0b07f3-py2.7.egg/ga4gh/repo_manager.py", line 200, in _moveFile
    os.rename(src, dst)
OSError: [Errno 18] Invalid cross-device link
```

jeromekelleher:
Solved in sql repo --- we no longer create symlinks or move files.

jeromekelleher:
Closed by #1166.

Issue #1120, opened 2016-04-19
Repo manager not ingesting references and reference sets correctly


macieksmuga:
The repo manager currently seems to expect a single FASTA file as input for its `add-referenceset` command, and creates a corresponding full reference set for it with a single JSON file for the first record encountered. This is incorrect regardless of how many records the FASTA file contains.

Since ingesting novel reference sets is not at present a critical function of the server (most of the time, we'll just be using copies of pre-ingested GRCh37 or 38 reference sets), I propose we solve this fully in the upcoming repo manager SQL refactor #1104, along with a proper solution to #955

diekhans:
Are the servers we are currently deploying using split FASTAs?

Does this impact compliance?

macieksmuga:
@diekhans: Yes, and no.
😄

david4096:
If this is going to be fixed in a future version we should list it as a known issue in our documentation somewhere.

jeromekelleher:
This is solved in the current sql repo branch.

jeromekelleher:
Closed by #1166.

PR #1166, opened 2016-04-28
Sql repo

jeromekelleher:
This PR changes master to use the sql repo, and removes support for file system based data repositories. 

Issues closed by this PR:

- #1149, #1030, #689, #751 Mandatory reference set link enforced by foreign key constraints.
- #1127 Files no longer copied/moved
- #1120   #955, Reference ingestion code replaced, now required a single FASTA and no JSON.
- #1087  Object names are now all optional arguments, so that any bad guesses can be remedied by the administrator. However, there is certainly room for improvement in the guessing algorithm.
- #1064 No external programs are now called.
- #1052 No new files are created.
- #1047 No longer directly relevant, but not all metadata values are currently settable via the CLI. See for example #1150.
- #1059 No longer relevant, as check has been changed to ``verify`` and rewritten. However, dealing with inconsistent VCF files passed as part of a variant set is something we need to tackle.
-  #1049 No longer relevant. Reference import should be much improved now.
- #1045 #970 Full support for URL backed ReadGroupSets and VariantSets.
- #1038 No longer relevant.
- #1019 No longer relevant. We should provide a HOWTO for importing references as part of our repository administration tutorial though.

This is a large change affecting all developers and users, so please review and vote.


macieksmuga:
-1 
This is just a temporary hold: I'm wildly in favor of this PR going in as soon as possible (with many thanks and kudos to @jeromekelleher, @dcolligan, @david4096  and everybody else who helped this come to be so quickly). I'll lift it the moment the compliance-related issues for this milestone are resolved, which shouldn't take long.

dcolligan:
@macieksmuga could you link the issues you are waiting to be resolved before you lift your block?  I agree we want to get this PR in to master as soon as possible.

jeromekelleher:
Thanks @macieksmuga, totally agree with resolving compliance issues before merging to master.

I'm checking out for a few days; have fun!

macieksmuga:
@dcolligan I've been marking all the relevant issues with the `sql data Repo` milestone. I may find a couple of further ones by the end of the day.

docs/configuration.rst
```diff
@@ -5,222 +5,47 @@ Configuration
...
-
-+++++
-Reads
-+++++
-
-A dataset can contain many ReadGroupSets, and each ReadGroupSet contains
-a number of ReadGroups. The ``reads`` directory contains a number of BAM
-files, each of which corresponds to a single ReadGroupSet. ReadGroups are
-then mapped to the ReadGroups that we find within the BAM file.
+The repository in the GA4GH reference server defines how your data is organised. The
```
david4096: Typo: data "are" organized.

docs/configuration.rst
```diff
@@ -5,222 +5,47 @@ Configuration
...
+repository itself is a SQLite database, which contains information about your
+datasets, reference sets and so on. Bulk data (such as variants and reads)
+is not stored in database, but instead accessed directly from the primary
+data files at run time. The locations of these data files is entirely up
+to the administrator.
```
david4096: ...but in some cases still requires write access.

docs/configuration.rst
```diff
@@ -5,222 +5,47 @@ Configuration
...
-    $ ga4gh_repo init path/to/datarepo
+    $ ga4gh_repo init path/to/repo.db
+
 
 +++++++
-check
+verify
 +++++++
 
 Performs some consistency checks on the given data repository to ensure it is
```
david4096: We should be more specific about what this is and why one would run it.

docs/configuration.rst
```diff
@@ -260,18 +75,17 @@ Destroys a dataset in the given repository with a given name.
 
 .. code-block:: bash
 
-    $ ga4gh_repo remove-dataset path/to/datarepo aDataset
+    $ ga4gh_repo remove-dataset path/to/repo.db aDataset
```
david4096: The optional arguments should be documented as well.

docs/configuration.rst
```diff
@@ -281,20 +95,20 @@ Removes a given reference set from a given data repository.
 
 .. code-block:: bash
 
-    $ ga4gh_repo remove-referenceset path/to/datarepo aReferenceSet
+    $ ga4gh_repo remove-referenceset path/to/repo.db aReferenceSet
 
 ++++++++++++++++
-add-ontologymap
+add-ontology
```
david4096: This has gone back and forth once or twice, just making sure we land where want to be "map" or not "map".

docs/configuration.rst
```diff
@@ -281,20 +95,20 @@ Removes a given reference set from a given data repository.
...
-Adds an Ontology Map, which maps identifiers to ontology terms, to 
+Adds an Ontology Map, which maps identifiers to ontology terms, to
 the repository. Ontology maps are tab delimited files with an
 identifier/term pair per row.
 
 
 .. code-block:: bash
 
-    $ ga4gh_repo add-ontologymap path/to/datarepo path/to/aOntoMap.txt
+    $ ga4gh_repo add-ontologymap path/to/repo.db path/to/aOntoMap.txt
```
david4096: `add-ontology`

docs/configuration.rst
```diff
@@ -326,18 +140,19 @@ Removes a read group set from a given data repository and dataset.
...
+    $ ga4gh_repo remove-readgroupset path/to/repo.db aDataset aReadGroupSet
 
 +++++++++++++++
 add-variantset
 +++++++++++++++
 
 Adds a variant set directory to a given data repository and dataset.  The
-directory should contain file(s) with extension ``.vcf.gz``. If a variant set is annotated it will be added as both a variant set and a variant annotation set.
+directory should contain file(s) with extension ``.vcf.gz``. If a variant set
+is annotated it will be added as both a variant set and a variant annotation set.
```
david4096: Are we providing indexes for them?

ga4gh/backend.py
```diff
@@ -496,15 +496,10 @@ def variantAnnotationSetsGenerator(self, request):
...
-        results = []
-        for annset in dataset.getVariantAnnotationSets():
-            try:
-                variantSetId = request.variantSetId
-            except ValueError:
-                variantSetId = ""
-            if str(annset._variantSetId) == str(variantSetId):
-                results.append(annset)
-        return self._objectListGenerator(request, results)
+        variantSet = dataset.getVariantSet(request.variantSetId)
```
david4096: I'm pretty sure this isn't correct to the schemas since the annotation set and the variant set are not necessarily in the same dataset. Although the implementation does place them in the same dataset currently, this was due to being backed by files. Now we shouldn't have to make the same assumptions when relating variant and annotation sets.

I would discourage writing code that solidifies the assumption. The SQL repo gives us a nice path for setting foreign keys, we can take advantage of it by removing the relational assumptions from our python classes. This makes separation of concerns much more clear and allows each protocol to interact with the other's via a simple API.

To fulfill a searchVariantAnnotationSets request we will find all annotationSets that have the given variantSetId, as opposed to the variantSet that has had these annotation sets added to them. This means we only look at one table when responding to the request and can handle meta-dataset searches properly. The SQL backing moves us towards better satisfying the schemas.

In general, I believe we are complicating our picture by maintaining the class-based container model. Being able to reference objects directly in a table will allow us to remove this complexity. Am I wrong to assume `datasets.getVariantSet` will be translated to SQL call? I leave this message here in hopes that we will maintain a single data structure in SQL as opposed to a tree of classes, and I know that some of these are stop-gap measures.

I realize some of the changes to the variants datamodel needed to be made to add in the values that were being populated by the metadata JSON. Could you provide justification for the other refactors, could they be moved into a separate PR? What do we gain?

ga4gh/cli.py
```diff
@@ -1510,448 +1521,595 @@ def configtest_main(parser=None):
...
+    Returns the filename of the specified path without its extensions.
+    This is usually how we derive the default name for a given object.
+    """
+    if len(filePath) == 0:
+        raise ValueError("Cannot have empty path for name")
+    fileName = os.path.split(os.path.normpath(filePath))[1]
+    # We need to handle things like .fa.gz, so we can't use
+    # os.path.splitext
+    ret = fileName.split(".")[0]
+    assert ret != ""
```
david4096: Are there any characters that will cause us problems downstream? It looks like SQLite stores unicode strings alright, but what about our ID scheme?

ga4gh/cli.py
```diff
@@ -1510,448 +1521,595 @@ def configtest_main(parser=None):
...
                 print("Aborted")
 
+    def _updateRepo(self, func, *args, **kwargs):
+        """
+        Runs the specified function that updates the repo with the specified
+        arguments. This method ensures that all updates are transactional,
+        so that if any part of the update fails no changes are made to the
+        repo.
+        """
+        # TODO how do we make this properly transactional?
```
david4096: It's a database, so it has mechanisms in place that make each step transactional and depending on journaling mode, "rollback-able".

If you wanted to be extra safe, add a key to the system table that gets flipped during open and end of commit.

diekhans: David Steinberg <notifications@github.com> writes:
> If you wanted to be extra safe, add a key to the system table that gets flipped
> during open and end of commit.

Trust your database or use another one...

ga4gh/datarepo.py
```diff
@@ -179,42 +262,826 @@ def __init__(
...
+        def __str__(self):
+            return "{}.{}".format(self.major, self.minor)
+
+    version = SchemaVersion("0.1")
+    systemKeySchemaVersion = "schemaVersion"
+    systemKeyCreationTimeStamp = "creationTimeStamp"
+
+    def __init__(self, fileName):
+        super(SqlDataRepository, self).__init__()
+        self._dbFilename = fileName
```
david4096: Does it make sense to enforce, or add a `.db` extension?

`repo init repo`, `repo init repo.db`

ga4gh/cli.py
```diff
@@ -1510,448 +1521,595 @@ def configtest_main(parser=None):
...
+        self._repo.open(datarepo.MODE_WRITE)
+        try:
+            func(*args, **kwargs)
+            self._repo.commit()
```
david4096: It might be nice to return the result of the operation. There is no feedback when a referenceset or readgroupset is added.

ga4gh/cli.py
```diff
@@ -1510,448 +1521,595 @@ def configtest_main(parser=None):
...
+        self._repo.verify()
 
-class RemoveReferenceSetRunner(AbstractRepoCommandRunner):
+    def addOntology(self):
+        """
+        Adds a new Ontology to this repo.
+        """
+        self._openRepo()
+        name = self._args.name
+        if name is None:
```
david4096: For all the types it appears there is no catch for blank name. If it's intended to allow for this it's fine, but I noticed there are checks for empty string in the getNameFromPath so I wanted to make sure this is the desired functionality.

`python repo_dev.py add-readgroupset l a ga4gh-example-data/HG00096.bam --name ""`

ga4gh/datarepo.py
```diff
@@ -179,42 +262,826 @@ def __init__(
...
+            raise exceptions.DuplicateNameException(dataset.getLocalId())
+
+    def removeDataset(self, dataset):
+        """
+        Removes the specified dataset from this repository. This performs
+        a cascading removal of all items within this dataset.
+        """
+        sql = "DELETE FROM Dataset WHERE id=?"
+        cursor = self._dbConnection.cursor()
+        cursor.execute(sql, (dataset.getId(),))
```
david4096: Would be great if we signalled to the user success or failure of the operation.

ga4gh/cli.py
```diff
@@ -1510,448 +1521,595 @@ def configtest_main(parser=None):
...
+                indexFile = dataUrl + ".bai"
+        name = self._args.name
+        if self._args.name is None:
+            name = getNameFromPath(dataUrl)
+        readGroupSet = reads.HtslibReadGroupSet(dataset, name)
+        readGroupSet.populateFromFile(dataUrl, indexFile)
+        referenceSetName = self._args.referenceSetName
+        if referenceSetName is None:
+            # Try to find a reference set name from the BAM header.
+            referenceSetName = readGroupSet.getBamHeaderReferenceSetName()
```
david4096: This is a nice UI addition. I think we would be safest to remove this and require a reference set ID when adding a readgroupset. Why are we requiring that the reference be available if you just want to serve reads?

ga4gh/cli.py
```diff
@@ -1510,448 +1521,595 @@ def configtest_main(parser=None):
...
+    def addVariantSetNameArgument(cls, subparser):
+        subparser.add_argument(
+            "variantSetName",
+            help="the name of the variant set")
+
+    @classmethod
+    def addFeatureSetNameArgument(cls, subparser):
+        subparser.add_argument(
+            "featureSetName",
+            help="the name of the variant set")
```
david4096: type "feature set"

ga4gh/cli.py
```diff
@@ -1510,448 +1521,595 @@ def configtest_main(parser=None):
...
+            "Add a variant set to the data repo based on one or "
+            "more VCF files. ")
+        addVariantSetParser.set_defaults(runner="addVariantSet")
+        cls.addRepoArgument(addVariantSetParser)
+        cls.addDatasetNameArgument(addVariantSetParser)
+        addVariantSetParser.add_argument(
+            "dataFiles", nargs="+",
+            help=(
+                "The VCF/BCF files representing the new VariantSet. "
+                "These may be specified either one or more paths "
```
david4096: Nice!

ga4gh/datamodel/__init__.py
```diff
@@ -548,40 +546,5 @@ def sanitizeString(cls, attr, attrName):
...
-        """
-        Sets the creationTime and accessTime for this file system based
-        DatamodelObject. This is derived from the ctime of the specified
-        directoryPath.
-        """
-        ctimeInMillis = int(os.path.getctime(directoryPath) * 1000)
-        self._creationTime = ctimeInMillis
-        self._updatedTime = ctimeInMillis
-
-    def _scanDataFiles(self, dataDir, patterns):
```
david4096: Glad to see this go :)

ga4gh/datamodel/datasets.py
```diff
@@ -213,16 +207,20 @@ def __init__(
             numAlignments=1, numFeatureSets=1):
         super(SimulatedDataset, self).__init__(localId)
         self._description = "Simulated dataset {}".format(localId)
+        # TODO create a simulated OntologyTermMap
+        sequenceOntology = ontologies.OntologyTermMap("sequence_ontology")
+        # TODO add some terms into the simualated sequence ontology
```
david4096: typo simulated

ga4gh/datamodel/ontologies.py
```diff
@@ -59,7 +79,6 @@ def getGaTermByName(self, name):
         term = protocol.OntologyTerm()
         term.term = name
         term.id = self.getId(name)
-        # TODO set source name smarter
         term.sourceName = self._sourceName
         # TODO how do we get the right version?
         term.sourceVersion = None
```
david4096: Probably need to add options when adding the ontology map. `--sourceVersion`.

ga4gh/datamodel/reads.py
```diff
@@ -586,6 +622,8 @@ class SimulatedReadGroup(AbstractReadGroup):
     def __init__(self, parentContainer, localId, randomSeed, numAlignments=2):
         super(SimulatedReadGroup, self).__init__(parentContainer, localId)
         self._randomSeed = randomSeed
+        self._numAlignedReads = self._parentContainer.getNumAlignedReads()
+        self._numUnalignedReads = 0
```
david4096: Is there a particular reason why this is set to 0?

ga4gh/datamodel/references.py
```diff
@@ -361,75 +399,80 @@ class HtslibReferenceSet(datamodel.PysamDatamodelMixin, AbstractReferenceSet):
...
+        fastaFile = self.getFastaFile()
+        for referenceName in fastaFile.references:
+            reference = HtslibReference(self, referenceName)
+            # TODO break this up into chunks and calculate the MD5
+            # in bits (say, 64K chunks?)
+            bases = fastaFile.fetch(referenceName)
+            md5checksum = hashlib.md5(bases).hexdigest()
+            reference.setMd5checksum(md5checksum)
+            reference.setLength(len(bases))
+            self.addReference(reference)
```
david4096: Noting that it seems we have lost the ability to add fastas that cover a single contig.

ga4gh/datamodel/references.py
```diff
@@ -361,75 +399,80 @@ class HtslibReferenceSet(datamodel.PysamDatamodelMixin, AbstractReferenceSet):
...
+        Populates this reference set from the values in the specified DB
+        row.
+        """
+        self._dataUrl = row[b'dataUrl']
+        self._description = row[b'description']
+        self._assemblyId = row[b'assemblyId']
+        self._isDerived = bool(row[b'isDerived'])
+        self._md5checksum = row[b'md5checksum']
+        self._ncbiTaxonId = row[b'ncbiTaxonId']
+        self._sourceAccessions = json.loads(row[b'sourceAccessions'])
```
david4096: I'm not sure sourceAccessions are ever set for single references. https://github.com/ga4gh/server/issues/1185

ga4gh/datamodel/references.py
```diff
@@ -361,75 +399,80 @@ class HtslibReferenceSet(datamodel.PysamDatamodelMixin, AbstractReferenceSet):
...
-        return pysam.FastaFile(dataFile)
+    def populateFromRow(self, row):
+        """
+        Populates this reference from the values in the specified DB row.
+        """
+        self._length = row[b'length']
+        self._isDerived = bool(row[b'isDerived'])
+        self._md5checksum = row[b'md5checksum']
+        self._ncbiTaxonId = row[b'ncbiTaxonId']
+        self._sourceAccessions = json.loads(row[b'sourceAccessions'])
```
david4096: I don't believe these can be set anywhere.

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -158,14 +158,26 @@ class AbstractFeatureSet(datamodel.DatamodelObject):
     """
     compoundIdClass = datamodel.FeatureSetCompoundId
 
-    def __init__(
-            self, parentContainer, localId, referenceSetId=None):
+    def __init__(self, parentContainer, localId):
         super(AbstractFeatureSet, self).__init__(parentContainer, localId)
-        self._referenceSetId = referenceSetId
```
david4096: Can you explain why you opted to attach the reference object to the classes as opposed to the identifier?

ga4gh/datamodel/sequenceAnnotations.py
```diff
@@ -303,15 +314,41 @@ class Gff3DbFeatureSet(AbstractFeatureSet):
...
+        """
+        self._dbFilePath = dataUrl
         self._db = Gff3DbBackend(self._dbFilePath)
 
+    def populateFromRow(self, row):
+        """
+        Populates the instance variables of this FeatureSet from the specified
+        DB row.
+        """
+        self._dbFilePath = row[b'dataUrl']
```
david4096: Should get the source-uri too.

ga4gh/datamodel/references.py
diekhans: Only allowing one fasta per ReferenceSet is a fine restriction.

ga4gh/cli.py
```diff
@@ -1510,448 +1521,595 @@ def configtest_main(parser=None):
...
+        """
+        self._openRepo()
+        dataset = self._repo.getDatasetByName(self._args.datasetName)
+        name = getNameFromPath(self._args.filePath)
+        featureSet = sequenceAnnotations.Gff3DbFeatureSet(
+            dataset, name)
+        referenceSetName = self._args.referenceSetName
+        if referenceSetName is None:
+            raise exceptions.RepoManagerException(
+                "a reference set name must be provided")
```
david4096: a *A

ga4gh/backend.py
```diff
@@ -574,7 +569,8 @@ def variantAnnotationsGenerator(self, request):
         compoundId = datamodel.VariantAnnotationSetCompoundId.parse(
             request.variantAnnotationSetId)
         dataset = self.getDataRepository().getDataset(compoundId.datasetId)
-        variantAnnotationSet = dataset.getVariantAnnotationSet(
+        variantSet = dataset.getVariantSet(compoundId.variantSetId)
+        variantAnnotationSet = variantSet.getVariantAnnotationSet(
```
macieksmuga: From discussion with @diekhans and @david4096, this seems to go against a common use case in which a variant annotation set would come from a different dataset than the variant set.

ga4gh/datamodel/ontologies.py
```diff
@@ -5,26 +5,43 @@
...
+        specified file.
+        """
+        self._dataUrl = dataUrl
+        self._readFile()
+
+    def populateFromRow(self, row):
+        """
+        Populates this Ontology using values in the specified DB row.
+        """
+        self._dataUrl = row[b'dataUrl']
```
david4096: Get the source version as well.

tests/end_to_end/test_repo_manager.py
```diff
@@ -33,6 +33,7 @@ def _runCmd(self, cmd, *args):
         command = ["--loud", cmd, self.tempdir] + list(args)
         cli.repo_main(command)
 
+    @unittest.skip("Skip test until repo manager is updated")
     def testEndToEnd(self):
         self._runCmd("init")
         self._runCmd("add-ontologymap", paths.ontologyPath)
```
david4096: Just a name problem here ( I believe ).

david4096:
Good stuff! Will we be turning the tests back on before merging? Left a dusting of comments and we can work with this before we move on.

From what I can tell, it's going to be difficult to capture the variety of annotated VCFs with internal server mechanics. It makes sense to me to move that effort to a better isolated bit of code.

macieksmuga:
#1189 removed my temporary hold: the code in this PR now passes compliance modulo two known exceptions in Reads.

kozbo:
Anything else that needs to be changed in this PR before the merge?

dcolligan:
There's all of @david4096 's comments...

(There's also some PRs of mine but I can just re-make them against master once this is merged if they don't get in in time.  We shouldn't wait on that.)

ga4gh/cli.py
david4096: https://github.com/ga4gh/server/pull/1188

ga4gh/datamodel/reads.py
jeromekelleher: I think it's for a non-nullable field, so we have to put something in it. 0 seems like the best value in this situation.

docs/configuration.rst
jeromekelleher: Depends on who you ask... we should rephrase this in any case if we're going for the `registry db` nomenclature.

docs/configuration.rst
jeromekelleher: We would need to be precise about how and when here. It would just be confusing to say that we'll sometimes need write access without fully qualifying.

docs/configuration.rst
jeromekelleher: I think the time for these comments was in the #1158 PR @david4096. This was never meant to be a full and final update of the docs. The stated intention was to make them less inaccurate. We definitely want to do a full run through of the docs once this has been merged, but I don't think it makes sense to block the whole PR on it.

docs/configuration.rst
jeromekelleher: As discussed in https://github.com/ga4gh/server/pull/1117, and in preparation for #1147 we want ``Ontology``.

docs/configuration.rst
jeromekelleher: We should create an issue holding the documentation requirements. This is not related to the actual changes in the PR.

ga4gh/backend.py
jeromekelleher: The schema says nothing on the topic, since there is no data set associated with  the variant annotation set object. So, as far as I'm concerned, this is the only correct interpretation of the schemas. 

I agree that we should flatten the object hierarchy, and translate GET requests directly into SQL requests. This will be done once we've changed the ID system (probably post protobuf, to try and minimise disruption).

If the current code works and doesn't break anything, please move discussion to an issue. We want to include these changes ASAP.

ga4gh/cli.py
jeromekelleher: Yes, I realise this @david4096, I was wondering how we set up our API to use these data base features.

ga4gh/datarepo.py
```diff
@@ -179,42 +262,826 @@ def __init__(
...
+        """
+        Closes this repo.
+        """
+        if self._openMode is None:
+            raise ValueError("Repo already closed")
+        self._openMode = None
+        self._dbConnection.close()
+        self._dbConnection = None
+
+    def verify(self):
```
jeromekelleher: It needs to be much more sophisticated, https://github.com/ga4gh/server/issues/1170

ga4gh/datarepo.py
jeromekelleher: I don't think so --- there's no particular reason for a DB extension, and this is unix.

david4096:
Just a quick note, none of my comments are blocking. I have an open PR on the branch that we could merge first https://github.com/ga4gh/server/pull/1196 .

ga4gh/cli.py
david4096: Issue for providing success feedback https://github.com/ga4gh/server/issues/1200

ga4gh/datamodel/sequenceAnnotations.py
jeromekelleher: Because in VariantSets and ReadGroupSets we need the actual reference objects to look things up. I opted the object reference here over the ID for uniformity.

ga4gh/cli.py
jeromekelleher: Because we need to create a linkage between readgroupsets and referencesets. There is a ReferenceSet field in ReadGroupSet which we must fill. If deriving the reference set from the BAM header proves to be problematic we can make it a mandatory argument.

ga4gh/cli.py
jeromekelleher: Please open an issue on this.

ga4gh/cli.py
jeromekelleher: Please open an issue outlining the tests we need.

ga4gh/backend.py
jeromekelleher: Fair enough @macieksmuga, but the schema doesn't actually support this at the moment. I'm just trying to simplify things right now, we can make them more complicated when the need arises.

ga4gh/datamodel/ontologies.py
jeromekelleher: It's not in the DB, and there's no place to set it in the CLI. See #1147.

ga4gh/cli.py
jeromekelleher: Done.

ga4gh/datamodel/sequenceAnnotations.py
jeromekelleher: Where does this come from? Please open an issue if this is not available.

ga4gh/datamodel/datasets.py
jeromekelleher: done.

ga4gh/cli.py
jeromekelleher: Done.

docs/configuration.rst
jeromekelleher: Done.

jeromekelleher:
I think I've addressed all the actionable comments here, and hopefully we can track the remaining with some issues. I've opened a PR #1207 for these review fixes, which hopefully we can merge quickly.

Are we ready to vote?

david4096:
I closed my other PR, I'm all for merging!

macieksmuga:
+1 for merging. I just re-ran the full compliance suite, and the only tests that fail are the expected ones on reads.

jeromekelleher:
Great, thanks @macieksmuga and @david4096. I'll try and get an update to sort out the latest merge problems pushed soon, and hopefully we can merge into maste quickly after that.

dcolligan:
+1 for merge

jeromekelleher:
OK, once #1208 has been merged I think we're ready to click the big green button!

jeromekelleher:
-1, blocking this until we fix the multi readgroup search problem pointed out in  #1208.

ga4gh/cli.py
david4096: Got it! https://github.com/ga4gh/server/issues/1211

ga4gh/cli.py
david4096: https://github.com/ga4gh/server/issues/1210

macieksmuga:
@jeromekelleher assuming PR #1212 passes muster, this should lift your -1

jeromekelleher:
Great, thanks @macieksmuga. Let's see what @dcolligan thinks of #1212, and take it from there. Hopefully we can merge today.

jeromekelleher:
OK,  #1212 has been merged so I think we're good to go. Retracting my previous -1.

@dcolligan, what do you think? Ready to push the button?

dcolligan:
I am now pushing the button.

dcolligan:
@jeromekelleher let the great issue-closing begin

jeromekelleher:
Woohoo! Thanks @dcolligan, time for a closeathon!

PR #1189, opened 2016-04-29
Minor fixes to sql_repo branch

david4096:
This PR includes the changes I needed to make to get the compliance data loaded.

I believe there was a dangling db.commit in insertReferenceSet, which was removed.

The type of ncbiTaxonId needed to be set to integer.

Variant annotation sets were receiving their localId from the annotation type, changed back to the name.

Lastly, updated the prepare compliance data script to work with the new repo.


david4096:
Note: there is one compliance test that has been solved in master and not this branch.

macieksmuga:
+1 Ran the compliance tests against this, all work with the exception pointed out by @david4096 above.
Looks great!

Issue #950, opened 2016-02-26
Returning calls in 1000 genomes is slow


david4096:
When there is a VCF with many callsets in it, responses can become slow.

jeromekelleher:
Need some more details here @david4096... what's your pagesize in particular? Are you sure it's not the client side JSON decoding that's the bottleneck? We have some perf benchmarks for this sort of thing in ``server_benchmark.py`` if you want to have a look. They might need to be updated a bit though; see #711.

kozbo:
I was running some tests against the 1000 Genomes VCF files (2504 samples) and I verified the slow results. I ran some tests with different page sizes and the amount of time to return each page increased linearly with the number of variants included in each page. I also tried changing my search variants to use `callSetIds=[]` but that made no difference hoping that I could avoid the callSetId searching in convertVariant, but that didn't work either because we don't check for the empty callset. So then we did some profiling and we found that we were spending a lot of time in base64 (which is probably why the CPU was so busy in top). Once we passed in the empty array and also modified the code in variants.py to not search if the length of the callSetId array is 0, and then my search ran very quickly. All this will see explained in @david4096 's PR coming soon.

We learned a lot about profiling the server so I thought I would capture the tricks here to start.
No surprise, you cannot easily profile from the Apache environment. Also we turned off the flask debugger and we also had to prevent the use of the reloader. Here is the resulting command line we used:
`python -m cProfile -s cumulative -o profile.out ko-server/server_dev.py -H 100.70.26.73 --dont-use-reloader --config-file /srv/ga4gh/config.py`

We also found a neat way to investigate the output of the profiler via a browser.
`pip install cprofilev
cprofilev -f profile.out -a 100.70.26.73`

jeromekelleher:
Excellent, thanks @kozbo. The variant search code definitely needs some attention, I'm sure there's quite substantial improvements we can make using the profiling.

david4096:
Closed with #1191

PR #1191, opened 2016-04-30
Improvements to the variants API

david4096:
Most noticeable when the call list is large.

I got at worst a 2x speedup. When an empty callset list is sent you'll see much faster but YMMV.

Replaced the hacked phase set implementation with pysam native. Found that pysam and pyvcf disagree on whether a single genotype in a variant set should be called phased. See demonstration [here](https://gist.github.com/david4096/997fc3ecb00783ba6775acf15750eec4).

Helps with #950

### DIFF
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -33,36 +33,6 @@ def isUnspecified(str):
     return str == "" or str is None
 
 
-def convertVCFPhaseset(vcfPhaseset):
-    """
-    Parses the VCF phaseset string
-    """
-    if vcfPhaseset is not None and vcfPhaseset != ".":
-        phaseset = vcfPhaseset
-    else:
-        phaseset = "*"
-    return phaseset
-
-
-def convertVCFGenotype(vcfGenotype, vcfPhaseset):
-    """
-    Parses the VCF genotype and VCF phaseset strings
-    """
-    phaseset = None
-    if vcfGenotype is not None:
-        delim = "/"
-        if "|" in vcfGenotype:
-            delim = "|"
-            phaseset = convertVCFPhaseset(vcfPhaseset)
-        if "." in vcfGenotype:
-            genotype = [-1]
-        else:
-            genotype = map(int, vcfGenotype.split(delim))
-    else:
-        genotype = [-1]
-    return genotype, phaseset
-
-
 class CallSet(datamodel.DatamodelObject):
     """
     Class representing a CallSet. A CallSet basically represents the
@@ -109,6 +79,7 @@ class AbstractVariantSet(datamodel.DatamodelObject):
         self._callSetIdMap = {}
         self._callSetNameMap = {}
         self._callSetIds = []
+        self._callSetIdToIndex = {}
         self._creationTime = None
         self._updatedTime = None
         self._referenceSet = None
@@ -185,6 +156,7 @@ class AbstractVariantSet(datamodel.DatamodelObject):
         self._callSetIdMap[callSetId] = callSet
         self._callSetNameMap[callSet.getLocalId()] = callSet
         self._callSetIds.append(callSetId)
+        self._callSetIdToIndex[callSet.getId()] = len(self._callSetIds) - 1
 
     def addCallSetFromName(self, sampleName):
         """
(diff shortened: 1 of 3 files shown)


ga4gh/datamodel/variants.py
```diff
@@ -123,6 +124,7 @@ def addCallSet(self, sampleName):
         self._callSetIdMap[callSetId] = callSet
         self._callSetNameMap[sampleName] = callSet
         self._callSetIds.append(callSetId)
+        self._callSets[callSet.getId()] = len(self._callSetIds) - 1
```
diekhans: The name _callSets implies that this contains CallSet objects and is confusing. maybe _callSetIdToIndex?

ga4gh/datamodel/variants.py
```diff
@@ -685,15 +687,19 @@ def convertVariant(self, record, callSetIds):
...
+                    call = protocol.Call()
+                    call.callSetId = callSetId
+                    call.callSetName = callSet.getSampleName()
+                    index = self._callSets[callSetId]
+                    genotypeData = sampleData[index].split(
+                             ":")[0]  # REMOVAL
+                    call.genotype, call.phaseset = convertVCFGenotype(
+                             genotypeData, call.phaseset)
+                    variant.calls.append(call)
         variant.id = self.getVariantId(variant)
```
diekhans: +1 very important, sans comments above.

We still need to seriously look at performance.

ga4gh/datamodel/variants.py
```diff
@@ -685,15 +687,19 @@ def convertVariant(self, record, callSetIds):
...
+                if callSet is not None:
+                    call = protocol.Call()
+                    call.callSetId = callSetId
+                    call.callSetName = callSet.getSampleName()
+                    index = self._callSets[callSetId]
+                    genotypeData = sampleData[index].split(
+                             ":")[0]  # REMOVAL
+                    call.genotype, call.phaseset = convertVCFGenotype(
+                             genotypeData, call.phaseset)
+                    variant.calls.append(call)
```
diekhans: I am not a coder on this, so feel free to ignore; however I find it much easier to read code if functions are created rather than nesting loops inside of ifs

ga4gh/datamodel/variants.py
```diff
@@ -625,36 +597,25 @@ def _addDataFile(self, filename):
...
-
-        ###########################################
-        # call.genotype = list(pysamCall.allele_indices)
-        # call.phaseset = pysamCall.phaseset
-        ###########################################
-
-        call.genotypeLikelihood = []
+    def _convertGaCall(self, callSet, pysamCall):
+        phaseset = None
+        if pysamCall.phased is True:
```
jeromekelleher:
```python
if pysamCall.phased:
     # stuff
```

ga4gh/datamodel/variants.py
```diff
@@ -680,20 +641,14 @@ def convertVariant(self, record, callSetIds):
...
-        sampleIterator = 0  # REMOVAL
-        if callSetIds is not None:
-            for name, call in record.samples.iteritems():
-                if self.getCallSetId(name) in callSetIds:
-                    genotypeData = sampleData[sampleIterator].split(
-                        ":")[0]  # REMOVAL
-                    variant.calls.append(self._convertGaCall(
-                        record.id, name, call, genotypeData))  # REPLACE
-                sampleIterator += 1  # REMOVAL
+        if callSetIds is not None and len(callSetIds) > 0:
```
jeromekelleher: When could ``callSetIds`` be None here? I don't think that's a valid input. We can just ``assert`` that callSetIds is not None I think.

ga4gh/datamodel/variants.py
```diff
@@ -680,20 +641,14 @@ def convertVariant(self, record, callSetIds):
...
+            for callSetId in callSetIds:
+                callSet = self.getCallSet(callSetId)
```
jeromekelleher: Why would callSet be None here? This seems like an error condition to me. We shouldn't silently hide error conditions, this will cause very hard to find bugs. Much better to fail loudly with an exception.

jeromekelleher:
This is a much needed cleanup, thanks @david4096. I think we can simplify things a bit by identifying our error cases more clearly and raising exceptions when nasty things happen. 

Re the pysam/pyvcf issue, can you give us some examples of where they disagree?

david4096:
@jeromekelleher the one example I found in our data you look at here https://gist.github.com/david4096/997fc3ecb00783ba6775acf15750eec4

jeromekelleher:
I'm not sure what to make of this phasing issue @david4096 --- this looks like an error in the input vcf to me, since we have '0' rather than '0/0', '0/1' etc in the genotype column, unlike the rest of the file. I could be wrong though. In any case, it seems like it's an awkward corner case, and I'm happy to assume that pysam has got it right. So, I would work around this in the tests by:
```python
if len(genotype.split("/")) == 1:
     # assert phased is true
else:
     # Compare our value for phased with pyvcf
```

How does this sound?

david4096:
@jeromekelleher Works for me! Updated!

jeromekelleher:
Excellent, thanks @david4096. This is a super important update, I'm excited to see how our benchmarks look after we merge.

+1.

jeromekelleher:
Can you rebase this please @david4096? We want to get this in ASAP.

david4096:
Sure updated!

jeromekelleher:
Looks like we're out of date again. Can you rebase and merge please @david4096?

Issue #1147, opened 2016-04-26
Ontology handling update required


jeromekelleher:
The handling of the ontology term mapping in the sql reqo is currently quite brittle, and needs some attention. The most important issue that we need to understand is where the ontology term mapping file comes from, and how we version it. Ideally, I think it would be good if we could get rid of the intermediate text file, and create the term name -> ID mapping as a table in the DB. 

Once this is done, we should set up a proper relational link between variant sets and ontologies using foreign key constraints, with a properly designed table.

Any thoughts or discussion on this would be much appreciated.

sarahhunt:
@jeromekelleher - the SO is available here:
https://github.com/The-Sequence-Ontology/SO-Ontologies/blob/master/so-xp.obo

There is a python OBO parser. There is a data-version date in the file, but I'm not sure how this relates to the version number on the browser: http://www.sequenceontology.org/browser/obob.cgi

@keilbeck manages the SO so would be able to advise on versioning.

jeromekelleher:
Thanks a million @sarahhunt, this is hugely useful. I'm sure we can build solid ontology support based on this!

jeromekelleher:
Here's a snipped using the ontology library above:
```python
from __future__ import division
from __future__ import print_function
from __future__ import unicode_literals

import orangecontrib.bio.ontology

filename = "so-xp.obo"
ontology = orangecontrib.bio.ontology.OBOOntology(filename)
for j, term in enumerate(ontology.terms()):
    print(term.id, term.name)
    if j == 10:
        break

print()
print("Looking up term names...")
name = "sequence_variant_affecting_polypeptide_function"
term = ontology.term_by_name(name)
print(term)
```

Gives
```
SO:0000000 Sequence_Ontology
SO:0000001 region
SO:0000002 sequence_secondary_structure
SO:0000003 G_quartet
SO:0000004 interior_coding_exon
SO:0000005 satellite_DNA
SO:0000006 PCR_product
SO:0000007 read_pair
SO:0000008 gene_sensu_your_favorite_organism
SO:0000009 gene_class
SO:0000010 protein_coding

Looking up term names...
[Term]
id: SO:1000117
name: sequence_variant_affecting_polypeptide_function
comment: OBSOLETE: This term was deleted as it conflated more than one term. The alteration is separate from the effect.
synonym: "mutation affecting polypeptide function" RELATED []
synonym: "sequence variant affecting polypeptide function" EXACT []
is_obsolete: true
[...]

PR #1196, opened 2016-05-02
Updates to sql_repo branch

david4096:
Added documentation to configuration for optional arguments and fixed a few typos.

Changing references of `add-ontology` to `add-ontologymap` and reinstating some end to end tests. Removed calls to `destroy`.

The tests for the repo need to be for a single reference set, so the test BAM files have been switched to those from thousand genomes and NCBI37.


docs/configuration.rst
```diff
@@ -40,8 +41,8 @@ the first command you run.
 verify
 +++++++
 
-Performs some consistency checks on the given data repository to ensure it is
-well-formed.
+Checks that the data pointed to in the repository works and
+there aren't any broken URLs or missing files.
```
macieksmuga: plural data? Alternately:
```
Checks that the data files pointed to in the repository exist and are readable in the locations specified by file paths or URLs.
```

macieksmuga:
+1 These look good to me (modulo minor wording edit above). All tests pass cleanly.

jeromekelleher:
-1

The reasoning for Ontology has been discussed: https://github.com/ga4gh/server/pull/1117 and https://github.com/ga4gh/server/issues/1147

macieksmuga:
I'm officially confused.

david4096:
@jeromekelleher updated, thanks!

tests/unit/test_cli.py
```diff
@@ -425,7 +425,8 @@ def testAddOntology(self):
 
     def testRemoveOntology(self):
         ontologyName = "the_ontology_name"
-        cliInput = "remove-ontology {} {}".format(self.repoPath, ontologyName)
+        cliInput = "remove-ontology {} {}".format(
```
jeromekelleher: Why the split? Looks like a fairly gratuitous change...

tests/end_to_end/test_repo_manager.py
```diff
@@ -5,14 +5,11 @@
...
 
-import mock
 import os
-import shutil
 import tempfile
 import unittest
 
 import ga4gh.cli as cli
-import ga4gh.exceptions as exceptions
 import tests.paths as paths
```
jeromekelleher: These changes are already covered in pr #1198 I think

docs/configuration.rst
```diff
@@ -164,6 +199,33 @@ Removes a variant set from a given data repository and dataset.
 
     $ ga4gh_repo remove-variantset path/to/repo.db aDataset aVariantSet
 
+++++++++++++++
+add-featureset
+++++++++++++++
+
+Adds a feature set to the repo created using `scripts/generate_gff3_db.py`.
```
jeromekelleher: I'm not sure how helpful this actually is --- we don't distribute the ``generate_gff3_db.py`` script, so this is only going to cause confusion. We'll need to make this a first class CLI program, and document it separately. This needs to be done in a follow up issue, where we look at the requirements for importing featuresets properly and figure out the workflow for the administrator.

jeromekelleher:
I'm afraid I don't see any of these changes as helping us much @david4096. The CLI documentation for repo manager needs a complete overhaul, and the changes here don't really improve it much. There's no point in adding in documenting a couple of options here and there; we need to comprehensively document the interface. It would be good if we could use something like [sphinx-argparse]
(http://sphinx-argparse.readthedocs.io/en/latest/) to do this, so that we only document the CLI once, in the source.

Issue #1180, opened 2016-04-28
Add tests to ``tests/unit/test_datarepo.py`` to verify behaviour when incorrect indexes are passed


jeromekelleher:
Assumes #1166 has been merged.

The ``add-variantset`` and ``add-referenceset`` functions in repo manager allow the option of specifying index files to go with the data URLs provided. We should add some tests to check that we correctly catch problems such as providing files that (a) aren't actually indexes; and (b) are indexes but are for the wrong file. It might be tricky to do this with pysam, but we should figure out what we need to do to detect this situation and raise an informative error message.

dcolligan:
The `add-referenceset` command doesn't allow specification of indexes... presumably you meant `add-readgroupset`?

jeromekelleher:
Yes, sorry, that was a typo.

david4096:
Related https://github.com/ga4gh/server/issues/1210

dcolligan:
Fixed in #1214

PR #1197, opened 2016-05-03
Test for incorrect data file indexes

dcolligan:
Issue #1180

### DIFF
--- a/ga4gh/datamodel/reads.py
+++ b/ga4gh/datamodel/reads.py
@@ -211,8 +211,13 @@ class AlignmentDataMixin(datamodel.PysamDatamodelMixin):
         # not throw an error if the index is missing.
         if not os.path.exists(self._indexFile):
             raise exceptions.FileOpenFailedException(self._indexFile)
-        return pysam.AlignmentFile(
-            self._dataUrl, filepath_index=self._indexFile)
+        try:
+            return pysam.AlignmentFile(
+                self._dataUrl, filepath_index=self._indexFile)
+        except IOError as exception:
+            # IOError thrown when the index file passed in is not actually
+            # an index file... may also happen in other cases?
+            raise exceptions.DataException(exception.message)
 
 
 class AbstractReadGroupSet(datamodel.DatamodelObject):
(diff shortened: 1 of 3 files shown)


ga4gh/datamodel/reads.py
```diff
@@ -209,8 +209,13 @@ def openFile(self, dataFile):
...
             raise exceptions.FileOpenFailedException(self._indexFile)
-        return pysam.AlignmentFile(
-            self._dataUrl, filepath_index=self._indexFile)
+        try:
+            return pysam.AlignmentFile(
+                self._dataUrl, filepath_index=self._indexFile)
+        except IOError:
+            # IOError thrown when the index file passed in is not actually
+            # an index file... may also happen in other cases?
+            raise exceptions.DataException()
```
jeromekelleher: There should be a file and index name passed to the exception here so we can make a sensible error message. Is there any useful information encoded in the IOError exception thrown back? It might make sense to pass this along as a supllemental message as well.

jeromekelleher:
What works well at the moment, and what happens when a bad index is passed? It looks like you're skipping the tests where we pass one...

tests/unit/test_repo_manager.py
```diff
@@ -681,3 +682,80 @@ def testFeatureSetDelete(self):
...
+
+    def _testWithIndexPath(self, indexPath):
+        cmd = (
+            "add-variantset {} {} {} --referenceSetName={} -I {}").format(
+                self._repoPath, self._datasetName, paths.vcfPath1,
+                self._referenceSetName, indexPath)
+        with self.assertRaises(exceptions.NotIndexedException):
+            self.runCommand(cmd)
+
+    def testNonexistantIndexFile(self):
```
david4096: Typo existent

dcolligan:
@jeromekelleher updated

Yeah I don't know if there's a way in the skipped tests to detect if a well-formed but incorrect index is passed without iterating through the file and getting an error.  I don't know if it's something we can detect at file-open time.

dcolligan:
Closing; the new PR is #1197 (which is opened against `master` instead of the now-merged `sql_repo`)

PR #1199, opened 2016-05-03
explicitly set boolean to True/False

saupchurch:
sqlite stores booleans as ints (0 or 1).  This explicitly sets boolean value to True/False rather than relying on python's type conversion which work fine internally but was causing the response JSON to contain the int value instead of a boolean value

### DIFF
--- a/ga4gh/datamodel/rna_quantification.py
+++ b/ga4gh/datamodel/rna_quantification.py
@@ -70,7 +70,7 @@ class ExpressionLevel(AbstractExpressionLevel):
         self._expression = record["expression"]
         self._quantificationGroupId = record["quantification_group_id"]
         # sqlite stores booleans as int (False = 0, True = 1)
-        self._isNormalized = record["is_normalized"] != 0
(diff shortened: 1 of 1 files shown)


ga4gh/datamodel/rna_quantification.py
```diff
@@ -69,7 +69,8 @@ def __init__(self, parentContainer, record):
         self._annotationId = record["annotation_id"]
         self._expression = record["expression"]
         self._quantificationGroupId = record["quantification_group_id"]
-        self._isNormalized = record["is_normalized"]
+        # sqlite stores booleans as int (False = 0, True = 1)
+        self._isNormalized = record["is_normalized"] != 0
```
jeromekelleher:
```python
bool(record["is_normalized"] )
```
is a bit more explicit

saupchurch: that it is - will make the change

jeromekelleher:
Thanks @saupchurch, +1.

PR #1208, opened 2016-05-04
Sql merge update

jeromekelleher:
Brings sql repo branch up to date with master. Some minor merge mess management here, but there shouldn't be any real changes.


jeromekelleher:
OK, I think we can merge this fairly quickly as it's just a merge update. @david4096, @dcolligan or @macieksmuga, can you take a quick look at merge if you're happy/online?

macieksmuga:
testing now...

macieksmuga:
Passes compliance with the exception of the two reads tests as before. Was this merge expected to incorporate Danny's enhancement for searching across multiple `ReadGroup`s within a single `ReadGroupSet`?

jeromekelleher:
Yes, I think that went in a while ago, didn't it? This is just merging new commits from master into the sql repo branch, I think something fairly major would need to go wrong if we lost commits like that along the way.

macieksmuga:
Hmm. I still keep getting the server error:
```
[...]
  File "/Users/maciek/Documents/dev/maciek-ga4gh-server/ga4gh/backend.py", line 160, in _search
    return self._parentContainer.getReadAlignments(
AttributeError: 'HtslibReadGroupSet' object has no attribute 'getReadAlignments'
```
when I issue the following multi-readgroup request:
```
curl -XPOST -H"Content-Type:application/json" -d'{"readGroupIds": ["WyJicmNhMSIsInJncyIsIkhHMDAwOTYiLCJTUlIwNjI2MzQiXQ", "WyJicmNhMSIsInJncyIsIkhHMDAwOTYiLCJTUlIwNjI2MzUiXQ", "WyJicmNhMSIsInJncyIsIkhHMDAwOTYiLCJTUlIwNjI2NDEiXQ"], "referenceId": "WyJoZzM3IiwicmVmX2JyY2ExIl0", "start": 0, "end": 150, "pageSize": null, "pageToken": null}' localhost:8000/reads/search
```

jeromekelleher:
OK, looks like we might have another problem not covered by the unit tests that your tests are provoking. This probably got lost in the merge somewhere, and I assumed it was working since no test cases failed. Good catch @macieksmuga, thanks.

The issues doesn't actually affect this particular merge though, so I think we should go ahead with this PR ASAP. I'll put a block on the sql repo PR until we've solved the problem.

macieksmuga:
Fine with me. +1 for merging in this PR. I'll create an issue for it so we don't lose this case.

PR #1214, opened 2016-05-05
Test for incorrect data file indexes

dcolligan:
Issue #1180 

This is the same code as #1197, except the PR is made against `master` instead of the (now-merged) `sql_repo`


jeromekelleher:
Thanks. I'm not sure what works here and what doesn't; can you give a quick recap on where this is at?

dcolligan:
Same comment as the previous thread:

"Yeah I don't know if there's a way in the skipped tests to detect if a well-formed but incorrect index is passed without iterating through the file and getting an error. I don't know if it's something we can detect at file-open time."

jeromekelleher:
Sorry, I missed that.

+1.

PR #1215, opened 2016-05-05
Removes needless conversion.

jeromekelleher:
Removes no-op type checks from the ncbiTaxonId code path. The type of the value in the DB is integer, so the value within python must be either an integer or None. Also removes the call to `setNcbiTaxonId` as none of the other attributes did this, and there is no real difference between between ``_ncbiTaxonId`` and the rest.  Calling setters is good practise, but this should be done uniformly within the method on all attributes,  not piecemeal on one.

### DIFF
--- a/ga4gh/datamodel/references.py
+++ b/ga4gh/datamodel/references.py
@@ -184,10 +184,7 @@ class AbstractReferenceSet(datamodel.DatamodelObject):
         which do not belong to the modeled species, e.g.  EBV in a
         human reference genome.
         """
-        if self._ncbiTaxonId is not None:
-            return int(self._ncbiTaxonId)
-        else:
-            return None
+        return self._ncbiTaxonId
 
     def toProtocolElement(self):
         """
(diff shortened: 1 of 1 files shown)


macieksmuga:
+1 
This passes compliance just fine. I assume the removed code was there to catch the case where a backend incorrectly records the NCBI Taxon ID as a string. I believe I've seen these kinds of conversions in other places as well, I take this as a call to remove such code when we see it, and  assume the database provides the correct type?

jeromekelleher:
>  I believe I've seen these kinds of conversions in other places as well, I take this as a call to remove such code when we see it, and assume the database provides the correct type?

Yes, I think we should. As a general principle, types should be checked on the way in to an object, so that an error can be propagated back to the source of the problem. Silently 'fixing' bad types on the way out (i.e. in getters) is a bad idea because this will cause the problems to be propagated further down stream and lead to difficult to trace problems. Having the wrong type inside an instance variable is a programming error, and it should lead to an exception.

jeromekelleher:
@dcolligan, can you take a look and merge if you're happy please?

jeromekelleher:
@macieksmuga, @david4096, can one of you merge this please (if you're happy with it)?

Issue #1217, opened 2016-05-06
Clarify nomenclature


jeromekelleher:
Currently we're using the work 'repository' to refer to the sum total of all data within the server (i.e., datasets, reference sets, etc) as well as the database we use to store information about the objects concerned. This is probably a bit confusing. Several people have proposed  'registry' as an alternative, but it's not clear to me if this is just the DB (good IMO), or as a replacement for the entire repository concept (bad, IMO). It's important that we decide on our nomenclature and use it consistently throughout the project.

@macieksmuga, @david4096, you are the main 'registry' advocates here; can you outline your definitions of the various terms, and see if we can all agree on them? I'm happy to suppress my instinctive distaste for 'registry' and adopt it, I just want to be sure we all agree on what we're talking about first!

macieksmuga:
@jeromekelleher thanks for continuing this discussion. The way you informally define the terms is in line with my understanding: "Repository" refers to the sum total of all the data, while "registry" refers just to the SQL database that contains all metadata and pointers to the actual data objects. Thing is, by going in this direction, we're no longer creating a "repository" proper - the raw data files can now be fairly arbitrarily located, and the entity we're managing is just the registry of said data.

And, for what it's worth, I understand and share in your hesitation at reusing the word "registry", a term that was associated with a lot of problems due to data corruption and mismatches between the database and the actual resources it was supposed to be managing. But that's exactly the same pattern we're adopting here, so it may be useful to keep in mind that while there are many advantages to it, there are also some frustrations we'll probably stumble across.

jeromekelleher:
OK, good, I think we're more-or-less in agreement then. I think it's still useful to have a word to refer to the 'total sum of all data' in the server, as this is what the client interacts with ultimately. The client doesn't know or care how this (I'm going with singular here in the CS sense @david4096!) data is organised; it just knows that there is a server at an IP address, and that this server represents a repository of data.

From an administration perspective, this idea of a repository still exists, but it's defined by the registry database. The registry DB defines what the datasets, referenceSets, etc, are and where to find their backing files/URLs/databases/whatever. Does this make sense? Are we happy to go forward with admin documentation based on this? Please shout out now if you're not happy with this before I start actually writing the docs!

david4096:
No objections here. I've put together the latest set of commands I used to stand up the 1kg server, which you may find useful.

https://gist.github.com/david4096/fa83f6bbe85b7f3342accdbaa4e7d9d1

[...]

PR #1218, opened 2016-05-06
Added sphinx-argparse based repo manager docs.

jeromekelleher:
This is an update to use sphinx-argparse to automatically include documentation about command line arguments. The output isn't ideal, but it's a lot better than trying to maintain two versions of the same documentation.

Here's what it [looks like](http://ga4gh-server-fork.readthedocs.io/en/repo-cli-docs/configuration.html#data-repository) on RTD.

Note that this is not intended as a definitive, final update on these docs. It's just trying out this new technology to see if it's something we want to use to build the final, definitive version of the docs. We're going to need to rephrase a bunch of stuff based on the outcome of #1217.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -35,6 +35,7 @@ extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.intersphinx',
     'sphinx.ext.todo',
+    'sphinxarg.ext',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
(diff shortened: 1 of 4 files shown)


dcolligan:
Definitely a step in the right direction

jeromekelleher:
Thoughts from anyone else? @david4096, @macieksmuga? If no one objects I think we should merge this much, and then look to update the docs properly in a follow up PR.

david4096:
Neat!

reece:
I like it! I didn't know about the sphinx argparse extension -- thanks for teaching me something!

jeromekelleher:
Any +1s here so we can merge?

kozbo:
looks great to me +1

jeromekelleher:
Great, thanks @kozbo. The branch should be up to date now. Can someone push the green button when CI finishes please?

Issue #1164, opened 2016-04-27
SQL repo referenceSets search returns number for boolean field


macieksmuga:
From my quick and dirty testing of the `sql_repo` branch:
```
curl -XPOST -H"Content-Type:application/json" -d'{}' localhost:8000/referencesets/search 

{"nextPageToken": null,"referenceSets": [{"name": "hg37", "sourceURI": null, "assemblyId": null, "sourceAccessions": [], "ncbiTaxonId": null, "isDerived": 0, "id": "WyJoZzM3Il0", "md5checksum": "12827085103f42e97428dc05a7f26fec", "description": ""}]}                                                                                                                   
```
Per schemas, `isDerived` should be either `true` or `false`.

jeromekelleher:
Good catch @macieksmuga --- this is also a bug in our unit tests. Does this get past the Avro validate function? I think we need a test case that loads up our test data repo,  pulls out all the objects and checks if they are valid instances of the schema. Should be quick enough to do.

jeromekelleher:
Partial fix for this in #1165 to try and push the sql repo merge ahead. This issue shouldn't be closed until we have test coverage of the issue though.

dcolligan:
@jeromekelleher can we close this now?

david4096:
Yep, thanks!
```
http post 1kgenomes.ga4gh.org/referencesets/search pageToken=0
HTTP/1.1 200 OK
Connection: Keep-Alive
Content-Encoding: gzip
Content-Length: 294
Content-Type: application/json
Date: Wed, 13 Jul 2016 16:42:39 GMT
Keep-Alive: timeout=5, max=100
Server: Apache/2.4.7 (Ubuntu)
Vary: Accept-Encoding

{
    "nextPageToken": "",
    "referenceSets": [
        {
            "assemblyId": "",
            "description": "NCBI37 assembly of the human genome",
            "id": "WyJOQ0JJMzciXQ",
            "isDerived": false,   <------------
            "md5checksum": "54e0bb53844059bb7152618fc927cfa9",
            "name": "NCBI37",
            "ncbiTaxonId": 9606,
            "sourceAccessions": [],
            "sourceUri": "ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/technical/reference/phase2_reference_assembly_sequence/hs37d5.fa.gz"
        }
    ]
}
```

Issue #1213, opened 2016-05-05
New end-to-end test required


jeromekelleher:
Recent issues have shown some weaknesses in our test coverage (see e.g. #1209, #1164), where we're only discovering some issues when the compliance suite is run against the server. The root cause of this is that we're not really exercising our production data interfacing layer (i.e., where we interact with BAM, VCF and FASTA) in our end-to-end tests.

I think we could resolve this by adding a new end-to-end test, that uses the test data repository and runs a bunch of queries on it. Rough outline:

- Create a SqlDataRepository based on ``tests/data/repo.db``, and a ``LocalClient`` that uses this repo.
- Run a bunch of queries, exercising as much of the query interface as possible on all the objects. It's fine for these queries to be exhaustive, we don't have that much data in the test repo. This should also verify with page sizes, ensuring we always get back the same result set for different page sizes.

This should probably be done in a series of PRs, as there might be quite a lot to do in one go. References seem like a good place to start. We will probably find that our test data is inadequate to test all the query interfaces, so we should update the test data as necessary.

kozbo:
@jeromekelleher It seems like the ideal way to do actual end to end tests is to use the repo manager to put data into the repository and then queries to pull data back out again. Is this what you are suggesting here or is the repo.db a pre-ingested database?

jeromekelleher:
This is true @kozbo, and we should also do what you're talking about I think. However, what I had in mind here was just using the pre created repo.db that is checked into Git. For this test I think we need to concentrate on stressing the query interface, so we get into all the little cracks in the data interfacing code.

But yeah, another end-to-end test where we bung a bunch of stuff that we know about into the repo using repo manager, and then do queries on the server via a local client to make sure that the stuff we expect to find is something we want.

dcolligan:
Fixed in #1219

PR #1219, opened 2016-05-06
New end-to-end-ish test for references

dcolligan:
Issue #1213 

@jeromekelleher not at all sure if this is what you were getting at

### DIFF
--- a/tests/unit/test_data_interface.py
+++ b/tests/unit/test_data_interface.py
@@ -1,0 +1,269 @@
+"""
+Tests the data interfacing layer for various objects
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import unittest
+
+import ga4gh.client as client
+import ga4gh.datarepo as datarepo
+import ga4gh.backend as backend
+import tests.paths as paths
+import tests.utils as utils
+
+
+class TestInterfacingLayer(unittest.TestCase):
+    """
+    Test fetching objects via the db and via the client return
+    identical results
+    """
+    def _assertEqual(self, dbObj, pbObj):
+        dbToPb = dbObj.toProtocolElement()
+        self.assertEqual(dbToPb, pbObj)
+
+    def _getPageSizes(self, num):
+        if num <= 0:
+            return []
+        return [None, 1, 2, 3, num - 1, num, num + 1]
+
+    def _testGetMethod(self, repoIteratorMethod, clientGetMethod):
+        repoObjs = repoIteratorMethod()
+        for repoObj in repoObjs:
+            id_ = repoObj.getId()
+            clientObj = clientGetMethod(id_)
+            self._assertEqual(repoObj, clientObj)
+
+    def _testSearchMethod(self, repoIteratorMethod, clientSearchMethod):
+        repoObjs = list(repoIteratorMethod())
+        numRepoObjs = len(repoObjs)
+        pageSizes = self._getPageSizes(numRepoObjs)
+        for pageSize in pageSizes:
+            self._client.setPageSize(pageSize)
+            clientObjs = list(clientSearchMethod())
+            for repoObj, clientObj in utils.zipLists(repoObjs, clientObjs):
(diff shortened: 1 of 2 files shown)


jeromekelleher:
This looks like a great start @dcolligan. I think we can build this out by splitting it up a bit, and adding tests for the ``search`` code paths. What we have here is a combined test for GET reference sets and GET references. We should split this into to separate methods, testGetReferenceSets, testGetReferences.

We can then add another method, testSearchReferenceSets. We can do something easy then like check that each returned object is valid, and then also check that we get back the same list of objects for a few different page sizes (say, 1, 2, 3, n - 1, n, n +1), where n is the number of reference sets.

david4096:
This will also close https://github.com/ga4gh/server/issues/1164, correct?

jeromekelleher:
Yes, this should also close #1164; thanks for keeping an eye on these @david4096!

dcolligan:
@jeromekelleher updated

tests/unit/test_data_interface.py
```diff
@@ -0,0 +1,70 @@
...
+        self._repo.open(datarepo.MODE_READ)
+        self._backend = backend.Backend(self._repo)
+        self._client = client.LocalClient(self._backend)
+
+    def testGetReferenceSets(self):
+        repoReferenceSets = self._repo.getReferenceSets()
+        for repoReferenceSet in repoReferenceSets:
+            referenceSetId = repoReferenceSet.getId()
+            referenceSet = self._client.getReferenceSet(referenceSetId)
+            self._assertValid(referenceSet, protocol.ReferenceSet)
```
jeromekelleher: There's no real point in asserting validity of the JSON any more as protobuf does this. Anyway, this is a bit weak: we should be testing for deep equality here. Is this _exactly_ the same reference set?

Same comment applies to the other tests.

dcolligan: Trying to determine the best way to do a deep compare... did we get rid of the capability to serialize a datamodel object to json?  Is there a convenient way to transform a datamodel object into the corresponding pb object?

jeromekelleher: ``toProtocolElement()``?

dcolligan: Also a deep compare would seem to require snake caseing all the attributes... otherwise how to compare a ncbi_taxon_id to a ncbiTaxonId?  Are we ready to do that?  Should we do that?

dcolligan: Ah, thanks.  I'm still getting the cobwebs out...

dcolligan:
@jeromekelleher updated... I've stubbed out the remaining client method tests... let me know if this looks good and I'll fill the rest of them out

jeromekelleher:
Perfect.

dcolligan:
@jeromekelleher updated

tests/unit/test_data_interface.py
```diff
@@ -0,0 +1,277 @@
...
+            self._client.getVariantAnnotationSet)
+
+    def testGetFeatureSet(self):
+        self._testGetMethod(
+            self._repo.allFeatureSets, self._client.getFeatureSet)
+
+    def testGetFeature(self):
+        repoFeatures = self._repo.allFeatures()
+        for repoFeature in repoFeatures:
+            repoFeature = repoFeature[0]
```
jeromekelleher: This looks odd --- why do we need to do this?

jeromekelleher:
Awesome, great work. I agree on shelving the Feature search problem for now. Let's merge this much and make an issue for whatever is happening there.

+1

tests/unit/test_data_interface.py
dcolligan: the features iterator in `sequenceAnnotations.py:Gff3DbFeatureSet.getFeatures` returns a (feature, nextPageToken) tuple ... it has the nextPageToken logic at the wrong layer of abstraction ... this is probably the same problem in the other test as well

dcolligan:
Created issue #1288

Issue #1179, opened 2016-04-28
``tests/unit/test_faultydata.py`` should be reinstated


jeromekelleher:
Assumes #1166 has been merged.

The ``tests/unit/test_faultydata.py`` is intended to check that we get the correct behaviour when we try to insert bad data into our repo. The tests currently in the module will need some refactoring, and should be reinstated. We should also aim to add quite a few more tests here, as dealing properly with dodgy inputs is one of the most important usability criteria for bioinformatics software, and will really help our users do their job better and more efficiently.

dcolligan:
Fixed in #1221

PR #1221, opened 2016-05-06
Re-enable faulty data tests

dcolligan:
Issue #1179

(Removed tests where the exception the test was expecting did not exist anymore)


tests/unit/test_faultydata.py
```diff
@@ -30,63 +29,47 @@ def getFullPath(self, localId):
...
     localIds = ["inconsist_meta"]
 
-    @unittest.skip("Skipping until DB based repo refactor is done")
     def testInstantiation(self):
         for localId in self.localIds:
             path = self.getFullPath(localId)
-            variantSet = variants.HtslibVariantSet(
-                self.dataset, localId, path, None)
+            variantSet = variants.HtslibVariantSet(self.dataset, localId)
+            variantSet.populateFromDirectory(path)
```
jeromekelleher: Should this exception be thrown at ``populate`` time? What makes the most sense from the perspective of running the CLI import? What happens when you try to run the CLI ``add-variantset`` when you have inconsistent metadata within a variant set?

We need to structure this so that we're testing the code in the production path.

jeromekelleher:
Looks good, thanks @dcolligan. I think we need to look a bit at how the code is being called in the production code path, and make sure that we're calling into the same hooks that are being used there.

tests/unit/test_faultydata.py
dcolligan: The exception is thrown whenever `checkConsistency` is called.  Should we move the `checkConsistency` calls to be run whenever data files are added?

jeromekelleher: Yeah, I'd say we should. Otherwise, this method is never called at all, right?

dcolligan:
@jeromekelleher rebased

jeromekelleher:
+1.

david4096:
Thanks @dcolligan +1!

PR #1224, opened 2016-05-12
Updated Dockerfile and Docker deployment process

macieksmuga:
Builds and configures SQL-repo-based GA4GH reference server from master branch instead of from a PiPY point release. This PR also simplifies the structure of the Docker repo (removes the "variant" dockerfiles for different base images, settles on an ubuntu standard). 

A start at a help file for using this updated Docker deployment process can be found at https://github.com/BD2KGenomics/ga4gh-integration/wiki/Setting-up-data-for-a-docker-instance


deploy/001-ga4gh.conf
```diff
@@ -0,0 +1,13 @@
+<VirtualHost *:80>
+    ServerAdmin webmaster@localhost
+    DocumentRoot /var/www/html
+    ErrorLog ${APACHE_LOG_DIR}/error.log
+    CustomLog ${APACHE_LOG_DIR}/access.log combined
+    WSGIDaemonProcess ga4gh python-path=/srv/ga4gh/ga4gh-server-env/lib/python2.7/site-packages python-eggs=/var/cache/apache2/python-egg-cache
```
david4096: Fix here, processes/threads and no virtual environment.

Dockerfile
```diff
@@ -1,21 +1,53 @@
...
 ## Configurable to use a local dataset
-## Based on mod_wsgi-docker
-## Results of this build are available from Dockerhub as afirth/ga4gh_server_apache:prod
 ############################################################
-FROM grahamdumpleton/mod-wsgi-docker:python-2.7-onbuild
+FROM ubuntu
 
-# File Author / Maintainer
-MAINTAINER Alastair Firth
+# Originally created by Steve Hershman GitHub @hershman
```
jeromekelleher: We should add a line giving credit to Alastair here also.

Dockerfile
```diff
@@ -1,21 +1,53 @@
...
+COPY deploy/application.wsgi /srv/ga4gh/application.wsgi
+COPY deploy/config.py /srv/ga4gh/config.py
+
+# Configure apache to serve GA4GH site
+WORKDIR /etc/apache2/sites-enabled
+RUN rm -f 000-default.conf && ln -s /etc/apache2/sites-available/001-ga4gh.conf 001-ga4gh.conf
+
+# Create a database for the server repository
+WORKDIR /srv/ga4gh/server
+RUN python repo_dev.py init /srv/ga4gh/repo.db
```
jeromekelleher: Use ``ga4gh_repo`` here rather than the development script.

Dockerfile
```diff
@@ -1,21 +1,53 @@
...
+
+# Open port 80
+EXPOSE 80
+
+# Start server when container starts
+CMD while true; do sleep 1; done
```
jeromekelleher: Why are we doing this?

jeromekelleher:
Looks good @macieksmuga, thanks for taking this on. Two questions:

1. Why are we cloning from git? Surely we'd want to use the stable released version from pypi for deployments?
2. How do we get data into this? Ideally I guess we'd want to pass in a registry DB as an argument, but there are still issues about how we still need some shared file system space.

macieksmuga:
@jeromekelleher to answer your questions:
1. Docker naturally sources the code from GitHub, there should be no reason to try to get the code from pypi - that subverts the close github/docker integration. But you're right that the `git clone` command there seems weird and redundant. I'm aiming to remove it if possible - then the docker container will be tagged with the corresponding release name as specified in github.

2. Please refer the documentation I wrote up on loading data into a dockerized server instance: https://github.com/BD2KGenomics/ga4gh-integration/wiki/Setting-up-data-for-a-docker-instance

Dockerfile
macieksmuga: I need the docker container to be running some continuous process to have it up as a daemon. Running Apache in the foreground (as before) was an issue, as _restarting_ Apache in the container upon loading in the data via the repo manager would cause the container to shut down. 

But I agree that what's there now is a total hack: I queried our local docker slack channel for a better solution to this problem. In the meantime I'll discuss this in the comment above.

Dockerfile
macieksmuga: I in principle, but I ended up not reusing any of Alastair's code, as far as I could tell.

jeromekelleher: I don't think it matters if we use the code or not, he still contributed to the development of it. You could argue that git tracks this perfectly well, so we don't need to ACK Steve either. I'm fine with either deleting the ACK entirely, or ACKing both Steve and Alastair.

jeromekelleher:
Thanks @macieksmuga, that all makes sense. Can you ping me when you want more review on this please?

Dockerfile
macieksmuga: Added him in :)

macieksmuga:
One further question mostly for @david4096 - I'm currently creating and managing the repo.db file INSIDE the docker container's filesystem. This means that if an updated docker image is pulled and run, the entire configuration has to be re-run. It's just as easy for me to create and maintain this repo.db file in the host's filesystem, linked to within the container via a mapping. Any good reason I _shouldn't_ do this?

jeromekelleher:
Sounds like the right thing to do to me @macieksmuga, keeping the configuration outside of the registry DB outside the container should make it a lot easier to administer --- we'll be wanting to do live updates on this in the not-too-distant future.

jeromekelleher:
Also, while we're thinking about the Docker setup @macieksmuga --- have you any thoughts about spinning this into its own repo? This might make CI easier to manage.

david4096:
I don't think the host should be expected to have GA4GH installed, that defeats the point of containerization. If we suggest that pattern as normal practice we introduce the complexity on the administrator that the versions of their manager and the software in the instance remain in line. If you add to this the expectation that relative paths are expanded, working from outside the container will likely be pointing at the incorrect path.

jeromekelleher:
Hah, yes, excellent point @david4096! Best keep the DB within the container then.

macieksmuga:
@david4096 Storing the repo DB outside the container does not in any way require any GA4GH code to live on the host – I think there's a misunderstanding here. Additionally, I agree with @jeromekelleher's original sentiment regarding live updates, which would be made considerably easier if the repo DB lived outside the container.

macieksmuga:
@david4096 @jeromekelleher this branch is now ready for final review - I updated the setup so that 

1. The repo.db is stored on the host's system: Thus, updating the ga4gh server image with the latest version will preserve all the host's configuration and metadata info.
2. Removed that hacky keep-alive loop at the end of the `Dockerfile`, instead relying on multiple separate runs of the image, one run for each repo configuration command, and a final run to spin up the fully configured server in daemon mode.

jeromekelleher:
Looks good to me @macieksmuga.

+1 after the commits are squashed.

david4096:
+1 merging, I was able to create the docker image, use the repo manager from outside the docker instance, and run some queries against a variant set.

To note, the permissions on the host system need to allow for read and write by the user within the docker instance. The docker instance was able to read the UID from the host's filesystem, so this may not be as simple as adding `www-data` group permissions on the host system. https://github.com/BD2KGenomics/ga4gh-integration/wiki/Making-a-docker-image https://github.com/BD2KGenomics/ga4gh-integration/wiki/Setting-up-data-for-a-docker-instance

PR #1225, opened 2016-05-12
Ontologies revisited

jeromekelleher:
Updates discussed in #1147.

This is WIP and not yet ready for review. I'm posting this PR so people can see that the work is ongoing.

Comments are welcome though!


jeromekelleher:
Ready for review!

This update makes significant changes to ontology handling:

1. Change to using an OBO file as input, and using an Ontology object
   internally rather than a OntologyTermMap. Refactored ontology
   handling code in general.
2. Add a ontologyId field to the FeatureSet and AnnotationSet tables
   with a foreign key constraint on the Ontology table.
3. Add a -O 'option' to the add-featureset command which allows an admin
   choose the ontology they wish to associate with a given FeatureSet.
4. Changed the semantics of adding variant annotations sets so that it
   is now optional we want to add a VAS when annotations are detected.
   Now an annotation set will only be created if the -a option is
   provided. Also added the -O option to add-variantsets so the ontology
   can be specified.
5. Added the scripts/build_test_data.py scipt to build test data db.

**TODO** (In a follow up PR)

- Make the UI a bit slicker so that we don't force the admin to choose an ontology when there's only one SO instance on the server.

This change will affect all of us (particularly @macieksmuga and @david4096), so please review the code and try out the UI.

jeromekelleher:
Review please...

macieksmuga:
Reviewing :) there's a bunch of background stuff I was reading up on...

jeromekelleher:
As a workaround to avoid the very heavy dependencies that goatools pulls in, I've copied the source of ``obo_parser.py`` from goatools into the source here. I _think_ this is OK license-wise. As far as I can tell, goatools is BSD licensed, and (according to the Apache foundation) using a BSD licensed file within and Apache licensed project is OK: 

http://www.apache.org/legal/resolved.html#category-a

Not entirely sure that this is the right thing to do here, but it seems better than pulling in scipy, numpy, statsmodels and pandas as dependencies for no reason.

Anyone have opinions on this?

david4096:
Great to see the custom ontology format go. I opened this PR against your branch https://github.com/jeromekelleher/server/pull/6 with the two things I found. `prepare_compliance_data` needed to be updated (along with the test data in compliance https://github.com/ga4gh/compliance/pull/191), and the Fisher package required numpy, which had not yet been installed. +1 when these changes, or some version of them, are accepted!

```
Collecting fisher (from goatools==0.6.4->-r requirements.txt (line 9))
  Downloading fisher-0.1.4.tar.gz (45kB)
    100% |████████████████████████████████| 45kB 13.3MB/s
    Complete output from command python setup.py egg_info:
      File "<string>", line 20, in <module>
      File "/private/var/folders/34/qtyrnvz12l13zc9nmltj_c3r0000gp/T/pip-build-vyK9BB/fisher/setup.py", line 2, in <module>
        import numpy as np
    ImportError: No module named numpy
```

jeromekelleher:
Great, thanks @david4096! I think you must have pulled an out-of-date version of this PR when you were checking it, as I've taken out the dependency on goatools (which pulled in all sorts of crazy stuff, as you saw) and just copied the actual parser code in (see the last comment above). I'll update the PR to include your changes to the compliance data.

jeromekelleher:
@macieksmuga, any thoughts here? We really want to release the server soon, and this PR is the main thing we need to merge first.

@kozbo, any thoughts on including the source of obo_parser.py in our source tree? It's not something I'm hugely keen on, but it seems like the lesser of a bunch of evils.

kozbo:
@jeromekelleher I feel the same way as you I think, that this is a uncomfortable work-around to the problem, but it is the less uncomfortable than pulling in all the extra libs. Deconstructing libs runs the risk of losing the advantage of easy updates, but I am not sure we will need any other functionality here. The code you brought in is basically the same as what we would have to write ourselves, and the licensing seems OK, so I am fine with this approach. +1

macieksmuga:
Tested and works as promised. I'm not convinced this is the best solution to the ontology issue - perhaps it would be better to have some interface to an external ontology service - but it's an improvement on what's there now.

jeromekelleher:
@macieksmuga, agreed, we can look at adding remote ontology support in the future. No reason we can't do it. However, I think a local OBO will be a good option for a lot of users, so we should support that too.

OK, great. Since we've got two +1s and +0 (as I'm interpreting @macieksmuga's vote) I think we should merge and get ready to release.

PR #1236, opened 2016-05-20
Ontology docs

jeromekelleher:
1. Updates docs for ontologies.
2. Bumps version number for example data so that demo will still work.

Minor refactor of the docs to put `list` before `verify`.


david4096:
+1 for documentation PR merging

PR #1237, opened 2016-05-20
Fix for layering violations in VA code.

jeromekelleher:
The variant annotations code was adding extra fields to the protocol
elements and using these for control flow. This results in serious
problems for protobuf which does not allow this. These changes fix the issue
by passing a (variant, annotation) pair around.

This should not be merged until after 0.3.0 is tagged.


jeromekelleher:
This has already been merged into the protobuf branch, so I'm going to close this PR.

PR #1240, opened 2016-05-23
Force absolute path when adding close #1204

### LINKED ISSUE #1204: SQL repo needs absolute paths
The full path must be used when adding a file because the server may be running from a different location than the database. Even if the file exists according to `ga4gh_repo`, without an absolute path the installed server may not be able to find it.

I made this mistake adding a feature set to the 1kgenomes server and the failure wasn't detected leading to a 500 when issuing requests against that featureset.

david4096:
Forces absolute path when adding a file to the registry. Also test that absolute paths are returned even when relative paths are provided.

### DIFF
--- a/ga4gh/cli.py
+++ b/ga4gh/cli.py
@@ -1630,10 +1630,11 @@ class RepoManager(object):
         """
         self._openRepo()
         name = self._args.name
+        filePath = os.path.abspath(self._args.filePath)
         if name is None:
-            name = getNameFromPath(self._args.filePath)
+            name = getNameFromPath(filePath)
         ontology = ontologies.Ontology(name)
-        ontology.populateFromFile(self._args.filePath)
+        ontology.populateFromFile(filePath)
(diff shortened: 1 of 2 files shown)


ga4gh/cli.py
```diff
@@ -1741,7 +1748,7 @@ def addVariantSet(self):
             indexSuffix = ".tbi"
             # TODO support BCF input properly here by adding .csi
             indexFiles = [filename + indexSuffix for filename in dataUrls]
-
+        indexFiles = map(lambda x: os.path.abspath(x), indexFiles)
```
jeromekelleher: You don't need the lambda function here, just use
```python
indexFiles = map(os.path.abspath, indexFiles)
```
There's a few more instances of this below too.

Nice use of functional tools though!

jeromekelleher:
Great PR @david4096 --- I would have an easy job if all PRs were this clean and tidy!

+1

macieksmuga:
+1 Thanks @david4096!

PR #1242, opened 2016-05-25
Convert avro to protobuf on the server

irene-lshift:
This updates the protobuf work done previously with the latest changes from master.


requirements.txt
```diff
@@ -29,20 +29,39 @@ future==0.15.2
...
 oic==0.7.6
 pyOpenSSL==0.15.1
 lxml==3.4.4
+protobuf==3.0.0b1.post1
+# Requirements for test OIDC provider.
+cherrypy==3.2.4
+pyaml==15.03.1
+Jinja2==2.7.3
+yubico-client==1.9.1
+pyjwkest==1.0.1
```
david4096: Development dependencies and installation dependencies have been split into two files. This file currently contains duplicate entries causing the `pip install -r requirements.txt` to fail with double requirement errors. The entries only needed for development should be moved to the other file: `dev-requirements.txt`.

david4096:
Hi @irene-lshift would you please make this PR against the protobuf topic branch? We would like to be able to accept your changes without changing the state of master.

tests/datadriven/test_variant_annotations.py
```diff
@@ -180,10 +180,10 @@ def _verifyVariantAnnotationsEqual(
...
                 for treff, gaEffect in zip(
                         transcriptEffects,
-                        gaVariantAnnotation.transcriptEffects):
-                    self.assertEqual(gaEffect.alternateBases, treff['alt'])
-                    self.assertEqual(gaEffect.featureId, treff['featureId'])
-            self.assertIsNotNone(gaVariantAnnotation.transcriptEffects)
+                        gaVariantAnnotation.transcript_effects):
+                    self.assertEqual(gaEffect.alternate_bases, treff['alt'])
+                    self.assertEqual(gaEffect.feature_id, treff['featureId'])
+            # self.assertIsNotNone(gaVariantAnnotation.transcript_effects)
```
jeromekelleher: Delete this? Can't be None through proto I guess?

tests/unit/test_backends.py
```diff
@@ -190,24 +190,24 @@ def toProtocolElement(self):
                 return self
 
         self.request = FakeRequest()
-        self.request.pageToken = None
-        self.numObjects = 3
-        self.objects = [FakeTopLevelObject() for j in range(self.numObjects)]
+        self.request.page_token = None
```
jeromekelleher: Shouldn't this be the empty string?

tests/unit/test_simulated_stack.py
```diff
@@ -16,6 +16,9 @@
 import ga4gh.datamodel.sequenceAnnotations as features
 import ga4gh.frontend as frontend
 import ga4gh.protocol as protocol
+from ga4gh.allele_annotation_service_pb2 import \
```
jeromekelleher: These two imports shouldn't be needed; just use ``protocol.SearchVariantAnnotationSetsResponse`` and ``protocol.GAException``

jeromekelleher:
Looks great @irene-lshift, thanks! I have some very minor comments above. 

I agree with @david4096, let's remake this PR against the protobuf branch so we can merge it quickly without worrying about getting a wider consensus. My view would be that we should move over to this code ASAP though.

irene-lshift:
Yes we can make this PR against the protobuf brunch. It just might be a bit harder to review due to all the new code that's been merged from master.

tests/unit/test_backends.py
irene-lshift: Yes it should, I was wondering how does this work (as protobuf doesn't allow None assignments and that'd normally be an error), but here we're using a FakeRequest object instead. I'll update it.

tests/datadriven/test_variant_annotations.py
irene-lshift: Yes, this both can't be None and also gives an error "TypeError: Can only compare repeated composite fields against other repeated composite fields."
I'll replace this with an assertion over the length of this list.

tests/unit/test_schemas.py
```diff
@@ -1,390 +0,0 @@
...
-        for cls in protocol.getProtocolClasses():
-            for member in cls.__slots__:
-                if cls.isEmbeddedType(member):
-                    instance = cls.getEmbeddedType(member)()
-                    self.assertIsInstance(instance, protocol.ProtocolElement)
-                else:
-                    self.assertRaises(KeyError, cls.getEmbeddedType, member)
-
-
-class SearchResponseBuilderTest(SchemaTest):
```
david4096: I believe some of these tests that are not specific to Avro should be maintained. SearchResponseBuilder is still used in this PR.

irene-lshift:
The change I made to the requirements.txt / dev-requirements.txt seems to have made Travis unhappy (even though passing locally), I'm taking a look at it.

jeromekelleher:
Thanks @irene-lshift, this is great. I agree that it will be harder to review on the protobuf branch, so it was very helpful to make this PR. I'm happy to merge this into the protobuf branch now.

@david4096, @macieksmuga, @dcolligan, @diekhans, @kozbo, what do you think? There's some work to be done to bring this up to release level --- performance is poor (but it's pretty obvious why, and it should be easy to fix), and there's some hard questions to answer about installation, given that proto3 isn't on pypi yet, as far as I know.  The sooner we deal with these problems the better, and having the code in master will avoid more painful merges.

tests/unit/test_schemas.py
jeromekelleher: Yes, these are pretty good tests actually, we should keep them. I think they can be ported fairly easily by replacing getTypicalInstance with by just instantiating the class_(), and getting rid of ``getRandomInstance()``. So, whereever you see ``getTypicalInstance(class_)`` replace it with ``class_()``, and delete any getRandomInstance you see. That should make ``SearchResponseBuilderTest`` fairly portable.

david4096:
@jeromekelleher Re; protobuf https://pypi.python.org/pypi/protobuf/3.0.0b2

macieksmuga:
The obvious performance issue I noticed is the call to re-generate the full response JSON on _every_ call to `addValue` in the `SearchResponseBuilder`, so that its size can be compared against `_maxResponseLength`. I'd put a high priority on optimizing this bit, as it has a serious effect on search requests with larger page sizes.

kozbo:
@jeromekelleher I agree that #1243 should be our top priority. We saw the performance issues yesterday when @david4096 ran some quick tests but I haven't seen a show stopper yet. @macieksmuga found a likely culprit as mentioned earlier. Is that the same "obvious" reason that you saw? Anyway, I don't see any reason not to merge and then optimize, as far as performance goes. I will leave the more detailed code review to those who are more qualified.

macieksmuga:
@kozbo agree. Let's merge first into the ga4gh/server:protobuf branch, then optimize there.

david4096:
I've been able to confirm that the compliance tests are passing. The failure is a known unimplemented feature.
screen shot 2016-05-25 at 11 36 01 am

The requirements.txt changes also worked for me. Great!

jeromekelleher:
@macieksmuga, yes this is exactly the problem. Converting to JSON for every element added is hugely expensive. We'll be able to sort this out in no time, I'm sure.

@david4096, thanks this is fantastic news that protobuf is in pypi! Given this, I think we should be able to get this code into a releasable state in a matter of days. 

So, it looks like we have consensus then? Let's merge this into the protobuf branch as soon as the final few wrinkles have been ironed out in this PR, fix the performance bottleneck (in a follow up PR), and then merge to master. Hopefully we can merge this into the protobuf branch tomorrow.

jeromekelleher:
@irene-lshift, we definitely need to be at the latest protobuf version. The older ListValue JSON was incorrect in this version of protobuf. We need to fix any tests that are failing under this version.

jeromekelleher:
@irene-lshift OK, I see the problem here. We just need to rerun the ``scripts/build_test_data.py``. This has the older JSON embedded inside in the DB, which is causing failures. I've just run it there and it all works fine.

jeromekelleher:
Re performance, replacing 
```python
return len(toJson(self._protoObject))
```
with
```python
return self._protoObject.ByteSize()
```
dramatically improves things, and brings us back to where we were before more-or-less (I think). We'll need to rephrase the documentation a bit, but this gives us the same functionality (i.e., prevent memory starvation DOS attacks, provide a lever to approximately manage latency when the client doesn't provide a page size), and is cheap. It's certainly good enough to keep us going for now.

Anyway, this doesn't directly reflect on the PR, I just wanted to see what the effect of this change would be.

irene-lshift:
Thanks @jeromekelleher, I updated to the latest version of protobuf and reran `scripts/build_test_data.py`

jeromekelleher:
Great, thanks @irene-lshift. How does porting SearchResponseBuilderTest look? I think we're ready to merge into the protobuf branch after this is done.

Also, would you mind squashing down the last bunch of commits please? There's no need to go through all of them, just the last group back to @peterolph's pep8 fix. These should squash cleanly. It's not worth the bother of squashing older commits.

irene-lshift:
I tried to find out why was this file originally deleted and found this earlier conversation: https://github.com/ga4gh/server/pull/799#discussion_r48847207

It seems that this used to test the integrity of our custom serialisation / deserialisation, but protobuf guarantees this?

jeromekelleher:
Yes that's true @irene-lshift, most of the tests just check the integrity of the conversion process. However, I missed the ``SearchResponseBuilderTest`` when looking at it the first time, which tests the process of building response objects. It would be great if we could port this test to using protobuf. Apologies for going back on what I said in #790.

Given the notes about about getting rid of  ``getTypicalInstance(class_)`` and ``getRandomInstance`` I think it should be a quick job. If it is a lot of work, we can add an issue on it, and port the test later.

irene-lshift:
We restored `SearchResponseBuilderTest`, how does this look?

jeromekelleher:
It looks great, thanks @irene-lshift!

+1 to merge to the protobuf branch. 

I realise now that I got confused, and that PR #1242 has different contents to this one. Just so I'm clear: if we merge #1242, then we'll be able to merge this PR onto the protobuf branch? Can we just merge the contents of this PR onto the protobuf branch, or do we need this two-step process?

jeromekelleher:
Nope, looks like I got confused about being confused! Ignore the last comment. OK, since we're done reviewing this code, I think we can close this PR and move over to #1242. @kozbo, would you mind doing this, if you agree we're ready?

jeromekelleher:
#1243 has been merge so I'm going to close this PR. Thanks to everyone involved, we've nearly completed the Avro purge!

PR #1244, opened 2016-05-26
Proto perf

jeromekelleher:
Brings server_benchmark up to date, and improves response generation efficiency for protobuf.

Before:
```
jk@holly$ python scripts/server_benchmark.py --profile none WyIxa2ctcDMtc3Vic2V0IiwidnMiLCJtdm5jYWxsIl0
1.929188
```
After:
```
jk@holly$ python scripts/server_benchmark.py --profile none WyIxa2ctcDMtc3Vic2V0IiwidnMiLCJtdm5jYWxsIl0
0.392116
```

### DIFF
--- a/scripts/server_benchmark.py
+++ b/scripts/server_benchmark.py
@@ -11,17 +11,20 @@ import pstats
 import argparse
 import cProfile
 
-import ga4gh.backend as backend
-import ga4gh.protocol as protocol
-import ga4gh.datarepo as datarepo
-
 import guppy
+
+import utils
+utils.ga4ghImportGlue()
+import ga4gh.backend as backend  # noqa
+import ga4gh.protocol as protocol  # noqa
+import ga4gh.datarepo as datarepo  # noqa
 
 
 class HeapProfilerBackend(backend.Backend):
-    def __init__(self, dataDir):
-        dataRepository = datarepo.FileSystemDataRepository(dataDir)
-        super(HeapProfilerBackend, self).__init__(dataRepository)
+    def __init__(self, registryDb):
+        repo = datarepo.SqlDataRepository(registryDb)
+        repo.open(datarepo.MODE_READ)
+        super(HeapProfilerBackend, self).__init__(repo)
         self.profiler = guppy.hpy()
 
     def startProfile(self):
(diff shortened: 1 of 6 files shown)


jeromekelleher:
Can I get some review please @macieksmuga, @david4096, @dcolligan?

scripts/server_benchmark.py
```diff
@@ -11,17 +11,20 @@
...
-import ga4gh.datarepo as datarepo
-
 import guppy
 
+import utils
+utils.ga4ghImportGlue()
+import ga4gh.backend as backend  # noqa
+import ga4gh.protocol as protocol  # noqa
+import ga4gh.datarepo as datarepo  # noqa
+
```
macieksmuga: I know that #noqa cuts out flake8 warnings, but why is this necessary here?

scripts/server_benchmark.py
```diff
@@ -11,17 +11,20 @@
...
 
 class HeapProfilerBackend(backend.Backend):
-    def __init__(self, dataDir):
-        dataRepository = datarepo.FileSystemDataRepository(dataDir)
-        super(HeapProfilerBackend, self).__init__(dataRepository)
+    def __init__(self, registryDb):
+        repo = datarepo.SqlDataRepository(registryDb)
+        repo.open(datarepo.MODE_READ)
+        super(HeapProfilerBackend, self).__init__(repo)
```
macieksmuga: I assume this is general cleanup, not immediately related to the `maxResponseLength` -> `maxBufferSize` issue?

macieksmuga:
Looks good and necessary, but I see some fixes not directly related to the performance fix. Perhaps bundle this PR together with #1245 as a general miscellaneous fixes PR?

scripts/server_benchmark.py
jeromekelleher: It's because of the ``utils.ga4ghImportGlue()`` call. Pep8 doesn't want you to put in imports after code.

scripts/server_benchmark.py
jeromekelleher: This is part of the first commit, which brings server_benchmark up to date. The PR makes more sense if you look at the commits separately. I didn't see much point in making two PRs, as this would slow things down. I should have made it more clear that there was two separate things happening here though.

jeromekelleher:
I don't think we gain much from bundling this together with #1245, these are two closely related changes. I couldn't profile performance until I brought server_benchmark up to date, and the second change uses this benchmark to show that performance has been greatly improved.

macieksmuga:
+1 Sounds good! - if you don't mind quickly updating this branch, I'll merge it in.

jeromekelleher:
Thanks @macieksmuga, updated.

PR #1245, opened 2016-05-26
Tidy up various protobuf loose ends.

jeromekelleher:
- Bump the DB version number as the JSON stored in the DB is not 100%
  compatible.
- Update the example data and version number in the docs.
- Update the test DB with new version number.

After this and #1244 are merged to the protobuf branch, I see no reason not to merge into master, as we should release-ready.


kozbo:
+1

macieksmuga:
+1

PR #1255, opened 2016-06-03
Fixes #715

### LINKED ISSUE #715: Add simulated VariantSetMetadata and tests
The ga4gh.backend.SimulatedBackend class is very useful as it allows us to run a server without any data. We can simulate arbitrarily large data sets which is essential for testing. The SimulatedBackend is used extensively for testing, and allows us to test corners of the API which might be difficult to exercise using file based data.

The simulators are currently quite limited, however, Here, we want to add some simple metadata to our simulated variantsets and callsets. These are defined in the [schemas](https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/variants.avdl).

- Edit ``ga4gh/datamodel/variants.py``, and find ``SimulatedVariantSet``. Add some simulated metadata. The metadata items don't have to be particularly realistic, just try to give a reasonable coverage of the sorts of data we might expect to see in these fields. 

- Put the simulated metadata in ``self._metadata``, and remove the getMetadata function from the SimulatedVariantSet class. Move the definition for ``getMetadata in the HtslibVariantSet class to the AbstractVariantSet class, and document following the usual conventions. Make sure that ``self._metadata`` is defined in the contructor for ``AbstractVariantSet``. (This is a refactoring cleanup) 

- Edit ``tests/unit/test_simulated_stack.py`` and find ``verifyVariantSetsEqual``. Update to check the metadata on the backend object. 

- Also add some simulated values to the ``info`` field on CallSets, and add tests to make sure it's correctly translated.

andrewjesaitis:
- Adds metadata for simulated VariantSets
- Adds info onto Callset object

It looked like metadata was being added to the VariantSet twice (variants.py:220-222). I removed the later, but I wasn't sure if the intent was exactly the same as in the extend call above. In any case, how it was would fail in testing.

Also, double check if how I'm adding the info dict onto CallSet is how you want to do this. I wasn't really sure what you had in mind, so I just took a stab at it by building out the CallSet class with the field.

### DIFF
--- a/ga4gh/datamodel/variants.py
+++ b/ga4gh/datamodel/variants.py
@@ -42,6 +42,10 @@ class CallSet(datamodel.DatamodelObject):
     """
     compoundIdClass = datamodel.CallSetCompoundId
 
+    def __init__(self, parentContainer, localId):
+        super(CallSet, self).__init__(parentContainer, localId)
+        self._info = {}
+
     def populateFromRow(self, row):
         """
         Populates this CallSet from the specified DB row.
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Great cleanup, thanks @andrewjesaitis!

+1

PR #1258, opened 2016-06-07
RNA API server implementation

macieksmuga:
Corresponds to schemas PR https://github.com/ga4gh/schemas/pull/630


ga4gh/backend.py
```diff
@@ -654,6 +656,82 @@ def featureSetsGenerator(self, request):
             request, dataset.getNumFeatureSets(),
             dataset.getFeatureSetByIndex)
 
+    def rnaQuantificationGenerator(self, request):
+        """
+        Returns a generator over the (rnaQuantification, nextPageToken) pairs
+        defined by the specified request.
+        """
+        dataset = self.getDataRepository().getDataset(request.dataset_id)
+        if len(request.rna_quantification_id) is 0:
```
jeromekelleher: It doesn't matter in this case (0 is a singleton), but it's generally better to use '==' when comparing numbers.

jeromekelleher:
I haven't checked this out and played with it, but the code looks  excellent. I think this will have no problems merging into master, once the final schema details have  been ironed out.

kozbo:
+1

tests/datadriven/__init__.py
```diff
@@ -277,6 +277,7 @@ def assertValid(self, protocolClass, jsonDict):
         Asserts that the specified JSON dictionary is a valid instance
         of the specified protocol class.
         """
+        print(jsonDict)
```
dcolligan: this should be taken out

dcolligan:
`serverconfig.py` should probably be modified to include variables for RNA, and those should be passed through to the simulator (even if we aren't doing complex simulation testing at the moment, it's still good to have a useful bare-bones simulator)

ga4gh/backend.py
```diff
@@ -316,6 +316,8 @@ def __init__(self, dataRepository):
         self._defaultPageSize = 100
         self._maxResponseLength = 2**20  # 1 MiB
         self._dataRepository = dataRepository
+        self._featureIdMap = {}
```
dcolligan: these two lines should be taken out for now... they're not used anywhere AFAICT

saupchurch:
+1

dcolligan:
I'm +1, but we'll need to merge master in before merging this into master, of course: #1350

dcolligan:
Ok, merge is done

jeromekelleher:
+1

PR #1259, opened 2016-06-07
SqlAlchemy POC

jeromekelleher:
This is a very early work in progress, and not intended for review. The PR is intended to show the direction that the SQLAlchemy refactor is going in, and hopefully get some early feedback from those that are interested.

The main idea is that we pull out our clunky object model for all of the objects in our registry, and use the SQLAlchemy library to do the heavy work for us. 

## What works

- Rough outline model of Datasets, ReferenceSets and References.
- Polymorphism within the DB for HtslibReferenceSets and SimulatedReferenceSets. We can store either in the DB, and we instantiate the correct object at run time when we need to access information for a specific information. This is the key requirement for more flexible storage backends.
- Search and GET requests for references, referencesets and datasets. The paging algorithm has been reworked to use SQL natively, and is much cleaner.
- Native support for multiple databases for the registry. I've tested this on Sqlite and Postgres and it works great on both. 
- Updated the repo manager CLI to use the new interface for the subset of data types supported. The repo manager now takes a DB URL in the form of a SQLAlchemy connection string. 
- Online updates to the DB. Updates to the registry from the CLI are now reflected in the server without needing a server restart.

## TODO

- A fair bit of refactoring and shuffling around of code.
- Complete the rest of the data model (ReadGroupSets, VariantSets, etc).
- Fix test cases.
- Add some way of testing the postgres support using Travis.

## Questions

- Do we really need to worry about ID collisions between tables? I.e., does it actually matter if we have a  Dataset and a VariantSet with the same ID? I don't think we need uniqueness to fulfill the requirements of the server, and it adds a fair bit of complexity ensuring that they are unique.
- How should we model things like ReferenceSet.source_accesssions? At the moment we embed some JSON into the table and don't worry about it. However, the right thing to do here would seem to be to model it with a table. It would surely be more forward-compatible to do this, if we ever need to get clever about accession numbers. 
- More generally, do we go for a full-on SQL model of all objects (like Experiment, VariantSetMetadata, etc), or do we continue using embedded JSON stored in a table?


diekhans:
Reactions:

 - If we are going to store data as well have registry database,
   would a document store make more sense?  Then again, postgress
   has a document store facility:

       - https://blog.andyet.com/2016/02/04/postgres-9.5-document-store-hybrid/

 - If we are storing GA4GH record in blobs, it would be good to
   research storing the binary protobuf.  This would reduce the
   size and might be faster to decode into memory.

 - We need to have a design that handles being a REST front-end
   to the UCSC Browser database.  The UCSC Browser database will
   not map well to SqlAlchemy, as it mostly highly denormalized
   and doesn't use foreign keys.  It's closer to htslib in SQL
   than a true relational database.  It's doubtful that we can
   afford to do an ETL of the subset of the data that can be
   represented in GA4GH.   SqlAlchemy can't be the abstraction
   layer if we are going to support this.

We need to know the goals before we decided on the
implementation.

andrewjesaitis:
One thing that is worth thinking about is that Postgres (and SQLAlchemy) support JSON fields so you can dump python dicts into the field and when you pull them out they are proper dictionaries. Plus you can index into them which is handy.

I think the big trade-off is that the datamodel is no longer explicit so if you are joining on them, you lose a lot of the performance that relational dbs offer in query planning. It's probably not a huge deal for references directly since I doubt references will ever number more than a few hundred/thousand (though is could be a concern if query via the accession for other models). Also lost is the ability to know what attributes define the object and which are required and which are optional. Thus, the code working with this data needs to be robust.

So I would say favor tables for object with defined set of keys and where the values would be used in queries, but in cases where it is strictly metadata storage dump it into a json field.

Overall, I'm a big fan of this, but don't know enough about the UCSC Browser schema that @diekhans mentioned to know what the overhead of the transform would be.

jeromekelleher:
Thanks for the feedback @diekhans and @andrewjesaitis, this is very helpful.

@andrewjesaitis, I tend to agree with you. I think we should lean towards using tables where possible in the interest of forward compatibility as well. We may not have queries on parts of the data model at the moment, but hopefully we will have them in the future. It'll be a lot easier to accommodate this if we already have the information we need in tables.

@diekhans One thing that I should clear up here is that SqlAlchemy is _only_ for the registry. The actual implementations of the data sources for the particular objects (e.g. VariantSet) is separate. This is an important separation of concerns for me; we a well defined SQL based approach to manage the high-level data model, as this is what it excels at. For bulk data storage, there's any number of other things we can use. I don't think we should try to do both (metadata/data model management and bulk data storage) at the same time.

In this new approach, we can have any number of different types of (e.g.) VariantSet. Right now we have SimulatedVariantSet, and HtslibVariantSet, and the polymorphism between these is taken care of by SqlAlchemy. At query time, we create an instance of the correct type of object, which has all the information it needs to be able to run the query. So, for example, we might have UcscBrowserVariantSet which has a URL to access the information, and whatever IDs are necessary to map into the remote data model. Paging will be done at the level of the data source, so if there is a mapping  between GA4GH objects and the data source we should be able to accommodate it. 

So, this change is _not_ about changing how we access bulk data within a set, it's about changing how we store and access information about the sets themselves. It's also about creating a framework to allow us to be vastly more flexible about where and how we obtain the data that's within these sets. Does this make sense?

diekhans:
Oh, and on the id.  From the API perspective, I have never been
happy that id is server-local (really more like a pointer) and
that the true object data is is this ill-defined name think.

It's probably the right thing to do, but feels like that API
punted on a very important bioinformatics issue.

It makes me nervous to no have id unique within a server and
require context.  Maybe I am just being a purist.

How about explain the id trades off a bit more, maybe we can
come up with a good solution.

diekhans:
Yes, this makes a lot of sense...

I haven't looked at the change, so the object boundaries were
not clear.  This makes a lot sense for containers.

jeromekelleher:
Excellent, glad to hear it @diekhans! 

Yeah, I hear you on the ID front. As I see it, there are two approaches we can take for ensuring repository-wide unique IDs:

1. Create a global ID table for all the externally visible objects in the registry, and rejection sample until we get a unique one when we're creating a new object. Pros: easy; cons: an extra table, and maintenance overhead (we have to remember to delete the IDs from this table).

2. Create a unique prefix for each externally visible object type in the registry. Since we'd like to use integers for IDs if at all possible (within the registry, at least), we could use the top 4 bits of the 32bit value for this. Pros: no extra overhead, and easy to maintain; Cons: mildly tricky to implement.

I'm assuming at the moment that individual IDs should be generated randomly to avoid any potential information side channels in the future. Since we're going to be storing sensitive data I think we should take this seriously.

diekhans:
Although approach is to use UUIDs.  Although it does generate
128 bit ids, I am not sure how big this impact actually is. .

They are not guaranteed to be unique, but I think we can live
with the probably of the server group being destroyed by a meteor
being hight than a collision.

The high-bit approach sits well with me.  It might be cleaner to
do it as two columns that form the key.

david4096:
Am I right to understand that the compound ID scheme will go away with this PR, at least for objects where we haven't required it?

jeromekelleher:
Some progress made:

1. Decent support for metadata across references, variants and reads.
2. Good support for variants with _vastly_ simpler paging algorithm.
3. Partial support for reads using same paging algorithm. (I haven't fully convinced myself this is correct yet though...)

Here's the current state of the registry DB schema. I've tried to follow the letter of the protocol here for forwards compatibility, except for one or two minor cases. It would be great if SQLers could have a look at this to make sure it all makes sense! 

dbschema

jeromekelleher:
@david4096 Yes, CompoundId almost entirely goes away now. We still need it for variants and reads, but we'll be able to simplify the code a lot. Maybe 128 UUIDs are the way to go, but it just feels a bit wasteful to be joining on such large IDs. I can definitely be convinced here though.

diekhans:
I hope any require compound id support still gets handle through
polymorphism vs having to separate id paths.

Given all the advantages of UUIDs, I would like to see how much
impact they have for rejecting them based on worry.  If stored
at a number rather than a string, they are represented as two words
on a 64-bit computer.

I normally didn't worry about going from an one to two word
representation on a 16-bit computer which were orders of
magnitude slower.  Does it really matter here?

jeromekelleher:
> I hope any require compound id support still gets handle through
> polymorphism vs having to separate id paths.
Not sure I follow here @diekhans, can you elaborate a bit? Basically we have VariantCompoundId and ReadCompoundId which contain enough information to uniquely identify a variant/read within a respository. These are used in two places: (1) for GET variant; and (2) in the new paging algorithm for reads/variants.

>Given all the advantages of UUIDs, I would like to see how much
> impact they have for rejecting them based on worry.  If stored
> at a number rather than a string, they are represented as two words
> on a 64-bit computer.
It's not quite as simple as that though, as SqlAlchemy doesn't have a native 128 bit number type, so we'll have to do some fiddling around to make it work across DBs. Not necessarily a big deal, but there should be some real benefit before taking on this complexity. Being able to treat IDs directly as ints in Python code rather than bytes is handy as well. What are the benefits of UUIDs, other than making the probability of collision small enough that it can be disregarded (which is probably also true of a uniformly generated 64 bit number)?

Anyway, it'll be easy to change this once we have the code working. All external ID columns are allocated by a single function, so it'll be simple to change it. Right now we're just generating a random 31 bit value and not worrying about collisions. We should definitely look at this properly when the rest of the code is up to scratch.

diekhans:
point #1 is just to keep an object for an id, as opposed to
baking any assumptions into the code

point #2 the benefit of UUIDs is they can be used cross table,
database, servers with some centralized authority to allocate
them.  I have never seen an random number algorithm that can do
this with 64-bit numbers.  Although I discussion of 64-bit host
ids generated with timestamps, I would like to see an academic
paper on an algorithm before trusting it.  Also, UUIDs are the
defacto standard why to generate unique ids.

although #1 hides it all anyway.

Python can handled both 128-bit ints and has a native UUID type.
Postgress has a UUID type, sqllite and mysql can have int(128).
It probably breaks on Sql Server  ;-)

See
        http://docs.sqlalchemy.org/en/latest/core/custom_types.html#backend-agnostic-guid-type

on how to do a custom UUID type.

The disadvantage of UUIDs is they you don't have any information
to identify where the object is (table, file line, URL, etc)
Having id has a tuple (CompoundId) does this.

I guess this comes down to the properties we want id to have.
How about explicitly documenting (i.e. a comment) the CompoundId
class with the motivation??

pgrosu:
@diekhans Would this publication be something along the lines of what you might be looking for?

T. Nishimura, _Tables of 64-bit Mersenne Twisters_ ACM Transactions on Modeling and Computer Simulation 10. (2000) 348--357.  

The publication is available here for download:

http://dl.acm.org/citation.cfm?id=369540

The 64-bit implementation of the code is available here for download:  

http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/emt64.html

jeromekelleher:
OK, UUIDs are good, but there are some issues:

1. We don't need globally unique IDs,  just unique IDs within a repository. I don't see any point in generating a globally unique ID for say, a Dataset. What good does it do? This tells us nothing about its contents, and has no meaning across different repositories. You can say the same for ReadGroupSets and VariantSets, etc. Does it matter if there are ID clashes between repositories? The probability of this is really small, and unless ID formats are mandated by the schema, there are no guarantees anyway. 
2. Which specific UUID form are you in favour of? If it's a UUID1 or UUID3, then we're leaking a lot of information which might or might not turn out to be an information side channel. (I think we should keep this in mind since we'll be dealing with sensitive data.)  It it's UUID4 then we're basically just generating a 122 bit uniform random number, most likely by drawing from ``/dev/urandom``. The collision probabilities are just based on this source being uniform. There's no special  special sauce here, we can use exactly the same code to generate our own random numbers. A 64 bit random number will effectively be the same thing as a UUID4. The probability of a collision within a repository is about the same as dropping a ball bearing from a jet randomly and hitting, say, an egg sandwich (to paraphrase Douglas Adams). And we can make sure it doesn't happen pretty easily by using the top byte as a table specific value.

So, yes UUIDs solve some problems, but they don't solve everything. As you pointed out, we won't be able to tell what type of object an ID came from if we use UUID.  Also, I'm not sure that the problem they do solve (global uniqueness) is a problem that we actually need to solve.

Anyway, I think we should put this discussion off until after the code is ready for use and we can try out the various options. It'll be very easy to change, as I've deliberately isolated all the columns that generate externally visible IDs. The GUID SqlAlchemy type looks great, and would be easy to plug in.

In more progress in the code itself, I've just fixed the long standing issue with search over multiple read groups (within a read group set). It turned out to be pretty straightforward under the new organisation.

diekhans:
Agree we don't have a federated API and no need to support global
uniqueness.

My reason for suggesting UUIDs is that is the standard way to
generate uniques ids..  The probability is much higher that a
meteor will destroy your data center than a UUID collides.  The
logical is do it the common way unless the common way doesn't
work.

However, if you want the id to have actual meaning other than a
strong yet independent uniqueness property, it's the wrong tool.

I am trying to simulate discussion; not pushing anything.

Nice on fixing the read group stuff!!

jeromekelleher:
OK, this is starting to come together, so it would be really helpful to get some early review on the code as it is. 

## Main changes:

- Registry DB locations are now specified with a SQLAlchemy connection string, e.g., ``sqlite:///file.db`` or ``postgres://server.com/database``. 
- The entire ``ga4gh.datamodel`` package will be removed, and is being refactored into the new ``ga4gh.datasource`` package. The datasource package currently contains two modules ``simulator`` and ``htslib``. The idea is to localise all the code that accesses (or generates) the data in a similar way so that it can share functionality.
- The paging functionality for reads and variants has been pushed down into the datasource layer. This has greatly simplified these code paths, and allowed us to simply implement multi-readgroup searches.
- Top level metadata is now entirely modelled in SQL using the SqlAlchemy ORM layer. All queries should be now done on-line against a fully updateable DB.
- To fit in with the way SqlAlchemy is best used, there has been a substantial style change. All attributes on objects in the registry are now accessed via public attributes. Also all new code is using ``snake_case`` where possible. Test code still uses mostly camelCase in an effort to keep the diff getting too massive.

## TODO

- Update the CompoundId structure, and greatly simplify it. 
- Port ontologies
- Port Variant Annotations code.
- Port Sequence annotations code.
- Improve documentation and general code layout.
- Come up with a way of unit testing Postgres support (particularly for repo manager interface).

jeromekelleher:
It would be good to get some feedback on this @dcolligan --- would you mind having a quick look through?

ga4gh/backend.py
```diff
@@ -6,9 +6,18 @@
...
-import ga4gh.datamodel as datamodel
 import ga4gh.exceptions as exceptions
 import ga4gh.protocol as protocol
+import ga4gh.registry as registry
+
+# We must import these modules before queries are run, or we won't be
+# able to load the polymorphic types at run time. These should
+# probably be imported in the frontend so that we are more
+# configurable and open the possibility of a plugin architecture.
+import ga4gh.datasource.simulator  # noqa
```
dcolligan: why the noqa's here?

ga4gh/backend.py
```diff
@@ -446,137 +455,148 @@ def _noObjectGenerator(self):
...
+        #     if include:
+        #         results.append(obj)
+        # return self._objectListGenerator(request, results)
+
+    def readGroupSetsGenerator(self, request, responseBuilder):
+        # """
+        # Returns a generator over the (readGroupSet, nextPageToken) pairs
+        # defined by the specified request.
+        # """
+        # dataset = self.getDataRepository().getDataset(request.dataset_id)
```
dcolligan: you probably already know this, but a quick way to comment out blocks of code is just to turn it into a triple-quoted string

ga4gh/backend.py
```diff
@@ -446,137 +455,148 @@ def _noObjectGenerator(self):
...
+        query = self._registry_db.get_variant_sets_search_query(request)
+        self._run_sql_query(request, query, responseBuilder)
 
     def variantAnnotationSetsGenerator(self, request):
         """
         Returns a generator over the (variantAnnotationSet, nextPageToken)
         pairs defined by the specified request.
         """
-        compoundId = datamodel.VariantSetCompoundId.parse(
+        compoundId = datamodel.VariantSetCompoundId.parse(  # noqa
```
dcolligan: noqa?

ga4gh/backend.py
jeromekelleher: We're not actually referring to this package anywhere in the module, so pep8 gets upset. If we leave them out, SQL Alchemy breaks. Pr

ga4gh/backend.py
jeromekelleher: https://github.com/tpope/vim-commentary

'gcc' is proper musle-memory now! This is messy though, I'm just keeping the code around while I'm porting it. It'll all be cleaned up before a proper PR.

ga4gh/datasource/htslib.py
```diff
@@ -0,0 +1,915 @@
...
+                reference = HtslibReference(reference_name)
+                bases = fasta_file.fetch(reference_name)
+                reference.md5checksum = hashlib.md5(bases).hexdigest()
+                reference.length = len(bases)
+                reference.fasta_url = fasta_url
+                self.references.append(reference)
+        # Now that we have all the references, set the md5checksum
+        self._compute_md5checksum()
+
+##################################################
```
dcolligan: Might make sense to break up this file into a package as well... these big block comments are always an indication that a file ought to be broken up...

ga4gh/backend.py
jeromekelleher: Most of these noqa's are there to keep pep8 happy while the code is incompletely ported. In this example datamodel.VariantSetCompoundId doesn't exist any more, and pep8 freaks out.

ga4gh/datasource/htslib.py
jeromekelleher: I was thinking the same thing. There's still some work to be done to factor out some common code, but I definitely agree that this should be split up once it's fairly solid and stable.

ga4gh/datasource/simulator.py
```diff
@@ -0,0 +1,472 @@
...
+import hashlib
+
+import sqlalchemy
+
+import ga4gh.registry as registry
+import ga4gh.exceptions as exceptions
+import ga4gh.protocol as protocol
+
+
+#####################################################################
```
dcolligan: same comment about file breakup here

dcolligan:
Well this is a massive change, and a work in progress, but I see the motivation and the benefits.  Having a bunch of SQL strings (that needed to be updated with every field change) was getting cumbersome.  On-line updates that don't require a server restart is a win.  Multiple DB support is a win.  Greater extensibility is a win.  Using well-tested third party libraries over custom code is almost always a good idea.  

As far as answers to the questions you posed at the outset, I would favor stuffing fields in a JSON blob until there is a proven motivation for needing to normalize them in some kind of relational schema.  IDs: registry-unique would be nice (if strictly unnecessary), globally-unique seems unnecessary at this point.  64 bits should be sufficient.

diekhans:
I don't know if it was resolved it the POC, however the GA4GH API specification does
require that ids be unique across the server

from https://github.com/ga4gh/schemas/blob/master/doc/source/api/apidesign_intro.rst
- ids are unique within the scope of the server instance

jeromekelleher:
I've kicked this can down the road a bit @diekhans. This will be easy to sort this out once the code is merged and stable. For now, we're just generating random 32 bit numbers for every externally visible object. For practical purposes this means that IDs are unique within a server instance until we sort this out properly.

diekhans:
Has performance testing been done using the SQL offset/limit
mechanism for paging?

I am concern that it will be slow to get progressively slower
subsequent page request, especially for queries that are not
returning all rows in physical order.  Such as a range search
for variants.

jeromekelleher:
Not at scale, no. I would have thought that this would be fine once the query is using an index? All queries from the registry DB should use an index, as a matter of principle.

More specialised queries like range searches over variants are outside the scope of this PR.

diekhans:
Even if it's index, it has to play back the query and run it up
to the desired offset.  This slows down the larger the index
gets.  See

   https://developer.wordpress.com/2014/02/14/an-efficient-alternative-to-paging-with-sql-offsets/
   http://sqlperformance.com/2015/01/t-sql-queries/pagination-with-offset-fetch

This will should be fine if the data is small, as with most
containers.  A couple of random mysql trials showed significant
slowdown with an offset of 40000 vs 1.

your mileage may vary

jeromekelleher:
Hmm, that _is_ interesting! OK, this is something we'll have to consider. If we're searching on primary keys in table order then we're fine, as we can use the ID as the page token and resume from there. If not, then we'll have to think of something else.

Thanks for the heads-up @diekhans, I had not considered this issue...

dcolligan:
Closing for now.  We can open another branch if we decide to pursue this.

PR #1262, opened 2016-06-09
Resolves #1176: Adds created and updated to VariantAnnotationSets

### LINKED ISSUE #1176: Parse annotation creation times from the VCF header
Assumes #1166 has been merged.

Annotation programs often provide the creation time of the annotations in VCF headers. We should use this information to set the creation and update times in for VariantAnnotationSets when we call ``populateFromFile``. These are then stored in the DB.

andrewjesaitis:
This adds created and update times to variant annotation sets in the metadata db. Since this updates the db schema, a new example data tarball will be needed. I updated the example data urls in the docs for this.

This pushes the setting to the created and updated attributes down to the subclasses. `_creationTime` will be set by the `created` or `fileDate` field in the VCF header. `_updatedTime` remains the current datetime.


andrewjesaitis:
Really odd. I've seem to be generating absolute paths in the db. I thought it was due to a quirk on Linux, but it does it on my mac too. Is there a trick to calling `scripts/build_test_data.py`?

jeromekelleher:
We did change the paths generated in the repo manager to use fully qualified paths recently. Sorry, I forgot that this would have an impact here...

In any case, we should really be storing the DB in git at all when we can use this script to create it in Travis. It's probably easiest to just delete the DB, and add a call to the script in .travis.yml. We will need devs to update their DB on a fairly regular basis though, which may or many not be popular.

We should also probably add a command line option to the CLI to suppress the expansion to full paths, as there will surely be occasions when we don't want this to happen.

jeromekelleher:
Sorry @andrewjesaitis, I didn't see these updates until now. Looks good to me.

+1.

jeromekelleher:
Merging; thanks @andrewjesaitis.

PR #1267, opened 2016-06-16
BioMetadata Protocol

david4096:
Adds the BioSample and Individual objects in coordination with https://github.com/ga4gh/schemas/pull/636 https://github.com/ga4gh/compliance/pull/195.

This allows Calls and Read Groups to be filtered by a specific `bioSampleID`. Both individuals and biosamples can be searched by name, and biosamples can be further filtered by individual ID. The two new endpoints added are `/individuals` and `/biosamples`.

Minimum changes to the variants and reads APIs have been made to support this feature. Providing a bioSampleId is optional. This same pattern could be applied to future objects as well.

Known issues:

In practice one needs an individual ID when adding a biosample record to take complete advantage of the metadata linkages. An example of this has been provided in `prepare_compliance_data`.

A bioSampleId can modify the read group set record at query time, which can be confusing, in the future we may consider a separate endpoint for searching read groups.


ga4gh/backend.py
```diff
@@ -633,16 +701,18 @@ def callSetsGenerator(self, request):
...
-            return self._singleObjectGenerator(callSet)
+        results = []
+        for obj in variantSet.getCallSets():
+            include = True
+            if request.name:
+                if request.name != obj.getLocalId():
+                    include = False
+            if request.bio_sample_id:
+                if request.bio_sample_id != obj.getBioSampleId():
+                    include = False and include
```
jeromekelleher: Isn't the ``and include`` redundant here? False && x == False.

jeromekelleher:
This is truly excellent @david4096, nice, simple, straightforward code. I haven't tried it out and played with the new protocol yet, but I'm confident this code will go into master without any problems. Great work.

docs/configuration.rst
```diff
@@ -168,6 +168,50 @@ Adds a reference set used in the 1000 Genomes project using the name
...
+   :func: getRepoManagerParser
+   :prog: ga4gh_repo
+   :path: add-individual
+   :nodefault:
+
+**Examples:**
+
+.. code-block:: bash
+
+    $ ga4gh_repo add-individual registry.db dataset1 HG00096 '{"individualId": "abc"}'
```
macieksmuga: You probably wanted `'{"id": "abc"}'` there

macieksmuga:
+1 modulo documentation quibble. Also, `ga4gh_repo list` doesn't list the added biosamples or individuals (may be nice to add these features). I ran and tested the code with the compliance suite test data - all works as expected.

macieksmuga:
Could we one more +1 for this, to allow merging? @jeromekelleher @diekhans @calbach @dcolligan, anybody else you can think of. Thanks!

skeenan:
@mbaudis

mbaudis:
+1

docs/api.rst
```diff
@@ -25,6 +25,17 @@ Types
 
 .. todo:: Add links to the upstream documentation for the GA4GH types.
 
+
```
jeromekelleher: Please remove this section --- this will just print out confusing protobuf stuff into the docs.

jeromekelleher:
Great work @david4096, +1 modulo (very) minor issue with the docs.

docs/api.rst
kozbo: @jeromekelleher quick clarification: do you mean the todo section or the following BioMetadata section?

PR #1279, opened 2016-07-05
Mac OS X Installation Instructions

david4096:
Reopening with Bettie's commits and bumping the version number of the example data. We have used these directions successfully a couple of times now.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -325,59 +325,120 @@ This will set the environment variable which is read by config.py
 
 You can then get logs from the docker container by running ``docker logs (container)`` e.g. ``docker logs ga4gh_demo``
 
------------------------
-Deployment on Mac OS X
------------------------
-To deploy on Mac OS X, you need to install libraries and header code for `Python 2.7 <https://www.python.org/download/releases/2.7/>`_. It will be a lot easier if you have `Homebrew <http://brew.sh/index.html>`_, which is called the missing package manager for OS X, installed first. To install Homebrew, please paste the following at a Terminal prompt ($):
+----------------------------------------------
+Installing the development version on Mac OS X
+----------------------------------------------
+
+**Prerequisites**
+
+First install libraries and header code for
+`Python 2.7 <https://www.python.org/download/releases/2.7/>`_.
+It will be a lot easier if you have `Homebrew <http://brew.sh/index.html>`_,
+the "missing package manager" for OS X, installed first.
+To install Homebrew, paste the following at a Terminal prompt ($):
 
 .. code-block:: bash
 
   $ /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
 
-1. Now you can use ``brew install`` to install libraries for Python.
+Now use ``brew install`` to install Python if you don't have Python 2.7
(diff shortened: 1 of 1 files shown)


jeromekelleher:
This looks good, but it seems like it's getting a bit complicated with HomeBrew. Has anyone tried this with conda? It might be easier to install OpenSSH using conda, and this might get rid of the need for editing .profile and so on.

jmarshall:
What is it that is requiring OpenSSH?

david4096:
I've run into the problem twice with OpenSSL, however, the last couple of installs I've done have not required that change. I believe that portion could be safely removed for normal installations.

docs/installation.rst
```diff
@@ -324,3 +324,120 @@ To enable DEBUG on your docker server, call docker run with ``-e GA4GH_DEBUG=Tru
...
+Run tests to verify the install:
+
+.. code-block:: bash
+
+  (ga4gh-env) $ python scripts/run_tests.py
+
+If you deactivated the virtualenv, ``source ga4gh-env/bin/activate``
+to reactivate the ga4gh-env virtualenv.
+ 
+Download and extract the example data:
```
jeromekelleher: Maybe we could put in a link to the tutorial here rather than spell this out again? It's pretty tedious updating these version numbers all the time.

docs/installation.rst
```diff
@@ -324,3 +324,120 @@ To enable DEBUG on your docker server, call docker run with ``-e GA4GH_DEBUG=Tru
...
+.. code-block:: bash
+
+  (ga4gh-env) $ pip install -r dev-requirements.txt
+
+You may encounter the error "ssl.h not found"; if so, you will have to
+``brew install openssl`` after deactivating the virtualenv:
+
+.. code-block:: bash
+
+  (ga4gh-env) $ brew install openssl
```
jeromekelleher: Is ga4gh-env still active here? This contradicts the last sentence above.

docs/installation.rst
```diff
@@ -324,3 +324,120 @@ To enable DEBUG on your docker server, call docker run with ``-e GA4GH_DEBUG=Tru
...
+
+If the error persists, you need to tell the compiler about the newly
```
jeromekelleher: Unless this is a common problem I think we should remove it.

jeromekelleher:
@jmarshall we need crypto for the OIDC authentication support. I think it's OpenSSL all right though, not OpenSSH.

jmarshall:
Duh mutual typo, OpenSSL indeed.

I guess the requirement comes from the dependency on the `cryptography` package, and perhaps others.  All this comes down to the removal of `<openssl/*.h>` from the OS X 10.11 SDK that was shipped with Xcode 7.  So I was wondering whether there was a story for these dependencies to use CommonCrypto on OS X rather than the there-"deprecated" OpenSSL.  And indeed there is: pyca/cryptography#858 though it appears to be a bit more complicated than that and something of a work in progress.  But perhaps rather than writing this out here we can refer to [their instructions](https://cryptography.io/en/latest/installation/#building-cryptography-on-os-x).

david4096:
Thanks @jmarshall and @jeromekelleher updated to remove the OpenSSL part.

jeromekelleher:
Thanks, +1.

PR #1283, opened 2016-07-07
rna topic branch update

saupchurch:
this update brings the RNA topic branch in line with master.  Also has a couple of small changes to:
1) reflect move of Program message from reads.proto to metadata.proto
2) handles RSEM transcript input data
3) make tests pass


david4096:
+1 merging great progress!

PR #1285, opened 2016-07-08
process_schemas Make schema path required, don't infer path

### LINKED ISSUE #1268: Process schemas assumes directory structure
When running in various environments we should not assume the relative directory of schemas versus the server. As suggested by Gabriel, we should just accept an explicit path to the schemas as this is more clear and reliable.

`scripts/process_schemas.py`

        script_path = os.path.dirname(os.path.realpath(__file__))
        server = os.path.realpath(os.path.join(script_path, ".."))
        schemas = os.path.join(server, args.schema_path)

david4096:
Closes #1268 @gabrielsaldana

### DIFF
--- a/scripts/process_schemas.py
+++ b/scripts/process_schemas.py
@@ -25,7 +25,7 @@ class ProtobufGenerator(object):
     def run(self, args):
         script_path = os.path.dirname(os.path.realpath(__file__))
         server = os.path.realpath(os.path.join(script_path, ".."))
-        schemas = os.path.join(server, args.schema_path)
+        schemas = os.path.realpath(args.schemapath)
 
         if not os.path.exists(schemas):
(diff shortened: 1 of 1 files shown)


scripts/process_schemas.py
```diff
@@ -106,8 +106,8 @@ def main():
     parser.add_argument(
         "version", help="Version number of the schema we're compiling")
     parser.add_argument(
-        "--schema-path", default="../schemas",
-        help="Path to schemas. Default is ../schemas")
+        "--schema-path",
+        help="Path to schemas.", required=True)
```
jeromekelleher: Required 'options' make for a weird UI --- can we make this a positional argument instead please?

david4096:
@jeromekelleher updated!

PR #1299, opened 2016-07-20
Fix VA search bug

### LINKED ISSUE #1294: Search Variant annotations bug
Thanks @Kusdhill!

> I found a bug in the **searchVariantAnnotations** function where duplicate transcript effects are returned.

<a href="https://gist.github.com/Kusdhill/770c2754687a29a199e4e4f44d8691d0" target="_blank">This code</a> gives the following output of ID's:

> c8222b12ee2dafeeaaba593cc44cb3c6
**cc90d63e3a5b7e14bc2954414b0cf649**
2824090d5107a819c2e29bbc823c09e5
d7179d31c57b5bcc59d34997447ab1b2
c8222b12ee2dafeeaaba593cc44cb3c6
**cc90d63e3a5b7e14bc2954414b0cf649**

As you can see in the output, the second and last ID are the same.

The issue can be found <a href="https://github.com/ga4gh/server/blob/master/ga4gh/backend.py#L302" target="_blank">here</a>. Extending the two lists adds duplicate entries into the new list which can be problematic.

david4096:
During the protobuf conversion extend was incorrectly used. This bug caused repeated transcript effect records to appear in results. Fixing this bug revealed that the simulated stack tests for VA were incorrect and generated transcript effect records for the same alternate base repeatedly. This PR removes the uncecessary outer loop.

A test has been added to guard for this. Close #1294

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -290,15 +290,17 @@ class VariantAnnotationsIntervalIterator(IntervalIterator):
 
     def _removeNonMatchingTranscriptEffects(self, ann):
         newTxE = []
+        oldTxE = ann.transcript_effects
         if len(self._effects) == 0:
             return ann
-        for txe in ann.transcript_effects:
+        for txe in oldTxE:
             add = False
             for effect in txe.effects:
                 if self._matchAnyEffects(effect):
                     add = True
             if add:
                 newTxE.append(txe)
+        ann.ClearField('transcript_effects')
         ann.transcript_effects.extend(newTxE)
         return ann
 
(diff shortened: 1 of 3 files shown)


jeromekelleher:
+1.

We really should have tests for this sort of thing from real data also, but I think we need to tackle this as part of another update. Namely, I think we should change the current data driven tests framework to be an end-to-end test. That is, the client has the data (VCF file, BAM file, etc) that corresponds to a given ReadGroupSet, VariantAnnotationSet, etc, and then makes queries it knows should succeed or fail. This would plug a big gap in our current test coverage I think.

Anyway, that's off-topic on this thread.

Issue #1304, opened 2016-07-26
Read group getStats checks unset values


david4096:
I'm having some trouble loading stats for an `HtslibReadGroupSet` using the `datarepo` because of errors like these.

```
Adding HG03270
Traceback (most recent call last):
  File "1kgenomes.py", line 181, in <module>
    main()
  File "/Users/david/server-1/scripts/utils.py", line 79, in wrapper
    result = func(*args, **kwargs)
  File "1kgenomes.py", line 175, in main
    repo.insertReadGroupSet(read_group_set)
  File "/Users/david/server-1/ga4gh/datarepo.py", line 902, in insertReadGroupSet
    statsJson = json.dumps(protocol.toJsonDict(readGroupSet.getStats()))
  File "/Users/david/server-1/ga4gh/datamodel/reads.py", line 321, in getStats
    stats.aligned_read_count = self._numAlignedReads
AttributeError: 'HtslibReadGroupSet' object has no attribute '_numAlignedReads'
```

It seems like the stats are at both the readgroupset and readgroup level but the `getStats` function expects values to be set on readgroupsets. I added these lines to the instantiation of `AbstractReadGroupSet`:

```
        self._numAlignedReads = -1
        self._numUnalignedReads = -1
```

PR #1307, opened 2016-07-26
Add initial counts to read group set

david4096:
Note, I'm not sure what the intended behavior is here. It seems like stats are on both the RG and RGS level, but they derive their data from the same place. This PR simply initializes values checked by `getStats` to `-1`.

Closes https://github.com/ga4gh/server/issues/1304


dcolligan:
+1

PR #1309, opened 2016-07-26
Reference list bases post request

david4096:
In an effort to make the API consistent this PR implements the list reference bases endpoint as a POST request (https://github.com/ga4gh/schemas/pull/647), as opposed to a get request with query parameters. This helps to make the API more consistent by making POST requests used for anything that requires paging.

`GET /reference/abcd123/bases?start=0&end=100&pageToken=3`
  becomes 
`POST /reference/abcd123/bases {start: 0, end: 100, pageToken: "3"}`

A couple of bits of code that were noted for removal have been cleared out with this. Namely, as suggested in the TODO, we can pass around request objects without hand editing a parameter dictionary.

- Change tests to use post endpoint
- Simplify testing code path
- Change routes for list ref bases
- Add to client

### DIFF
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -923,7 +923,7 @@ class Backend(object):
         self.endProfile()
         return responseString
 
-    def runListReferenceBases(self, id_, requestJson):
+    def runListReferenceBases(self, requestJson):
         """
         Runs a listReferenceBases request for the specified ID and
         request arguments.
@@ -939,7 +939,6 @@ class Backend(object):
                     protocol.ListReferenceBasesRequest)
             except protocol.json_format.ParseError:
                 raise exceptions.InvalidJsonException(requestJson)
-        request.reference_id = id_
         compoundId = datamodel.ReferenceCompoundId.parse(request.reference_id)
         referenceSet = self.getDataRepository().getReferenceSet(
             compoundId.reference_set_id)
--- a/ga4gh/backend.py
+++ b/ga4gh/backend.py
@@ -923,23 +923,34 @@ class Backend(object):
         self.endProfile()
         return responseString
 
-    def runListReferenceBases(self, id_, requestArgs):
+    def runListReferenceBases(self, id_, requestJson):
         """
         Runs a listReferenceBases request for the specified ID and
         request arguments.
         """
-        compoundId = datamodel.ReferenceCompoundId.parse(id_)
+        # In the case when an empty post request is made to the endpoint
+        # we instantiate an empty ListReferenceBasesRequest.
+        if not requestJson:
+            request = protocol.ListReferenceBasesRequest()
+        else:
+            try:
+                request = protocol.fromJson(
+                    requestJson,
+                    protocol.ListReferenceBasesRequest)
+            except protocol.json_format.ParseError:
+                raise exceptions.InvalidJsonException(requestJson)
+        request.reference_id = id_
+        compoundId = datamodel.ReferenceCompoundId.parse(request.reference_id)
         referenceSet = self.getDataRepository().getReferenceSet(
             compoundId.reference_set_id)
-        reference = referenceSet.getReference(id_)
-        start = _parseIntegerArgument(requestArgs, 'start', 0)
-        end = _parseIntegerArgument(requestArgs, 'end', reference.getLength())
+        reference = referenceSet.getReference(request.reference_id)
+        start = request.start
+        end = request.end
         if end == 0:  # assume meant "get all"
             end = reference.getLength()
-        if 'pageToken' in requestArgs:
-            pageTokenStr = requestArgs['pageToken']
-            if pageTokenStr != "":
-                start = _parsePageToken(pageTokenStr, 1)[0]
+        if request.page_token:
+            pageTokenStr = request.page_token
+            start = _parsePageToken(pageTokenStr, 1)[0]
 
         chunkSize = self._maxResponseLength
         nextPageToken = None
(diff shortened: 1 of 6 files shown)


dcolligan:
+1

ga4gh/backend.py
```diff
@@ -782,21 +782,34 @@ def runSearchRequest(
         self.endProfile()
         return responseString
 
-    def runListReferenceBases(self, id_, requestArgs):
+    def runListReferenceBases(self, id_, request):
```
jeromekelleher: It's confusing using ``request`` here to refer to the incoming json and the parsed protocol object. Perhaps ``requestJson`` as the parameter?

ga4gh/backend.py
```diff
@@ -782,21 +782,34 @@ def runSearchRequest(
...
         """
         Runs a listReferenceBases request for the specified ID and
         request arguments.
         """
-        compoundId = datamodel.ReferenceCompoundId.parse(id_)
+        if not request:
```
jeromekelleher: When would the input parameter be Falsy? Need to document this behaviour.

ga4gh/backend.py
```diff
@@ -782,21 +782,34 @@ def runSearchRequest(
...
+                raise exceptions.InvalidJsonException(request)
+        request.reference_id = id_
+        compoundId = datamodel.ReferenceCompoundId.parse(request.reference_id)
         referenceSet = self.getDataRepository().getReferenceSet(
             compoundId.reference_set_id)
-        reference = referenceSet.getReference(id_)
-        start = _parseIntegerArgument(requestArgs, 'start', 0)
-        end = _parseIntegerArgument(requestArgs, 'end', reference.getLength())
+        reference = referenceSet.getReference(request.reference_id)
+        start = _parseIntegerArgument(
```
jeromekelleher: It's a bit wasteful converting to JSON twice in two lines here. Don't we already have the JSON as the input parameter anyway?

jeromekelleher: Also, why bother? Can't you just say, ``start = request.start``?

ga4gh/backend.py
```diff
@@ -782,21 +782,34 @@ def runSearchRequest(
...
+        end = _parseIntegerArgument(
+            protocol.toJsonDict(request),
+            'end', reference.getLength())
         if end == 0:  # assume meant "get all"
             end = reference.getLength()
-        if 'pageToken' in requestArgs:
-            pageTokenStr = requestArgs['pageToken']
+        if request.page_token:
+            pageTokenStr = request.page_token
             if pageTokenStr != "":
```
jeromekelleher: This has already been tested for being the empty string, you can just call ``start = _parsePageToken(pageTokenStr, 1)[0]``

ga4gh/client.py
```diff
@@ -685,11 +686,11 @@ def _run_get_request(self, object_name, protocol_response_class, id_):
             response.text, protocol_response_class)
 
     def _run_list_reference_bases_page_request(self, id_, request):
```
jeromekelleher: If we're using a POST request, then do we need this special casing for list reference bases at all? We should be able to reuse the existing machinery for all the other requests.

jeromekelleher:
The code looks good, but I think we could make it fit into the existing patterns better. If the upstream changes are accepted then this is fine.

I'm +0 on making the change --- I don't think it makes much difference whether we use a GET of a POST, but this is a long standing wart.

david4096:
Thanks @jeromekelleher let's remove this wart! Some of the ugly code paths are removed and I've renamed the variables as you've requested.

ga4gh/backend.py
```diff
@@ -782,23 +782,34 @@ def runSearchRequest(
...
         return responseString
 
-    def runListReferenceBases(self, id_, requestArgs):
+    def runListReferenceBases(self, id_, requestJson):
         """
         Runs a listReferenceBases request for the specified ID and
         request arguments.
         """
-        compoundId = datamodel.ReferenceCompoundId.parse(id_)
+        # In the case when an empty post request is made to the endpoint
```
jeromekelleher: This can't happen --- the body is required to be a valid JSON document by the spec.

In any case, we shouldn't be doing this here. You should be able to reuse the machinery used by all the other POST requests to take care of this. Start at the frontend and get rid of the special case for ListReferenceBases.

Should this still be called ListReferenceBases?

jeromekelleher:
I agree it's a wart, but it's one that's been deployed. This needs to be filled out at the schema level before this PR can be accepted. There must be a meaningful dialog before a breaking change such as this is made. Then, the implementation here should be carefully reworked to get rid of all special cases for the ListReferenceBases endpoint. 

I'm putting a -1 on this until the schema changes have been accepted, since this can technically be merged otherwise.

ga4gh/backend.py
david4096: Yes, I believe it is still a `ListReferenceBases` request. Do you think it is a good idea for endpoints that have all optional fields (like datasets/search) require a JSON be posted? Range is optional in this request. 

I always found this choice awkward and inconsistent with other APIs I have used. I have to make an empty JSON in order to list datasets in the reference server.

ga4gh/client.py
david4096: It appears the existing machinery hardcodes the search ending, and so I believe you'd be introducing machinery to handle the URL to the other functions, or leave them functionally isolated as they are. I'm happy to approach it either way.

david4096:
Looks like we got some +1's! https://github.com/ga4gh/schemas/pull/647

david4096:
bump :)

jeromekelleher:
If this is approved upstream I have no objections here of course. Consider my -1 cancelled.

dcolligan:
Looks like this is all but agreed upon (but still strangely unmerged...) in the schemas repo.  Merging...

PR #1318, opened 2016-08-02
Remove feature group

saupchurch:
removes feature group per schema PR #672


dcolligan:
these commits should be squashed

saupchurch:
hopefully I did this right

ga4gh/client.py
```diff
@@ -827,8 +799,7 @@ def __init__(self, backend):
             "individuals": self._backend.runGetIndividual,
             "rnaquantificationsets": self._backend.runGetRnaQuantificationSet,
             "rnaquantifications": self._backend.runGetRnaQuantification,
-            "expressionlevels": self._backend.runGetExpressionLevel,
-            "featuregroups": self._backend.runGetFeatureGroup
+            "expressionlevels": self._backend.runGetExpressionLevel
```
dcolligan: nitpick: it's good to keep the comma at the end of the list so only one line is changed in the diff, not 2 (and in a future diff where a line is added to this list for the same reason)

dcolligan:
looks good to me

+1

saupchurch:
ok - will resolve the nit and squash

saupchurch:
squashed and updated

PR #1321, opened 2016-08-03
support feature id list for expression search

saupchurch:
contains two fixes for expression search:
1) adds support for feature_ids
2) removes erroneous expression id requirement in rnaquantification back end


ga4gh/cli.py
```diff
@@ -797,11 +797,14 @@ class SearchExpressionLevelsRunner(AbstractSearchRunner):
     def __init__(self, args):
         super(SearchExpressionLevelsRunner, self).__init__(args)
         self._rnaQuantificationId = args.rnaQuantificationId
+        if len(args.feature_ids) > 0:
```
dcolligan: what happens if `len(args.feature_ids) <= 0`?  will it error in `run` because `self._featureIds` does not exist?

dcolligan:
looks good aside from my comment

david4096:
+1 merge at will

ga4gh/cli.py
saupchurch: good catch will fix

saupchurch:
fixed, squashed.  Can I get a renewal of the +1? @david4096 @dcolligan

ga4gh/cli.py
dcolligan: nit: we usually do this like 
```
self._feature_ids = []
if len(args.feature_ids) > 0:
    self._feature_ids = args.feature_ids.split(",")
```
Cuts down on one line and ensures something is assigned to the attribute, even if the things in the if/else clauses change

dcolligan:
+1

PR #1323, opened 2016-08-04
Fill todos

dcolligan:
(Kinda) depends on #1322

### DIFF
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -338,6 +338,12 @@ class TestClientArguments(unittest.TestCase):
 
     # def testFeatureSetsSearch(self):  # TODO
 
+    # def testExpressionLevelsGet(self): # TODO
+
+    # def testRnaQuantificationSetsGet(self): # TODO
+
+    # def testRnaQuantificationSetsSearch(self): # TODO
+
 
 class TestRepoManagerCli(unittest.TestCase):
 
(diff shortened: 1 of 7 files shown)


tests/unit/test_cli.py
```diff
@@ -81,13 +82,21 @@ class TestClientArguments(unittest.TestCase):
...
     def setUp(self):
         self.parser = cli.getClientParser()
 
-    # TODO we need a way to test parse failures. This is tricky because
-    # argparse calls sys.exit() on error, which we can't catch directly as
-    # an exception. Using mock to intercept this call would at least
-    # verify that an error has been raised.
+    def testParseFailure(self):
+        cliInput = "invalidCommand"
+        with utils.suppressOutput():
```
jeromekelleher: Why not reuse captureOutput here? We could also test that the stdout is emtpy and stderr is nonempty.

dcolligan: Because `parse_args` doesn't return, since it's throwing an exception

dcolligan: And we can't continue beyond that line, since `parse_args` (reasonably) assumes internally that the lines after a `sys.exit`do not execute, so if we make `sys.exit` a no-op, we get a different exception on the next line or whatever

jeromekelleher: OK, cool, makes sense.

jeromekelleher:
+1

PR #1324, opened 2016-08-04
Simulated tests for RNA

dcolligan:
(Kinda) depends on #1323 

Also should be merged after #1321, which fixes one of the RNA endpoints


saupchurch:
looks good
#1321 is merged so can re-enable the skipped test then +1

PR #1325, opened 2016-08-04
Fix features paging bug

david4096:
I had observed a 500 when searching for features because the nextpagetoken was being coerced from None to an integer. This sets the default to be `"0"`.

Add indexing by gene name and name to database generation script. This improves query performance of the features API. This patch has been applied to the 1kgenomes gencode instance.

Set help to be printed by default for gff3 script.


kozbo:
+1 @dcolligan does this look good to you?

dcolligan:
Anything that fixes a 500 is fine.  The Features code (paging?) has a lot more issues than this PR addresses.

+1

PR #1326, opened 2016-08-05
Data interface tests for RNA

dcolligan:


tests/unit/test_data_interface.py
```diff
@@ -267,3 +282,59 @@ def testSearchReferences(self):
...
                     self._assertEqual(repoReference, reference)
+
+    def testSearchRnaQuantificationSets(self):
+        self._testSearchMethodInContainer(
+            'getRnaQuantificationSets',
+            self._client.search_rna_quantification_sets,
+            self._repo.getDatasets())
+
+    def testSearchRnaQuantifications(self):
+        # TODO Can't use because of weird search_rna_quantifications signature
```
dcolligan: @saupchurch the search_rna_quantifications signature has both dataset_id and rna_quantification_set_id as parameters... why not just the later since it contains the former?  That is, dataset_id seems redundant to me (we should change the schema if we delete it).

tests/unit/test_data_interface.py
```diff
@@ -267,3 +282,59 @@ def testSearchReferences(self):
...
+                    self._client.search_rna_quantifications(
+                        datasetId, rnaQuantificationSetId))
+                for repoRnaQuantification, clientRnaQuantification in \
+                        utils.zipLists(
+                        repoRnaQuantifications, clientRnaQuantifications):
+                    self._assertEqual(
+                        repoRnaQuantification, clientRnaQuantification)
+
+    def testSearchExpressionLevels(self):
+        # TODO Can't use this because of weird getExpressionLevels signature
```
dcolligan: @saupchurch Why does `SqliteRnaQuantification.getExpressionLevels` take a rnaQuantificationId as a parameter?  Couldn't it just use the class instance's id?

dcolligan: @saupchurch also, it doesn't seem like `getExpressionLevels` is doing on-demand generation of objects (i.e. using the `yield` keyword to return objects) ... or am I missing something?

tests/unit/test_data_interface.py
saupchurch: good catch - this is an artifact of an older version of the code where rnaQuantification was the top level

tests/unit/test_data_interface.py
saupchurch: good points.  In fact, `expressionLevelsGenerator` in `backend.py` gets the rnaQuantificationId out of the object just to pass it to `getExpressionLevels` so yes is cleaner to do that within `SqilteRnaQuantification`

and yes, should probably use `yield`

dcolligan:
@saupchurch Updated.  Can you review and merge if acceptable?

saupchurch:
looks great! merging

Issue #1330, opened 2016-08-05
Expression Level paging does not work


dcolligan:
When searching expression levels, the `next_page_token` of the response is not set correctly.  Indeed, it always is set to empty, regardless of the query.  Therefore, the maximum number of pages returned is always 1.

The reason is `expressionLevelsGenerator` uses `_objectListGenerator` to iterate through the results it returns via `rnaQuantification.getExpressionLevels`.  Unfortunately, `rnaQuantification.getExpressionLevels` uses the pageSize to create a sql statement using a LIMIT query which only returns as many results as were requested for that particular page.  Therefore, when `_objectListGenerator` reaches the end of the returned results, it sees that there are no more objects to return on a future page, and does not set `next_page_token`.

I would not be surprised if this is a common problem affecting all search methods on objects which use a sqlite database as a backing store.  We will need a new interface in `backend.py` to deal with this kind of thing.

dcolligan:
Fixed in #1332

There will likely be similar problems with other db-backed data types, so a refactor would probably be a good idea the next time this comes up.

PR #1332, opened 2016-08-08
Fix paging for expression levels search

dcolligan:
Issue #1330

Also:
- Added tests for sqliteBackend
- Refactored RNA to make bulk db fetches on-demand
- Renamed some methods

The paging strategy introduced here should be generalized to other similar sql-based endpoints in a future refactor.

@jeromekelleher


jeromekelleher:
+1

PR #1338, opened 2016-08-10
Align RNA data with existing repository

david4096:
In order to make feature IDs in the RNA expression levels point to features within an existing repository three changes were made. The rnaseq2ga script now accepts a repository path.

First, when writing expression levels to the database features are searched within the first feature set in the first dataset found. Features are matched based on name and the resulting feature ID is written to the database.

Second, the signature of the usage of this script in prepare compliance data was changed.

And third, the rna data model needed to be modified to read from the feature ID column. It was reporting the 'id' for both name and featureId.

@saupchurch @dcolligan


dcolligan:
+1

PR #1343, opened 2016-08-11
RNA Update

david4096:
Fix threshold to be closed to reflect schemas.

Use read group set name from control file to find read group IDs.


dcolligan:
+1

saupchurch:
+1 merging

Issue #936, opened 2016-02-23
G2P needs a simulator


jeromekelleher:
We need some simple way of simulating G2P data so that we can include tests for it in the standard layout.

This should include the standard tests in ``tests/test_simulated_stack.py``.

dcolligan:
There is a basic simulator in the G2P branch now, so this issue can be reconsidered as "G2P needs a better simulator that tests more than main code paths"

PR #1348, opened 2016-08-15
Improved test coverage

bwalsh:
* simulated tests , data interface tests, simulated stack


ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -39,22 +39,18 @@ def __init__(self, parentContainer, localId, randomSeed):
...
+        fpa = protocol.FeaturePhenotypeAssociation()
+        fpa.phenotype_association_set_id = self._parentContainer.getId()
+        fpa.id = ""
+        fpa.feature_ids.extend(['featureId'])
+        fpa.evidence.extend([protocol.Evidence()])
+        fpa.phenotype.MergeFrom(protocol.PhenotypeInstance(id='phenotypeId'))
+        fpa.description = "description"
+        fpa.environmental_contexts.extend(
+            [protocol.EnvironmentalContext(id='environmentalContextId')])
+        return [fpa]
```
dcolligan: So the simulated `getAssociations` returns an array of length one in all cases?  That's not how the non-simulated `getAssociations` works, does it?  We want the operation of both to be as close as possible -- the point here is to simulate the return of real data, without having to load real data in for testing.  See `ga4gh/datamodel/reads.py` for a good example.

We can also vary what gets returned by use of a random number generator (see the previously mentioned example), but if and how we use that depends on what data we want to simulate and what conditions we want to test.  You know the data model better than me, so that's your call.

Another thing we want to do is to allow the user to control the simulated variables in the server config.  I just pushed a PR that does this for RNA: #1349  You should create variables in `serverconfig.py` that determine e.g. the number of `SimulatedPhenotypeAssociationSet`s that get created (I always like 2 as a default) in the simulated configuration.

tests/unit/test_simulated_stack.py
```diff
@@ -1147,8 +1147,44 @@ def testGetBioSample(self):
...
+                    responseData = self.sendSearchRequest(
+                        path, request,
+                        protocol.SearchPhenotypesResponse)
+                    for clientPhenotype in responseData.phenotypes:
+                        self.assertEqual(clientPhenotype, repoAssoc.phenotype)
+
+    def testSearchGenotypePhenotypes(self):
+        path = "/genotypephenotypes/search"
+        for dataset in self.dataRepo.getDatasets():
+            for repoPaSet in dataset.getPhenotypeAssociationSets():
```
dcolligan: You can replace this double loop with a single loop by using `allPhenotypeAssociationSets`

tests/unit/test_simulated_stack.py
```diff
@@ -1147,8 +1147,44 @@ def testGetBioSample(self):
...
+            responseData = self.sendSearchRequest(
+                path, request,
+                protocol.SearchPhenotypeAssociationSetsResponse)
+            for clientPaSet in responseData.phenotype_association_sets:
+                self.assertTrue(clientPaSet.id in repoPaSetIds)
 
-    # TODO def testSearchGenotypePhenotypes(self):
+    def testSearchPhenotypes(self):
+        path = "/phenotypes/search"
+        for dataset in self.dataRepo.getDatasets():
```
dcolligan: same comment about the double loop here

bwalsh:
@dcolligan Addressed your comments above.  Let me know what you think.

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -35,26 +35,46 @@ def toProtocolElement(self):
...
+                associations.append(self._makeSimulatedAssociation(i))
+        elif isinstance(request, protocol.SearchPhenotypesRequest):
+            associations.append(self._makeSimulatedAssociation(
+                _id=0, phenotype_id=request.id))
         else:
-            return []
+            test_phenotype_ids = ["phenotype-0"]
+            if len(request.phenotype_ids) > 0:
+                test_phenotype_ids = request.phenotype_ids
+            print(test_phenotype_ids)
```
dcolligan: we shouldn't be printing anything in the server

ga4gh/datamodel/genotype_phenotype.py
```diff
@@ -35,26 +35,46 @@ def toProtocolElement(self):
 
 
 class SimulatedPhenotypeAssociationSet(AbstractPhenotypeAssociationSet):
-    def __init__(self, parentContainer, localId, randomSeed):
+    def __init__(self, parentContainer, localId, randomSeed,
+                 numAssociations=2):
```
dcolligan: `numAssociations` should be able to be set in `serverconfig.py`.  There should be a variable for that and also a variable for the number of `SimulatedPhenotypeAssociationSet`s in a dataset.  These variables should be passed through `frontend`, `backend`, down to the appropriate level.

We do this so we can easily tweak the simulator from the config -- without having to dive into the depths of the implementation -- for testing various conditions.

bwalsh: @dcolligan thanks as usual for your review.  I'm assuming I would follow the handoff between frontend and dataRepository found [here](https://github.com/ga4gh/server/blob/g2p/ga4gh/frontend.py#L235-L236)

dcolligan: Yes, that's exactly what I'm talking about.  Although there are a few other handoffs, in addition to that one, at lower levels.

bwalsh: @dcolligan  [commit 6a3d879](https://github.com/ga4gh/server/pull/1348/commits/6a3d879845982419c2481e7fcc5f308cb749be19) should address this.  Thanks, -b

dcolligan: there should be another similar variable in `serverconfig` that is used to control `numPhenotypeAssociationSets`

jeromekelleher:
Looks good to me, +1. I think we can close #936 after this?

ga4gh/serverconfig.py
```diff
@@ -31,6 +31,8 @@ class BaseConfig(object):
     SIMULATED_BACKEND_NUM_REFERENCES_PER_REFERENCE_SET = 1
     SIMULATED_BACKEND_NUM_ALIGNMENTS_PER_READ_GROUP = 2
     SIMULATED_BACKEND_NUM_READ_GROUPS_PER_READ_GROUP_SET = 2
+    SIMULATED_BACKEND_NUM_PHENOTYPE_ASSOCIATIONS = 2
+    SIMULATED_BACKEND_NUM_PHENOTYPE_ASSOCIATION_SETS = 1
```
dcolligan: let's make this default to 2

dcolligan:
getting there... these commits will need to be squashed and rebased on top of `g2p` before they can be merged

bwalsh:
closing. superseded by of #1357

PR #1363, opened 2016-08-26
Add default pagesize to client

### LINKED ISSUE #1362: Client sends 0 page size
When attempting to use the client against another GA4GH server it was found that it defaults to send a pageSize of 0. `{u'pageToken': u'', u'pageSize': 0}` are sent with the first request. This bug caused the client to continually request empty pages.

david4096:
Set the default page size to 10. The client was sending a pagesize of 0 by default. Close #1362

### DIFF
--- a/ga4gh/client.py
+++ b/ga4gh/client.py
@@ -20,7 +20,7 @@ class AbstractClient(object):
     """
 
     def __init__(self, log_level=0):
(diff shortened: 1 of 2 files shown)


david4096:
Development on the client happening [here](http://ga4gh-client.readthedocs.io).

PR #1368, opened 2016-09-09
add rna import to repo manager WIP

saupchurch:
WIP for comments

new and legacy TODO's to be addressed shortly

open question:
should `rnaseq2ga` be left as is under `ga4gh` or should everything be extracted and moved to `datarepo.py`?


ga4gh/rnaseq2ga.py
```diff
@@ -140,11 +137,11 @@ def writeExpression(self, rnaQuantificationId, quantfile):
...
             featureName = fields[self._featureCol]
-            dataset = self._dataRepo.getDatasets()[0]
-            featureSet = dataset.getFeatureSets()[0]
             featureId = ""
-            for feature in featureSet.getFeatures(name=featureName):
-                featureId = feature[0].id
+            if featureSets is not None:
+                for featureSet in featureSets:
+                    for feature in featureSet.getFeatures(name=featureName):
+                        featureId = feature[0].id
```
david4096: It would be good to break here once you find one.

ga4gh/rnaseq2ga.py
```diff
@@ -246,78 +243,40 @@ def writeExpressionTable(writer, data):
...
+    elif rnaType == "rsem":
+        writer = RsemWriter(rnaDB, featureType)
+    else:
+        raise Exception("Unknown RNA file type: {}".format(rnaType))
+    readGroupSetName = readGroupSetNames.strip().split(",")[0]
+    featureSetIds = ""
+    readGroupIds = ""
+    if dataset:
+        featureSetIdList = []
+        for annotationId in annotationIds.split(","):
```
david4096: As these are feature set names we ought to remove mention of IDs and replace with featureSetNames.

ga4gh/rnaseq2ga.py
```diff
@@ -246,78 +243,40 @@ def writeExpressionTable(writer, data):
...
-    args = parseArgs()
-    rnaseq2ga(args.inputDir, args.outputFile, args.controlFile)
-
-
-if __name__ == '__main__':
-    main()
+            featureSetIdList.append(featureSet.getId())
+        featureSetIds = ",".join(featureSetIdList)
+        # TODO: multiple readGroupSets
+        readGroupSet = dataset.getReadGroupSetByName(readGroupSetName)
```
david4096: It would be good to gracefully fail if not found. I guess we get an uncaught exception?

david4096:
Thanks @saupchurch a very welcome addition! Given how heavily it uses the repo manager functions I can imagine it being in datarepo and perhaps easier to maintain. I'm happy leaving it separate, however.

dcolligan:
It would be good to have the RNA code follow the patterns that the rest of the code uses, for consistency.

jeromekelleher:
This looks like a great addition. This seems to be something that we will need to do more (converting input in various formats), so perhaps we should have a ``ga4gh.repo`` package to wrap up all this sort of stuff?

saupchurch:
I like that idea @jeromekelleher - will refactor

saupchurch:
pushed commit that creates a basic `repo` module for the rna importers will remove WIP from this PR following additional testing to be completed shortly

ga4gh/repo/rnaseq2ga.py
```diff
@@ -1,37 +1,25 @@
...
-
-
-def data_repo(path):
-    dataRepository = datarepo.SqlDataRepository(path)
-    dataRepository.open(datarepo.MODE_READ)
-    return dataRepository
+SUPPORTED_RNA_INPUT_FORMATS = ["cufflinks", "kallisto", "rsem"]
 
 
 class RNASqliteStore(object):
```
dcolligan: @saupchurch just fyi we try and camel-case acronyms in variable names (so it's easier to tell where one word ends and the other begins); this should be `RnaSqliteStore`; I'll make the change (and other similar ones) in my branch

Issue #1290, opened 2016-07-14
Feature(Sets) tests missing


dcolligan:
There should be tests for feature(sets) where there are not in:

```
tests/unit/test_cli.py
tests/end_to_end/test_client_json.py
tests/unit/test_client.py
```

(see TODOs added in #1289)

david4096:
+1

dcolligan:
Fixed in #1369

PR #1369, opened 2016-09-12
Feature(Sets) tests

dcolligan:
- added dataset info arg to cli for repo manager
- added force arg to build_test_data script
- changed default value for feature set search arg

Issue #1290


david4096:
+1

Issue #1375, opened 2016-09-14
Fix testSearchFeatures in test_client_json.py


dcolligan:
In #1370 I harmonized the pagination logic of features with the rest of the code base.  This fixed `testSearchFeatures` in `tests/end_to_end/test_client_json.py`.  However, rebasing it on top of the G2P changes re-broke the test.  @bwalsh , could you figure out why the test is failing (it is currently skipped now, of course)?

bwalsh:
@dcolligan  this should fix it.

dcolligan:
The test fix part looks good (I suspected it was something to do with ordering).  Someone else will have to comment on the version bump.  Usually it's good to make a PR for each logically distinct issue.

dcolligan:
Fixed in #1376

PR #1376, opened 2016-09-14
Update schema version.  Address #1375

bwalsh:
This PR contains one feature, one bug fix:
* generated code: update schema version  and latest rna 
* fix: support refactored SearchFeatures #1375


dcolligan:
The test fix part looks good (I suspected it was something to do with ordering). Someone else will have to comment on the version bump. Usually it's good to make a PR for each logically distinct issue.

kozbo:
I don't have a problem with this one given that the second change is just a string change. @bwalsh please remerge with master. +1

Issue #1371, opened 2016-09-13
Unify RNA and Feature paging logic


dcolligan:
Right now, the same code is used in the RNA and Feature code in `backend.py` to do paging with a SQL backing store.  Unify these into some class hierarchy.

dcolligan:
Fixed in #1377

PR #1377, opened 2016-09-15
Unify expression level and feature paging logic

dcolligan:
Also create backend_iterators file

Issue #1371


saupchurch:
+1

jeromekelleher:
Very nice, definite +1. I'm not so sure about ``backend_iterators.py`` though, as it's a bit generic; perhaps ``paging.py`` would be a more concrete description of what's going on in here?

dcolligan:
@jeromekelleher updated with `paging.py`

PR #1378, opened 2016-09-15
Fix a bug in loading RNA feature IDs for prepare compliance data

david4096:
A bug was causing the prepare compliance data script to fail for me. Added a break to avoid unnecessary resetting of the feature.


saupchurch:
+1

bwalsh:
+1

dcolligan:
Yeah, this was a result of #1370 and wasn't caught because there are no tests for the script... similar issues might be lurking around in other untested code.  Should be easy to fix if so.

PR #1381, opened 2016-09-19
Add cli package

dcolligan:
Issue #847


ga4gh/cli/common.py
```diff
@@ -0,0 +1,88 @@
+"""
```
jeromekelleher: It might be cleaner to put this stuff into ``cli/__init__.py``. Then you just have `import ga4gh.cli as cli` in every module rather than importing ``cli.common as cli_common``.

jeromekelleher:
Definitely a good move. +1 --- comments above are just suggestions.

repo_dev.py
```diff
@@ -1,7 +1,7 @@
 """
 Simple shim for running the repo program during development.
 """
-import ga4gh.cli
+import ga4gh.cli.repomanager as cli_repomanager
```
dcolligan: On the other hand (coupled with your other suggestion above), this would create circular references

PR #1405, opened 2016-10-03
RNA Repo manager

david4096:
Incorporates commits from https://github.com/ga4gh/server/pull/1368 to allow multiple RNA quantifications to be added to a Set. Introduces the `init-rnaquantificationset` command. @dcolligan @saupchurch

To test:

    python repo_dev.py init-rnaquantificationset ga4gh-example-data/registry.db rnaseq.db
    python repo_dev.py add-rnaquantification rnaseq.db ~/compliance/test-data/rna_brca1.tsv kallisto ga4gh-example-data/registry.db brca1 gencodev19 HG00096
    # And then again with the same data
    python repo_dev.py add-rnaquantification rnaseq.db ~/compliance/test-data/rna_brca1.tsv kallisto ga4gh-example-data/registry.db brca1 gencodev19 HG00096
    # then add the set to the registry
    python repo_dev.py add-rnaquantificationset ga4gh-example-data/registry.db -R hg37 brca1 rnaseq.db --name test


dcolligan:
I'm in favor of this only as a stopgap measure, as previously discussed.  I'd prefer the repo manager to always be operating on the registry, not other data stores.  In the long run the `add-rnaquantificationset` and init-rnaquantificationset` commands should probably be merged.

kozbo:
I have run this through it's paces and other than a problem with supporting a new type of RSEM file this is working. I will file a separate issue for that problem as it has nothing to do with the effort here.  +1 and merging

Issue #1407, opened 2016-10-04
G2P: client <--> frontend path mismatches


bwalsh:
When using the client, there are mismatches between the client and front end

ga4gh.client:404 {"errorCode": 523835337, "message": "The request path was not found"}

dcolligan:
Fixed in #1408

PR #1408, opened 2016-10-04
align paths between the client and front end

bwalsh:
This PR addresses a mismatch between the paths used in the client and the frontend
Bug fixes:  #1407


dcolligan:
Huh.  Surprised testing didn't catch this.

PR #1409, opened 2016-10-04
Refactor "configure" function.

dkucsc:
I came across the 'configure' function, and it seems like it does important work, and is also not documented.  It's a long function too.

I read the whole thing, and found that this portion can be isolated out cleanly, and so I did just that.

### DIFF
--- a/ga4gh/frontend.py
+++ b/ga4gh/frontend.py
@@ -197,28 +197,8 @@ def reset():
     app.config.from_object(configStr)
 
 
-def configure(configFile=None, baseConfig="ProductionConfig",
-              port=8000, extraConfig={}):
-    """
-    TODO Document this critical function! What does it do? What does
-    it assume?
-    """
-    file_handler = StreamHandler()
-    file_handler.setLevel(logging.WARNING)
-    app.logger.addHandler(file_handler)
-    configStr = 'ga4gh.serverconfig:{0}'.format(baseConfig)
-    app.config.from_object(configStr)
-    if os.environ.get('GA4GH_CONFIGURATION') is not None:
-        app.config.from_envvar('GA4GH_CONFIGURATION')
-    if configFile is not None:
-        app.config.from_pyfile(configFile)
-    app.config.update(extraConfig.items())
-    # Setup file handle cache max size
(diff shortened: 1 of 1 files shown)


ga4gh/cli/repomanager.py
```diff
@@ -1,6 +1,8 @@
 """
 repo manager cli
 """
+
+
```
dcolligan: Don't put spaces like this between the top comment and the import lines.  It's unnecessary and it creates a diff for files that you didn't really make any changes in, making the PR harder to review.  Please fix that for all other files below as well when you resubmit.

dcolligan:
Also you have style issues that are breaking the build.  Use `scripts/run_tests.py` to run the tests as travis does

david4096:
@dkucsc I like what you've done with the configure function, thanks!

Travis CI runs flake8 using the following command, so you can see the style guidelines `flake8 *.py tests ga4gh scripts --exclude=ez_setup.py,ga4gh/*_pb2.py,ga4gh/datamodel/obo_parser.py`.

ga4gh/cli/repomanager.py
dkucsc: Oh I did that for some particular/intentional reasons.  I would put a single space between lines of code in a function (sometimes I won't even do that! - a 1-line hash comment can most times be read _as_ a space between source code lines), and I would use 2 spaces to separate top-level functions, as well as all of the other top-level things in a single Python file, like the top comment, constant/global/capitalized variables, and any initialization code that's not in a top-level function, including the `if __name__ == '__main__'` stuff at the bottom.

That's what most Python files I've ever seen looked like, and in that order: the top comment, the import lines, global variables and initialization code (like `logger = logging.getLogger()`), a mix of top-level function and class definitions, and the usual `if __name__ == '__main__'` stuff at the bottom if you want.  And then I've grown over time to separate those _large_ things by 2 spaces, and within those large things, separate the __small__ things you're working with there with 1 space.

But sure, I'll go ahead and make the changes.

dkucsc:
Ok, after some work, I was able to improve this pull request.

Does Travis automatically run its tests when I update the branch?  I hope so.  Please let me know how this works for you folks.

PR #1412, opened 2016-10-05
datarepo.py refactor

dcolligan:
In preparation for adding validation


david4096:
@dcolligan can we close based on switch to ORM?

dcolligan:
I still think the refactor here would be a good idea, but we can close since we're not planning to merge anytime soon (and to implement similar ideas now it might be better to start from scratch rather than do the rebase...)

PR #1416, opened 2016-10-06
convert rnaseq2ga to use csv module

saupchurch:
solves the problem of RSEM output columns changing with datatype, improves readability and is overall better


kozbo:
working now, thanks. +1

PR #1424, opened 2016-10-12
fix repo command for remove-rnaquantificationset

kozbo:
was using the wrong names for a field and a function


dcolligan:
committed in #1426

PR #1429, opened 2016-10-21
fix for failing testVariantSetMetadata test

saupchurch:
when the `vcf` module read the header info it included a hidden character at the end of the `ANN` field (presumably a new line)


saupchurch:
was an issue with pysam version on my end- using the pinned 0.9.0 version makes everything agree

PR #1434, opened 2016-10-24
Use ga4gh-common for setup and utils

dcolligan:
@david4096 CI is passing now; could you review this?


tests/end_to_end/client.py
```diff
@@ -9,7 +9,7 @@
 import subprocess
 import tempfile
 
-import tests.utils as utils
+import ga4gh_common.utils as utils
```
dcolligan: Yes, but as we've discussed, this will be a later step

PR #1441, opened 2016-10-25
performance fixes for rna import

### LINKED ISSUE #1436: Skip featureId lookup optionally
When adding an RNA quantification to an RNA quantification set, the ReadGroupSet and FeatureSet arguments should be optional via a flag.

saupchurch:
Resolves #1436 
Increases `batchSize` to 2000 and restricts `featureSetName` lookup to the list passed into `rnaseq2ga()`

With these changes a test dataset load went from 6 minutes (37K entries with feature lookup) to 12 seconds (37K entries no feature lookup) on an underpowered laptop


david4096:
Uh, yes please!

PR #1446, opened 2016-10-27
Make server depend on ga4gh-schemas package

dcolligan:


david4096:
Super excited to see this PR, -8,257 LOC!!! It sounds like you're going to stage all of the dependencies and then work out the namespacing before we release?

dcolligan:
The namespacing changes (if they work, which I think they will) will be the last change of the package reshuffle.

PR #1450, opened 2016-11-01
Add grab_wheel script

dcolligan:
Makes testing against multiple packages that we're also developing (our situation now) easier


david4096:
`pip install git+ssh://git@github.com/ga4gh/ga4gh-common.git`?

dcolligan:
Oh wow I didn't know pip had these options.  It also satisfies the local case (`pip install ../ga4gh-common`) as well as the remote case (`pip install git+git://github.com/ga4gh/ga4gh-common.git@0c49de512`).  Closing ...

PR #1454, opened 2016-11-02
Hierarchical namespace

dcolligan:


dcolligan:
@david4096 so it looks like we've hit a problem here.  nose doesn't seem to like our namespace packages (see travis output).  This might be the same bug: https://github.com/nose-devs/nose/issues/856

dcolligan:
@david4096 a few options

1) abandon hierarchical namespace plan
2) patch nose and use patched version in testing until they ship a new version
3) find a nose alternative (they seem to have similar problems... ) https://github.com/ga4gh/server/pull/1459 and https://github.com/ga4gh/server/pull/1460

david4096:
Ugh, of course everything works except the testing framework LOL! I'm going to try to bang at it for a bit too.

dcolligan:
@david4096 looks like nosetests actually works if we launch it differently: `python -mnose tests` (I guess this uses python's loader instead of nose's); still a lot of problems with the tests though.  They might all be fixable, we'll see...

It's just a bummer that this doesn't work:
```
>>> import ga4gh.common
>>> import ga4gh.schemas
>>> import ga4gh.client
>>> import ga4gh.server
  File "<stdin>", line 1, in <module>
ImportError: No module named server
```

dcolligan:
@david4096 Although all the Travis tests pass, I still don't want to merge this into master because it breaks development.  By that I mean if you try to run the tests locally they won't work because some tests rely on the above-mentioned mechanism (`import ga4gh.server`) to run.  Specifically, this applies to some of the end-to-end tests that invoke the shim scripts like `server_dev.py`.  A lot of the scripts won't work as well.  There might be some way to fix this, but if we can't, it makes the namespace packages a lot less attractive.  Do you have any ideas of how to get around this?  There might be some import trick that I am unaware of.

david4096:
Perhaps for local packages for the server you could do relative imports.

Another hack is stuffing things onto the path:

```
import sys
sys.path.append("./ga4gh/server")
```

And I imagine if you use `importlib` you can do whatever you want.

If you can't make a solution you're satisfied with after another crack, we'll live with underscores, but please do see if you can't get someone on stackoverflow or something to chime in in the meantime. I think this is more than a nice-to-have in the longview.

http://stackoverflow.com/questions/72852/how-to-do-relative-imports-in-python

dcolligan:
So the trick was to use the namespace module's `__path__` [attribute](https://docs.python.org/2/tutorial/modules.html#packages-in-multiple-directories):
```
>>> import ga4gh.server
  File "<stdin>", line 1, in <module>
ImportError: No module named server
>>> import ga4gh
>>> ga4gh.__path__.insert(0, 'ga4gh')
>>> import ga4gh.server
>>>
```
I've put this in a file called `dev_glue.py` for convenience:
```
>>> import ga4gh.server
  File "<stdin>", line 1, in <module>
ImportError: No module named server
>>> import dev_glue
>>> import ga4gh.server
>>>
```

david4096:
Nice!!!

PR #1456, opened 2016-11-03
Delete unused scripts

### LINKED ISSUE #1291: scripts/update_data.py data needs to be updated
The example data that `scripts/update_data.py` pulls does not contain a `repo.db` file.  We should either delete (if this script has been superseded by a better method of creating example data) or update the script.

### LINKED ISSUE #1453: Delete unused scripts
https://github.com/ga4gh/server/blob/master/scripts/generate_fasta.py

Generates random FASTA files

https://github.com/ga4gh/server/blob/master/scripts/split_fasta.py

Was added as part of our ETL process on FASTA before better pysam support.

Remove `update_data.py` will close https://github.com/ga4gh/server/issues/1291

david4096:
Close #1291
Close #1453


dcolligan:
+1

Issue #667, opened 2015-10-02
Flask config is awful


jeromekelleher:
Configuring the server is a massive pain at the moment. We used the Flask configuration mechanisms originally because it seemed sensible to reuse as much as possible. Our configuration mechanism right now is very complicated, allowing for variables to be set via class configurations, files and even environment variables. It has almost no test coverage, and it's hard to see how this could be improved using the current flask-based approach.

It's looking like we might be better of just rolling our own config instead of using Flask's methods. This would be a good extended hackathon topic, if people were interested in taking it on. Basically, we would like to see a sane and easy to follow configuration mechanism that is usable both from the command line development server and for deployment in Apache, Docker, etc.

david4096:
https://github.com/ga4gh/server/pull/1409 helps

david4096:
Specifying a configuration file when starting a server doesn't seem like too onerous a task. Perhaps instead of having *Configuration classes in the code base we could convert these to example configuration files that would set the environment variables and be loaded by tests? What do you think @dcolligan ?

dcolligan:
I don't really see the (dis)advantage of a config file vs `serverconfig.py` -- seems like about the same thing to me.  I think we want to avoid environment variables wherever possible.

david4096:
But we don't need both, eh? Like we could remove `serverconfig.py` and just use config files?

david4096:
Ran into this error trying to use pydev with a class based config:

https://gist.github.com/david4096/2e138db96cf103b8650bbe380255d254

PR #1458, opened 2016-11-03
Remove class based configuration

david4096:
There are currently too many ways of setting a configuration: flags, classnames, environment variables, and python files. This PR removes the class based configuration in favor of just using the file-based configuration method throughout. These configuration files have been added under the `ga4gh/config` directory.

It also removes two flags from the CLI server script in favor of adding them as configuration variables. These variables have been added to the documentation.

Helps to address #667

Currently fails a few tests around server starting that assumed a default flask configuration (I think).


dcolligan:
Isn't the drawback here that we lose inheritance for configs?  I.e. if I want a config that is one attribute different from, say, base I need to copy-pasta the whole file and then add my attribute to the new config?

david4096:
@dcolligan Yes, you're right. I took that out. The one time I have had to support this it was because there were too many command line options. Having a class based configuration hierarchy doesn't do much for us and just hides practical details from people rolling out the server.

david4096:
Will take another stab at this again with only changing the way we accept environment variables.

PR #1460, opened 2016-11-04
Namespace pytest

dcolligan:
try pytest


david4096:
Looks like pytest needs to be installed on travis first.

Noting that sphinx also doesn't like this but there is a fix in 1.5 https://github.com/sphinx-doc/sphinx/issues/2949

david4096:
@dcolligan Nose is deprecated basically and pytest is getting active development, so I like where this is going. Maybe the most expedient solution would be to roll back the hierarchical namespace until we switch to pytest, although trading them should be fairly painless. I'll leave it to your discretion.

PR #1470, opened 2016-11-17
1452 Auth0

### LINKED ISSUE #1452: Authenticated API access through key
I would like to know if I can configure GA4GH server to accept only authenticated requests through one or more keys. I tested the [OpenID Connect Providers tutorial](http://ga4gh-reference-implementation.readthedocs.io/en/latest/configuration.html#openid-connect-providers) but it is not exactly what I want because I understood that anyone who have a Google account will be able to access the API. Is there another method for adding authentication to GA4GH API?

I also tested https://github.com/ga4gh/server/tree/master/oidc-provider but it raises the following error:

```
  File "src/run.py", line 1, in <module>
    from provider.server.server import main
  File "/home/welliton/GA4GH/server/oidc-provider/simple_op/src/provider/server/server.py", line 16, in <module>
    from cherrypy.wsgiserver.wsgiserver2 import WSGIPathInfoDispatcher
ImportError: No module named wsgiserver2
```

david4096:
This PR will add login using the Auth0 OAuth service. Auth0 allows a developer to connect to a variety OpenID providers, or create username/password accounts. This requires having an Auth0 account.

This is not meant to be a complete security solution.

The proposed changes:

* `/login` page that renders the Auth0 "lock" login flow.
* Function decorator that will check the authorization header of incoming requests to protected endpoints. (Should this include the landing page?)
* `/callback` endpoint that handles the return from Auth0.
* Authorized emails provided via configuration variable.
* `/logout` endpoint that removes the token from the cache.

Sessions are logged in for the duration of the key (a web token setting in Auth0) or when the server restarts.

This is an alternative to the current login flow. Will close #1452


ga4gh/server/serverconfig.py
```diff
@@ -93,6 +93,20 @@ class GoogleOidcConfig(ProductionConfig):
     OIDC_CLIENT_SECRET = "XXX"
 
 
+class Auth0Config(DevelopmentConfig):
+    """
+    Auth0 configuration
+    """
+    AUTH0_ENABLED = True
```
kozbo: Will you be changing the default to AUTH0_ENABLED = False for the final release?

kozbo:
this looks very exciting, can't wait to try it out!

ga4gh/server/serverconfig.py
david4096: This is a configuration setting for testing. To use it you provide `-c Auth0Config` at the command line.

david4096:
TODO: Currently will not work in a production environment as the `SimpleCache` is not shared between processes.

david4096:
Started using the file system cache and added some tests. The test coverage of the inner functions could be improved, and the callback generator refactored for readability.

The testing apparatus needs to delete the cache when finished. What is the expected server behavior regarding logins across restarts? Perhaps we should add a configuration variable to invalidate the cache?

ejacox:
tests/unit/test_views.py fails if it is run by itself, but not if it is run with the other unit tests

david4096:
@ejacox thanks fixed that. The scope for setting CORS wasn't correct.

david4096:
Start a protected server using my Auth0 credentials.

```
git checkout david4096/1452_auth0
virtualenv myenv
source myenv/bin/activate
pip install -r dev-requirements.txt -c constraints.txt
python server_dev.py -c TestAuth0Config
```

Visit [http://localhost:8000](http://localhost:8000) and observe the error page. Visit /login and use Google ID to login. You will be able to generate a token.

Then in a python REPL:

```
from ga4gh.client import client
c = client.HttpClient("http://localhost:8000")
c.search_datasets().next() # Observe an error since you haven't set the token
key = "YOURTOKEN"
c = client.HttpClient("http://localhost:8000", id_token=key)
c.search_datasets().next() # Observe an error since you're not authorized
```

Now make a `config.py` that you can use to authorize yourself.

```
AUTH0_ENABLED = True
SECRET_KEY = "super_secret"
AUTH0_SCOPES = "openid email"
AUTH0_CALLBACK_URL = "http://localhost:8000/callback"
AUTH0_HOST = "david4096.auth0.com"
AUTH0_CLIENT_ID = "7PyT5eQRRdaSxgM4hEYz04wzncNiXpqH"
AUTH0_CLIENT_SECRET = \
    "2OkzpU7LMYMllXf7cd027WFcbBpb4b_p3iNRaUfsYOQgxiYM_puwRxCddCRy_RtV"
AUTH0_AUTHORIZED_EMAILS = "kosborn2@ucsc.edu"
```

Then start the server using `python server_dev.py -f /absolute/path/to/config.py`. You should now be able to visit the page, login, get a token and then access the client methods.

david4096:
Going to update so that we don't give the Oauth token when they're not authorized to the server.

kozbo:
Ran some basic tests including support for multiple email addresses. Working for me.

kozbo:
It would be nice if the original landing page resolved once authentication is complete.

david4096:
Will now link you to the landing page after login. @kozbo

kozbo:
Thanks @david4096 that works great!

PR #1474, opened 2016-11-23
Change to protobuf3.1, add constraints.txt

### LINKED ISSUE #1471: Move to protobuf 3.1
Protobuf 3 has left beta as 3.1 which moves some private functions we use. We should move to this stable release and update requirements as needed.

david4096:
Remove private API calls, close #1471 

Note pointing to a development branch of the client. PR for same change in client can be found [here](https://github.com/ga4gh/ga4gh-client/pull/35)

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,0 +1,2 @@
+git+git://github.com/david4096/ga4gh-client.git@protobuf31#egg=ga4gh_client
+git+git://github.com/david4096/schemas.git@protobuf31#egg=ga4gh_schemas
--- a/constraints.txt
+++ b/constraints.txt
@@ -5,10 +5,10 @@
 # packaging script. This allows us to maintain dependencies across
 # repositories.
 #
-# For example, during development the we use the development version
+# For example, during development we use the development version
 # of the schemas with the line:
 #
-# git+git://github.com/ga4gh/ga4gh-schemas.git@master#egg=ga4gh_schemas
+# git+git://github.com/ga4gh/schemas.git@master#egg=ga4gh_schemas
 #
 # When one has proposed a schema modification, they can change this
 # constraint to point at the branch of their proposed change:
@@ -17,5 +17,12 @@
 #
 # Then, when the change has been accepted at the schemas repository
 # you can change this to point back at the master branch!
+#
+# When making a release, these constraints are changed to point at
+# the released dependency.
+#
+# ga4gh-client==0.1.0
+# ga4gh-schemas==0.1.0
+#
 git+git://github.com/david4096/ga4gh-client.git@protobuf31#egg=ga4gh_client
 git+git://github.com/david4096/schemas.git@pipfixes#egg=ga4gh_schemas
--- a/constraints.txt
+++ b/constraints.txt
@@ -18,4 +18,4 @@
 # Then, when the change has been accepted at the schemas repository
 # you can change this to point back at the master branch!
 git+git://github.com/david4096/ga4gh-client.git@protobuf31#egg=ga4gh_client
-git+git://github.com/david4096/schemas.git@protobuf31#egg=ga4gh_schemas
+git+git://github.com/david4096/schemas.git@pipfixes#egg=ga4gh_schemas
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,2 +1,21 @@
+# This file allows our dependencies to be against specific versions
+# or git tags. Each line is similar to a requirements.txt line, 
+# except that when installing the dependencies, the constraints.txt
+# is converted into the `dependency_links` of the setuptools
+# packaging script. This allows us to maintain dependencies across
+# repositories.
+#
+# For example, during development the we use the development version
+# of the schemas with the line:
+#
+# git+git://github.com/ga4gh/ga4gh-schemas.git@master#egg=ga4gh_schemas
+#
+# When one has proposed a schema modification, they can change this
+# constraint to point at the branch of their proposed change:
+#
+# git+git://github.com/david4096/schemas.git@protobuf31#egg=ga4gh_schemas
+#
+# Then, when the change has been accepted at the schemas repository
+# you can change this to point back at the master branch!
 git+git://github.com/david4096/ga4gh-client.git@protobuf31#egg=ga4gh_client
 git+git://github.com/david4096/schemas.git@protobuf31#egg=ga4gh_schemas
--- a/constraints.txt
+++ b/constraints.txt
@@ -24,5 +24,5 @@
 # ga4gh-client==0.1.0
 # ga4gh-schemas==0.1.0
 #
-git+git://github.com/david4096/ga4gh-client.git@protobuf31#egg=ga4gh_client
-git+git://github.com/david4096/schemas.git@pipfixes#egg=ga4gh_schemas
+git+git://github.com/ga4gh/ga4gh-client.git@master#egg=ga4gh_client
+git+git://github.com/ga4gh/schemas.git@master#egg=ga4gh_schemas
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,10 +12,9 @@
 # way to resolve our internal dependencies. During development,
 # the constraints file will point at the current master branch 
 # of the respective module.
--c constraints.txt
+ga4gh-client
 ga4gh-common==0.0.5
 ga4gh-schemas
-ga4gh-client
 
 Werkzeug==0.11.5
 MarkupSafe==0.23
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,8 @@
 
 ga4gh-common==0.0.5
 ga4gh-schemas==0.0.8
-ga4gh-client==0.0.3
+#ga4gh-client==0.0.3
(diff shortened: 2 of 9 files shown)


david4096:
For the method used to specify the remote repository to work with setuptools, I think we need to modify the setup.py. This PR points to an open issue about how best to make updates the server, which should be reflected in the client. http://stackoverflow.com/questions/18026980/python-setuptools-how-can-i-list-a-private-repository-under-install-requires

david4096:
I added the `constraints.txt` that resolves the dependencies for working with other repositories/branches. It seems the cleanest approach, in that when working on feature branches one can specify the git repository in the `constraints.txt` file and then install using `pip install -r dev-requirements.txt -c constraints.txt`. Before merging into master the constraints will be removed.

ejacox:
+1

tools/travis-install-protoc.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+# ensure that the specified version of protoc is installed in
+# /tmp/proto$PROTO_VERSION/bin/protoc, which maybe cached
+
+# bash tools/travis-install-protoc.sh 3.0.0-beta-3.1
+#   
+
+
+# make bash more robust.
+set -beEux -o pipefail
```
dcolligan: This is some black magic here...

Did you get this script from somewhere?  If so, put a link to the source in the header comment.  If not, add a comment here about what this line is doing

constraints.txt
```diff
@@ -0,0 +1,2 @@
+git+git://github.com/david4096/ga4gh-client.git@protobuf31#egg=ga4gh_client
```
dcolligan: It seems to me that the constraints files checked into master should always be empty.  Ensure this by checking in an empty file, committing, and then adding a line for `constraints.txt` to the `.gitignore` file, and committing again.

If you think the constraints files should not always be empty in master, I would like to hear why.  It seems like then we would have a master branch that has dependencies on non-master branches of ga4gh code, which doesn't seem right.

requirements.txt
```diff
@@ -9,8 +9,8 @@
 # that are dependencies of libraries listed in the next section
 
 ga4gh-common==0.0.5
-ga4gh-schemas==0.0.8
-ga4gh-client==0.0.3
+ga4gh-schemas
```
dcolligan: I really don't like this.  This means if we release a version of schemas that is incompatible with server (of any version that doesn't pin its requirements exactly), we will be breaking those installs.  We should always exactly pin versions of ga4gh code in requirements files, in my view.

Are you doing this because the constraints feature will not work otherwise?  If so, we have at least a couple options:

1) Have people that want to install with the constraints file (that is, only developers) edit the pins out of the requirements file and then do another pip install with the constraints.  This seems acceptable, since they're editing pip-related files (i.e. `constraints.txt`) anyways

2) Create another requirements file which doesn't have ga4gh versions pinned which can be used with the constraints files

constraints.txt
andrewjesaitis: I may be missing the reasoning behind splitting the versions out of requirements and I can see the reasoning behind @dcolligan comment, but it seems to me that the best placeholder would be the versions required to build the current version of the server. However, the versions checked into master should probably reference official ga4gh repos. Then you can just reference the constraints file in requirements file and it should be buildable with a simple `pip install -r requirements.txt`.

requirements.txt
andrewjesaitis: I agree with this. See my suggestion for referencing the constraints file in requirements file to pin the versions.

constraints.txt
david4096: Yeah, good catch both!

I like the idea of keeping the constraints file out of the tree and it being part of the development documentation. The problem I ran into was getting the dependencies to resolve against pinned versions installed as part of `install_requires`. You have to create a properly formed [egg fragment](http://stackoverflow.com/questions/11835396/why-egg-foo-when-pip-installing-from-git-repo) at the end of the url, which was a task I didn't really want to hand down to other developers.

The alternative was to unpin the version in requirements and force a dependency against just the name in the fragment you see. That suggests that I was headed towards @andrewjesaitis' solution, which would have the master branch `constraints.txt`, when possible, against a released version of the schemas. The entries in constraints have the same syntax as in the `requirements.txt`. 

We don't do a release every time something is merged into the schemas, so the `constraints.txt` should probably be against schemas/master until a release. As long as PRs that have dependencies on other repositories are merged within a similar time, a release coordinator will easily be able to pick versions that are stable across our dependencies.

To do this, first the release coordinator makes a schema release with a tagged version. They then modify the `requirements.txt` of the client to point at the newly released version of schemas and remove the contents of the `constraints.txt`. They then release the client and repeat the same process in the server. I think this is more or less the same onerousness of the task that assembling the egg fragment would be, but done only once per release by the release coordinator, and could be automated.

I am trying to optimize the development experience of modifying the schemas and sharing an implementation. This requires nailing down the release process for the PyPi package. If we did keep the constraints file empty in master, wouldn't we need to release the package to pip with every schema change that needed implementation?

requirements.txt
david4096: @dcolligan It should be possible to pin the version in the requirements.txt for released versions. But since the schemas aren't updated in pypi with every git commit, we need a better way to manage working against the master branch of schemas, which in the past was maintained by careful runs of `process_schemas.py`. 

This is modeled in a way with what is done with the schemas, which use github repos as a dependency via a git tag. The compliance/master runs against schemas/master, so if schemas/master changes compliance breaks. This is not currently the case with the server, which this will change.

So I would say you're both right! For released versions I think we will select specific versions in the requirements and remove the constraints.txt. The master-development versions point at the running master branch of their dependency.

tools/travis-install-protoc.sh
david4096: Haha agreed. I lifted it from the schemas. https://github.com/ga4gh/schemas/blob/master/tools/travis-install-protoc.sh Maybe @diekhans would like to explain this bash-fu.

constraints.txt
dcolligan: Here's another solution that makes things easier for developers: keep the `requirements.txt` file as is and put into the `constraints.txt` file commented-out lines (with the egg fragment) of what would be there if a developer wanted to point to git branches instead of released packages.  The developer would only have to uncomment the line (or, at worst, it would be more obvious what to do to modify it).

I would be against adding more steps to the release process, if at all possible.  Manual steps in releases = inevitable screwups.  Indeed, the closer we can get to a one-button release (as advocated in https://www.amazon.com/Building-Scalable-Web-Sites-Applications/dp/0596102356 ), the better.

david4096: Thanks @dcolligan ! You're right that we could do that, but I don't think any thing I suggested can't be automated as part of a release. I did add more comments to the files to make it clear how to use them.

Our internal dependencies shouldn't have stuck versions in the `requirements.txt` unless it is a released version. This gives us a way to use git to manage our dependencies in the same way we do with compliance. To pin the version of schemas in the requirements is a misrepresentation of the facts, because if the schemas master moves ahead server/master should break. 

We could do rolling releases to pypi to bump the version number with each accepted PR. Then the requirements.txt would reflect the state of development.

dcolligan: "Our internal dependencies shouldn't have stuck versions in the requirements.txt unless it is a released version."

So the release manager is going to have to remember this, and track the versions down that s/he is using, on every release?

"We could do rolling releases to pypi to bump the version number with each accepted PR"

This seems like somewhere between ambitious and problematic.  Pypi releases are for everyone (i.e. end users), not just for developers.  In any event, it's a different conversation than what this thread is addressing.

david4096:
Although it was passing tests, I had a problem when using setup.py directly to install the server. The generated protobuf weren't being added to the egg properly. I have opened a schemas PR to fix that here https://github.com/ga4gh/schemas/pull/751 . This PR now includes that change as a dependency.

When making a release of the server, edit the `constraints.txt` to point at a released version on PyPi just as you would in a requirements.txt. A comment to this effect has been placed in the constraints.txt file. Since this file is used by both `pip` and `setuptools` I think we should check it into the tree.

To optimize the development process, we will make it so the server depends on the current schemas/master, so that we don't need to wait for releases to keep up with the pace of schemas development. Ideally, we'll have a failing build for the time between merging PR's. This is the same way we are operating the compliance suite.

The side effect of this is that at any point in history are not easily be able to reconstruct the state of schemas from server, you only have the git tag and the date. This probably won't be a problem if we keep our releases small. It could be avoided if we chose a specific commit hash of the schemas in the constraints whenever schema altering changes were verified in the server.

constraints.txt
david4096: We could add a script that before a release adds the pinned versions and then runs register/upload.

The main thing here is whether we are ok with having the development version pointing at schemas/master as a dependency. This has the nice feature that changes in the schemas are breaking in the server. Although the schemas shouldn't move forward without a working implementation, this gives us better guarantees of that dependency in code.

david4096:
When the client PR at https://github.com/ga4gh/ga4gh-client/pull/35 merges I can update the dependency here to be on ga4gh-client/master

PR #1476, opened 2016-11-29
Fixed OIDC Authentication

Ellmen:
There were a few bugs in the OIDC implementation that prevented it from storing the user sessions properly.

### DIFF
--- a/ga4gh/server/frontend.py
+++ b/ga4gh/server/frontend.py
@@ -307,7 +307,7 @@ def configure(configFile=None, baseConfig="ProductionConfig",
         # If we are testing, then we allow the automatic creation of a
         # redirect uri if none is configured
         redirectUri = app.config.get('OIDC_REDIRECT_URI')
-        if redirectUri is None and 'TESTING' in app.config:
+        if redirectUri is None and app.config.get('TESTING'):
             redirectUri = 'https://{0}:{1}/oauth2callback'.format(
                 socket.gethostname(), app.myPort)
         app.oidcClient.redirect_uris = [redirectUri]
@@ -317,7 +317,7 @@ def configure(configFile=None, baseConfig="ProductionConfig",
 
         # We only support dynamic registration while testing.
         if ('registration_endpoint' in app.oidcClient.provider_info and
-           'TESTING' in app.config):
+           app.config.get('TESTING')):
             app.oidcClient.register(
                 app.oidcClient.provider_info["registration_endpoint"],
                 redirect_uris=[redirectUri])
--- a/ga4gh/server/frontend.py
+++ b/ga4gh/server/frontend.py
@@ -282,14 +282,15 @@ def configure(configFile=None, baseConfig="ProductionConfig",
     app.backend = _configure_backend(app)
     if app.config.get('SECRET_KEY'):
         app.secret_key = app.config['SECRET_KEY']
-    else:
-        app.secret_key = 'super_secret_CHANGE_ME'
+    elif app.config.get('OIDC_PROVIDER'):
+        raise exceptions.ConfigurationException(
+            'OIDC configuration requires a secret key')
     if app.config.get('CACHE_DIRECTORY'):
         app.cache_dir = app.config['CACHE_DIRECTORY']
     else:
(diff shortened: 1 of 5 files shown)


Ellmen:
The tests failed because they are manually trying to change a property of the tokenMap which I have removed. I have replaced the tokenMap with a shelve python db (included in the standard python library).

Currently, the sessions are being stored in a property of the app variable which isn't shared among the different threads. This means the sessions aren't properly stored between requests, giving a "Not Authenticated" error. The fix is to use a database that all the threads can read and write to.

david4096:
@Ellmen nice! In putting together [this PR](https://github.com/ga4gh/server/pull/1470) I used the flask cache to store login information across threads. This should avoid maintaining a temporary file in the tree for sessions and might be easier to test against.

Ellmen:
@david4096 cool, that works! I'll use the cache and update the tests.

Ellmen:
@david4096 did you ever get the SimpleCache working? I'm having the same problem as before (different memory across different apache threads)

david4096:
@Ellmen ugh no. You're absolutely correct that the SimpleCache only works in single process environments! I think you can change it out for the FileSystemCache stored in a temp directory pretty easily?

`class werkzeug.contrib.cache.FileSystemCache(cache_dir, threshold=500, default_timeout=300, mode=384)`

http://werkzeug.pocoo.org/docs/0.11/contrib/cache/#cache-systems

Ellmen:
@david4096 great that works. Using a cache has the advantage of cleaning up old sessions anyway so it's probably a little better than using shelve. I pushed the new version.

ga4gh/server/frontend.py
```diff
@@ -279,16 +279,22 @@ def configure(configFile=None, baseConfig="ProductionConfig",
...
-    app.secret_key = os.urandom(SECRET_KEY_LENGTH)
+    if app.config.get('SECRET_KEY'):
+        app.secret_key = app.config['SECRET_KEY']
+    else:
+        app.secret_key = 'super_secret_CHANGE_ME'
+    if app.config.get('CACHE_DIRECTORY'):
+        app.cache_dir = app.config['CACHE_DIRECTORY']
+    else:
+        app.cache_dir = '/tmp/ga4gh'
+    app.cache = FileSystemCache(app.cache_dir, threshold=500, default_timeout=300, mode=384)
```
david4096: We probably want to make the cache timeout longer. In a way. Just noting... the threshold is a crude way to limit number of allowed sessions.

Ellmen: Ya the timeout isn't too big a deal because after it times out, it can automatically request a new token from the provider. The threshold might be a little low though. I'm happy to increase both.

ga4gh/server/frontend.py
```diff
@@ -279,16 +280,23 @@ def configure(configFile=None, baseConfig="ProductionConfig",
     app.serverStatus = ServerStatus()
 
     app.backend = _configure_backend(app)
-    app.secret_key = os.urandom(SECRET_KEY_LENGTH)
+    if app.config.get('SECRET_KEY'):
+        app.secret_key = app.config['SECRET_KEY']
+    else:
+        app.secret_key = 'super_secret_CHANGE_ME'
```
david4096: You can't generate a new secret key for every process, so thanks for fixing that. Instead of defaulting to a hardcoded value I'd like to say that `SECRET_KEY` is a required configuration item for OIDC support. Failing here if that is not set is acceptable behavior.

david4096:
This is great, thanks Ellmen! I think we ought to add a configuration variable to the `docs/configuration.rst` such that OIDC mode requires the `SECRET_KEY` be set. If you can remove that default and add it to LocalOidcConfig in `serverconfig.py`, I'd be happy to merge!

Ellmen:
Alright, will do!

david4096:
+1 splendid, we'll see if any other review comes in.

Ellmen:
Oops I think you're right. This one _should_ pass.

ejacox:
+1 Thanks. Looks great.

Issue #1480, opened 2016-11-30
Document name generation rules for datarepo


david4096:
- [ ] Add an Architecture diagram showing the registry DB, the server, the data container objects (like HtslibVariantSet, etc), etc, to the documentation for the data repository.

- [ ] Explain how the datamodel is structured and how names are used as local identifiers.

- [ ] Link to name and identifier section from datarepo/add-variantset

Issue #1479, opened 2016-11-30
Document gff3 parsing into FeatureSet


david4096:
The process for generating feature sets needs to be better documented and included with the `demo.rst`.

PR #1481, opened 2016-11-30
Data repository docs to its own file (doc updates)

### LINKED ISSUE #1161: Document database write access
As @david4096 pointed out in #1158 we might need to allow for write permissions for the web server when pointing it at a repository DB file. See [here](https://www.sqlite.org/tempfiles.html) for information on sqlite temp files.

It seems a bit wasteful though, since these files are all for transational updates and so on, and we really just want to open the DB in read-only mode when running within the webserver. This would avoid this problem, and also give us some extra strength-in-depth security wise in terms of data modification through SQL injection attacks.

Unfortunately, there doesn't seem to be an easy way to do this in [Python < 3.4](http://stackoverflow.com/questions/10205744/opening-sqlite3-database-from-python-in-read-only-mode).

Thoughts?

### LINKED ISSUE #1229: Documentation does not tell how to install from source
Where should we document how to install from source? installation.rst is already a bit confusing, but it is a common use case to build from source during development `python setup.py install`. Perhaps in `development.rst`?

### LINKED ISSUE #1473: Move repository documentation to its own file
Configuration and repo commands are in a single file. Splitting them into two files would make it more clear - configuration vs. administration.

### LINKED ISSUE #1477: Running the demo: "ga4gh-example-data/registry.db' does not exist"
Following the demo instructions [here](http://ga4gh-reference-implementation.readthedocs.io/en/latest/demo.html) I get this error:

    ga4gh.exceptions.RepoNotFoundException: Database file 'ga4gh-example-data/registry.db' 
    does not exist.  Either change the configuration to point to a valid file or create one 
    using the repo manager.`

I've checked and the "registry.db" is missing in the tar file. Do I have to create it first? How?

Thanks.

david4096:
* Point at new example data close #1477 
* Move datarepo documentation it its own file close #1473 
* Add note for filesystem permissions close #1161 
* Installing from source close #1229 
* Removed TODOs and made issues #1479 #1480 
* Closed a TODO by moving OIDC demo down a section

See updated docs here http://server-1.readthedocs.io/en/1477_egdata/


ejacox:
+1

kozbo:
I built the docs and read through the new organization and verified that the many issues listed in this ticket are addressed. Looks good.
+1

PR #1483, opened 2016-12-01
Improve read group set generator close #1316

### LINKED ISSUE #1316: Current read group set generator slow
Listing read group sets in the 1kgenomes server becomes intractably slow. A more intelligent readgroupset generator function is needed.

david4096:
Avoids building the entire list before sending back a response. The same should probably be done to the RNA quantifications generator, but would require adding a get-by-index functionality to the SqliteRnaQuantificationSet.

We should strive to use SQL to handle paging and iterating. When the number of items gets very large, this solution, though fast at request time, will take longer and longer to start as we load much of the SQLite database into memory.

Close #1316

### DIFF
--- a/ga4gh/server/backend.py
+++ b/ga4gh/server/backend.py
@@ -202,14 +202,30 @@ class Backend(object):
         defined by the specified request.
         """
         dataset = self.getDataRepository().getDataset(request.dataset_id)
-        results = []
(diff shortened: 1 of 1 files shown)


ejacox:
It is much faster.
+1

kozbo:
+1

PR #1485, opened 2016-12-02
Use ORM for datarepo

david4096:
This replaces the inline SQLite3 code in datarepo to models using the [peewee](http://docs.peewee-orm.com/en/latest/) python module. This will allow us to potentially connect to MySQL or postgresql databases, though that will be addressed in a later PR. It also provides a framework of extensions, one of which we could use to write database migrations to allow registries to be upgraded as necessary. This will also make it easier to remove the memory based backend, as the simple controllers using these models could be used to perform paging and fulfillment of requests.

* Add peewee to requirements.txt
* Add new `models.py` containing representations of the tables
* Alter `datarepo.py` to remove inline SQL
* Modify `populateFromRow` class functions to load data from objects as opposed to dictionaries

This is not a rethinking of the data repository, it simply generalizes the way it interacts with databases. Because of that, some features and class methods of its original design have been left intact, for example, `repo.commit` doesn't do anything. These should be removed, but probably as part of an effort to remove the in memory datarepo representation.

By using this interface directly, the repo manager might be able to be simplified, as CSV database extensions for `peewee` are also available. These changes should eventually be reflected in how the RNA Quantification and Sequence Annotations are backed.


david4096:
As pointed out by @ejacox , this PR adds new classes that are parallel to the datamodel classes. This is undesirable as it creates another hierarchy that a developer has to manage when wanting to instantiate a datamodel (or peewee model) object.

This PR is intended to clear the way for better handling of relational data and removing the in-memory data base. An option presented by @ejacox is to add the peewee fields to our data model objects and then making them inherit from the peewee BaseModel. If this works, it could remove a lot of duplicated code.

Although I'm not attached to this code in particular, it presents a step wise direction towards an interchangeable registry backing. It is meant to cause as little disruption to the current code base while removing inline SQL.

ejacox:
Some notes following a discussion with @kozbo and @david4096.

Pros:
- simple
- adds ability to change the backend database
- cleaner in the code, especially if the parallel object models are merged
Cons:
- doesn't do migration (SQLAlchemy might)
- slower
- less powerful than SQLAlchemy

I timed the build_test_data.py script. It is 20% slower with peewee. Queries are not necessarily slower.

What about bulk insertions, as with RNA recompute?

Overall, as we do not anticipate complex queries, peewee should be sufficient.

ORM Links:
http://pythoncentral.io/sqlalchemy-vs-orms/
https://www.reddit.com/r/Python/comments/4tnqai/choosing_a_python_ormpeewee_vs_sqlalchemy/
http://stackoverflow.com/questions/36946874/peewee-vs-sqlalchemy-perfomance

ga4gh/server/repo/models.py
```diff
@@ -0,0 +1,278 @@
+"""
```
dcolligan: Add ``__future__`` lines

dcolligan:
We've avoided using acronym variable names (e.g. `fs` for a feature set) thus far in the code and that is a practice I would like to continue.  Otherwise, this looks good.

david4096:
Thanks @dcolligan updated with your suggestions.

Travis bug is causing it to not report it's state. You can see the successful travis build for now [here](https://travis-ci.org/ga4gh/server/builds/190762391) I expect it to be intermittent and that a future merge will rerun it allowing this to be mergeable.

ga4gh/server/datamodel/bio_metadata.py
```diff
@@ -146,15 +146,15 @@ def toProtocolElement(self):
                 gaIndividual.info[key].values.add().string_value = value
         return gaIndividual
 
-    def populateFromRow(self, row):
+    def populateFromRow(self, ind):
```
dcolligan: I prefer using the full name... otherwise, what is `ind`?  Index?  Indent?  Indicator?  Individual?  It's nice to have context provided.  A bit more typing, yes, but I think it's worth it, and it's the style that we have in our code hitherto.

dcolligan: Same comment for other shortened variable names below.

david4096: Missed a couple, my apologies :1st_place_medal:

david4096:
I cannot get CI to report status here. Gonna try closing and reopening

david4096:
Addressed @dcolligan's code style concerns. Can we merge this today @ejacox @kozbo ? It's going to cause conflicts the standing PRs that work with the repo.

Alternatively we can do the schema changes that are ready to merge first: https://github.com/ga4gh/server/pull/1513 https://github.com/ga4gh/server/pull/1521 https://github.com/ga4gh/server/pull/1513

Thoughts?

Issue #1486, opened 2016-12-06
G2P Features search accepts regex patterns


david4096:
Although we can block for the internal server error in https://github.com/ga4gh/server/issues/1379 , the g2p endpoint still parses and responds to regex patterns. The schemas for SearchFeaturesRequests use strict matching on string fields, and so this is a bug in terms of our schemas compliance.

PR #1499, opened 2016-12-07
catches malformed regex and throws a new type of exception

### LINKED ISSUE #1379: G2P Features search bug
When I post a request to features/search with `name: "*"` using the compliance dataset I get an Internal Server Error.

```
http post localhost:8000/features/search featureSetId=WyJicmNhMSIsImdlbmNvZGV2MTkiXQ name="*"
```

ejacox:
For a features search, if a search field contains a malformed regular expression, a server error was produced. This fix catches the regular expression error and throws a new exception that returns a "400 BAD REQUEST" with the message "Malformed regular expression"

Close #1379

### DIFF
--- a/ga4gh/server/datamodel/genotype_phenotype_featureset.py
+++ b/ga4gh/server/datamodel/genotype_phenotype_featureset.py
@@ -6,10 +6,12 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import re
 import bisect
 import rdflib
 from rdflib import RDF
 
+import ga4gh.server.exceptions as exceptions
 import ga4gh.server.protocol as protocol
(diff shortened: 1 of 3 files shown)


david4096:
Awesome! It's just a flake styling error in exceptions that's holding this one up. 

Aside, I don't think there's another good way to handle this without altering how the RDF is queried to enforce string matching, as opposed to allowing regex for the features search endpoint. https://github.com/ga4gh/server/issues/1486

For completeness, you might construct a test in `tests/end_to_end/test_g2p.py` You can construct a query like in `testGenotypeSearchByName` then use a malformed regex pattern and assert it returns your exception. +1 mergeable when it passes

ejacox:
Formatting fixed and test added.

PR #1501, opened 2016-12-09
Minor docker updates, fix readthedocs

### LINKED ISSUE #1500: Deployment/installation via pip
Moving the server to its own namespace seems to have had the odd effect that when installing via `python setup.py install`, it is not added to the `ga4gh` module. The current deployment process relies on all of the ga4gh to be importable in the current path, which setup install does not perform. Instead, setup install creates the egg package and links the binary, so the server can be used, but its modules are not importable.

This could be a bug or a feature, but it has the side effect that when installing for deployment, we use `pip`, which moves the server to the `ga4gh` module so it can be imported. The documentation needs to be updated to reflect this. Our `.travis.yml` shows how setup and pip can work together to install a distribution.

```
  - pip install pip --upgrade
# RDFLib requres html5lib;  html5lib requires setuptools version 18.5 or above;
  - pip install setuptools==25.1.0
  - python setup.py sdist
  - pip install dist/ga4gh*.tar.gz
```

To close this issue, update the documentation to reflect installation via pip, as opposed to setup.py directly.

david4096:
The existing install pattern using `python setup.py install` does not install the server to the `ga4gh` namespace in the same way `pip` does. This PR adjusts the documentation to reflect proper installation practices in the Dockerfile, and updates the Dockerfile to include the compliance dataset by default.

Also removes some orphaned documentation.

If you have docker installed you can test it out with `docker run -d -p 8000:80 david4096/server-1:1500_docs`. This will start a server at [http://localhost:8000](http://localhost:8000) hosting the compliance dataset.

I noted that there is a problem with how the G2P feature sets are being loaded in that the path is included in the localId. This causes compliance to fail on some brittle tests on the features protocol. 

Close #1500

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,7 +57,7 @@ EXPOSE 80
 # Prepare container for deployment
 # The directory that the user will land in when executing an interactive shell
 WORKDIR /srv/ga4gh/server
-RUN python scripts/prepare_compliance_data.py -o ga4gh_example_data
+RUN python scripts/prepare_compliance_data.py -o ../ga4gh-compliance-data
 
 # Default action: Bring up a webserver instance to run as a daemon
 CMD ["/usr/sbin/apache2ctl", "-D", "FOREGROUND"]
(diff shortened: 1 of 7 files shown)


kozbo:
Works for me.

PR #1502, opened 2016-12-09
Patch bug for list printing

### LINKED ISSUE #1472: repomanager: List out of range error on readGroupIds when using "list" command
run this on the 1kgenomes server:
```
python repo_dev.py list /run/shm/1000Genomes/registries/repo.db
```
and see the following error
```
  File "/srv/ga4gh0_6_0a8pre/ga4gh/cli/repomanager.py", line 1053, in repo_main
    RepoManager.runCommand(args)
  File "/srv/ga4gh0_6_0a8pre/ga4gh/cli/repomanager.py", line 1031, in runCommand
    runMethod()
  File "/srv/ga4gh0_6_0a8pre/ga4gh/cli/repomanager.py", line 126, in list
    self._repo.printSummary()
  File "/srv/ga4gh0_6_0a8pre/ga4gh/datarepo.py", line 257, in printSummary
    quant._readGroupIds[0],
IndexError: list index out of range
```
Note that the list program displays all data until it gets to printing out the quantifications in the first quantification set, and then it fails.  

There are no readgroupIDs set because we don't have the RNA reads for this RNA data.

david4096:
The readgroupIds were being iterated into directly when at times there are no read group IDs associated with an RNA quantification.

Close #1472

### DIFF
--- a/ga4gh/server/datarepo.py
+++ b/ga4gh/server/datarepo.py
@@ -254,8 +254,8 @@ class AbstractDataRepository(object):
                         print(
                             "\t\t", quant.getLocalId(),
                             quant._description,
-                            quant._readGroupIds[0],
-                            quant._featureSetIds[0], sep="\t")
+                            ",".join(quant._readGroupIds),
+                            ",".join(quant._featureSetIds), sep="\t")
 
     def allReferences(self):
         """


kozbo:
+1 minor change/fix, LGTM

PR #1511, opened 2016-12-19
fix attempt for RTD build dependancies

kozbo:


david4096:
+1 interested to see if it works. The other thing you can try is setting a version in the egg fragment greater than any released version perhaps. https://readthedocs.org/projects/ga4gh-reference-implementation/builds/4819430/

Issue #1478, opened 2016-11-30
API server returns string type to all attributes


Welliton309:
I am testing GA4GH API server and I observed that the server returns `string` type (quoted values) for all attributes, even for attributes that the schemas documentation says other type. For example the attributes `start` and `end` should be `long` ([reference](https://ga4gh-schemas.readthedocs.io/en/latest/schemas/variants.proto.html#protobuf.Variant)) but in the JSON these values are represented as string.

``` bash
ga4gh_client variants-search http://1kgenomes.ga4gh.org/ --variantSetId WyIxa2dlbm9tZXMiLCJ2cyIsInBoYXNlMy1yZWxlYXNlIl0 -r 1 -s 15902 -e 15903 -O json | jq .
```

``` json
{
  "id": "WyIxa2dlbm9tZXMiLCJ2cyIsInBoYXNlMy1yZWxlYXNlIiwiMSIsIjE1OTAyIiwiNWIzYjk4MzNmYmZiMGM0NzVlMGNhMTVlOTU5YTUwNWIiXQ",
  "referenceName": "1",
  "alternateBases": [
    "GC"
  ],
  "names": [
    "rs557514207"
  ],
  "info": {
    "DP": [
      "7012"
    ],
    "EX_TARGET": [
      "True"
    ],
    "NS": [
      "2504"
    ],
    "EAS_AF": [
      "0.868099987507"
    ],
    "VT": [
      "INDEL"
    ],
    "EUR_AF": [
      "0.465200006962"
    ],
    "SAS_AF": [
      "0.532700002193"
    ],
    "AC": [
      "2209"
    ],
    "AA": [
      "ccc|CC|CCC|deletion"
    ],
    "AF": [
      "0.441094011068"
    ],
    "AFR_AF": [
      "0.0430999994278"
    ],
    "AMR_AF": [
      "0.414999991655"
    ],
    "AN": [
      "5008"
    ]
  },
  "variantSetId": "WyIxa2dlbm9tZXMiLCJ2cyIsInBoYXNlMy1yZWxlYXNlIl0",
  "end": "15903",
  "calls": [],
  "created": "0",
  "updated": "0",
  "referenceBases": "G",
  "start": "15902"
}
```

Is this correct? Should the client applications parse the JSON? I also expected that INFO values should be integer or float (i.g. `DP` and `AFR_AF`) because these attributes are unpredictable (may or may not have in VCF file).

david4096:
[...]

PR #1521, opened 2017-01-10
Better type representation in attributes

### LINKED ISSUE #1520: Implement improved attributes
This will replace the info field with a polymorphic field we can use to interchange arbitrary unstructured / loosely typed data.

https://github.com/ga4gh/schemas/pull/700

david4096:
Close #1520 by refactoring usage of the `info` field to an Attributes message, which can contain nested keys in addition to the current features of the `info` field. A version of this message was in use in sequence annotations. This PR uses the same model for tag bags throughout the API.

This refactor is complete, however, to reap the benefit, the field will need to be added to some remaining messages and made configurable through the repo manager.

I think there might be an opportunity to simplify the variant set metadata messages with a similar polymorphism.

stretch
* tagstorm input
* validation
* figure out how to set and persist for RNA quantifications (not managed in the same way)


david4096:
Now better set variant attributes to close #1478

david4096:
@ejacox @andrewjesaitis This is ready for review along with https://github.com/ga4gh/compliance/pull/241 

It still faces a bit of a usability problem when specifying the attributes message at input (see example in [compliance test-data](https://github.com/ga4gh/compliance/pull/241/files#diff-f64ef433a0f11f030423aa8a4cac94c2)) and the test coverage could be improved.

ga4gh/server/datamodel/bio_metadata.py
```diff
@@ -46,9 +45,10 @@ def toProtocolElement(self):
             individual_id=self.getIndividualId(),
             name=self.getName(),
             disease=disease)
-        for key in self.getInfo():
-            for value in self.getInfo()[key]['values']:
-                biosample.info[key].values.add().string_value = value
+        attributes = self.getAttributes()
+        for key in attributes:
+            biosample.attributes.attr[key] \
+                .values.extend(protocol.encodeValue(attributes[key]))
```
andrewjesaitis: Wonder if this pattern should be abstracted onto the `datamodel` class since it's used in all the toProtocolElement calls?
Just something like:
```
def serializeAttributes(self, msg):
   attributes = self.getAttributes()
   for key in attributes:
       msg.attributes.attr[key].values.extend(protocol.encodeValue(attributes[key]))
```

david4096:
@ejacox Please take a look when you have a chance. I think this PR just makes the info field we were already using more accessible across the board.

I tried the basic, "can I pass in JSON" use case and it will work fine. Protobuf will hopefully hide the nasty serialization bits so we can focus on making clients that allow us to easily set the attributes. I think an interactive metadata editing functionality for the repo manager is where we want to be. For now, at least all the items can have an arbitrary tag bag associated with them.

```
python repo_dev.py init a.db
python repo_dev.py add-dataset a.db test -A '{"my_string_key": "my_string_value", "my_numeric_value": 23, "my_float_value": 3.234}'
```

We will need to address handling the tag bag at the application level, but I think protobuf is doing its job here of decoupling the interchange from representation. Using the python API you can do more, but for just attaching an arbitrary JSON this is the best we have to offer right now.

I'd like to have discussions about the direction of the Repo API in this release cycle, especially since we'll be working off an ORM. If possible, I'd like to get these feature merged so that we can move onto other schema PRs.

Note, @ejacox, I like your idea of consolidating the metadata related items into a single field (`.metadata.created, .metadata.attributes`) It might be a challenge getting good consensus for that while we're in the middle of these features.

In general, it is problematic if your API needs to rely on a tag bag for any use cases other than interchange. If there are use cases that are NOT application specific for keys in the tag bag, we should work to move them into the main API. For example, if everyone uses the `gene` tag of a variant message, we might consider adding that as an optional named field. By presenting a field that can have arbitrary hierarchy, we're admitting a kind of defeat. We should work closely with application developers to find the fields they need for their analyses.

PR #1522, opened 2017-01-11
Fixup demo_example script

andrewjesaitis:
Just noticed that this script was out of date with respect to some of the newer changes. Minor stuff, but it works now.

Updated client path and attribute names.


david4096:
Thanks! +1

PR #1523, opened 2017-01-12
Update ontology term

### LINKED ISSUE #1487: Implement updated ontology term
The ontology term representation is getting a little lighter as suggested in https://github.com/ga4gh/schemas/pull/694 . This issue is to track the development of that change in the server.

david4096:
The ontology term has fewer fields and has had a field renamed. This will make it easier to work with in practice and removes redundant fields.

`term.id` becomes `term.term_id`, while `source_version` and `source_name` have been removed.

Close #1487


david4096:
Needs a compliance test before it can be merged.

PR #1527, opened 2017-01-19
Allow empty documents to post

david4096:
When sending a request to datasets/search, sending an empty document without a mime-type could send back reasonable results, but instead we say `Method Not Allowed`. Making the first request to the API _easier_, since its a post, will lower the barrier to entry. `http post localhost:8000/datasets/search` with no data will respond with the same as `{}` with a mime-type set.


david4096:
@dcolligan or @ejacox Mind giving an opinion, review? It's a tiny change. I like this because I can post an empty body and still get back a response from datasets/search.

This would be a problem if we had multiple mime-types responding from our POST endpoints. We don't so I think it's safe to assume application/json.

dcolligan:
Sounds like a good idea, but there should be a positive test in the test suite for this.

david4096:
Thanks @ejacox @dcolligan. I'll add a positive test before merging.

PR #1531, opened 2017-01-20
remove the --user

### LINKED ISSUE #1525: Remove the `--user` parameter for the pip install in the docs
In the github workflow section under development

kozbo:
closes #1525


david4096:
We want to keep the `--user` on that part, which installs pip to the site packages in case they don't have it. However, we want to remove the flag when we are installing our own dependencies in a virtual environment. See https://github.com/ga4gh/server/pull/1532/files

PR #1534, opened 2017-01-20
Update demo.rst

ksens:
* Updated a broken link
* Added a `wget` command to make doc uniform
* Show that `rm` *WGS* is a separate command


david4096:
Thanks a bunch @ksens !

kozbo:
-1 Actually that is the correct way to import the client in the new ga4gh-client namespace. Think that these docs are a little ahead of the released code. We are in the middle of a release now which should be consistent with this doc.

kozbo:
yes, +1 on this

PR #1535, opened 2017-01-20
Update demo.rst

ksens:
* Fixed error in Python code snippet

### DIFF
--- a/docs/demo.rst
+++ b/docs/demo.rst
@@ -273,7 +273,7 @@ around 3GB. Next, we will add the reference set.
 .. code-block:: bash
 
     $ ga4gh_repo add-referenceset registry.db /full/path/to/hs37d5.fa.gz \
-      -d “NCBI37 assembly of the human genome” --ncbiTaxonId 9606 --name NCBI37 \
+      -d "NCBI37 assembly of the human genome" --ncbiTaxonId 9606 --name NCBI37 \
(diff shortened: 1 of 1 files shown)


docs/demo.rst
```diff
@@ -273,7 +273,7 @@ around 3GB. Next, we will add the reference set.
 .. code-block:: bash
 
     $ ga4gh_repo add-referenceset registry.db /full/path/to/hs37d5.fa.gz \
-      -d “NCBI37 assembly of the human genome” --ncbiTaxonId 9606 --name NCBI37 \
+      -d "NCBI37 assembly of the human genome" --ncbiTaxonId 9606 --name NCBI37 \
```
david4096: Good catch!

docs/demo.rst
```diff
@@ -185,7 +185,7 @@ performed above, we can use the following code:
 
     from __future__ import print_function
 
-    import ga4gh.client.client as client
+    import ga4gh.client as client
```
david4096: The signature has changed in the latest version. I personally prefer `from ga4gh.client import client`, but the line you've removed is correct for the current development version. 0.3.4 released version uses the syntax you've provided.

kozbo:
-1 Actually that is the correct way to import the client in the new ga4gh-client namespace. Think that these docs are a little ahead of the released code. We are in the middle of a release now which should be consistent with this doc. (sorry for posting this in the wrong PR earlier. Having both PR's have the same name is confusing)

ksens:
@david4096 do take a look now.

david4096:
Thanks @ksens !

PR #1538, opened 2017-01-23
Add ga4gh-common to constraints.txt

dcolligan:


dcolligan:
Ah, so this is a problem worth noting... if a downstream repo (like the server, here) tries to do development with some commits of upstream repos corresponding to releases (i.e. that have their dependencies pinned) then this problem -- one package looking for an exact version of a ga4gh library that we only have a dev version of and therefore failing -- can result.  Solution: remove all pins for ga4gh packages in upstream ga4gh repos and re-download packages.

dcolligan:
This now depends on fixing https://github.com/ga4gh/schemas/issues/797

PR #1541, opened 2017-01-23
Example data link

david4096:
Links to the proper example data in the notebook in scripts.

I tested that the 4.6 example data will load properly into the server. The only changes that might have affected the example data are the biosample refactor, and since the example data doesn't include any biometadata, the current example data appear to work.

```
wget https://github.com/ga4gh/server/releases/download/data/ga4gh-example-data_4.6.tar
tar -xf ga4gh-example-data_4.6.tar
python server_dev.py
```


david4096:
This file has been removed closing!

PR #1547, opened 2017-01-26
Remove protocol.py

dcolligan:
Issue https://github.com/BD2KGenomics/ga4gh-integration/issues/53


david4096:
@dcolligan last commit still has WIP, not sure if intentional

PR #1556, opened 2017-02-07
Peer service

david4096:
This PR implements the peer service https://github.com/ga4gh/schemas/pull/806 . It does not add any new dependencies, ~~but makes explicit the dependency the server has on the client for performing the announce request~~ (switched to using requests, which was already a dependency). ~~This could easily be replaced by using the requests module directly.~~

It is meant to present the minimum functionality to support implementing the schemas PR. Coordinated testing amongst nodes is not addressed. Atlhough interesting, making peer to peer connections and hostname resolution is not addressed.

Peers can be announced to a server by sending a `AnnouncePeerRequest` to `/announce` that looks like `{peer: {uri: "http://localhost"}}`. These will be added as a timestamped row in the registry so they can be evaluated for being added as potential peers.

```
sqlite> select * from announcement;
1|{"attr": {"version": {"values": [{"int32Value": 32}]}}}|http://1kgenomes.ga4gh.org|127.0.0.1|HTTPie/0.9.6|2017-02-28 16:48:47.515380
2|{}|http://1kgenomes.ga4gh.org|127.0.0.1|python-requests/2.7.0 CPython/2.7.12+ Linux/4.8.0-39-generic|2017-02-28 16:49:33.037886
3|{}|http://1kgenomes.ga4gh.org|127.0.0.1|python-requests/2.7.0 CPython/2.7.12+ Linux/4.8.0-39-generic|2017-02-28 16:49:43.286461
sqlite> 
```

Peers are managed via the repo manager, `ga4gh_repo add-peer http://1kgenomes.ga4gh.org` will add that peer to the registry. These peers are exposed via the `peers/list` endpoint (without rebooting the server). 

In order to support the feature of not reloading the server, some design decisions that expose the SQL of the data repository were made. My hope is that with this PR will help begin to form an opinion on how to separate the messages that require some compound ID (variant), with others that can use auto-incrementing IDs (dataset, peer, referenceset, and so on).

To gather more details about the AnnouncePeerRequest before logging it, the request is handled directly instead of via `handleFlaskPostRequest` breaking some design patterns.

TODO
* ~~Add repo manager CLI methods for listing and clearning announcements.~~
* ~~Repo manager documentation~~
* ~~Configuration setting documentation~~
* ~~Add an initial announce to peers at start.~~ Tested against 1kgenomes, will send an announce packet at server start for any initial peers listed. Setting the URL to announce is still a challenge because of the different configurations we work under.
* ~~Log the user-agent of the `AnnouncePeerRequest`.~~ ~~A JSON with a smattering of extra fields about the announce request are stored to help detect spam.~~ The submitter IP and user agent are stored in the database along with the announce request.
* ~~Protocol version should be set automatically.~~
* ~~Make peer and announcement datamodel objects.~~
* ~~Add end to end tests that use the client (send announcement, list peers, get info).~~
* ~~Add "initialpeers" file that is read at startup.~~


david4096:
S3 is down and so Travis is not working...

david4096:
This code is ready for review @dcolligan @ejacox . I can tell you the ugliest part is probably how announcements are handled. As we continue to develop the use cases on how we will work with announcements I expect to clean this up.

For now, you can add a peer `ga4gh_repo add-peer registry.db http://mypeer.org` and then call `ga4gh_client list-peers http://localhost:8000` to see your peer has been added. You can then remove that peer using `ga4gh_repo remove-peer registry.db http://mypeer.org`.

To test the announcement functionality, `ga4gh_client announce http://localhost:8000 http://mypeer.org`. Then open the registry using `sqlite3 ga4gh-example-data/registry.db` and then `select * from announcement;`. You should see a timestamped messaged that includes the user agent that made the announcement.

Finally, you can observe the protocol version the service provides using `ga4gh_client get-info http://localhost:8000`.

kozbo:
Exciting!

david4096:
Thanks @ejacox good questions.

That first one is a bug, the announce on init isn't currently working since I was trying to suppress its output incorrectly.

Those methods in abstract are only used during testing, yes. I hope we can remove this class entirely as it's in-memory model isn't useful.

It does look a bit ugly to have that code in frontend and I don't like the idea of using frontend as a catch-all for operations. In a couple of cases I needed to have the `app` around to get at the config, and I like isolating code that touches app to the configuration functions. I don't think we want the datamodel code to act directly on configuration values. 

Considering that, I would like to move it to another module and isolate how `app` is used to reduce the LOC in frontend. Sound ok?

ejacox:
I would say move what you can and otherwise mark it as being in the wrong place. We can refactor it later.

david4096:
Updated

Warning: Because of flask's use of the SimpleHTTPServer when running in debug mode, if you set the service itself as an initial peer it will result in a broken pipe.

david4096:
Although 1kgenomes is on a released version and so doesn't have the `announce` endpoint, you can see the requests being logged when the server is started.

`X.X.X.X - - [01/Mar/2017:16:35:38 -0800] "POST /announce HTTP/1.1" 404 271 "-" "python-requests/2.7.0 CPython/2.7.12+ Linux/4.8.0-39-generic"`

ga4gh/server/network/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import os
+import requests
+
+from ga4gh.server import datamodel, exceptions
```
david4096: I had you in mind when I wrote that line @dcolligan :). I do like that we have uniform import statements, I just find this syntax nicer to read. I'll change it back but continue to try to convince you that we could clean up the code base by switching to this style.

```
import ga4gh.server.datamodel.datasets as datasets
import ga4gh.server.datamodel.ontologies as ontologies
import ga4gh.server.datamodel.reads as reads
import ga4gh.server.datamodel.references as references
import ga4gh.server.datamodel.variants as variants
import ga4gh.server.datamodel.sequence_annotations as sequence_annotations
import ga4gh.server.datamodel.continuous as continuous
import ga4gh.server.datamodel.bio_metadata as biodata
import ga4gh.server.datamodel.genotype_phenotype as genotype_phenotype
import ga4gh.server.datamodel.genotype_phenotype_featureset as g2pFeatureset
import ga4gh.server.datamodel.rna_quantification as rna_quantification
```

Might become:

```

from ga4gh.server.datamodel import ontologies, reads, references, variants, \
    sequence_annotations, continuous,  biodata, genotype_phenotype, rna_quantification
```

I think it might actually be more PEPpy?

http://stackoverflow.com/questions/15011367/python-module-import-single-line-vs-multi-line

david4096:
Thanks @dcolligan for your review! All good suggestions. I am also going to add functionality for list and clearing announcements using the CLI.

ga4gh/server/datamodel/peers.py
```diff
@@ -0,0 +1,104 @@
...
+
+import json
+import re
+
+import ga4gh.server.exceptions as exceptions
+
+import ga4gh.schemas.protocol as protocol
+
+
+def isUrl(urlString):
```
david4096: Yeah, instead of a regex you can just catch an exception from `urllib`, I'll do that.

http://grokbase.com/t/python/python-list/018fex83ty/check-url-simply

```
import urllib

def checkURL(url):
try: urllib.urlopen(url)
except IOError: return 0
return 1

if __name__ == '__main__':
assert checkURL('http://slashdot.org')
assert not checkURL('http://foo')

HTH.
Alex.
```

ga4gh/server/network/__init__.py
dcolligan: Another consideration is searching through the code... it is much easier to grep for 'import ga4gh.server.datamodel.variants' to find all the places it is imported than... I'm not even sure what regex one would write for the alternatives.

ga4gh/server/network/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+        textFile.close()
+        ret = text.split("\n")
+    except:
+        if logger:
+            logger.warn("Couldn't load the initial "
+                        "peer list. Try adding a "
+                        "file named 'initial_peers.txt' "
+                        "to {}".format(os.getcwd()))
+        ret = []
+    # Remove lines that start with a hash or are empty.
```
david4096: https://github.com/ga4gh/server/issues/1595

PR #1557, opened 2017-02-07
Dependency management - relaxing requirements

david4096:
* Relax some version requirements (`==` becomes `>=`)

* I haven't seen many projects use the `pipedeptree` approach to defining dependencies. I understand that these packages should be able to maintain their own dependency tree, and we don't need to (and shouldn't) try to manage their dependencies for them.

* Remove the `humanize` dependency, which was used for one line on the landing page (server uptime nicely printed).

* The lxml requirement is only needed for testing https://github.com/ga4gh/server/blob/73e496bc2995a4066ca35411c6e4ac1c7097ea77/tests/end_to_end/test_oidc.py .

* sphinx-argparse is needed in RTD, but not for basic server usage.

Note: I did not relax the requirements for Flask as there appear to be some problems with how we are using the flask configuration for logging. A future PR will update our dependencies for Flask and Flask_cors.

Note, this will keep our software better up to date as CI will be able to find when one of our dependencies triggers a breaking upgrade. During release, it is possible to once again fix the requirements to a known working version of a module.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -29,6 +29,7 @@ snakefood==1.4
 lxml==3.4.4
 Sphinx==1.4.6
 sphinx_rtd_theme
+sphinx-argparse==0.1.15
 
 # Requirements for OIDC provider
 pyaml==15.03.1
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -26,6 +26,7 @@ PyVCF==0.6.7
 freezegun==0.3.6
 guppy==0.1.10
 snakefood==1.4
+lxml==3.4.4
 Sphinx==1.4.6
 sphinx_rtd_theme
 
--- a/constraints.txt
+++ b/constraints.txt
@@ -25,5 +25,5 @@
 # ga4gh-schemas==0.6.0a9post2
 #
 git+git://github.com/ga4gh/ga4gh-common.git@master#egg=ga4gh_common
-git+git://github.com/ga4gh/schemas.git@master#egg=ga4gh_schemas
+git+git://github.com/david4096/schemas.git@reqts#egg=ga4gh_schemas
 git+git://github.com/david4096/ga4gh-client.git@reqts#egg=ga4gh_client
--- a/constraints.txt
+++ b/constraints.txt
@@ -26,4 +26,4 @@
 #
 git+git://github.com/ga4gh/ga4gh-common.git@master#egg=ga4gh_common
 git+git://github.com/ga4gh/schemas.git@master#egg=ga4gh_schemas
-git+git://github.com/ga4gh/ga4gh-client.git@master#egg=ga4gh_client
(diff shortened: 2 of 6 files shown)


dev-requirements.txt
```diff
@@ -26,8 +26,10 @@ PyVCF==0.6.7
 freezegun==0.3.6
 guppy==0.1.10
 snakefood==1.4
+lxml==3.4.4
 Sphinx==1.4.6
 sphinx_rtd_theme
+sphinx-argparse==0.1.15
```
david4096: Do you think that should be in requirements? I consider the doc creation as the development dependencies. We can manage RTD build dependencies in a nicer way using the `environment.yml`. However, this PR is untested on RTD.

david4096:
Thanks both! And thanks for the link @dcolligan. We definitely want to strike a balance between keeping our code base active and delivering packages we trust will run. 

I think that the approach of adding `pipedeptree` is nice for giving better guarantees, and I don't have a problem with making that part of the release process. I think that CI makes this situation easier to maintain, a nightly build failure will let us know which dependency updated. This is a better developer position to be in than guessing which modules should receive updates.

If we are using modules in ways that break when *their* dependencies change, we should try to remove those bits of code. Those are like private API calls and we want the testing machinery to reveal them. By relaxing the requirements we might more information about how to maintain the software. Not to mention bug fixes and performance improvements in our dependencies.

Do you think we could try a cycle or two with pipedeptree removed and see if we get any spurious problems? Managing dependencies in python is a nightmare so I think in practice many people will dockerize their environment anyway.

dev-requirements.txt
dcolligan: Er, the client has `sphinx-argparse` in `dev-requirements.txt`, so moving it seems fine

dcolligan:
Saying this is hurting the cloud software portion of my soul, but sure, we can try the relaxed requirement strategy.  Having the nightly CI builds at least gives us worst-case 24 hours notice of a problem.  On the other hand, we will probably have to make more bugfix releases this way to respond to problems with dependencies shipping versions that break our code.

kozbo:
l like the idea of pinning our dependencies at release time for the very reason that @dcolligan expressed. We should know which versions we work with and then enforce them on our releases to assure that our installations work as promised.

david4096:
I'll leave it up to y'all! From my position it's information we're not getting because our dev version has the versions pinned for all it's requirements. I understand it's a little more work for the release. You'd run `pip list` and set the version for any packages you want to set.

From @dcolligan in #1007 
> I know that this strategy does have downsides, including not getting the most updated packages, with their bugfixes and features. However, this is a cost that I am willing to pay, and -- more to the point -- it's not the priority.

Also, worth noting that you point out a couple of bad apples. `yubico-client`, `pyjwkest` are both modules that are part of Lshift's OIDC module, `yubico` being just for tests. I think this might be a case of having to overthink things because of these packages. I want to (automatically) know which packages aren't being managed well, so we can deprecate them.

PR #1559, opened 2017-02-07
Add prepare compliance data to travis

david4096:
A bug in the prepare compliance test data wasn't caught by the test machinery. This adds it to Travis so that `prepare_compliance_data` will run against the current compliance/master. Thanks @ejacox !


dcolligan:
Hm, we're implicitly adding a dependency here on the compliance repository.  Do we really want to do that?

ejacox:
Can this test be put in compliance?

david4096:
Great makes more sense over there, thanks both! I think making the implicit dependence there is fair (what good is the compliance suite if we can't prepare the data).

https://github.com/ga4gh/compliance/issues/246

Issue #1334, opened 2016-08-08
Add info to transcript effect record 


david4096:
To go with https://github.com/ga4gh/schemas/pull/648

kozbo:
@david4096  #648 is merged. closing

david4096:
Let's keep this open unless the feature is implemented in the server 👍

andrewjesaitis:
Started working on this tonight.

david4096:
Awesome!

I just started implementing an upgraded tag bad (info field to attributes). The field is added across the board but I haven't addressed the "other keys" in annotated VCFs yet. Take a look at https://github.com/ga4gh/server/pull/1521 . It's pretty painless to switch back and forth between the two. I think we're slowly converging toward a type reflection method that will work for read tags, VCF tags, and random metadata.

david4096:
Closed with https://github.com/ga4gh/server/pull/1562

PR #1562, opened 2017-02-10
Fix for #1334: Refactor transcript annotation

andrewjesaitis:
* Places extra data in attributes field
* Simplify code path for annotation parsing


ga4gh/server/datamodel/variants.py
```diff
@@ -1040,6 +1057,33 @@ class HtslibVariantAnnotationSet(AbstractVariantAnnotationSet):
     Class representing a single variant annotation derived from an
     annotated variant set.
     """
+    CSQ_FIELDS = ("alternate_bases", "gene", "feature_id",
```
david4096: I aspire to have these tuples as part of the registry editor.

It should be possible to easily write a VCF->ANN plugin that accepts a tuple or dictionary that can effectively parse a ANN field.

ga4gh/server/datamodel/variants.py
```diff
@@ -37,6 +38,34 @@ def isUnspecified(str):
...
+def deepGetAttr(obj, path):
+    """
+    Utility function which is similiar to std lib's getattr, but works
+    to resolve a dot-delimited path on an object. If path is not found
+    an `AttributeError` will be raised.
+    """
+    return reduce(getattr, path.split('.'), obj)
+
+
+def deepSetAttr(obj, path, val):
```
david4096: These could probably go into protocol in schemas. I think this pattern of dot access for dictionaries is a sane way to work with data when you don't want to make a new class in python, and mocks the protocol buffers and peewee models well.

This is what I did for the peer service
```
class Struct:
    def __init__(self, **entries):
        self.__dict__.update(entries)

###

map(lambda x: Struct(**x), self._peers[offset:offset + limit])
```
https://github.com/ga4gh/server/pull/1556/files#diff-02d20abc7e4d68c8ae177064ca45176aR49

ga4gh/server/datamodel/variants.py
andrewjesaitis: Yeah, I thought the same thing since we are mixing data with code here. But, we'll save it for an annotation plugin refactor.

ga4gh/server/datamodel/variants.py
andrewjesaitis: Cool. I'm moving these over to schemas then.

Here's the recursive version:

```python
def deepGetAttr(obj, path):
    first, sep, rest = path.partition(".")
    obj = getattr(obj, first)
    if rest:
        obj = deepGetAttr(obj, rest)
    return obj
```

Seems like the recursive version is a little bit uglier. Any specific benefit you were thinking of?

andrewjesaitis:
Also, how does it work when a PR in server depends on a PR in schemas? Should I commit my constraints.txt file that points to my version of schemas that has he edited protocol.py?

dcolligan:
Here's how I would advise doing it:
- push a branch to ga4gh/schemas and open a PR on it
- open a PR with your contstraints.txt file line for schemas pointing to the remote schemas branch
- presumably both the schemas and the server changes are going in, and the server depends on schemas, so merge the schemas PR first, once it is approved
- change your server PR to use the default version of constraints.txt (that is, the one that points to all the master branches)
- merge the server PR

ga4gh/server/datamodel/variants.py
david4096: Sorry, failed to communicate myself properly how you've approached it is fine. Sending in dot notation as a string scares me a little bit :).

I was thinking something [like this](https://gist.github.com/david4096/cdf8717768ee6d33d6424347759949e8), but don't let it hold up this PR.

david4096:
Hey we'll merge this next @andrewjesaitis !

PR #1564, opened 2017-02-10
speed ups to rna quant ingest

### LINKED ISSUE #1549: Improve performance of rnaseq2ga module
We have developed new code that implements the same functionality as rnaseq2ga.rnaseq2ga() but is MUCH faster. This issue asks that the new faster version replace the old version without changing any of the parameters. @saupchurch

ejacox:
I made the expression id an auto increment field and sped up reading the expression data from the file. Closes #1549

### DIFF
--- a/ga4gh/server/repo/rnaseq2ga.py
+++ b/ga4gh/server/repo/rnaseq2ga.py
@@ -89,7 +89,7 @@ class RnaSqliteStore(object):
         take a long time.
         """
 
-        sql = '''CREATE INDEX feature_id_index 
+        sql = '''CREATE INDEX feature_id_index
                  ON Expression (feature_id)'''
         self._cursor.execute(sql)
         self._dbConn.commit()
--- a/ga4gh/server/repo/rnaseq2ga.py
+++ b/ga4gh/server/repo/rnaseq2ga.py
@@ -4,7 +4,6 @@ from __future__ import unicode_literals
 
 import sqlite3
 import csv
-import uuid
 
 import ga4gh.server.exceptions as exceptions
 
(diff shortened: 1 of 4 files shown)


ejacox:
Added fix to oidc problem also to close #1566.

ejacox:
@david4096 I removed uuid id generation, which was very slow. The auto increment was just for convenience. The ids can be generated by incrementing a counter in the code, too. On reflection, this will need to be changed to assign ids incrementally by RnaQunatifiction in order to have reproducible ids. The primary key will need to be changed to be the id column and the rna_quantification_id.

ejacox:
I changed the id as described above and added an ingest test to increase the code coverage.

ga4gh/server/repo/rnaseq2ga.py
```diff
@@ -157,6 +190,7 @@ def writeExpression(self, rnaQuantificationId, quantfilename,
                               rawCount, score, units, confidenceLow,
                               confidenceHi)
                 self._db.addExpression(datafields)
+                expressionId += 1
```
david4096: Cool! Do you think this same idea will work for adding expressions in parallel (ignoring sqlite)?

tests/datadriven/test_rna_quantification.py
```diff
@@ -175,3 +180,22 @@ def testSearchExpressionLevelsWithFeatureIds(self):
...
+        """
+        tempDir = tempfile.mkdtemp(prefix="ga4gh_rna_quant",
+                                   dir=tempfile.gettempdir())
+        dbName = os.path.join(tempDir, "rnaQuantDB")
+        storeDb = rnaseq2ga.RnaSqliteStore(dbName)
+        storeDb.createTables()
+
+        testTsvFile = os.path.join(
+                            paths.testDataDir,
+                            "datasets/dataset1/rnaQuant/rsem_test_data.tsv")
```
david4096: This test is really just guaranteeing the whole this doesn't go belly up, which is a great improvement!

ga4gh/server/repo/rnaseq2ga.py
ejacox: Yes, since each file will be processed within it's own process/thread.

Issue #1567, opened 2017-02-13
Proposal: test to prevent bad constraints.txt files from being merged into master


dcolligan:
Here's something I've been thinking about.  There seems to be a constant danger of merging a modified constraints.txt file into master.  I propose we create a test that prevents this from happening:

- check in a file, `constraints.txt.default` that is a constraints.txt file that points to all of the master branches -- that is, what the `constraints.txt` file's contents in master should always be
- add a unit test that checks whether `constraints.txt.default`'s contents are identical to `constraint.txt`'s, and fail if not

Pro: would prevent check-ins of erroneous constraint.txt files to master
Con: would cause a general test failure whenever constraint.txt is modified

Opinions?  @david4096 @andrewjesaitis

andrewjesaitis:
I think this is a great idea. I guess the alternative would be to add `constraints.txt` to `.gitignore`. I guess it depends on which tests you want to fail when submitting a pair of PR's to schemas and server.

If we unit test `constraints.txt`, I think the flow would be like:
1. Push PR to schemas - schemas test PASS
1. Push PR to server with constraints.txt pointing to schemas PR branch - server tests FAIL (because of constraints file differs)
1. Schemas PR merged
1. Update server PR with `constraints.txt` pointing at schemas@master - server tests PASS

With the `.gitignore`:
1. Push PR to schemas - schemas test PASS
1. Push PR to server; modified constraints.txt not pushed - server tests FAIL (because of missing dependency)
1. Schemas PR merged - server tests PASS

Problems with the `.gitignore` is that
* Can't easily push working demos that rely on modified `constraints.txt`
* Not sure how you kick Travis CI to rerun tests when a schemas PR is merged

I vote `constraints.txt` unit test.

ejacox:
I think the approach sounds reasonable. Instead of a .default file, could git be used to check if the file changed? 'git diff' maybe

dcolligan:
Fixed in #1568

dcolligan:
Parallel PRs in

schemas: https://github.com/ga4gh/schemas/pull/817

client: https://github.com/ga4gh/ga4gh-client/pull/62

PR #1568, opened 2017-02-13
Prevent check-ins of non-default constraints.txt

dcolligan:
Issue #1567

### DIFF
--- a/constraints.txt.default
+++ b/constraints.txt.default
@@ -1,0 +1,29 @@
+# This file allows our dependencies to be against specific versions
+# or git tags. Each line is similar to a requirements.txt line, 
+# except that when installing the dependencies, the constraints.txt
+# is converted into the `dependency_links` of the setuptools
+# packaging script. This allows us to maintain dependencies across
+# repositories.
+#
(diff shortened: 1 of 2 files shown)


david4096:
It's a nice feature to be able to see if all the tests pass and something is merge-able. There isn't a huge problem of checking in the constraints because each PR that affects the dependencies will be forced to rewrite it to test anyway.

If there's a way to make this guarantee without causing all of CI to report failure, I'd consider it.

dcolligan:
Having the `constraints.txt` file entries not point to master is an easy thing to get wrong.  Is is an annoying thing to merge a non-default `constraints.txt` file into master because someone has to make a commit to undo it.

I don't think there is a way to do this without CI failure, since the whole point is to fail CI if the guarantees we want to enforce on the code do not hold.

david4096:
OK works for me! It's a good guarantee that we always only merge things that have had their dependencies merged.

I like being able to see what is passing by the green light even when the constraints has changed. But I can always check the Travis details, and you can always mention if something is ready for merge!

PR #1572, opened 2017-02-15
fixed problem

ejacox:
The rnaseq2ga pull request (#1564) caused a problem with compliance. This fixes it.


david4096:
Thanks @ejacox !

PR #1573, opened 2017-02-15
Bigwig

ejacox:
Adds server code to implement the continuous sequence annotation service as part of ga4gh/schemas#769.


david4096:
Really stellar code! Just took a look and it follows the existing patterns clearly.

Noting that the only failing test is `test_constraints`.

requirements.txt
```diff
@@ -53,6 +53,8 @@ requests==2.7.0
 oic==0.7.6
 pyOpenSSL==0.15.1
 lxml==3.4.4
+pyBigWig==0.3.2
+numpy==1.11.3
```
david4096: Does it get used?

ga4gh/server/datamodel/continuous.py
```diff
@@ -0,0 +1,430 @@
...
+        """
+        Read a bigwig file and return a protocol object with values
+        within the query range.
+
+        This method uses the bigWigToWig command line tool from UCSC
+        GoldenPath. The tool is used to return values within a query region.
+        The output is in wiggle format, which is processed by the WiggleReader
+        class.
+
+        There could be memory issues if the returned results are large.
```
david4096: This is a very clever usage of the existing tools to support the BigWig format using the reader you wrote! You note the issues of running a subprocess. It seems like you opted to use the existing python tools, which I think is a very sane choice. Would you like to leave this in for future feature support? Is it used?

requirements.txt
ejacox: Not anymore. Removed.

david4096:
Would you please add an entry to the templates/index.html for the continuous sets?

david4096:
I had to install the `libcurl4-gnutls-dev` to successfully install PyBigWig. I wasn't able to replicate this issue uninstalling the package though. PyBigWig seems to want to have `curl-config` around for accessing remote BigWig files.

david4096:
Congratulations @ejacox this is a sizable addition to the protocol and does a good job at making some of Jim Kent (and other's) work available to a wider community!

Issue #1569, opened 2017-02-15
Deploy from travis


david4096:
It is possible to automatically release from successful travis builds using https://docs.travis-ci.com/user/deployment/pypi/ .

dcolligan:
This came up a while back but we elected not to do it; I forget why.  What do you see as the advantages / disadvantages of this?

david4096:
Basically, it's less time thinking about how to do a release and more time writing code. Having successful master merges release from Travis is a nice way to do continuous release.

Having continuous/nightly builds instead of manual alpha builds would let us test out the PyPi machinery between tagged releases without any work. If Travis never overwrites the tagged releases, it can be a good source for alpha builds without any real risk of harm (other than filling pypi with a lot of quickly obsoleted releases).

dcolligan:
It could be worth investigating, but the only step this really saves us is uploading the source tarball to pypi, which is one less step of N, I suppose.

dcolligan:
I'm not sold on the nightly alpha builds idea, but I do think auto-releases from Travis is a good idea.  The only drawback, so far as I can see, is that the release to pypi will not get done for up to 24 hours, since the push will be triggered by a cron job, not a PR build (since the release is tagged only after the release commit is merged into master).  If we have a requirement of pushing to pypi immediately after a release, then we shouldn't go this route, but so far as I know that's not a concern.  (In the event of an emergency, we can always fall back on manual uploading steps.)

dcolligan:
Er, that drawback can be mitigated as well since Travis builds can be manually triggered http://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit

kozbo:
Is it a common pattern to store all of our intermediate builds on Pypi? Will this confuse our Python users as to which version to download?

dcolligan:
Automatic deployments from ga4gh-common are working as of this release: https://github.com/ga4gh/ga4gh-common/releases/tag/v0.0.6

dcolligan:
Just for future reference, the command I found that worked was:

`$ travis setup pypi --repo ga4gh/ga4gh-common --debug --force` (modulo the appropriate repo name)

The `travis encrypt` command that they recommend was a bit funky.

david4096:
Very cool. So you tag a release and then cron will put it in PyPi for you?

dcolligan:
[...]

PR #1576, opened 2017-02-23
Automate pypi deployment

dcolligan:
https://github.com/ga4gh/server/issues/1569


david4096:
Thanks, one less thing to think about! +1

Issue #1491, opened 2016-12-07
Add quality and filters to variant


david4096:
This PR improves the fidelity of our VCF/variant representation by adding named fields for the filters. This will require interpreting the filters field of the pysam record. Since it doesn't offer any new queries, just new data, this change could be fairly small in lines of code.

https://github.com/ga4gh/schemas/pull/777

david4096:
Closed with https://github.com/ga4gh/server/pull/1577

PR #1577, opened 2017-02-24
add filters

ejacox:
Server code for PR ga4gh/schemas#777 adding fields for variant filters.

### DIFF
--- a/tests/datadriven/test_variants.py
+++ b/tests/datadriven/test_variants.py
@@ -83,6 +83,7 @@ class VariantSetTest(datadriven.DataDrivenTest):
         self._vcfVersion = metadata["fileformat"]
         self._infos = vcfReader.infos
         self._formats = vcfReader.formats
+        self._filters = vcfReader.filters
         self.vcfSamples = vcfReader.samples
         for record in vcfReader:
             self._reference_names.add(record.CHROM)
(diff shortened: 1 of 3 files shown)


tests/datadriven/test_variants.py
```diff
@@ -192,6 +193,14 @@ def _verifyVariantCalls():
                     self.assertEqual(str(alt1), str(alt2))
             else:
                 self.assertEqual(gaVariant.alternate_bases, alt)
+            if not gaVariant.filters_applied:
```
david4096: Are you confident these tests do run for some of our example data? There isn't a good way to guarantee this path is hit per test, since not all VCF will have filters. Maybe we could guarantee that for the suite this path is hit? If there is a file that demonstrates this behavior in the tests we might mention it in a comment for now.

The compliance VCF has all `PASS` I think https://github.com/ga4gh/compliance/blob/master/test-data/brca1_1kgPhase3_variants.vcf .

ejacox: There are variants with no filters applied (marked with '.') in tests/data/datasets/dataset1/variants/WASH7P_annotation/WASH7P_annotation.vcf.gz.

david4096: Thanks, I made this issue so that we can circle back to it. https://github.com/ga4gh/server/issues/1589

kozbo:
closes #1491

PR #1578, opened 2017-02-24
Document auto-releases, constraint instructions

dcolligan:
@david4096 after these changes are approved I will make similar changes in the constraints.txt of the other repositories.


david4096:
+1 Thanks for fixing this up.

PR #1596, opened 2017-03-02
Make referencesets optional when possible

david4096:
The repo manager requires that a reference set is provided when it is really only a metadata item from the perspective of the variants, features, and RNA. For this reason, we can make it an optional flag at the repo manager.

* ~~Allow nullity for referencesetid field in models~~
* ~~Safely access the referenceset name in a few places~~
* Make repo manager safely access referenceset for featuresets, RNA, annotations

We should strongly consider a PR at the schemas that removes the `referenceId` as a required field of a reads search, as that is the only endpoint that can't implement this feature.


ga4gh/server/datarepo.py
```diff
@@ -221,9 +221,13 @@ def printSummary(self):
                         sep="\t")
             print("\tVariantSets:")
             for variantSet in dataset.getVariantSets():
+                referenceSetName = ""
```
david4096: These should be refactored into one method.

ga4gh/server/cli/repomanager.py
```diff
@@ -591,6 +585,18 @@ def func():
         self._confirmDelete(
             "RnaQuantificationSet", rnaQuantSet.getLocalId(), func)
 
+    def safeGetReferenceSetByName(self, referenceSetName):
+        """
+        Gets a reference set by name from the repository throwing a useful
+        exception if it is not found.
```
david4096: I think this is a noop

PR #1605, opened 2017-03-08
Bio age

ejacox:
Adds a bio age field ga4gh/schemas#849


david4096:
+1 minimum change required to support the schema feature.

Issue #1457, opened 2016-11-03
Deprecate converters, move to own repo


david4096:
@dcolligan I believe these should be in the client. Opinions?

https://github.com/ga4gh/server/blob/master/ga4gh/cli/ga2sam.py
https://github.com/ga4gh/server/blob/master/ga4gh/cli/ga2vcf.py
https://github.com/ga4gh/server/blob/master/ga2vcf_dev.py
https://github.com/ga4gh/server/blob/master/ga2sam_dev.py
https://github.com/ga4gh/server/blob/db3f7bd35944b7d0f510c62760eaa071bf53d7d8/tests/unit/test_converters.py
https://github.com/ga4gh/server/blob/master/ga4gh/converters.py

They're all client side applications I believe.

dcolligan:
I was thinking about this, too... I could see an argument either way.  I say we hold off on it for the moment.  We can't do it without at least some refactoring, since `converters.py` imports some things from `datamodel`.  Also the tests use some server code as well.  Let's wait until after #1449 is fixed, at least.

david4096:
No argument here! It looks to me like the parts of the datamodel it uses are just bytemaps for all the SAM flags. Those seem like they could be moved to common easily.

dcolligan:
A few things:

1) Moving the converters to the client repo would have us either have us not test the converters fully in the client repo, and keep the tests (which depend on server) in server, or import the server package in client for the purpose of testing.  I suppose either is possible.

2) The flag classes in datamodel can not be moved to common because they have a dependency on the protocol, which lives in schemas.  The lowest we could push them down, therefore, would be to schemas.  Does it make sense to have these classes in the schemas repo?  I could see arguments either way.

3) How does the potential moving of the datamodel to its own repo affect this issue?

david4096:
Thanks @dcolligan!

1. My preference is: import the server in the client dev-requirements. It's actually quite nice that we can ship a fully functional server for test.
2. Which flags module? The read SAM flags? I think that it makes sense to provide functions that convert byte values to protobuf enum, which is not necessarily how you might use the feature in the server. Also, in general, for esoteric values like this that only take a few lines, one can err on the side of redundancy.
3. I think since the converters decouple via the client, it would simply change the testing apparatuses. If we put ga4gh-pysam into it's own repository then tests of the pysam transalation feature of the server would have to specifically import it. But you'll still have to import a whole server in order to test it.

[...]

PR #1610, opened 2017-03-09
Remove converters

dcolligan:
Issue #1457

Deleting these from the server repo, which everyone seems to agree is a good idea... the issue of where to put them, if anywhere, can be solved elsewhere.


david4096:
Thanks for doing this. We'll definitely get back here https://github.com/ga4gh/ga4gh-server/issues/83

+1

PR #1612, opened 2017-03-10
Remove dead code with the help of vulture

david4096:
https://pypi.python.org/pypi/vulture revealed some dead code. You will need to rebuild your test data.

There is also orphaned code that is under test but doesn't present any features that vulture doesn't catch.

@ejacox note I removed the fillEnd functions if you planned on keeping them for something.


ga4gh/server/repo/rnaseq2ga.py
```diff
@@ -82,22 +82,6 @@ def batchAddExpression(self):
             self._dbConn.commit()
             self._expressionValueList = []
 
-    def createIndices(self):
```
ejacox: This should be left in. It can be used by clients importing this file.

ejacox:
It's OK to remove fillEnd, but keep createIndices in rnaseq2ga.

david4096:
Thanks @ejacox . Did I miss where the code that uses the indices is? You use it in an external script?

david4096:
I also removed the `.whiskey` directory that worked with an old Dockerfile.

ga4gh/server/cli/__init__.py
```diff
@@ -9,14 +9,6 @@
 
 import ga4gh.schemas.protocol as protocol
 
-# the maximum value of a long type in avro = 2**63 - 1
-# (64 bit signed integer)
-# http://avro.apache.org/docs/1.7.7/spec.html#schema_primitive
-# TODO in the meantime, this is the max value pysam can handle
-# This should be removed once pysam input sanitisation has been
-# implemented.
-AVRO_LONG_MAX = 2**31 - 1
```
dcolligan: heh, we always forgot to get rid of this ugly wart

ga4gh/server/repo/models.py
```diff
@@ -86,7 +79,6 @@ class Biosample(BaseModel):
     individualAgeAtCollection = pw.TextField(null=True)
 
     class Meta:
-        db_table = 'Biosample'
```
dcolligan: hm, are these `db_table` attributes meaningful, even though they are not used by us? (i.e. used as a directive to some other library)

david4096: They were an extra string that named the tables. Now they are named by the class name (less code same functionality). It does require you rebuild the test data since they may change.

The answer is no, there isn't another library that chooses the tables in our registries by name, but this may affect any ETL scripts that act on the registry by table name.

ga4gh/server/cli/__init__.py
david4096: Protobuf's JSON serialization has a similar problem it works around, in that it doesn't serialize longs as Javascript numbers (that max at 2**57 -1 or something), but as strings.

Issue #1565, opened 2017-02-10
Update Auth0 documentation


Welliton309:
I am trying to figure out how to configure a working example of GA4GH server with Auth0 authentication service. I am using the latest release version (0.3.5). Because Auth0 service should be configured via web browser, I created a step-by-step tutorial with my progress. I think the following pictures and text may help others. The problem I found is at the end of this text.

Follow the [Deployment on Apache](http://ga4gh-reference-implementation.readthedocs.io/en/latest/installation.html#deployment-on-apache) instructions to install the reference implementation server.

At Apache configuration, you should use the root path instead of `/ga4gh` because the server is using absolute path when redirecting web address.

```
WSGIDaemonProcess ga4gh \
    processes=10 threads=1 \
    python-path=/srv/ga4gh/ga4gh-server-env/lib/python2.7/site-packages \
    python-eggs=/var/cache/apache2/python-egg-cache
WSGIScriptAlias / /srv/ga4gh/application.wsgi

<Directory /srv/ga4gh>
    WSGIProcessGroup ga4gh
    WSGIApplicationGroup %{GLOBAL}
    Require all granted
</Directory>
```

# Configuring Auth0 service
First login or sign up in Auth0 website: https://auth0.com/

# Creating Client
On tab `Client` click in `Create Client`. Give a name for your Client and choose the `Non Interactive Clients` client type. Click `Create`.

![create_client](https://cloud.githubusercontent.com/assets/645272/23128816/0c753c58-f75f-11e6-94a2-c89d6bc1b8b6.png)

In `Settings` tab copy the `Domain`, `Client ID` and `Client Secret` data.

![get_client_secret_domain](https://cloud.githubusercontent.com/assets/645272/23128957/9eb48510-f75f-11e6-9487-74684023331f.png)

These data will be used to set the following server configuration values ([reference](http://ga4gh-reference-implementation.readthedocs.io/en/latest/configuration.html#configuration-values)):
- `AUTH0_HOST` with `Domain`;
- `AUTH0_CLIENT_ID` with `Client ID`;
- `AUTH0_CLIENT_SECRET` with `Client Secret`.

Fill `Allowed Callback URLs`, `Allowed Logout URLs` and `Allowed Origins (CORS)` with the web address of server endpoint plus the related path.

![allow_address](https://cloud.githubusercontent.com/assets/645272/22839443/9a62f994-efb0-11e6-8533-ed93db81871e.PNG)

# Creating API
[...]

PR #1620, opened 2017-03-10
Add Auth0 docs and ~b64decoding

david4096:
Thank you SOOO much @Welliton309 for writing these docs. The community could really benefit from easier authentication/authorization solutions!

This PR fixes a problem found in #1565 that Auth0 changed the formatting of their client secrets (no longer b64 encoded). This PR fixes that to support all future Auth0 clients.

The links to the images are hosted by github's cache, which is probably not a great solution and they should be checked in.

TODO

~~Move images to static directory.~~


kozbo:
+1, and merging

PR #1634, opened 2017-03-29
Protobuf 3.2.0

ljdursi:
Bumps protobuf version to 3.2.0, which means C++ implementation on Linux and so faster/reduced memory usage. Addresses issue ga4gh/ga4gh-server#1623.

The C++ implementation is fussier than the python version - this means that pythonic things like conversion between string types doesn't happen automatically, so that strings passed to protobuf routines have to be byte strings, not unicode strings.  Floating point comparisons are done at float32 precision.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ sudo: false
 cache:
   directories:
     - $HOME/.cache/pip
-    - /tmp/proto3.1.0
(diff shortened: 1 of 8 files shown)


david4096 (approved):
Updates requirements and tests, +1 thanks!

PR #1640, opened 2017-10-03
Added deprecation notice

rishidev:
At a GA4GH Meeting it was decided to formally halt work on the Reference Server

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -1,6 +1,11 @@
 
 .. image:: http://genomicsandhealth.org/files/logo_ga.png
 
+=================================================
+THIS REPOSITORY HAS BEEN DEPRECATED FOR NEW WORK
+=================================================
+Due to a GA4GH move to produce smaller, tightly defined, standards in smaller bites, future work on the reference server implementation will cease. This repository is no longer suitable as a place for future work. Please see the 'GA4GH website <https://ga4gh.org>' for further details on the GA4GH Connect operational structure (which includes details on Work Streams), lists of existing standards and contact details should you have further queries. 
+
 ==============================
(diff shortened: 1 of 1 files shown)


ljdursi:
?? It seems like it might have been useful to wait for the plenary to make that decision?

kozbo:
I agree, this does deserve some discussion at the Plenary, though effectively, development has already halted. My team has been reassigned with other priorities. My team believes that the reference server is not meant to be a product. It provides one example of how one might implement a server to support the GA4GH data APIs, but we do not feel that it is a supportable development path. There are much more flexible and light weight approaches that can be taken.  

The UCSC team has shifted its attention to studying the best ways to collect and provide access to RNA data via the HCA GA4GH driver project. We are also working on the  GA4GH DREAM challenge for establishing APIs for reproducible workflows.

The opinions above do not prevent someone else from picking up the torch to extend or support the reference server.

ljdursi:
@kozbo - I agree with what you’ve said; I’d just like to see deprecation (and merging of PRs) held off until the code base could be deprecated in favor of something else.

rishidev:
At a meeting on the future direction of the GA4GH GitHub and related matters, we had a discussion which was unanimous in how to proceed with the GitHub reference schema/server/test suites going forward. GA4GH itself is reorganising as a standards organisation. The master/working branch are 1000+ commits out of sync which noone has the engineering resources to reconcile. The plan is to extract schemas that will need future development, and focus on smaller, focused implementations of these items. 

For full details see https://docs.google.com/document/d/1DKktWM7QiAisALxWddUqqTRmlC7zgl-5A0yqveXJljM/edit

